PR #5714, opened 2018-08-23
tests: new test for cifs-mount interface

sergiocazzolato:


tests/main/interfaces-cifs-mount/task.yaml
```diff
@@ -0,0 +1,70 @@
+summary: Ensure that the cifs-mount interface works.
+
+details: |
+    The cifs-mount interface allows mounting and unmounting CIFS filesystems.
+
+prepare: |
+    snap install test-snapd-cifs-mount
+
+    # Create linux and samba users 
+    sudo useradd -s /bin/true test-cifs
```
bboozzoo: The tests are executed by root anyway, no need to use sudo.

tests/main/interfaces-cifs-mount/task.yaml
```diff
@@ -0,0 +1,70 @@
...
+    sudo smbpasswd -a test-cifs -s < test-pass
+
+    # Create mount point
+    mkdir -p local_share /var/snap/test-snapd-cifs-mount/current/local_mount
+    chmod 755 local_share
+    touch local_share/test
+
+    # Configure samba
+    cp /etc/samba/smb.conf /etc/samba/smb.conf.back
+    echo "[data]" >> /etc/samba/smb.conf
```
bboozzoo: This should work too:
```cat <<EOF > /etc/samba/smb.conf
[data]
path=$(pwd)/local_share
writable=yes
browsable=yes
```
shellcheck will probably suggest something like this:
```
(echo "[data]"
echo "path=$(pwd)/local_share"
echo "writable=yes"
echo "browsable=yes" ) > /etc/samba/smb.conf
```

tests/main/interfaces-cifs-mount/task.yaml
```diff
@@ -0,0 +1,70 @@
...
+    echo "path=$(pwd)/local_share" >> /etc/samba/smb.conf
+    echo "writable=yes" >> /etc/samba/smb.conf
+    echo "browsable=yes" >> /etc/samba/smb.conf
+
+    systemctl restart smbd.service
+    systemctl restart nmbd.service
```
bboozzoo: You don't need to touch `nmbd`, IIRC it's only used for NetBIOS name resolution and shouldn't matter here.

tests/main/interfaces-cifs-mount/task.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+execute: |
+    echo "The interface is not connected by default"
+    snap interfaces -i cifs-mount | MATCH -- '^- +test-snapd-cifs-mount:cifs-mount'
+
+    echo "When the interface is connected"
+    snap connect test-snapd-cifs-mount:cifs-mount
+
+    echo "Then the snap is able to mount and umount a cifs fs"
+    test-snapd-cifs-mount.sh -c "mount -t cifs //127.0.0.1/data /var/snap/test-snapd-cifs-mount/current/local_mount -o user=test-cifs,passowrd=test"
```
bboozzoo: I thin this line and ones below should use `$SNAP_DATA`, eg:
```
test-snapd-cifs-mount.sh -c "mount -t cifs //127.0.0.1/data \$SNAP_DATA/local_mount -o user=test-cifs,passowrd=test"
```

bboozzoo: And you probably need to mkdir the mount point first too:
```
test-snapd-cifs-mount.sh -c "mkdir \$SNAP_DATA/local_mount"
```

sergiocazzolato:
@bboozzoo thanks for reviewing this, all the commest have been addressed.

bboozzoo:
```
2018-08-27 16:02:15 Error executing google:ubuntu-16.04-64:tests/main/interfaces-cifs-mount : 
-----
+ echo 'The interface is not connected by default'
The interface is not connected by default
+ MATCH -- '^- +test-snapd-cifs-mount:cifs-mount'
+ snap interfaces -i cifs-mount
+ echo 'When the interface is connected'
When the interface is connected
+ snap connect test-snapd-cifs-mount:cifs-mount
+ echo 'Then the snap is able to mount and umount a cifs fs'
Then the snap is able to mount and umount a cifs fs
+ test-snapd-cifs-mount.sh -c 'mount -t cifs //127.0.0.1/data $SNAP_DATA/local_mount -o user=test-cifs,passowrd=test'
/bin/sh: 1: mount: Permission denied
-----
...
[Mon Aug 27 16:02:14 2018] audit: type=1400 audit(1535385735.584:105): apparmor="DENIED" operation="exec" profile="snap.test-snapd-cifs-mount.sh" name="/bin/mount" pid=25232 comm="sh" requested_mask="x" denied_mask="x" fsuid=0 ouid=0
```

Maybe the interface should add (u)mount to the interface?
```
/usr/bin/mount ixr,
/usr/bin/umount ixr,
```

Otherwise when you have for eg. a simple Python app, you cannot really use the interface unless you poke the libc via ctypes or somesuch. @zyga @jdstrand wdyt?

niemeyer:
@bboozzoo This interface is paired with mount-observe to have access to reading the current mounts.

@sergiocazzolato You'll probably need that too.

tests/main/interfaces-cifs-mount/task.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+    # Create mount point
+    mkdir -p local_share
+    test-snapd-cifs-mount.sh -c "mkdir \$SNAP_DATA/local_mount"
+    chmod 755 local_share
+    touch local_share/test
+
+    # Configure samba
+    cp /etc/samba/smb.conf /etc/samba/smb.conf.back
+    { echo "[data]"; echo "path=$(pwd)/local_share"; echo "writable=yes"; echo "browsable=yes"; } >> /etc/samba/smb.conf
```
niemeyer:
```
$ echo -e "foo\nbar"
foo
bar
```

tests/lib/snaps/test-snapd-cifs-mount/snapcraft.yaml
```diff
@@ -0,0 +1,15 @@
+name: test-snapd-cifs-mount
+version: 1.0
+summary: Basic cifs-mount snap
+description: A basic snap which allows mounting and unmounting CIFS filesystems
+
+apps:
+  sh:
+    command: sh
+    plugs: [cifs-mount]
```
niemeyer: Probably needs mount-observe, per comments.

tests/main/interfaces-cifs-mount/task.yaml
```diff
@@ -0,0 +1,66 @@
+summary: Ensure that the cifs-mount interface works.
+
+details: |
+    The cifs-mount interface allows mounting and unmounting CIFS filesystems.
+
+prepare: |
+    snap install test-snapd-cifs-mount
```
niemeyer: Shouldn't this be installing the local snap?

tests/main/interfaces-cifs-mount/task.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+    # Restore samba configuration
+    smbpasswd -x test-cifs || true
+    userdel test-cifs
+    cp /etc/samba/smb.conf.back /etc/samba/smb.conf || true
+    systemctl restart smbd.service
+
+execute: |
+    echo "The interface is not connected by default"
+    snap interfaces -i cifs-mount | MATCH -- '^- +test-snapd-cifs-mount:cifs-mount'
```
niemeyer: Why the --?

bboozzoo: Not strictly needed here because `'^..` will be detected as part of the pattern and handled porperly. However, the pattern `MATCH '- +test..'` will fail because `'-` will be seen as a `-foo` switch. So it's either `MATCH -- <pattern>`. We could use `grep .. -e`  switch which makes grep expec `<pattern>` as the next argument.

tests/lib/snaps/test-snapd-cifs-mount/snapcraft.yaml
sergiocazzolato: @niemeyer , thanks for reviewing this. I tried by adding the mount-observe plug but still having the same denial. testlog: https://paste.ubuntu.com/p/X56wt4CHrw/

tests/main/interfaces-cifs-mount/task.yaml
sergiocazzolato: As I need some dependencies as part of this snap, I had to create a snap and upload it to the store.

tests/main/interfaces-cifs-mount/task.yaml
sergiocazzolato: done

bboozzoo:
> This interface is paired with mount-observe to have access to reading the current mounts.

sergiocazzolato:
@chipaca, in the test we are already using mount.cifs.
> test-snapd-cifs-mount.sh -c "mount.cifs //127.0.0.1/data1 \$SNAP_DATA/local_mount -o user=test-cifs,passowrd=test"

zyga:
```
+ smbpasswd -x test-cifs
/bin/bash: line 64: smbpasswd: command not found
```

It looks like we need to patch the test to install another package @sergiocazzolato

zyga:
```
When the interface is connected
+ snap connect test-snapd-cifs-mount:cifs-mount
+ snap connect test-snapd-cifs-mount:network-control
error: snap "test-snapd-cifs-mount" has no plug named "network-control"
```

zyga:
This is now failing on:
```
+ test-snapd-cifs-mount.sh -c 'mount.cifs //127.0.0.1/data1 $SNAP_DATA/local_mount -o user=test-cifs,passowrd=test'
/bin/sh: 1: mount.cifs: not found
```

@sergiocazzolato can you look at this next week please. Feels like we are missing something :)

sergiocazzolato:
It is blocked until we find a solution for the denial trying to access to /run/mount/utab

sergiocazzolato:
Closeing the PR until a solution to the denial is found

PR #5717, opened 2018-08-24
snapd: go into degraded mode when the selftest fails

mvo5:
If the selftest fails the daemon dies currently. This is not ideal
because the user will need to dig through journalctl to find the
reason for the failure.

This PR changes the behaviour of the daemon to go into "degraded"
mode instead. The daemon will run but will reply with an internal
error to all API commands.

This means that on a system with no squashfs support the result
will look like this:
```
$ snap install josm
error: selftest failed: cannot mount squashfs image using "squashfs": exit
       status 1
```
which means the user has an easier time to figure out what is going
on. Before we reported:
```
error: cannot communicate with server: Post http://localhost/v2/snaps/josm: dial unix /run/snapd.socket: connect: no such file or directory
```

If this looks good I will add more unit tests.


cmd/snapd/main.go
```diff
@@ -102,6 +98,17 @@ func run() error {
...
 		return err
 	}
+
+	// Run selftest now, if anything goes wrong with the selftest we go
+	// into "degraded" mode where we always report the given error to
+	// any snap client.
+	if err := selftest.Run(); err != nil {
+		degradedErr := fmt.Errorf("selftest failed: %s", err)
+		logger.Noticef("%s", degradedErr)
+		logger.Noticef("Entering degraded mode")
```
zyga: Nitpick, it's easier to grep for if this is on one line: `logger.Noticef("entering degraded mode: %s", degradedErr)`

daemon/daemon.go
```diff
@@ -350,6 +350,16 @@ func (d *Daemon) Init() error {
...
 }
 
+// DegradedMode puts the daemon into an degraded mode which will
+// always return the internal error given in the "err" argument.
+//
+// This is useful to report permanent errors to the client when
+// the daemon cannot work because e.g. a selftest failed.
+func (d *Daemon) DegradedMode(err error) {
+	d.router = mux.NewRouter()
+	d.router.PathPrefix("/").Handler(InternalError(err.Error()))
```
zyga: I wonder if we should support a least enough to run "snap version" here. Eventually we could do more things (essentially all read-only queries should work). What do you think?

bboozzoo: `snap debug selftest`. Jokes aside, exposing the status via `/v2/system-info` (and the actual endpoint here too) makes it useful to non-cli front ends.

mvo5:
@niemeyer Thanks for your review. I agree that getting out of the degraded mode is desired. I looked into it and its unfortunately not super straight forward. Once the routes are setup (including the degraded router) its not easy to change it again so this requires some extra work. One cheap option would be to rerun the selftests and if the work we simply restart the daemon, I outlined this in http://paste.ubuntu.com/p/4xnD8pQvyG/ - its not super nice this is why I have not committed it yet. Or maybe @chipaca has an idea how to reset the degraded mode after d.Start() was called.

chipaca:
@mvo5 is it possible to recover from either of those situations without a reboot?

mvo5:
@chipaca I think the squashfs one yes - e.g. by compiling the module and inserting it. But  it requires pretty special knowledge.

chipaca:
One thing I just thought of is that we talked about going into 'readonly' mode when disk space was low. This is different from 'degraded' mode. Maybe we should write these modes down in the forum somewhere.

mvo5:
@chipaca Hm, it sounds like we could merge those modes. I mean, all self-tests so far are compatible with read-only mode. If confinement is not available or squashfs is not available in both cases all the other operations are fine. So we can probably merge them as long as we communicate to the user what is doing on (i.e. why the system is in read-only mode).

cmd/snapd/main.go
```diff
@@ -84,16 +91,23 @@ func runWatchdog(d *daemon.Daemon) (*time.Ticker, error) {
...
 }
 
-func run() error {
-	t0 := time.Now().Truncate(time.Millisecond)
-	httputil.SetUserAgentFromVersion(cmd.Version)
+var selftestRetryDelay = 300 * time.Second
 
-	if err := selftest.Run(); err != nil {
-		return fmt.Errorf("cannot start snapd: %v", err)
+func doSelftest(d *daemon.Daemon) (t *time.Timer) {
```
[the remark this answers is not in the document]
mvo5: Thanks for these suggestions! Much nicer this way.

PR #5719, opened 2018-08-24
strutil: add new ParseByteSize

mvo5:
The new ParseValueWithUnit helper allows the user to write
"500kB" which will be converted into 500*1000. This will be
used in the rate-limit PR to support options like:

    snap set core refresh.rate-limit=500kB

Split out to make the review simpler.


strutil/strutil.go
```diff
@@ -111,3 +111,37 @@ func TruncateOutput(data []byte, maxLines, maxBytes int) []byte {
 	}
 	return data
 }
+
+// ParseValueWithUnit parses a value like 500kB and returns the number
+// in byte
+func ParseValueWithUnit(inp string) (int64, error) {
+	unitMultiplier := map[string]int{
+		"B":  1,
+		"kB": 1000,
```
zyga: Should we support `KB` in case someone just types that out of a habit? (regardless if they meant 1024 or 1000)

strutil/strutil_test.go
```diff
@@ -137,3 +137,40 @@ func (ts *strutilSuite) TestTruncateOutput(c *check.C) {
...
+	}{
+		{"0B", 0},
+		{"1B", 1},
+		{"400B", 400},
+		{"1kB", 1000},
+		{"900kB", 900 * 1000},
+		{"1MB", 1000 * 1000},
+		{"20MB", 20 * 1000 * 1000},
+		{"1GB", 1000 * 1000 * 1000},
+		{"31GB", 31 * 1000 * 1000 * 1000},
```
zyga: Hard drive vendors approve of this number ;-)

strutil/strutil.go
bboozzoo: `KiB`?

strutil/strutil.go
```diff
@@ -111,3 +111,37 @@ func TruncateOutput(data []byte, maxLines, maxBytes int) []byte {
...
+		"GB": 1000 * 1000 * 1000,
+		"TB": 1000 * 1000 * 1000 * 1000,
+		"PB": 1000 * 1000 * 1000 * 1000 * 1000,
+		"EB": 1000 * 1000 * 1000 * 1000 * 1000 * 1000,
+	}
+	if len(inp) < 2 {
+		return 0, fmt.Errorf("cannot parse %q: need a unit", inp)
+	}
+	// simple case, "1234B", read as byte
+	if lastCh := inp[len(inp)-1]; lastCh == 'B' {
```
stolowski: Unless I'm missing something, we will always hit this case since every unit ends with 'B', so we will (usually) unnecessarily try convert the string to int, it will fail so we will carry on with two-char suffix check.
Since two-letter units will be used most of the time I suppose, would it make sense to reverse and re-organise the two conditions (`unitMultiplier[unit]` lookup failure wouldn't be fatal in such case), or alternatively enhance this one to check if the inp[len(inp)-2] character is a digit to avoid ParseInt?

strutil/strutil_test.go
stolowski: Can you add TB/EB/PB checks as well?

strutil/strutil.go
zyga: I meant that lower case and upper case difference is pretty subtle and may be missed by people. For usability we should perhaps ignore case here.

strutil/strutil.go
zyga: Yes but then we don't do anything when err != nil (we just continue and then hit the remaining cases)

stolowski: Yes, what I'm trying to say is we will hit ParseInt error case 99% of the time if we check for 'B' suffix first.

stolowski: And to be clear, the logic as currently is is perfectly correct, just suggesting to slightly re-organize the checks so that we only attempt ParseInt when we know unit is already correct (after two-character lookup, and then 'B' suffix check).

strutil/strutil.go
```diff
@@ -111,3 +111,37 @@ func TruncateOutput(data []byte, maxLines, maxBytes int) []byte {
...
+		if asByte, err := strconv.ParseInt(inp[:len(inp)-1], 10, 64); err == nil {
+			return asByte, nil
+		}
+	}
+	// two char suffix
+	unit := inp[len(inp)-2:]
+	mul, ok := unitMultiplier[unit]
```
bboozzoo: Or the old-school method and you can drop the map:
```
val := uint64(123)
unit := "GB"
units := []string{"B", "KB", "MB", "GB", "TB", "PB"}
for i := 0; i < len(units) && unit != units[i]; i++ {
   val *= 1000
}
```

strutil/strutil.go
mvo5: Yeah, I was sitting on the fence about it but I think it does make sense to make the users life easier this way. I will make it ignore the case.

strutil/strutil.go
```diff
@@ -111,3 +111,37 @@ func TruncateOutput(data []byte, maxLines, maxBytes int) []byte {
...
+	unitMultiplier := map[string]int{
+		"B":  1,
+		"kB": 1000,
+		"MB": 1000 * 1000,
+		"GB": 1000 * 1000 * 1000,
+		"TB": 1000 * 1000 * 1000 * 1000,
+		"PB": 1000 * 1000 * 1000 * 1000 * 1000,
+		"EB": 1000 * 1000 * 1000 * 1000 * 1000 * 1000,
+	}
+	if len(inp) < 2 {
```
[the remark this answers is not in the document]
mvo5: My gut feeling is that we should enforce a unit. The error message tries to help steer the user into the right direction.

strutil/strutil.go
mvo5: I changed the order of the checks now.

strutil/strutil.go
```diff
@@ -111,3 +111,41 @@ func TruncateOutput(data []byte, maxLines, maxBytes int) []byte {
 	}
 	return data
 }
+
+// ParseValueWithUnit parses a value like 500kB and returns the number
+// in bytes. The case of the unit will be ignored for user convenience.
+func ParseValueWithUnit(inp string) (int64, error) {
```
niemeyer: The function name doesn't give a hint about what kind of value we have at hand (time, distance, length, etc). Perhaps _ParseByteSize_ or _ParseByteUnit_ or _ParseSizeUnit_?

strutil/strutil.go
```diff
@@ -111,3 +111,41 @@ func TruncateOutput(data []byte, maxLines, maxBytes int) []byte {
...
+		return val * int64(mul), nil
+	}
+
+	// check for the simple case, e.g. "1234B"
+	if lastCh := inp[len(inp)-1]; lastCh != 'B' {
+		return 0, fmt.Errorf("cannot use unit in %q: try 'kB' or 'MB'", inp)
+	}
+	asByte, err := strconv.ParseInt(inp[:len(inp)-1], 10, 64)
+	if err != nil {
+		return 0, err
```
niemeyer: This error will probably look awkward with a string like "200GiB".

mvo5: Nice catch, will fix.

PR #5720, opened 2018-08-24
interfaces/apparmor: do not downgrade confinement on arch with linux-hardened 4.17.4+

bboozzoo:
The `linux-hardened` kernel available from `community` repo has AppArmor enabled.
Since it's a recent kernel, there should be no need to downgrade confinement.

### DIFF
--- a/interfaces/apparmor/backend.go
+++ b/interfaces/apparmor/backend.go
@@ -446,6 +446,29 @@ func addUpdateNSProfile(snapInfo *snap.Info, opts interfaces.ConfinementOptions,
 	}
 }
 
+func downgradeConfinement() bool {
+	kver := osutil.KernelVersion()
+	switch {
+	case release.DistroLike("opensuse-tumbleweed"):
+		if cmp, _ := strutil.VersionCompare(kver, "4.16"); cmp >= 0 {
+			// As a special exception, for openSUSE Tumbleweed which ships Linux
+			// 4.16, do not downgrade the confinement template.
+			return false
+		}
+	case release.DistroLike("arch"):
+		if !strings.HasSuffix(kver, "-hardened") {
+			// how did we get here anyway?
+			return true
+		}
+		if cmp, _ := strutil.VersionCompare(kver, "4.17.4"); cmp >= 0 {
+			// linux-hardened 4.17.4+ has apparmor enabled
+			return false
+		}
+	}
+
+	return true
+}
+
 func addContent(securityTag string, snapInfo *snap.Info, opts interfaces.ConfinementOptions, snippetForTag string, content map[string]*osutil.FileState) {
 	// Normally we use a specific apparmor template for all snap programs.
 	policy := defaultTemplate
@@ -467,10 +490,7 @@ func addContent(securityTag string, snapInfo *snap.Info, opts interfaces.Confine
 		// so the meaning of the template would change across kernel
 		// versions and we have not validated that the current template
 		// is operational on older kernels.
-		if cmp, _ := strutil.VersionCompare(osutil.KernelVersion(), "4.16"); cmp >= 0 && release.DistroLike("opensuse-tumbleweed") {
-			// As a special exception, for openSUSE Tumbleweed which ships Linux
-			// 4.16, do not downgrade the confinement template.
-		} else {
+		if downgradeConfinement() {
 			policy = classicTemplate
 			ignoreSnippets = true
 		}
--- a/interfaces/apparmor/backend.go
+++ b/interfaces/apparmor/backend.go
@@ -456,16 +456,15 @@ func downgradeConfinement() bool {
 			return false
 		}
 	case release.DistroLike("arch"):
-		if !strings.HasSuffix(kver, "-hardened") {
-			// how did we get here anyway?
-			return true
-		}
-		if cmp, _ := strutil.VersionCompare(kver, "4.17.4"); cmp >= 0 {
-			// linux-hardened 4.17.4+ has apparmor enabled
-			return false
+		if strings.HasSuffix(kver, "-hardened") {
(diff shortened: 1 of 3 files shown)


Erick555:
> The linux-hardened kernel available from community repo has AppArmor enabled.

linux-hardened is in [extra repo](https://www.archlinux.org/packages/extra/x86_64/linux-hardened/), not community.

AppArmor is supposed to be enabled at compile time (not runtime) in [main Arch Linux kernel](https://git.archlinux.org/svntogit/packages.git/commit/?id=c75a915313f72924fa0a3ed45356f9e0ea488f3b) as well. Perhaps around 4.18.6.

interfaces/apparmor/backend.go
```diff
@@ -446,6 +446,29 @@ func addUpdateNSProfile(snapInfo *snap.Info, opts interfaces.ConfinementOptions,
...
+			// As a special exception, for openSUSE Tumbleweed which ships Linux
+			// 4.16, do not downgrade the confinement template.
+			return false
+		}
+	case release.DistroLike("arch"):
+		if !strings.HasSuffix(kver, "-hardened") {
+			// how did we get here anyway?
+			return true
+		}
+		if cmp, _ := strutil.VersionCompare(kver, "4.17.4"); cmp >= 0 {
```
mvo5: Hm, won't we need to do the version check first and then the harden check? If the version is `4.17.4` then the `downgradeConfinement` helper here would return "true" because it first checks for !HasSuffix("-hardened") and that is false so it returns early. This also indicates a test is missing :)

interfaces/apparmor/backend_test.go
```diff
@@ -601,6 +601,41 @@ func (s *backendSuite) TestCombineSnippetsOpenSUSETumbleweedOldKernel(c *C) {
 	c.Check(profile, testutil.FileEquals, "\n#classic"+commonPrefix+"\nprofile \"snap.samba.smbd\" (attach_disconnected) {\n\n}\n")
 }
 
+func (s *backendSuite) TestCombineSnippetsArchSufficientHardened(c *C) {
+	restore := release.MockAppArmorLevel(release.PartialAppArmor)
```
mvo5: Given that we now have a helper (downgradeConfinement) I wonder if we can put that into export_test.go and test it directly? Then we can much easier test the various combinations with e.g. a table driven test.

bboozzoo:
> linux-hardened is in extra repo, not community.

My bad. IIRC it was in community, but maybe I'm misremembering things.

> AppArmor is supposed to be enabled at compile time (not runtime) in main Arch Linux kernel as well. Perhaps around 4.18.6.

Do you have any links to the discussion? I've only located this https://bugs.archlinux.org/task/59733 but there's no useful information thee.

interfaces/apparmor/backend.go
```diff
@@ -446,6 +446,29 @@ func addUpdateNSProfile(snapInfo *snap.Info, opts interfaces.ConfinementOptions,
...
+	switch {
+	case release.DistroLike("opensuse-tumbleweed"):
+		if cmp, _ := strutil.VersionCompare(kver, "4.16"); cmp >= 0 {
+			// As a special exception, for openSUSE Tumbleweed which ships Linux
+			// 4.16, do not downgrade the confinement template.
+			return false
+		}
+	case release.DistroLike("arch"):
+		if !strings.HasSuffix(kver, "-hardened") {
+			// how did we get here anyway?
```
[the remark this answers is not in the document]
bboozzoo: As of now, `linux-hardened` is the only kernel in the repositories that has AppArmor enabled. The comment was to note that getting here with a different kernel is unexpected. If 4.18.6 arch kernel is really released with AppArmor this will need to be updated accordingly

Erick555:
> My bad. IIRC it was in community, but maybe I'm misremembering things.

Yes, it was in community but was moved to extra some time ago.

> Do you have any links to the discussion? I've only located this https://bugs.archlinux.org/task/59733 but there's no useful information thee.

AppArmor/SElinux were disabled due to dependency on Audit which had performance concerns. Recently Audit was enabled in linux 4.18 because [performance concerns were gone](https://bugs.archlinux.org/task/42954#comment172055). That opened way to enabling AppArmor/SElinux as well.

Erick555:
Unfortunately Arch kernel maintainer got cold feet and there will be no AppArmor support in main kernel: https://git.archlinux.org/svntogit/packages.git/commit/?id=62b80117614b44f4a17921b0d1cb9b9e6d08fa4e

bboozzoo:
@Erick555 saw that. My understanding is that there's far less userspace to rebuild for AppApparmor than for SELinux, but both were killed anyway. Well, maybe some other time then.

Erick555:
@bboozzoo There was yet another U-turn and now AppArmor is enabled in [linux 4.18.8](https://git.archlinux.org/svntogit/packages.git/commit/?id=c46609a4b0325c363455264844091b71de01eddc) (also [linux-zen](https://git.archlinux.org/svntogit/packages.git/commit/trunk?h=packages/linux-zen&id=3ab708208c8e13d5de08dabebc442d2d0be7a585)) which is released in stable Arch repo. I'm not sure if it's the end of road but let's hope it is.

bboozzoo:
@Erick555 thanks for tracking this so closely. Yeah, I also find this back and forth a bit confusing. Hopefully, it stays enabled in the kernel config for longer thatn just next minor release. As soon as apparmor moves from `community-testing` to `community` I plan to add it as build dependency for `snapd`. Users who enabled AppArmor at boot will benefit, while those who did not should not see any regressions either.

Erick555:
@bboozzoo apparmor is now in community. Also as `linux` has same apparmor support as `linux-hardened` then this workaround should be enabled for `linux` too.

PR #5721, opened 2018-08-24
interfaces: retain order of inserted security backends

zyga:
This patch makes the interface repository retain the order of inserted
security backends. This in turns makes the interface manager respect
that order (as it uses the repository to know about available backends).
This in turn means that the order defined in the interfaces/backends/
package, specifically the All variable there, matters.

The order defined there is:
 - systemd
 - seccomp
 - dbus
 - udev
 - mount
 - kmod
 - apparmor (optionally)

The prior order is essentially the same list, sorted alphabetically:

 - apparmor (optionally)
 - dbus
 - kmod
 - mount
 - seccomp
 - systemd
 - udev

This means that the apparmor module would always go first. This was okay
for most interfaces but unfortunately the gpio interface depends on the
systemd backend (specifically it needs to read a file that only exists
after the systemd backend has completed its task).

Forum: https://forum.snapcraft.io/t/snap-enable-failure/6307/

### DIFF
--- a/interfaces/backends/backends.go
+++ b/interfaces/backends/backends.go
@@ -37,6 +37,8 @@ var All []interfaces.SecurityBackend = backends()
 
 func backends() []interfaces.SecurityBackend {
 	all := []interfaces.SecurityBackend{
+		// Because of how the GPIO interface is implemented the systemd backend
+		// must be earlier in the sequence than the apparmor backend.
 		&systemd.Backend{},
 		&seccomp.Backend{},
 		&dbus.Backend{},
--- a/testutil/checkers.go
+++ b/testutil/checkers.go
@@ -300,3 +300,75 @@ func (c *syscallsEqualChecker) Check(params []interface{}, names []string) (resu
 	}
 	return true, ""
 }
+
+type intChecker struct {
+	*check.CheckerInfo
+	rel string
+}
+
+func (checker *intChecker) Check(params []interface{}, names []string) (result bool, error string) {
+	a, ok := params[0].(int)
+	if !ok {
+		return false, "left-hand-side argument must be an int"
+	}
+	b, ok := params[1].(int)
+	if !ok {
+		return false, "right-hand-side argument must be an int"
+	}
+	switch checker.rel {
+	case "<":
+		result = a < b
+	case "<=":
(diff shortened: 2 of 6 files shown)


interfaces/backends/backends.go
```diff
@@ -37,6 +37,8 @@ var All []interfaces.SecurityBackend = backends()
 
 func backends() []interfaces.SecurityBackend {
 	all := []interfaces.SecurityBackend{
+		// Because of how the GPIO interface is implemented the systemd backend
```
mvo5: :heart:

testutil/checkers.go
```diff
@@ -300,3 +300,75 @@ func (c *syscallsEqualChecker) Check(params []interface{}, names []string) (resu
...
+		result = a != b
+	case ">":
+		result = a > b
+	case ">=":
+		result = a >= b
+	default:
+		return false, fmt.Sprintf("unexpected relation %q", checker.rel)
+	}
+	if !result {
+		error = fmt.Sprintf("failed relation %d %s %d", a, checker.rel, b)
```
bboozzoo: Maybe something in the lines of: `relation %d %s %d is not true`

testutil/checkers_test.go
```diff
@@ -374,3 +374,29 @@ func (s *CheckersS) TestSystemCallSequenceEqual(c *check.C) {
...
+	c.Assert(2, IntNotEqual, 1)
+	c.Assert(2, IntGreaterThan, 1)
+	c.Assert(2, IntGreaterEqual, 2)
+
+	// Wrong argument types.
+	testCheck(c, IntLessThan, false, "left-hand-side argument must be an int", false, 1)
+	testCheck(c, IntLessThan, false, "right-hand-side argument must be an int", 1, false)
+
+	// Relationship error.
+	testCheck(c, IntLessThan, false, "failed relation 2 < 1", 2, 1)
```
bboozzoo: Then you'd have `relation 2 < 1 is not true`.

mvo5:
I will just cherry-pick 3890ebb5ffaecb316bd9a3787ef0b24b86b2a77f the rest is just fluff ;)

PR #5593, opened 2018-08-02
tests: new test for hostname-control interface

sergiocazzolato:

PR #5722, opened 2018-08-24
tests: test for the hostname interface

sergiocazzolato:
As previous PR #5593 was closed, this is a clean one with the test.

### DIFF
--- a/tests/main/interfaces-hostname-control/task.yaml
+++ b/tests/main/interfaces-hostname-control/task.yaml
@@ -1,0 +1,55 @@
+summary: Ensure that the hostname-control interface works
+
+details: |
+    The hostname-control interface allows to configure the system hostname
+
+prepare: |
+    echo "Install test hostname-control snap"
+    #shellcheck source=tests/lib/snaps.sh
+    . "$TESTSLIB/snaps.sh"
+    install_local test-snapd-hostname-control
+
+restore: |
+    rm -f ./*.error
+
+execute: |
+    echo "Then the plug is disconnected by default"
+    snap interfaces -i hostname-control | MATCH -- '- +test-snapd-hostname-control:hostname-control'
+    
+    echo "When the plug is connected"
+    snap connect test-snapd-hostname-control:hostname-control
+    
+    echo "And we have the hostname for the current instance"
+    hostname="$(hostname)"
+    
+    echo "And the hostname command can be used"
+    snap_hostname="$(test-snapd-hostname-control.sh -c "/bin/hostname")"
+    [ "$hostname" = "$snap_hostname" ]
+    
+    echo "And the /etc/hostname file can be written"
+    test-snapd-hostname-control.sh -c "echo $hostname > /etc/hostname"
+    
+    echo "And hostname can be set/set through dbus"
+    test-snapd-hostname-control.sh -c "hostnamectl" | MATCH "$hostname"
+    test-snapd-hostname-control.sh -c "hostnamectl set-hostname $hostname"
+    
+    echo "When the plug is disconnected"
+    snap disconnect test-snapd-hostname-control:hostname-control
+    
+    if [ "$(snap debug confinement)" = partial ]; then
+        exit 0
+    fi
+
+    echo "Then the hostname method cannot be accessed through dbus"
+    if test-snapd-hostname-control.sh -c "hostnamectl set-hostname $hostname" 2>access.error; then
+        echo "Expected permission error trying to set hostname with disconnected plug"
+        exit 1
+    fi
+    MATCH "Permission denied" < access.error
+
+    echo "And the /etc/hostname file cannot be written"
+    if test-snapd-hostname-control.sh -c "echo $hostname > /etc/hostname" 2>hostname.error; then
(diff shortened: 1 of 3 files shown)


tests/main/interfaces-hostname-control/task.yaml
```diff
@@ -0,0 +1,65 @@
...
+    . "$TESTSLIB/systems.sh"
+
+    echo "Then the plug is disconnected by default"
+    snap interfaces -i hostname-control | MATCH -- '- +test-snapd-hostname-control:hostname-control'
+    
+    echo "When the plug is connected"
+    snap connect test-snapd-hostname-control:hostname-control
+    
+    echo "And we have the hostname for the current instance"
+    hostname="$(hostname)"
```
mvo5: I think we can put this before the interface connection. AIUI it is allowed by default so it's confusing if we test it as if it was only available after we do the interface connection.

sergiocazzolato: Thanks for reviewing this @mvo5 , change already done

bboozzoo:
Only fails on Core 18 now:
```
2018-08-27 16:08:02 Error executing google:ubuntu-core-18-64:tests/main/interfaces-hostname-control : 
-----
+ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/systems.sh
++ hostname
+ hostname=localhost
+ echo 'Then the plug is disconnected by default'
Then the plug is disconnected by default
+ MATCH -- '- +test-snapd-hostname-control:hostname-control'
+ snap interfaces -i hostname-control
+ echo 'When the plug is connected'
When the plug is connected
+ snap connect test-snapd-hostname-control:hostname-control
+ echo 'The hostname command can be used'
The hostname command can be used
++ test-snapd-hostname-control.sh -c /bin/hostname
+ snap_hostname=localhost
+ '[' localhost = localhost ']'
+ is_classic_system
+ [[ ubuntu-core-18-64 != ubuntu-core-* ]]
+ return 1
+ echo 'And hostname can be set/set through dbus'
And hostname can be set/set through dbus
+ MATCH localhost
+ test-snapd-hostname-control.sh -c hostnamectl
+ test-snapd-hostname-control.sh -c 'hostnamectl set-hostname localhost'
Could not set property: Failed to set static hostname: Read-only file system
-----
```
Some denials too, but I don't think those affect `hostnamectl`:
```
+ grep DENIED
+ dmesg --ctime
[Mon Aug 27 16:08:02 2018] audit: type=1400 audit(1535386082.134:107): apparmor="DENIED" operation="open" profile="snap.test-snapd-hostname-control.sh" name="/proc/1/environ" pid=3914 comm="hostnamectl" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
[Mon Aug 27 16:08:02 2018] audit: type=1400 audit(1535386082.134:108): apparmor="DENIED" operation="capable" profile="snap.test-snapd-hostname-control.sh" pid=3914 comm="hostnamectl" capability=12  capname="net_admin"
+ echo '# seccomp denials (kills) '
```

tests/main/interfaces-hostname-control/task.yaml
```diff
@@ -0,0 +1,64 @@
...
+    snap_hostname="$(test-snapd-hostname-control.sh -c "/bin/hostname")"
+    [ "$hostname" = "$snap_hostname" ]
+
+    # On core systems /etc/hostname is not writable
+    if is_classic_system; then
+        echo "And the /etc/hostname file can be written"
+        test-snapd-hostname-control.sh -c "echo $hostname > /etc/hostname"
+    fi
+
+    echo "And hostname can be set/set through dbus"
```
niemeyer: set/set?

tests/main/interfaces-hostname-control/task.yaml
```diff
@@ -0,0 +1,64 @@
...
+restore: |
+    rm -f ./*.error
+
+execute: |
+    #shellcheck source=tests/lib/systems.sh
+    . "$TESTSLIB/systems.sh"
+
+    hostname="$(hostname)"
+
+    echo "Then the plug is disconnected by default"
```
niemeyer: s/Then the/The/?

sergiocazzolato:
@jdstrand could you please check the denial we are getting for core-18?
Thanks

jdstrand:
"test-snapd-hostname-control.sh -c 'hostnamectl set-hostname localhost'
Could not set property: Failed to set static hostname: Read-only file system"

This should be an indication that /etc/hostname is not writable. /etc/hostname should, AFAIK, be a symlink to /etc/writable/hostname. If it is not, that is part of the problem, however hostname_control.go needs to be updated to change this:
```
/etc/hostname w,
```
to
```
/etc/{,writable/}hostname w,
```

sergiocazzolato:
@jdstrand  I did the change in the interface and I see the same error https://paste.ubuntu.com/p/qz3XMBb7hQ/, so, should we need a change in the core image? 

@mvo5 what do you think?

sergiocazzolato:
I am mergin this because the original PR #5593 had all the approvals (and this PR change is identical).

PR #5726, opened 2018-08-27
release, interfaces: make snapd degrade gracefully when AppArmor userspace tooling is unavailable

bboozzoo:
This is an attempt to make snapd fail gracefully when the host system has AppArmor support in the kernel, but the userspace tooling is not present. Namely when `apparmor_parser` is unavaialble, do not fail hard, preventing installation/removal of snaps. Instead, degrade AppArmor support level to NoAppArmor and work with that.


bboozzoo:
Closing for now, we can always revisit if needed.

PR #5727, opened 2018-08-27
interfaces/browser-support, cmd/snap-seccomp: Allow read-only ptrace, for the Breakpad crash reporter.

jld:
Breakpad, which is used for crash reporting by Firefox, uses ptrace to
read register/memory state from the crashed process, but it doesn't
need to modify any state.  We can allow that without allowing ptrace
operations that write registers, which can be used to bypass security;
see https://lkml.org/lkml/2016/5/26/354.

See also the Firefox bug report:
https://bugzilla.mozilla.org/show_bug.cgi?id=1461848.

Caveats: This commit adds Breakpad support to the `browser_support`
interface rather than trying to factor it out; note that Breakpad
may have other requirements that happen to be satisfied by
`browser_support`.  Also, the snap-seccomp change includes only the
`PTRACE_*` constants needed here, not the full set that might appear in
syscall args.

---

Licensing note: Any copyright on this is owned by Mozilla, which has executed the corporate CLA; I can also file a CLA myself if necessary.


cmd/snap-seccomp/main.go
```diff
@@ -388,6 +388,18 @@ var seccompResolver = map[string]uint64{
...
+
+	// man 2 ptrace
+	"PTRACE_ATTACH":     syscall.PTRACE_ATTACH,
+	"PTRACE_DETACH":     syscall.PTRACE_DETACH,
+	"PTRACE_GETREGS":    syscall.PTRACE_GETREGS,
+	"PTRACE_GETFPREGS":  syscall.PTRACE_GETFPREGS,
+	"PTRACE_GETFPXREGS": syscall.PTRACE_GETFPXREGS,
+	"PTRACE_GETREGSET":  syscall.PTRACE_GETREGSET,
+	"PTRACE_PEEKDATA":   syscall.PTRACE_PEEKDATA,
+	// <linux/ptrace.h> and <sys/ptrace.h> have different spellings for PEEKUS{,E}R
```
zyga: Thank you for this comment.

cmd/snap-seccomp/main.go
```diff
@@ -388,6 +388,18 @@ var seccompResolver = map[string]uint64{
 	"NETLINK_RDMA":           C.NETLINK_RDMA,
 	"NETLINK_CRYPTO":         C.NETLINK_CRYPTO,
 	"NETLINK_INET_DIAG":      C.NETLINK_INET_DIAG, // synonymous with NETLINK_SOCK_DIAG
+
```
zyga: Please run `go fmt` on this file.

interfaces/builtin/browser_support.go
```diff
@@ -268,6 +268,22 @@ setpriority
...
 mknod - |S_IFCHR -
 mknodat - - |S_IFCHR -
+
+# The Breakpad crash reporter uses ptrace to read register/memory state
+# from the crashed process, but it doesn't need to modify any state; see
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1461848.
+#
+# These rules allow that but don't allow ptrace operations that
+# write registers, which can be used to bypass security; see
+# https://lkml.org/lkml/2016/5/26/354.
```
zyga: This is fine

cmd/snap-seccomp/main.go
jld: Thanks; the last time I wrote any Go was years ago, so I'd forgotten about `go fmt`.

jld:
@jdstrand Is there anything else I need to do here?  The CI run seems to have passed (after failing a few times from what looked like transient network problems unrelated to the patch).

chipaca:
About the CLA, I'm checking whether I can update the cla checker to say "if @mozilla.com then a'ight". If it takes me longer than expected to get an answer I might ask you to sign it yourself, to expedite things, but only if this gets a go-ahead from jdstrand before then.

interfaces/builtin/browser_support.go
```diff
@@ -257,6 +257,22 @@ setpriority
...
+ptrace PTRACE_ATTACH
+ptrace PTRACE_DETACH
+ptrace PTRACE_GETREGS
+ptrace PTRACE_GETFPREGS
+ptrace PTRACE_GETFPXREGS
+ptrace PTRACE_GETREGSET
+ptrace PTRACE_PEEKDATA
+ptrace PTRACE_PEEKUSER
```
jdstrand: Currently the corresponding apparmor rules that allow the above are in browserSupportConnectedPlugAppArmorWithSandbox, therefore these seccomp calls should be in browserSupportConnectedPlugSecCompWithSandbox, not browserSupportConnectedPlugSecComp as they are here.

cmd/snap-seccomp/main.go
```diff
@@ -388,6 +388,18 @@ var seccompResolver = map[string]uint64{
...
+	// man 2 ptrace
+	"PTRACE_ATTACH":     syscall.PTRACE_ATTACH,
+	"PTRACE_DETACH":     syscall.PTRACE_DETACH,
```
jdstrand: I'm a little surprised that you didn't need PTRACE_CONT here. It is fine to not have it since we are only concerned about Breakpad, but would it make sense to include it here and in the policy below for future-proofing? Not a blocker.

cmd/snap-seccomp/main.go
```diff
@@ -388,6 +388,18 @@ var seccompResolver = map[string]uint64{
 	"NETLINK_RDMA":           C.NETLINK_RDMA,
 	"NETLINK_CRYPTO":         C.NETLINK_CRYPTO,
 	"NETLINK_INET_DIAG":      C.NETLINK_INET_DIAG, // synonymous with NETLINK_SOCK_DIAG
+
+	// man 2 ptrace
```
jdstrand: Can you adjust this to be: `// man 2 ptrace (subset)`

mvo5:
Please also merge master to get the updated CLA check code :)

chipaca:
done :-)

interfaces/builtin/browser_support.go
bboozzoo: Moved

bboozzoo:
@jld pushed the changes requested by @jdstrand, please pull first before adding any new patches.

@jdstrand can you take another look?

PR #5699, opened 2018-08-21
overlord,store: support proxy settings internally too

mvo5:
This PR adds support to read the proxy settings from the core config. 

There is some more work in devicestate (which creates its own httputil.NewHTTPClient). If this looks reasonable so far I will attack devicestate.

PR #5729, opened 2018-08-28
many: rename ClientOpts to ClientOptions

mvo5:
In https://github.com/snapcore/snapd/pull/5699 it was pointed out that the httputil.ClientOpts should be named httputil.ClientOptions for consistency. This PR does the rename.

Build on top of https://github.com/snapcore/snapd/pull/5699


overlord/configstate/proxyconf/proxy.go
```diff
@@ -0,0 +1,57 @@
...
+	return &ProxySettings{st: st}
+}
+
+func (p *ProxySettings) Conf(req *http.Request) (*url.URL, error) {
+	p.st.Lock()
+	tr := config.NewTransaction(p.st)
+	p.st.Unlock()
+
+	var proxy string
+	err := tr.Get("core", fmt.Sprintf("proxy.%s", req.URL.Scheme), &proxy)
```
bboozzoo: Since we allow proxy per URL scheme, a test that verifies `proxy.https` is effective could be useful

PR #5733, opened 2018-08-28
snap/squashfs: improve error message from Build on mksquashfs failure

chipaca:
Without this change:

    $ snap pack test-snapd-complexion/
    error: cannot pack "test-snapd-complexion/": exit status 1

With this change:

    $ snap pack test-snapd-complexion/
    error: cannot pack "test-snapd-complexion/": mksquashfs call failed:
    -----
    mksquashfs: Compressor "xz" is not supported!
    mksquashfs: Compressors available:
            gzip (default)
    -----

### DIFF
--- a/snap/squashfs/squashfs.go
+++ b/snap/squashfs/squashfs.go
@@ -313,10 +313,11 @@ func (s *Snap) Build(buildDir string) error {
 			"-no-fragments",
 			"-no-progress",
 		).CombinedOutput()
-		if err == nil {
-			return nil
+		if err != nil {
+			return fmt.Errorf("mksquashfs call failed: %s", osutil.OutputErr(output, err))
 		}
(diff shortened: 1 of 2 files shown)


snap/squashfs/squashfs.go
```diff
@@ -304,15 +304,19 @@ func (s *Snap) Build(buildDir string) error {
...
 			"mksquashfs",
 			".", fullSnapPath,
 			"-noappend",
 			"-comp", "xz",
 			"-no-xattrs",
 			"-no-fragments",
 			"-no-progress",
-		).Run()
+		).CombinedOutput()
+		if err == nil {
```
mvo5: Hm, why not the usual pattern of `if err != nil {` here? Is there an advantage?

chipaca: I think the first iteration had more code below this check, so the "exit early" rule won

PR #5736, opened 2018-08-29
cmd: add systemd environment generator

mvo5:
We currently only ship an environment generator for the user
sessions. However we need a environment generator that runs
at system startup time and creates environment available for
units that run outside of the user context (e.g. the
cloud-init-service). This PR adds such a system generator.

This was already tried in #5226 but back then the environment.d approach
was taken. It turns out this approach only adds environment to the *user*
sessions not the system.


cmd/snapd-env-generator/main.c
```diff
@@ -0,0 +1,34 @@
...
+#include<stdio.h>
+
+#include "config.h"
+
+// Systemd environment generators work since version 233 which ships
+// in Ubuntu 17.10+
+int main(int argc, char **argv)
+{
+	char *path = getenv("PATH");
+	if (strstr(path, "/snap/bin") != NULL) {
```
bboozzoo: I think this should be `strstr(path, SNAP_MOUNT_DIR "/bin")`

cmd/Makefile.am
```diff
@@ -465,6 +465,14 @@ libexec_PROGRAMS += snapd-generator/snapd-generator
 snapd_generator_snapd_generator_SOURCES = snapd-generator/main.c
 snapd_generator_snapd_generator_LDADD = libsnap-confine-private.a
 
+##
+## snapd-env-generator
+##
+
+libexec_PROGRAMS += snapd-env-generator/snapd-env-generator
```
bboozzoo: We could have shipped a shell script :smile:

bboozzoo: Just for the record (before I forget about that again), ideally we should have something like:
```
systemdsystemenvgeneratordir=$(SYSTEMD_SYSTEM_ENV_GENERATOR_DIR)
systemdsystemenvgenerator_PROGRAMS = snapd-env-generator/snapd-env-generator
```
and drop the unnecessary `mv ../snapd/snapd-generator /usr/lib/systemd..`  in packaging.

As a side note, `pkg-config --print-variables systemd` does not list any variable that points to system env generator directory :-1: 

Obviously no reason to fix it in this PR.

mvo5: Yes, the first iteration of this PR did that. Its fine if we decide to do. However this program is run for each unit that systemd starts and given how simple it is what we need to do I think C is smaller/faster (and it is saying something if I select C!).

bboozzoo: Probably as much as ought to be done in C anyway :smile:

bboozzoo:
On Amazon Linux:
```
RPM build errors:
    File listed twice: /usr/lib/systemd/system/snapd.seeded.service
    File must begin with "/": %{_systemd_system_env_generator_dir}/snapd-env-generator
```

cmd/Makefile.am
```diff
@@ -7,6 +7,9 @@ dist_man_MANS =
 noinst_PROGRAMS =
 noinst_LIBRARIES =
 
+# FIXME: get this via something like pkgconf once it is defined there
+SYSTEMD_SYSTEM_ENV_GENERATOR_DIR=/usr/lib/systemd/system-environment-generators
```
zyga: Not perfect but we might use:
```
pkg-config systemd --varbiale=systemdsystemgenaratordir
```

cmd/Makefile.am
```diff
@@ -465,6 +468,15 @@ libexec_PROGRAMS += snapd-generator/snapd-generator
...
 snapd_generator_snapd_generator_LDADD = libsnap-confine-private.a
 
+##
+## snapd-env-generator
+##
+
+systemdsystemenvgeneratordir=$(SYSTEMD_SYSTEM_ENV_GENERATOR_DIR)
+systemdsystemenvgenerator_PROGRAMS = snapd-env-generator/snapd-env-generator
+
+snapd_env_generator_snapd_env_generator_SOURCES = snapd-env-generator/main.c
```
zyga: This is fine

cmd/snapd-env-generator/main.c
```diff
@@ -0,0 +1,37 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include<stdlib.h>
```
zyga: Please run `make fmt` here. The indenting is rather unexpected.

cmd/snapd-env-generator/main.c
```diff
@@ -0,0 +1,37 @@
...
+#include "config.h"
+
+// Systemd environment generators work since version 233 which ships
+// in Ubuntu 17.10+
+int main(int argc, char **argv)
+{
+        const char *snap_bin_dir = SNAP_MOUNT_DIR"/bin";
+
+	char *path = getenv("PATH");
+	if (strstr(path, snap_bin_dir) != NULL) {
```
zyga: Not a big deal but we could use strchr to iterate over PATH components and then use strcmp to ensure that we have a perfect fit. The current approach with `strstr` would match PATH that is `/snap/bin/baz:/usr/bin` for example.

cmd/snapd-env-generator/main.c
```diff
@@ -0,0 +1,37 @@
...
+
+#include "config.h"
+
+// Systemd environment generators work since version 233 which ships
+// in Ubuntu 17.10+
+int main(int argc, char **argv)
+{
+        const char *snap_bin_dir = SNAP_MOUNT_DIR"/bin";
+
+	char *path = getenv("PATH");
```
zyga: Please note that `getenv` may return `NULL`

cmd/snapd-env-generator/main.c
bboozzoo: So how about that shell script? :smile: If I'm not mistaken, the whole thing could be:
```
#!/bin/sh
snap_bin_path="@SNAP_MOUNT_DIR@/bin"
if [ -n "${PATH##*${snap_bin_path}}" -a -n "${PATH##*${snap_bin_path}:*}" ]; then
    echo PATH=$PATH:${snap_bin_path}
fi
```

cmd/snapd-env-generator/main.c
```diff
@@ -0,0 +1,37 @@
...
+#include<string.h>
+#include<stdio.h>
+
+#include "config.h"
+
+// Systemd environment generators work since version 233 which ships
+// in Ubuntu 17.10+
+int main(int argc, char **argv)
+{
+        const char *snap_bin_dir = SNAP_MOUNT_DIR"/bin";
```
bboozzoo: Didn't notice it before and indent will probably fix it, but it should be `SNAP_MOUNT_DIR "/bin"` (gcc used to yell about that).

cmd/Makefile.am
mvo5: When I run this (with the typo fixed) I get empty output.

cmd/snapd-env-generator/main.c
mvo5: Sure, done. Note that only a single line changed:
```
-        const char *snap_bin_dir = SNAP_MOUNT_DIR"/bin";
+       const char *snap_bin_dir = SNAP_MOUNT_DIR "/bin";
```

cmd/Makefile.am
bboozzoo: `systemdsystemgenaratordir` is a different thing. There was no var for system env generators dir, hence I did not bring it up.

bboozzoo: For reference:
```
maciek@galeon:~ pkg-config --print-variables systemd
containeruidbasemax
containeruidbasemin
dynamicuidmax
dynamicuidmin
systemgidmax
systemuidmax
catalogdir
modulesloaddir
binfmtdir
sysctldir
sysusersdir
tmpfilesdir
systemdshutdowndir
systemdsleepdir
systemdusergeneratordir
systemdsystemgeneratordir
systemduserunitpath
systemdsystemunitpath
systemduserconfdir
systemdsystemconfdir
systemduserpresetdir
systemduserunitdir
systemdsystempresetdir
systemdsystemunitdir
systemdutildir
prefix
pcfiledir

# not what we want
maciek@galeon:~ pkg-config --variable=systemdsystemgeneratordir systemd
/usr/lib/systemd/system-generators
```

zyga: Indeed, let's leave it as-is.

cmd/snapd-env-generator/main.c
zyga: What? did indent now decide to `#include<foo>`? Oh the horror :)

cmd/Makefile.am
```diff
@@ -465,6 +468,26 @@ libexec_PROGRAMS += snapd-generator/snapd-generator
...
+snapd_env_generator_snapd_env_generator_LDADD = libsnap-confine-private.a
+EXTRA_DIST += snapd-env-generator/snapd-env-generator.rst
+
+if HAVE_RST2MAN
+snapd-env-generator/%.7: snapd-env-generator/%.rst
```
zyga: ❤️ thank you!

PR #5739, opened 2018-08-29
interfaces/default: don't scrub with change_profile with classic

jdstrand:
The kernel performs its unsafe_exec routines under various circumstances
(see 'man 8 ld.so' (secure-execution) and 'man 5 apparmor.d' for
details). One such condition is on an AppArmor enabled system where a
process is running under a security label that is not 'unconfined' and
it uses libapparmor's aa_change_*() functions to change to another
profile.

The 'classic confinement' template allows use of the aa_change_*() API,
but it only has a bare 'change_profile' rule. Because the classic
template is already wide open, there is no meaningful security benefit
with scrubbing the environment with profile transitions (even in the
case of launching another snap, if the exec'd snap is in strict mode,
its confinement won't allow loading files outside of its confinement),
but scrubbing the environment causes unexpected side-effects as seen
with microk8s, docker and kubeflow.

As such, for the classic template, disable environment scrubbing when
using aa_change_*(). Importantly, these changes do *not* modify normal
secure-execution checks by the kernel (eg, for setuid). Eg,
LD_LIBRARY_PATH/etc is still scrubbed when executing snaps since they go
through snap-confine.

Because not all systems have a parser that understands the 'unsafe'
syntax, probe the parser to determine the support and conditionally add
to the classic template if the parser supports it. For completeness, add
the parser features to the system-key and sandbox features.

References:
- https://forum.snapcraft.io/t/ld-library-path-in-snapped-docker/6903/16
- https://github.com/docker/docker-snap/issues/30
- https://github.com/kubeflow/kubeflow/issues/1367


jdstrand:
The test errors all seem unrelated.

interfaces/apparmor/template.go
```diff
@@ -490,7 +490,7 @@ var classicTemplate = `
   /** pix,
 
   capability,
-  change_profile,
+  change_profile unsafe /**,
```
bboozzoo: Do you need a similar change in docker-support interface?

jdstrand: Yes, but it needs special handling that will be done in a separate PR.

jdstrand:
@zyga - I saw this error but didn't have time to look into it. I can say I was surprised that the core configure hook was using the classic template.

zyga:
@jdstrand on systems with partial apparmor everything gets a classic template.

ktsakalozos:
Is there any update on this PR?

carmine:
Any updates?

jdstrand:
@zyga - "@jdstrand on systems with partial apparmor everything gets a classic template."

Yes, but on those ignoreSnippets is set to 'true' so there shouldn't be a conflict there. I just refreshed with master. Let's see what happens.

interfaces/apparmor/template.go
jdstrand: FYI, I have a plan for docker-support that, while not ideal, will work. That will be in a separate PR.

jdstrand:
FYI, it looks like opensuse-42.3 is back so it is failing again. I pushed some changes for testing apparmor_parser features which verifies if 'change_profile unsafe ...' is a valid rule and if not, kicks it down to PartialApparmor. Then adjust the classic template to conditionally use 'change_profile unsafe' only with FullAppArmor. I'll add some tests for all of this in the next commits.

jdstrand:
@zyga - ok, I adjusted the PR for the following:

"Because not all systems have a parser that understands the 'unsafe'
syntax, probe the parser to determine the support and conditionally add
to the classic template if the parser supports it. For completeness, add
the parser features to the system-key and sandbox features."

jdstrand:
Ok, down to one failure which I really think is unrelated since it was killed due to taking too long:
```
2018-10-09 21:01:10 Failed tasks: 1
    - google:arch-linux-64:tests/main/snap-run
```

interfaces/apparmor/backend.go
```diff
@@ -572,13 +580,20 @@ func (b *Backend) SandboxFeatures() []string {
...
-	tags := make([]string, 0, len(features))
+	pFeatures := parserFeatures()
+	tags := make([]string, 0, len(features)+len(pFeatures))
 	for _, feature := range features {
 		// Prepend "kernel:" to apparmor kernel features to namespace them and
 		// allow us to introduce our own tags later.
 		tags = append(tags, "kernel:"+feature)
 	}
 
+	for _, feature := range pFeatures {
```
zyga: This is nice, thank you!

release/apparmor.go
```diff
@@ -133,3 +143,48 @@ func AppArmorFeatures() []string {
...
+}
+
+var requestedParserFeatures = []apparmorParserFeature{
+	{"unsafe", "change_profile unsafe /**,"},
+}
+
+// tryParser will run the parser on the rule to determine if the feature is
+// supported.
+func tryParser(rule string) bool {
+	cmd := exec.Command("apparmor_parser", "-p")
```
zyga: What will happen when `apparmor_parser` is not on PATH? Like on SELinux distributions? Specifically I think we should just return false without any errors.

zyga: Could you also use the `--preprocess` spelling rather than the short version. This is more discoverable to casual readers.

jdstrand: I didn't trace it all down, but spread wasn't grumpy about it. I added a LookupPath just in case just now.

release/apparmor.go
```diff
@@ -133,3 +143,55 @@ func AppArmorFeatures() []string {
...
+var requestedParserFeatures = []apparmorParserFeature{
+	{"unsafe", "change_profile unsafe /**,"},
+}
+
+// tryParser will run the parser on the rule to determine if the feature is
+// supported.
+func tryParser(rule string) bool {
+	parser := "apparmor_parser"
+
+	_, err := exec.LookPath(parser)
```
mvo5: AFAICT this function could be simplified a bit:
```
func tryParser(rule string) bool {
	cmd := exec.Command("apparmor_parser", "--preprocess")
	cmd.Stdin = bytes.NewBufferString(fmt.Sprintf("profile snap-test {\n %s\n}", rule))
	if _, err := cmd.CombinedOutput(); err != nil {
		return false
	}
	return true
}
```
Because we don't return anything different if apparmor_parser does not exist or does not support the feature we can skip the LookupPath bit.

release/apparmor.go
```diff
@@ -133,3 +143,55 @@ func AppArmorFeatures() []string {
...
+	if err != nil {
+		return false
+	}
+
+	cmd := exec.Command(parser, "--preprocess")
+	stdin, err := cmd.StdinPipe()
```
mvo5: Go just needs an io.Reader for stdin. So this can be simplified to something like: `cmd.Stdin = bytes.NewBufferString(fmt.Sprintf("profile snap-test {\n %s\n}", rule))` and the go func() can go with that change.

interfaces/system_key.go
```diff
@@ -118,6 +119,9 @@ func generateSystemKey() (*systemKey, error) {
 	// Add apparmor-features (which is already sorted)
 	sk.AppArmorFeatures = release.AppArmorFeatures()
 
+	// Add apparmor-parser-features (which is already sorted)
+	sk.AppArmorParserFeatures = release.AppArmorParserFeatures()
```
mvo5: Nice

release/apparmor.go
```diff
@@ -133,3 +143,55 @@ func AppArmorFeatures() []string {
...
+	_, err = cmd.CombinedOutput()
+	if err != nil {
+		return false
+	}
+	return true
+}
+
+// probeAppArmorParser returns a sorted list of apparmor features like
+// []string{"unsafe", ...}.
+func probeAppArmorParser() []string {
```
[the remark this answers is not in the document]
jdstrand: Looks fine. I saw it in init() so didn't break it out into a different test, but hey, the more tests the better. :)

release/apparmor.go
jdstrand: Heh, this was what was originally there but @zyga wanted something explicit the lookup path :) I don't personally care either way ;)

release/apparmor.go
jdstrand: Ok. Do you want this now or later?

jdstrand:
Thanks for merging this! :)

ktsakalozos:
Thank you for this work. Do we have an ETA on when this will land on production? Thank you.

PR #5746, opened 2018-08-31
wrappers: remove Wants=network-online.target

mvo5:
When generating units we always and unconditionally pull in
Wants=network-online.target currently. However as reported in

    https://forum.snapcraft.io/t/6063

this will slow down service startup by 2 minutes if any of the
network interfaces is not ready. Instead this PR changes it to
the network.target which is less strict and only means that the
systems network stack is up (and conversely that services should
be stopped before the network is stopped).

Note that this change exiting behaviour and might break existing
snaps. However arguable those are already broken because the
network may go up and down so network-online is not a good
gurantee for something that the software itself should deal
with.

### DIFF
--- a/systemd/systemd.go
+++ b/systemd/systemd.go
@@ -131,7 +131,7 @@ const (
 	ServicesTarget = "multi-user.target"
 
 	// the target prerequisite for systemd units we generate
-	PrerequisiteTarget = "network-online.target"
+	PrerequisiteTarget = "network.target"
 
 	// the default target for systemd socket units that we generate
 	SocketsTarget = "sockets.target"
--- a/wrappers/services_gen_test.go
+++ b/wrappers/services_gen_test.go
@@ -46,8 +46,8 @@ const expectedServiceFmt = `[Unit]
 # Auto-generated, DO NOT EDIT
 Description=Service for snap application snap.app
 Requires=%s-snap-44.mount
-Wants=network-online.target
-After=%s-snap-44.mount network-online.target
+Wants=network.target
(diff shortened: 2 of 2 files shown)


zyga:
Note: it is unfortunate we don't refresh wrappers like we do security profiles so this change will still take time to take effect.

zyga:
Some documentation about the network and network-online targets:
```
       network-online.target
           Units that strictly require a configured network connection should
           pull in network-online.target (via a Wants= type dependency) and
           order themselves after it. This target unit is intended to pull in
           a service that delays further execution until the network is
           sufficiently set up. What precisely this requires is left to the
           implementation of the network managing service.

           Note the distinction between this unit and network.target. This
           unit is an active unit (i.e. pulled in by the consumer rather than
           the provider of this functionality) and pulls in a service which
           possibly adds substantial delays to further execution. In contrast,
           network.target is a passive unit (i.e. pulled in by the provider of
           the functionality, rather than the consumer) that usually does not
           delay execution much. Usually, network.target is part of the boot
           of most systems, while network-online.target is not, except when at
           least one unit requires it. Also see Running Services After the
           Network is up[1] for more information.

           All mount units for remote network file systems automatically pull
           in this unit, and order themselves after it. Note that networking
           daemons that simply provide functionality to other hosts generally
           do not need to pull this in.

           Note that this unit is only useful during the original system
           start-up logic. After the system has completed booting up, it will
           not track the online state of the system anymore. Due to this it
           cannot be used as a network connection monitor concept, it is
           purely a one-time system start-up concept.

```

PR #5713, opened 2018-08-23
many: mount namespace mapping for parallel installs of snaps

bboozzoo:
The series add a support for making adjustments to mount namespace of parallel installed snaps and supersedes previous attempt from #5654.

This time, the update is performed by `snap-update-ns` via mount profiles.  The mapping of `/snap/foo_bar -> /snap/foo` and `/var/snap/foo_bar -> /var/snap/foo` are done via the snap mount profile. `$SNAP_USER_DATA` is handled using user mount profile. There is some complication getting the user's `$HOME` inside `snap-update-ns`. For this reason, the user mount profile includes `$USER_HOME_SNAP_DIR` which corresponds to `/home/joe/snap`, `snap-confine` passes `$SNAP_INSTANCE_USER_DATA` env var to `snap-update-ns` which is then used to recover/compute the proper value of `$USER_HOME_SNAP_DIR` for the given user.

The code includes all bits for mounts. However, the AppArmor bits only cover `snap-update-ns` execution. Thus installing snaps on systems with full confinement is possible, but various things inside the mount ns may not work.

There are 2 spread tests included in the branch. `parallel-install-basic` does basic installation, runs the app and verifies that data is written to correct location. `paalel-install-layout` adds layouts into the mix and verifies that the layout as preserved correctly. In both tests, the instance-keyed snap is installed in devmode on Ubuntu. This is expected to be fixed once the default profile is updated.

PR #5749, opened 2018-08-31
interfaces: make ifacetest.InstallSnap() take snap instance name

bboozzoo:
Make the helper easier to use when testing parallel snap instances code.

Offshoot off #5713


interfaces/ifacetest/backendtest.go
```diff
@@ -147,10 +147,15 @@ slots:
 // Support code for tests
 
 // InstallSnap "installs" a snap from YAML.
-func (s *BackendSuite) InstallSnap(c *C, opts interfaces.ConfinementOptions, snapYaml string, revision int) *snap.Info {
+func (s *BackendSuite) InstallSnap(c *C, opts interfaces.ConfinementOptions, instanceName, snapYaml string, revision int) *snap.Info {
 	snapInfo := snaptest.MockInfo(c, snapYaml, &snap.SideInfo{
 		Revision: snap.R(revision),
 	})
+
+	_, instanceKey := snap.SplitInstanceName(instanceName)
```
niemeyer: Per review in the prior PR, can't we make the name default to the one in the snapInfo, so we don't have to repeat it all the time?  This way most cases would become an empty string instead of repeating the same name which is then ignored here. Also, would be good to validate that when provided the name matches the one in the info instead of ignoring it, to prevent awkwardness.

bboozzoo: I recall an argument, in a PR adding something similar in other package, that making this explicit was better. Added anyway. Validation was already there.

PR #5751, opened 2018-08-31
tests: setting "storage: preserve-size" just for amazon-linux system

sergiocazzolato:


sergiocazzolato:
It is possible to apply this change now because spread has been updated.

PR #5756, opened 2018-09-03
interfaces/modem-manager: allow access to more USB strings

alfonsosanchezbeato:
Allow access to manufacturer and product USB strings.


mvo5:
Should this go into 2.35 (if we do a 2.35.2)?

alfonsosanchezbeato:
@mvo5 yes, please include this in 2.35 if possible - it is a low risk change and it will be good to have asap.

PR #5757, opened 2018-09-03
many: use IdleManager to decide when to stop and go into socket activated mode

mvo5:
This is a variant of https://github.com/snapcore/snapd/pull/5583 that uses a new IdleManager to encapsulate the code that checks when snapd can go into socket activated mode. The nice thing about this approach is that it integrates with the Ensure loop which means that our checks are synchronized with e.g. snapstate that may install snaps or seed snaps. The downside is that its not a perfect fit for a manager, it does e.g. provide no task handling. Feedback on the approach welcome!


bboozzoo:
```
Checking spelling errors
daemon/daemon_test.go:832:24: "happend" is a misspelling of "happened"
daemon/daemon_test.go:862:24: "happend" is a misspelling of "happened"
```

cmd/snapd/main.go
```diff
@@ -49,6 +49,12 @@ func init() {
 func main() {
 	cmd.ExecInSnapdOrCoreSnap()
 	if err := run(); err != nil {
+		if err == daemon.ErrRestartSocket {
+			// Note that we don't prepend: "error: " here because
+			// ErrRestartSocket is not an error as such.
+			fmt.Fprintf(os.Stdout, "%v\n", err)
```
zyga: Nitpick: perhaps we could add a comment that explains why the unexpected combination of fprintf + stdout is used (for testing)

cmd/snapd/main.go
```diff
@@ -49,6 +49,12 @@ func init() {
...
+			os.Exit(42)
```
zyga: Is 42 relevant?

mvo5:
Closing as discussed during the standup the direction looks reasonable so I'm folding this back into the original PR #5583

cmd/snapd/main.go
mvo5: Very much so, I'm adding a comment.

PR #5759, opened 2018-09-04
ifacestate: implementation of defaultDeviceKey function for hotplug

stolowski:
Implementation of the `defaultDeviceKey` helper for hotplug. This function tries to compute a unique key from attributes reported by udev event (many of which may be missing). The goal is to have a key that's _unique for given interface type_. 
Implementations of interfaces that implement hotplug will have a way to override default key and provide own function to compute the key if they want to (e.g. if they are aware of attributes specific to given class of devices).


overlord/ifacestate/hotplug.go
```diff
@@ -20,9 +20,34 @@
...
+	key := make([]string, len(attrGroups))
+	for i, group := range attrGroups {
+		for _, attr := range group {
+			if val, ok := devinfo.Attribute(attr); ok && val != "" {
+				key[i] = val
+				break
+			}
+		}
+	}
+	return strings.Join(key, "/")
```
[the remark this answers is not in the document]
stolowski: Yes, it's okay, in such case the implementation of the interface *must* implement the method that computes a key, from the attributes that the interface knows about (this optional method of the interface has this signature: `func HotplugDeviceKey(di *hotplug.HotplugDeviceInfo) (string, error)`. If we cannot compute the key and the interface doesn't provide `HotplugDeviceKey` method, we cannot support given device.

overlord/ifacestate/hotplug_test.go
```diff
@@ -0,0 +1,81 @@
...
+	c.Assert(key, Equals, "pcislot//modelenc/")
+
+	di, err = hotplug.NewHotplugDeviceInfo(map[string]string{
+		"DEVPATH":   "a/path",
+		"ACTION":    "add",
+		"SUBSYSTEM": "foo",
+	})
+	c.Assert(err, IsNil)
+	key = ifacestate.DefaultDeviceKey(di)
+	c.Assert(key, Equals, "///")
```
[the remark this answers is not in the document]
stolowski: Sure, I intend to reject such key as it's essentialy an empty key. But I think I'll adjust `DefaultDeviceKey` implementation to simply return "" in such case.

overlord/ifacestate/hotplug.go
```diff
@@ -20,9 +20,34 @@
...
+// The final key is composed of 4 attributes (some of which may be empty), separated by "/".
+var attrGroups = [][]string{
+	{"ID_V4L_PRODUCT", "NAME", "ID_NET_NAME", "PCI_SLOT_NAME"},
+	{"ID_VENDOR_ID", "ID_VENDOR", "ID_WWN", "ID_WWN_WITH_EXTENSION", "ID_VENDOR_FROM_DATABASE", "ID_VENDOR_ENC", "ID_OUI_FROM_DATABASE"},
+	{"ID_MODEL_ID", "ID_MODEL_ENC"},
+	{"ID_SERIAL", "ID_SERIAL_SHORT", "ID_NET_NAME_MAC", "ID_REVISION"},
+}
+
+func defaultDeviceKey(devinfo *hotplug.HotplugDeviceInfo) string {
+	key := make([]string, len(attrGroups))
```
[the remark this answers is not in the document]
stolowski: The idea is to have empty slots for attributes that are missing, e.g. if an attribute in the middle is missing, I'd like a key "A/B//D", this creates a bigger space of unique keys.

overlord/ifacestate/hotplug_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	. "gopkg.in/check.v1"
+)
+
+type hotplugSuite struct{}
+
+var _ = Suite(&hotplugSuite{})
+
+func (s *hotplugSuite) TestDefaultDeviceKey(c *C) {
+	di, err := hotplug.NewHotplugDeviceInfo(map[string]string{
```
mvo5: (drive-by) I would love to see a test that uses a copy of real device data to see what a key looks like in practice.

stolowski: Added 3 samples made of real devices. Note, these examples needs to be taken a with a grain of salt... it's a little bit of an uncharted territory, there may be devices which need special treatment (dedicated implementation of HotplugDeviceKey method on the interface, or extra attributes added to this defaultDeviceKey helper).

overlord/ifacestate/hotplug.go
```diff
@@ -20,9 +20,34 @@
...
 import (
+	"strings"
+
 	"github.com/snapcore/snapd/interfaces/hotplug"
 )
 
+// List of attributes that determine the computation of default device key.
+// Attributes are grouped by similarity, the first non-empty attribute within the group goes into the key.
+// The final key is composed of 4 attributes (some of which may be empty), separated by "/".
+var attrGroups = [][]string{
```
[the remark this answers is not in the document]
stolowski: Added.

overlord/ifacestate/hotplug.go
```diff
@@ -20,9 +20,47 @@
...
+	{"ID_MODEL_ID", "ID_MODEL_ENC"},
+	// Identifier
+	{"ID_SERIAL", "ID_SERIAL_SHORT", "ID_NET_NAME_MAC", "ID_REVISION"},
+}
+
+// defaultDeviceKey computes device key from the attributes of
+// HotplugDeviceInfo. Empty string is returned if too few attributes are present
+// to compute a good key. Attributes used to compute device key are defined in
+// attrGroups list above.
+func defaultDeviceKey(devinfo *hotplug.HotplugDeviceInfo) string {
```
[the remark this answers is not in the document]
stolowski: I agree, versioning makes sense. Also, +1 for using null instead of "/" as a separator, it could potentially lead to collisions indeed. I also changed the implementation to use sha checksums, the only downside of this that I see is it makes keys completely opaque, which makes debugging slightly harder.

One general and obvious issue I see is with reverts to old versions while new keys are in use in the state - we will not match affected devices.

overlord/ifacestate/hotplug.go
```diff
@@ -28,6 +30,60 @@ import (
...
+// The resulting key returned by the function has the following format:
+// <version>:<checksum> where checksum is the sha256 checksum computed over
+// select attributes of the device.
+func defaultDeviceKey(devinfo *hotplug.HotplugDeviceInfo, keyVersion int) string {
+	var key bytes.Buffer
+	found := 0
+	for _, group := range attrGroups[keyVersion] {
+		for _, attr := range group {
+			if val, ok := devinfo.Attribute(attr); ok && val != "" {
+				if key.Len() > 0 {
```
niemeyer: I wouldn't worry about that.. this would work fine, and be exactly the same length:

key \x00 value \x00

overlord/ifacestate/hotplug.go
```diff
@@ -28,6 +30,60 @@ import (
...
+				key.Write([]byte(val))
+				found++
+				break
+			}
+		}
+	}
+	if found < 2 {
+		return ""
+	}
+	return fmt.Sprintf("%d:%x", keyVersion, sha256.Sum256(key.Bytes()))
```
niemeyer: No need for the separator.. just `"0" + hex` would be enough. We know the first byte is the version.

overlord/ifacestate/hotplug.go
```diff
@@ -28,6 +30,60 @@ import (
...
+// defaultDeviceKey computes device key from the attributes of
+// HotplugDeviceInfo. Empty string is returned if too few attributes are present
+// to compute a good key. Attributes used to compute device key are defined in
+// attrGroups list above and they depend on the keyVersion passed to the
+// function.
+// The resulting key returned by the function has the following format:
+// <version>:<checksum> where checksum is the sha256 checksum computed over
+// select attributes of the device.
+func defaultDeviceKey(devinfo *hotplug.HotplugDeviceInfo, keyVersion int) string {
+	var key bytes.Buffer
```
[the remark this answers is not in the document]
stolowski: Indeed, thanks, simplified.

overlord/ifacestate/hotplug.go
stolowski: Done.

overlord/ifacestate/hotplug.go
stolowski: Done.

overlord/ifacestate/hotplug.go
```diff
@@ -28,6 +29,58 @@ import (
...
+				key.Write([]byte{0})
+				found++
+				break
+			}
+		}
+	}
+	if found < 2 {
+		return ""
+	}
+	return fmt.Sprintf("%d%x", keyVersion, key.Sum(nil))
```
niemeyer: First needs to be %x as well, and keyVersion needs to be < 16.

overlord/ifacestate/hotplug.go
```diff
@@ -28,6 +29,58 @@ import (
...
+// The resulting key returned by the function has the following format:
+// <version>:<checksum> where checksum is the sha256 checksum computed over
+// select attributes of the device.
+func defaultDeviceKey(devinfo *hotplug.HotplugDeviceInfo, keyVersion int) string {
+	found := 0
+	key := sha256.New()
+	for _, group := range attrGroups[keyVersion] {
+		for _, attr := range group {
+			if val, ok := devinfo.Attribute(attr); ok && val != "" {
+				// attribute name and value are separated by null character
```
niemeyer: That's exactly what the code below it says, and more clearly at that.

PR #5760, opened 2018-09-04
cmd/snap-update-ns: don't trespass on host filesystem in /etc and other places

zyga:
This large branch changes the snap-update-ns command to check for
_trespassing_, that is the act of writing in unexpected places to the host
filesystem. Before this bug is fixed trespassing would manifest itself as empty
files, empty directories or symlinks in places that are shared with the host,
typically `/etc`.

To combat this issue we're employing writable mimics, that is, privately
mounted tpmfs-es that contain the orignal content of a given directory created
using a set of bind mounts.

To know when we should construct a writable mimic we now keep track of
Restrictions structure, as a re-factored extension to the existing Secure
structure. While Secure keeps track of global state (set of tmpfs-es that were
mounted, set of white-listed paths where restrictions don't apply, such as
$SNAP_DATA) the Restrictions type keeps track of _local_ state of a single write operation.

It is possible that as we progress through some path we are constructing, we
will discover a tmpfs that as mounted by snapd (and we have records to confirm
that). If we manage to succesfully create a new directory on such privately
shared tmpfs filesystem we know that subsequent writes will stay private to the
mount namespace we operate on. This allows us to LiftRestrictions and carry
on creating things as we please.

There are some more tests that we can write (I have some on my home machine
that I haven't pushed here) but I wanted to do a first round of public review
after a few rounds of review with mborzecki.

NOTE: We know of several shortcomings but those are not new and are somewhat
special cases of a special case alreday. I will be addressing them separately
on top of this PR.

### DIFF
--- a/cmd/snap-update-ns/main.go
+++ b/cmd/snap-update-ns/main.go
@@ -151,6 +151,10 @@ func computeAndSaveChanges(snapName string, sec *Secure) error {
 		return fmt.Errorf("cannot load current mount profile of snap %q: %s", snapName, err)
 	}
 	debugShowProfile(currentBefore, "current mount profile (before applying changes)")
+	// Synthesize mount changes that were applied before for the purpose of the tmpfs detector.
+	for _, entry := range currentBefore.Entries {
+		sec.AddChange(&Change{Action: Mount, Entry: entry})
+	}
 
 	currentAfter, err := applyProfile(snapName, currentBefore, desired, sec)
 	if err != nil {
--- a/cmd/snap-update-ns/main.go
+++ b/cmd/snap-update-ns/main.go
@@ -148,12 +148,12 @@ func applyFstab(snapName string, fromSnapConfine bool) error {
 	// TODO: Handle /home/*/snap/* when we do per-user mount namespaces and
 	// allow defining layout items that refer to SNAP_USER_DATA and
 	// SNAP_USER_COMMON.
-	sec := &Secure{}
-	sec.AddUnrestrictedPrefixes("/tmp", "/var/snap", "/snap/"+snapName)
-	return computeAndSaveChanges(snapName, sec)
+	as := &Assumptions{}
+	as.AddUnrestrictedPrefixes("/tmp", "/var/snap", "/snap/"+snapName)
+	return computeAndSaveChanges(snapName, as)
 }
 
-func computeAndSaveChanges(snapName string, sec *Secure) error {
+func computeAndSaveChanges(snapName string, as *Assumptions) error {
 	// Read the desired and current mount profiles. Note that missing files
 	// count as empty profiles so that we can gracefully handle a mount
 	// interface connection/disconnection.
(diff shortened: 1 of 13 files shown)


cmd/snap-update-ns/secure.go
```diff
@@ -0,0 +1,187 @@
...
+func (as *Assumptions) RestrictionsFor(desiredPath string) *Restrictions {
+	if as.isRestricted(desiredPath) {
+		return &Restrictions{assumptions: as, desiredPath: desiredPath, restricted: true}
+	}
+	return nil
+}
+
+// TrespassingError is an error when filesystem operation would affect the host.
+type TrespassingError struct {
+	ViolatedPath string
```
bboozzoo: Maybe we can drop `Path` suffix as it's kind assumed already. BTW. `Violated` looks a bit awkward, wondering if there could be a nicer alternative.

cmd/snap-update-ns/utils.go
```diff
@@ -218,22 +196,27 @@ func (sec *Secure) MkPrefix(base string, perm os.FileMode, uid sys.UserID, gid s
 	}
 	iter.Next() // Advance iterator to '/'
 
-	const openFlags = syscall.O_NOFOLLOW | syscall.O_CLOEXEC | syscall.O_DIRECTORY
 	// Open the root directory and start there.
+	//
+	// NOTE: We don't have to check for possible trespassing on / here because
+	// we are going to check for it in sec.MkDir call below (which verifies
+	// that / is not violated)
```
bboozzoo: Reads a bit awkward. Maybe `.. which verifies that trespassing restrictions are not violated`?

cmd/snap-update-ns/utils.go
```diff
@@ -273,27 +258,39 @@ func (sec *Secure) MkDir(dirFd int, dirName string, name string, perm os.FileMod
...
+		// As soon as we find a place that is safe to write we can switch off
+		// the restricted mode (and thus any subsequent checks). This is
+		// because we only allow "writing" to read-only filesystems where
+		// writes fail with EROFS or to a tmpfs that snapd has privately
+		// mounted inside the per-snap mount namespace. As soon as we start
+		// walking over such tmpfs any subsequent children are either read-
+		// only bind mounts from $SNAP, other tmpfs'es  (e.g. one explicitly
+		// constructed for a layout) or writable places that are bind-mounted
+		// from $SNAP_DATA or similar.
+		rs.LiftRestrictions()
```
bboozzoo: Wondering if the caller should do it and instead `MkDir()` should return if the directory was made or not. Modifying this inside in non-obvious way makes me feel a bit uncomfortable.

cmd/snap-update-ns/main.go
```diff
@@ -128,13 +128,32 @@ func applyFstab(snapName string, fromSnapConfine bool) error {
...
+	// /snap/$SNAP_NAME needs to be there as the code that creates such mount
+	// points must traverse writable host filesystem that contains /snap/*/ and
+	// normally such access is off-limits. This approach allows /snap/foo
+	// without allowing /snap/bin, for example.
+
+	// TODO: Handle /home/*/snap/* when we do per-user mount namespaces and
+	// allow defining layout items that refer to SNAP_USER_DATA and
+	// SNAP_USER_COMMON.
+	as := &Assumptions{}
+	as.AddUnrestrictedPrefixes("/tmp", "/var/snap", "/snap/"+snapName)
```
bboozzoo: Can you add a TODO for parallel install? Eg:
```
TODO parallel-install: add instance specific path
```

cmd/snap-update-ns/secure.go
zyga: How about `Intended` and `Blocked`?

cmd/snap-update-ns/utils.go
zyga: I explicitly didn't want to do this. In an earlier iteration of this code I was passing and returning the restricted flag around and it was simply more confusing and convoluted to handle. Specifically this is an internal implementation detail of the restriction system and in fact, mkdir is the only place where the restriction can be safely lifted so it felt like pushing the responsibility to the caller.

zyga:
I'll mark this as blocked, until the prerequisites land.

zyga:
I applied comment changes to the next branch.

PR #5763, opened 2018-09-04
store: refactor tests so that they work as store_test package 

mvo5:
The tests in the store code are one of the few places left that
do not use the "$pkg_test" idiom. This caused me some confusion
so I did a (very mechnical) translation of the tests. This
probably should have a followup that refactors some bits and
moves bits into smaller and more targeted suites. But to make
reviews easier this tries to minimize the changes as much as
possible.

If this looks reasonable there is more to do still: details_v2_tests.go,
userinfo_test.go.


store/export_test.go
```diff
@@ -20,13 +20,43 @@
...
 
 var (
 	HardLinkCount = hardLinkCount
+	ApiURL        = apiURL
+	Download      = download
+
+	UseDeltas  = useDeltas
+	ApplyDelta = applyDelta
+
+	GetCurrentSnap    = currentSnap
```
zyga: Is there a clash with another symbol (CurrentSnap)?

store/store_test.go
```diff
@@ -471,7 +473,8 @@ func (s *storeTestSuite) TestDownloadRangeRequest(c *C) {
 func (s *storeTestSuite) TestResumeOfCompleted(c *C) {
 	expectedContentStr := "nothing downloaded"
 
-	download = nil
+	// FIXME: hu?
```
zyga: What?

store/export_test.go
mvo5: Yes, its rather unfortunate. I can do a followup for this. The "currentSnap" is a function and CurrentSnap is a type IIRC.

store/store_test.go
mvo5: Oh, sorry. Yeah, this one I commented out because I was not sure why it was done like this originally. The comment obviously sucks, I think I need to do a followup for this one.

store/export_test.go
```diff
@@ -43,7 +43,10 @@ var (
 	UseDeltas  = useDeltas
 	ApplyDelta = applyDelta
 
-	GetCurrentSnap    = currentSnap
+	// FIXME: there is a name clash, store has both "currentSnap"
+	// and "CurrentSnap" which do different things.
```
zyga: Thanks!

PR #5769, opened 2018-09-05
partition: remove unused runCommand

mvo5:
The "runCommand" function in the partition package is only used
in tests and trivially replaced with exec.Command(). This PR does
that and drops this old (and crufty) code.


zyga:
+1

PR #5768, opened 2018-09-05
selftest: actually run the kernel version selftest

mvo5:
The deadcode checker found that the kernel version test was not
used. This PR fixes this.

PR #5771, opened 2018-09-05
selftest: add test to ensure selftest.checks is up-to-date 

mvo5:
t was found in #5768 that a selftest was added but not run. To prevent issues like
this in the future a test is added that intospects the functions
of the selftest package and warns if any check* function is added
without also adding it to the checks that need to be run.

Based on https://github.com/snapcore/snapd/pull/5768


selftest/selftest_test.go
```diff
@@ -67,3 +75,45 @@ func (s *selftestSuite) TestRunNotHappy(c *C) {
...
+
+func (s *selftestSuite) TestUnexportedChecks(c *C) {
+
+	// collect what funcs we run in selftest.Check
+	var runCheckers []string
+	v := reflect.ValueOf(selftest.Checks)
+	for i := 0; i < v.Len(); i++ {
+		v := v.Index(i)
+		fname := runtime.FuncForPC(v.Pointer()).Name()
+		l := strings.Split(fname, ".")
```
zyga: Are we getting the last element of a sequence of dotted elements?

selftest/selftest_test.go
```diff
@@ -67,3 +75,45 @@ func (s *selftestSuite) TestRunNotHappy(c *C) {
...
+	var checkers []string
+	for _, fn := range goFiles {
+		f, err := parser.ParseFile(fset, fn, nil, 0)
+		c.Assert(err, IsNil)
+		ast.Inspect(f, func(n ast.Node) bool {
+			switch x := n.(type) {
+			case *ast.File:
+				return true
+			case *ast.FuncDecl:
+				name := x.Name.Name
```
zyga: This is okay but this is funny to look at :)

zyga:
@mvo5 I think the new check actually prevents running snapd

```
Job for snapd.service failed. See 'systemctl status snapd.service' and 'journalctl -xn' for details.
dpkg: error processing package snapd (--install):
 subprocess installed post-installation script returned error exit status 1
dpkg: snapd-xdg-open: dependency problems, but configuring anyway as you requested:
 snapd-xdg-open depends on snapd (= 1337.2.35.1~14.04); however:
  Package snapd is not configured yet.

Setting up snapd-xdg-open (1337.2.35.1~14.04) ...
dpkg: snap-confine: dependency problems, but configuring anyway as you requested:
 snap-confine depends on snapd (= 1337.2.35.1~14.04); however:
  Package snapd is not configured yet.

Setting up snap-confine (1337.2.35.1~14.04) ...
Processing triggers for man-db (2.6.7.1-1ubuntu1) ...
Errors were encountered while processing:
 snapd
```

mvo5:
This is blocked until the problem with trusty is fixed. Now that the selftest checks for the kernel version snapd will fail to start on trusty and this makes dpkg fail. So we really need something like "degraded boot" before this can land.

selftest/selftest_test.go
bboozzoo: How about:
```
pos := strings.LastIndexByte(fname, '.')
l := fname[pos+1:]
```

mvo5: Yes, I updated it to the suggestion from @bboozzoo

selftest/selftest_test.go
mvo5: It is :) Upstream code!

PR #5773, opened 2018-09-05
tests: avoid removing core snap on reset

sergiocazzolato:
This change is done to avoid removing the core snap, which is removed on
the reset script on ubuntu-core-18* systems.

This is making slow the the tests which are installing snaps depending
on the core snap, such as test-snapd-tools.

An example, the test install-closed-channel test is taking 12 seconds on
ubuntu-core-16 and 90 seconds on ubuntu-core-18 running in a local vm.


mvo5:
Nice one Sergio

PR #5775, opened 2018-09-05
tests: also run unit/gccgo in 18.04

chipaca:
Without this we were catching some weird gccgoisms only during
pkgbuild tests, which is very late.


chipaca:
FWIW I expect this to fail until we fix'em :-)

PR #5777, opened 2018-09-05
wrappers/services.go: don't start disabled services

anonymouse64:
This prevents `snap install` from starting up services that were disabled using `snapctl` during the install hook.

See forum post https://forum.snapcraft.io/t/using-snapctl-in-install-hook-to-disable-services/7214 for more details.

Additionally, I have a simple snap which can be used as a reproducer available here: https://github.com/anonymouse64/service-test


bboozzoo:
Some failures in unit tests:
```
FAIL: managers_test.go:664: mgrsSuite.TestHappyRemoteInstallAndUpgradeSvc

managers_test.go:703:
    c.Assert(chg.Status(), Equals, state.DoneStatus, Commentf("install-snap change failed with: %v", chg.Err()))
... obtained state.Status = 9 ("Error")
... expected state.Status = 4 ("Done")
... install-snap change failed with: cannot perform the following tasks:
- Start snap "foo" (42) services (cannot get service status: missing Id, Type, UnitFileState in â€˜systemctl showâ€™ output)

OOPS: 54 passed, 1 FAILED
--- FAIL: TestOverlord (9.17s)
FAIL
coverage: 90.1% of statements
exit status 1
FAIL	github.com/snapcore/snapd/overlord	9.357s
```

tests/main/snapctl-disable-services-install/task.yaml
```diff
@@ -0,0 +1,20 @@
+summary: Check that `snapctl stop --disable` actually stops services on install
+
+prepare: |
+    snap pack $TESTSLIB/snaps/snapctl-disable-services-install
```
bboozzoo: `"$TESTSLIB"`

tests/main/snapctl-disable-services-install/task.yaml
```diff
@@ -0,0 +1,20 @@
...
+execute: |
+    echo "Verify that the install hook finishes while installing"
+    if ! snap install --dangerous snapctl-disable-services-install_1.0_all.snap; then
+        echo "Expected snap install to be successful"
+        exit 1
+    fi
+
+    for service in "simple" "forking" do;
+        echo "Verify that the $service service isn't running"
+        snap services | MATCH "snapctl-disable-services-install\.$service\s\+disabled\s\+inactive"
```
bboozzoo: `"snapctl-disable-services-install\\.$service\\s\\+disabled\\s\\+inactive" `

tests/main/snapctl-disable-services-install/task.yaml
```diff
@@ -0,0 +1,20 @@
...
+    rm -f snapctl-disable-services-install_1.0_all.snap
+
+execute: |
+    echo "Verify that the install hook finishes while installing"
+    if ! snap install --dangerous snapctl-disable-services-install_1.0_all.snap; then
+        echo "Expected snap install to be successful"
+        exit 1
+    fi
+
+    for service in "simple" "forking" do;
```
bboozzoo: `for service in "simple" "forking" ; do`

wrappers/services.go
```diff
@@ -130,7 +130,20 @@ func StartServices(apps []*snap.AppInfo, inter interacter) (err error) {
 		}(app)
 
 		if len(app.Sockets) == 0 && app.Timer == nil {
-			services = append(services, app.ServiceName())
+			// check if the service is disabled, if so don't start it up
+			// this could happen for example if the service was disabled in
+			// the install hook by snapctl
+			statuses, err := sysd.Status(app.ServiceName())
```
bboozzoo: `Status()` is pretty invovled, just a thought but maybe a thin wrapper around `systemctl is-enabled`, eg `IsEnabled(serviceName string) (bool, error)`, would be useful.

Eg:
```
// IsEnabled checkes whether the given service is enabled
func (*systemd) IsEnabled(serviceName string) (bool, error) {
	_, err := systemctlCmd("--root", s.rootDir, "is-enabled", serviceName)
	if err == nil {
		return true, nil
	}
	if sysdErr, ok := err.(*Error); ok && sysdErr.msg == "disabled" {
		return false, nil
	}
	return false, err
}
```

tests/lib/snaps/snapctl-disable-services-install-hook/meta/hooks/install
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -e
+
+for service in simple forking do;
```
bboozzoo: `for service in simple forking ; do`

tests/lib/snaps/snapctl-disable-services-install-hook/meta/hooks/install
```diff
@@ -0,0 +1,9 @@
...
+    echo "Disabling $service service from install hook"
+    if [ "$(snapctl stop --disable snapctl-disable-services-install-hook.$service 2>&1)" != "" ]; then
```
bboozzoo: It's already running with `set -e`, so I believe the whole script could be this:
```
#!/bin/bash -e

for service in simple forking; do
    echo "Disabling $service service from install hook"
    snapctl stop --disable "snapctl-disable-services-install-hook.$service"
done
```

tests/lib/snaps/snapctl-disable-services-install-hook/meta/hooks/install
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -e
```
bboozzoo: Can you run shellcheck on this file?

tests/main/snapctl-disable-services-install/task.yaml
```diff
@@ -0,0 +1,20 @@
+summary: Check that `snapctl stop --disable` actually stops services on install
```
bboozzoo: While at the top of the repo, can you run `./spread-shellcheck tests/main/snapctl-disable-services-install/task.yaml`?

tests/main/snapctl-disable-services-install/task.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+prepare: |
+    snap pack $TESTSLIB/snaps/snapctl-disable-services-install
+
+restore: |
+    rm -f snapctl-disable-services-install_1.0_all.snap
+
+execute: |
+    echo "Verify that the install hook finishes while installing"
+    if ! snap install --dangerous snapctl-disable-services-install_1.0_all.snap; then
```
bboozzoo: Feel free to ignore it. I don't know about others, but I prefer the shell code to be terse, the `if` block can be replaced with this:
```
echo "Verify that the install hook finishes while installing"
snap install --dangerous snapctl-disable-services-install_1.0_all.snap
...
```
this is running with `set -e`, and the command is expected to succeed (otherwise we'd use `! snap install ...`).

tests/lib/snaps/snapctl-disable-services-install-hook/meta/snap.yaml
```diff
@@ -0,0 +1,9 @@
+name: snapctl-disable-services-install-hook
+version: 1.0
+apps:
+    simple:
+        daemon: simple
+        command: "sleep 60"
+    forking:
+        daemon: forking
+        command: sh -c "sleep 60 &"
```
zyga: Does job control work without a pty?

wrappers/services.go
zyga: I agree with Maciej here. Plus thanks to the spread test we will have nice coverage of various systemd versions across the history

zyga:
Some tests are failing, please run unit tests:
```
FAIL: managers_test.go:664: mgrsSuite.TestHappyRemoteInstallAndUpgradeSvc

managers_test.go:703:
    c.Assert(chg.Status(), Equals, state.DoneStatus, Commentf("install-snap change failed with: %v", chg.Err()))
... obtained state.Status = 9 ("Error")
... expected state.Status = 4 ("Done")
... install-snap change failed with: cannot perform the following tasks:
- Start snap "foo" (42) services (cannot get service status: missing Id, Type, UnitFileState in âsystemctl showâ output)

OOPS: 54 passed, 1 FAILED
--- FAIL: TestOverlord (9.17s)
```

tests/main/snapctl-disable-services-install/task.yaml
anonymouse64: Fixed.

tests/main/snapctl-disable-services-install/task.yaml
anonymouse64: Fixed.

tests/main/snapctl-disable-services-install/task.yaml
anonymouse64: Fixed.

tests/lib/snaps/snapctl-disable-services-install-hook/meta/hooks/install
anonymouse64: Yep, typo. Thanks

tests/lib/snaps/snapctl-disable-services-install-hook/meta/hooks/install
anonymouse64: Fixed

tests/lib/snaps/snapctl-disable-services-install-hook/meta/hooks/install
anonymouse64: Yep, just did that.

tests/main/snapctl-disable-services-install/task.yaml
anonymouse64: Done, fixed the issues with the for loop.

tests/main/snapctl-disable-services-install/task.yaml
anonymouse64: Agreed, it's cleaner that way.

wrappers/services.go
anonymouse64: I implemented this method basically as provided with a few changes. I'm also checking the exit code to be 1, which seems to be standard behavior for `systemctl is-enabled` to be as strict as possible with ensuring that the service is actually disabled.

tests/lib/snaps/snapctl-disable-services-install-hook/meta/snap.yaml
anonymouse64: I'm not sure, I originally was just testing with `yes`, but it occurred to me that if the test fails for some reason that `yes` process would just eat up CPU until it was killed which I'm not sure when that would happen. There may be a better service to try here that doesn't use up CPU but stays alive for a while

anonymouse64:
I will run the unit tests when able, I can't right now due to the `gopkg.in` outage, but when that's restored and I can update the deps locally I will run the unit tests and fix what I can.

anonymouse64:
After gopkg.in came back online, I was able to update the unit tests so they pass.
I ran the unit tests with `./run-checks --unit`, let me know if there is another way to test this locally.

wrappers/services.go
```diff
@@ -130,7 +130,17 @@ func StartServices(apps []*snap.AppInfo, inter interacter) (err error) {
...
 		if len(app.Sockets) == 0 && app.Timer == nil {
-			services = append(services, app.ServiceName())
+			// check if the service is disabled, if so don't start it up
+			// this could happen for example if the service was disabled in
+			// the install hook by snapctl
+			isEnabled, err := sysd.IsEnabled(app.ServiceName())
+			if err != nil {
+				return err
+			}
+
```
bboozzoo: Can you remove this empty line?

wrappers/services_test.go
```diff
@@ -223,7 +223,10 @@ func (s *servicesTestSuite) TestStartServices(c *C) {
 	err := wrappers.StartServices(info.Services(), nil)
 	c.Assert(err, IsNil)
 
-	c.Assert(s.sysdLog, DeepEquals, [][]string{{"start", filepath.Base(svcFile)}})
+	c.Assert(s.sysdLog, DeepEquals, [][]string{
+		{"--root", s.tempdir, "is-enabled", filepath.Base(svcFile)},
+		{"start", filepath.Base(svcFile)},
```
bboozzoo: It'd be nice to add a test for a case when the service has been disabled by the hook.

I've uploaded a small diff with all the changes here: https://paste.ubuntu.com/p/YzymnqMgMw/

tests/main/snapctl-disable-services-install/task.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+prepare: |
+    snap pack "$TESTSLIB"/snaps/snapctl-disable-services-install
+
+restore: |
+    rm -f snapctl-disable-services-install_1.0_all.snap
+
+execute: |
+    echo "Verify that the install hook finishes while installing"
+    snap install --dangerous snapctl-disable-services-install_1.0_all.snap
```
bboozzoo: You can drop `prepare` and `restore` sections and do:
```
# shellcheck source=tests/lib/snaps.sh
. "$TESTSLIB"/snaps.sh
install_local snapctl-disable-services-install
```

wrappers/services_test.go
bboozzoo: There is an unrelated problem with unit tests currently and service ordering, I'll fix that in master.

zyga:
Are we having missing files here?
```
error: cannot pack "/home/gopath/src/github.com/snapcore/snapd/tests/lib/snaps/snapctl-disable-services-install": open /home/gopath/src/github.com/snapcore/snapd/tests/lib/snaps/snapctl-disable-services-install/meta/snap.yaml: no such file or directory
```

tests/main/snapctl-disable-services-install/task.yaml
bboozzoo: Oh and the snap is named `snapctl-disable-services-install-hook`

wrappers/services_test.go
anonymouse64: Do you need me to add the `NewError` function you have in that diff?

wrappers/services.go
anonymouse64: Done.

tests/main/snapctl-disable-services-install/task.yaml
anonymouse64: Ah that explains @zyga's comment about the missing file... Thanks!

zyga:
This is failing on:
```
++ find /home/gopath/src/github.com/snapcore/snapd/tests/lib/snaps/snapctl-disable-services-install -name '*.snap'
find: '/home/gopath/src/github.com/snapcore/snapd/tests/lib/snaps/snapctl-disable-services-install': No such file or directory
+ SNAP_FILE=
+ snap install --dangerous ''
error: cannot install snap with empty name
-----
```

tests/main/snapctl-disable-services-install-hook/task.yaml
```diff
@@ -0,0 +1,12 @@
+summary: Check that `snapctl stop --disable` actually stops services on install
+
+execute: |
+    # shellcheck source=tests/lib/snaps.sh
+    . "$TESTSLIB"/snaps.sh
+    install_local snapctl-disable-services-install
```
zyga: Did you mean to say `...-hook` here? Specifically `snapctl-disable-services-install-hook`?

Btw, the general naming we are trying to go after is `test-snapd-foo`. This is how we name internal snaps that only exists for testing.

anonymouse64: Yes this is another typo... Also I wanted to follow the naming convention, but my initial attempt at:
`test-snapctl-disable-services-install-hook` is over the 40 character limit for snap names, so I removed the `test-`. Is there a more concise name I'm not seeing here?

wrappers/services_test.go
anonymouse64: Got it, this is done.

anonymouse64:
I seem to be stuck on getting the MATCH to work, isn't `MATCH` just the same as `grep -qe`? The CI says this fails:
```
+ MATCH 'test-snapd-svcs-disable-install-hook\.simple\s\+disabled\s\+inactive'
+ snap services
Service                                       Startup   Current
test-snapd-svcs-disable-install-hook.forking  disabled  inactive
test-snapd-svcs-disable-install-hook.simple   disabled  inactive
```

But when I run locally it works fine:
```
$ set -x
++ history -a
$ snap services | grep -qe "test-snapd-svcs-disable-install-hook\\.simple\\s\+disabled\\s\+inactive"
+ snap services
+ grep --color=auto -qe 'test-snapd-svcs-disable-install-hook\.simple\s\+disabled\s\+inactive'
++ history -a
$ echo $?
+ echo 0
0
++ history -a
```

tests/main/svcs-disable-install-hook/task.yaml
```diff
@@ -0,0 +1,12 @@
+summary: Check that `snapctl stop --disable` actually stops services on install
+
+execute: |
+    # shellcheck source=tests/lib/snaps.sh
+    . "$TESTSLIB"/snaps.sh
+    install_local test-snapd-svcs-disable-install-hook
+
+    for service in simple forking; do
+        echo "Verify that the $service service isn't running"
+        snap services | MATCH "test-snapd-svcs-disable-install-hook\\.$service\\s\+disabled\\s\+inactive"
```
bboozzoo: This works `MATCH "test-snapd-svcs-disable-install-hook\\.$service\\s+disabled\\s+inactive"`

tests/lib/snaps/test-snapd-svcs-disable-install-hook/bin/forking.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -e
+
+sleep 60 &
```
bboozzoo: missing newline

tests/lib/snaps/test-snapd-svcs-disable-install-hook/bin/simple.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -e
+
+sleep 60
```
bboozzoo: here as well

tests/main/svcs-disable-install-hook/task.yaml
anonymouse64: Indeed it does! This test is now passing. I think I was confused because the escaping here is really a bash string inside a yaml string...

tests/lib/snaps/test-snapd-svcs-disable-install-hook/bin/forking.sh
anonymouse64: Done

tests/lib/snaps/test-snapd-svcs-disable-install-hook/bin/simple.sh
anonymouse64: Done

anonymouse64:
The most recent test failure is no longer related to this test, and is instead from lxd:
```
2018-09-13 12:41:06 Failed tasks: 1
    - google:ubuntu-18.04-64:tests/main/lxd
error: unsuccessful run
```

tests/main/svcs-disable-install-hook/task.yaml
```diff
@@ -0,0 +1,12 @@
...
+execute: |
+    # shellcheck source=tests/lib/snaps.sh
+    . "$TESTSLIB"/snaps.sh
+    install_local test-snapd-svcs-disable-install-hook
+
+    for service in simple forking; do
+        echo "Verify that the $service service isn't running"
+        snap services | MATCH "test-snapd-svcs-disable-install-hook\\.$service\\s+disabled\\s+inactive"
+    done
+
```
bboozzoo: Stray tab?

anonymouse64: It appears that way, do you want me to delete it?

wrappers/services.go
```diff
@@ -130,7 +130,16 @@ func StartServices(apps []*snap.AppInfo, inter interacter) (err error) {
 		}(app)
 
 		if len(app.Sockets) == 0 && app.Timer == nil {
-			services = append(services, app.ServiceName())
+			// check if the service is disabled, if so don't start it up
+			// this could happen for example if the service was disabled in
+			// the install hook by snapctl
+			isEnabled, err := sysd.IsEnabled(app.ServiceName())
```
chipaca: why is this check here, instead of at the top of the loop, right after the check for `IsService`?

If there is a reason for it to be here, we also need to not start the sockets and timers of the app.

anonymouse64: Well, effectively since the above function is deferred, that gets run only after we exit, and so this check is effectively run right after the `IsService`. 

I wasn't quite sure about disabling the sockets and timers case as I'm not familiar with how systemd timers work with snap services.
* If a snap service has a timer, does it have both a systemd unit for the timer and the service? Should we disable both of those if the service was disabled by `snapctl`? Probably yes
* Similar question for the socket, are there separate systemd units for the socket and the service? Should both of those be disabled if the snap service was disabled by `snapctl`?
* Also what does it mean to just run `snapctl stop SERVICE_WITH_TIMER` without the `--disable` flag during the install hook? Do we simply not run/start the timer on install? 
* Same question for sockets, what should happen when a `snapctl stop SERVICE_WITH_SOCKET` is run during install hook?

anonymouse64:
@chipaca please see my response above...

@zyga @bboozzoo Should we hold this PR back on handling sockets/timers? With some more work I can handle those cases in a followup PR, but we have snaps that we need this feature to work with and this PR at the least doesn't break anything so I think it's safe to just handle the sockets/timers case later.

zyga:
I think it makes sense to handle timers and sockets since they are interconnected to services. At the same time I don't want to block this. I think it's @niemeyer call to decide.

systemd/systemd.go
```diff
@@ -352,6 +369,11 @@ type Error struct {
 	exitCode int
 }
 
+// NewError creates a systemd.Error, exported for mocks during testing
+func NewError(exitCode int, cmd []string, msg []byte) *Error {
```
niemeyer: If it's just for tests this should be in export_test.go. There's no reason right now for external logic to create these errors.

wrappers/services.go
niemeyer: The right mindset here is to think about the high-level concept we're dealing with. Most of these questions refer to the implementation instead. That is, for example, should a snap application that has a timer run when we call _snapctl stop_ on that application? Yes, I think it should.. the request was to stop what is running right now, not to disable it. Should it run when we disable it? That doesn't sound right.

wrappers/services_test.go
niemeyer: If you actually mock the systemctl command (there are examples of how to do this elsewhere), you can make it return the exact output necessary and the exact error code.  That's better than faking an internal error which might not be generated properly depending on bugs elsewhere.

systemd/systemd.go
anonymouse64: I guess I don't follow, we can't put this function inside the `wrappers/export_test.go` because we need to access private members of the systemd.Error struct and the `wrappers/export_test.go` package is not `systemd,` it's `wrappers`.The only place during testing we could put this function would be in `systemd/export_test.go`, which doesn't allow us to access this error while running the `wrappers` tests, which is the actual place we need this.

@bboozzoo thoughts on this problem? I think putting it in the systemd package as you suggested initially is the simplest solution here.

wrappers/services_test.go
anonymouse64: @niemeyer Actually, even if you use `MockSystemctl`, you still need to access the private structs of the `Error` struct when returning an error so you still don't avoid that problem. The only way way I see to avoid this is to change `systemctlCmd` not return an `*systemd.Error` and instead return something more generic that can be override outside the `systemd` package. 
The reason we have this problem is because we're testing things coming from the `systemd` package inside the `wrappers` package.

anonymouse64:
I have handled the sockets/timers case here, as well as fixed `snapctl stop --disable SERVICE_WITH_SOCKET` and `snapctl stop --disable SERVICE_WITH_TIMER`. I still need to add tests for these cases however.

systemd/systemd.go
bboozzoo: @niemeyer the problem here is that `systemctl is-enabled <foo>`, exits with error and prints `disabled` when the service is indeed disabled. To mock that nicely we need to generate `systemd.Error{exitCode:1, msg: []byte("disabled"}` in a different package.

I suppose we could rename this to `MockError()` to make it more explicit that this is only for the tests and restore the original code here https://github.com/snapcore/snapd/pull/5777/files#diff-42c07ec7ffa970857b4e0db2614e482bR56 Would that work for you?

wrappers/services_test.go
bboozzoo: @anonymouse64 this is the alternative suggested by @niemeyer:

```
diff --git a/systemd/systemd.go b/systemd/systemd.go
index 05e5f36d7..25f43748c 100644
--- a/systemd/systemd.go
+++ b/systemd/systemd.go
@@ -53,7 +53,7 @@ var systemctlCmd = func(args ...string) ([]byte, error) {
        bs, err := exec.Command("systemctl", args...).CombinedOutput()
        if err != nil {
                exitCode, _ := osutil.ExitCode(err)
-               return nil, NewError(exitCode, args, bs)
+               return nil, &Error{exitCode: exitCode, cmd: args, msg: bs}
        }
 
        return bs, nil
@@ -369,11 +369,6 @@ type Error struct {
        exitCode int
 }
 
-// NewError creates a systemd.Error, exported for mocks during testing
-func NewError(exitCode int, cmd []string, msg []byte) *Error {
-       return &Error{exitCode: exitCode, cmd: cmd, msg: msg}
-}
-
 func (e *Error) Error() string {
        return fmt.Sprintf("%v failed with exit status %d: %s", e.cmd, e.exitCode, e.msg)
 }
diff --git a/wrappers/services_test.go b/wrappers/services_test.go
index 5fdb1b8f3..88a0785ae 100644
--- a/wrappers/services_test.go
+++ b/wrappers/services_test.go
@@ -233,24 +233,34 @@ func (s *servicesTestSuite) TestNoStartDisabledServices(c *C) {
        info := snaptest.MockSnap(c, packageHello, &snap.SideInfo{Revision: snap.R(12)})
        svcFile := filepath.Join(s.tempdir, "/etc/systemd/system/snap.hello-snap.svc1.service")
 
-       r := systemd.MockSystemctl(func(cmd ...string) ([]byte, error) {
-               s.sysdLog = append(s.sysdLog, cmd)
-               sdcmd := cmd[0]
-               if len(cmd) >= 2 {
-                       sdcmd = cmd[len(cmd)-2]
-               }
-               if sdcmd == "is-enabled" {
-                       return nil, systemd.NewError(1, cmd, []byte("disabled\n"))
-               }
-               panic("unexpected systemctl command " + sdcmd)
-       })
-       defer r()
+       s.restorer()
+       r := testutil.MockCommand(c, "systemctl", `#!/bin/sh
+if [ "$1" = "--root" ]; then
+    shift 2
+fi
+
+case "$1" in
+    is-enabled)
+        if [ "$2" = "snap.hello-snap.svc1.service" ]; then
+            echo "disabled"
+            exit 1
+        else
+            echo "unexpected call $*"
+            exit 2
+        fi
+        ;;
+    *)
+        echo "unexpected call $*"
+        exit 2
+esac
+`)
+       defer r.Restore()
 
        err := wrappers.StartServices(info.Services(), nil)
        c.Assert(err, IsNil)
 
-       c.Assert(s.sysdLog, DeepEquals, [][]string{
-               {"--root", s.tempdir, "is-enabled", filepath.Base(svcFile)},
+       c.Assert(r.Calls(), DeepEquals, [][]string{
+               {"systemctl", "--root", s.tempdir, "is-enabled", filepath.Base(svcFile)},
        })
 }
 
```

Can you apply this and drop `NewError()` ?

anonymouse64: I see thanks for explaining. I'll add this shortly.

tests/main/svcs-disable-install-hook/task.yaml
bboozzoo: Please do.

anonymouse64:
As per discussion with @bboozzoo I am closing this for now and have started a topic discussion regarding what to do about sockets/timers here: https://forum.snapcraft.io/t/how-to-manage-services-with-sockets-timers/7904 because it's non-trivial. 

I am going to re-open this PR with just the bits that we need for disabling a service without any sockets/timers from the install hook because that case is currently a bug and easily understood what the proper behavior should be.

PR #5782, opened 2018-09-06
ifacestate: helpers for generating slot names for hotplug

stolowski:
Three helpers for generating slot names for hotplug - whenever we detect a new hotplug device, these functions will aid in generating a unique and clean slot name for it. I'm proposing all three in a single PR as they will be used in tandem:
- _suggestedSlotName_ inspects device info for known attributes and returns the value of the most suitable one for a slot name. The name is cleaned up to obey our slot naming rules. If no suitable attribute is found, it returns a fallback name (which in real use will just be the interface name). Note that the returned name will be an input for _ensureUniqueName_ described below.
- _cleanupSlotName_ implements actual cleanup of slot names.
- _ensureUniqueName_ appends a digit suffix (or increases existing one) to make a name unique according to passed predicate. In actual hotplug code it will use a lookup in the repo to have a name that's unique for system slots.
Note, hotplug interfaces will have a say on the names - they will be able to make their own name and completely ignore auto-generated ones.


overlord/ifacestate/hotplug.go
```diff
@@ -30,3 +34,115 @@ func (m *InterfaceManager) HotplugDeviceAdded(devinfo *hotplug.HotplugDeviceInfo
...
+var nameAttrs = []string{"NAME", "ID_MODEL_FROM_DATABASE", "ID_MODEL"}
+
+// suggestedSlotName returns the shortest name derived from attributes defined by nameAttrs, or
+// the fallbackName if there is no known attribute to derive name from. The name created from
+// attributes is cleaned up by cleanupSlotName function.
+// The fallbackName is typically the name of the interface.
+func suggestedSlotName(devinfo *hotplug.HotplugDeviceInfo, fallbackName string) string {
+	var candidates []string
+	for _, attr := range nameAttrs {
+		val, ok := devinfo.Attribute(attr)
```
zyga: While scanning this I was confused by the use of `val` and `name` below. I think using a single variable name would be easier to follow, at least for me.

overlord/ifacestate/hotplug.go
```diff
@@ -30,3 +34,115 @@ func (m *InterfaceManager) HotplugDeviceAdded(devinfo *hotplug.HotplugDeviceInfo
...
+			name := cleanupSlotName(val)
+			if name != "" {
+				candidates = append(candidates, name)
+			}
+		}
+	}
+	if len(candidates) == 0 {
+		return fallbackName
+	}
+	shortestName := candidates[0]
```
zyga: Neat

overlord/ifacestate/hotplug.go
```diff
@@ -30,3 +34,115 @@ func (m *InterfaceManager) HotplugDeviceAdded(devinfo *hotplug.HotplugDeviceInfo
...
+		}
+	}
+	// make sure the name doesn't end with a dash
+	if len(out) > 0 && out[len(out)-1] == '-' {
+		out = out[:len(out)-1]
+	}
+	return string(out)
+}
+
+var nameAttrs = []string{"NAME", "ID_MODEL_FROM_DATABASE", "ID_MODEL"}
```
bboozzoo: With this attribute set, is it possible that if udev devices DB is updated the slot name will change?

overlord/ifacestate/hotplug.go
```diff
@@ -30,3 +34,115 @@ func (m *InterfaceManager) HotplugDeviceAdded(devinfo *hotplug.HotplugDeviceInfo
...
+		var err error
+		suffixNumValue, err = strconv.ParseUint(string(pname[suffixIndex+1:]), 10, 32)
+		if err != nil {
+			suffixIndex = end
+		}
+	}
+
+	// increase suffix value until we have a unique name
+	for {
+		suffixNumValue++
```
bboozzoo: There's no example how ensureUniqueName is combined with the other two, but won't we end up depending on order of registration and possibly a slot ending up with a different name after a reboot?

stolowski: These are low-level helpers  that will be used to generate names when we see a device for the first time. It can change only for slots that are not connected.  As soon as a slot gets connected (by the user, or by auto-connect where applicable), the name "sticks" and gets associated with given device by its *device key* (see #5759 for default device key computation), and it will be remembered in the state (and isUnique implementation in one of the followup PRs will make sure it's not taken by any other slot).

overlord/ifacestate/hotplug.go
stolowski: No, see the explanation below.

overlord/ifacestate/hotplug.go
```diff
@@ -30,3 +34,115 @@ func (m *InterfaceManager) HotplugDeviceAdded(devinfo *hotplug.HotplugDeviceInfo
...
+			charCount++
+			if charCount >= maxLen {
+				break Loop
+			}
+		case unicode.IsDigit(c) && charCount > 0:
+			out = append(out, c)
+			dash = false
+			charCount++
+			if charCount >= maxLen {
+				break Loop
```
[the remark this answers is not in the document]
stolowski: Yeah, me too, learned it not long time ago in a review of my other PR ;)

overlord/ifacestate/hotplug.go
```diff
@@ -30,3 +34,115 @@ func (m *InterfaceManager) HotplugDeviceAdded(devinfo *hotplug.HotplugDeviceInfo
...
+		out = out[:len(out)-1]
+	}
+	return string(out)
+}
+
+var nameAttrs = []string{"NAME", "ID_MODEL_FROM_DATABASE", "ID_MODEL"}
+
+// suggestedSlotName returns the shortest name derived from attributes defined by nameAttrs, or
+// the fallbackName if there is no known attribute to derive name from. The name created from
+// attributes is cleaned up by cleanupSlotName function.
```
niemeyer: That's something a bit awkward to say here. The documentation of a function shoud point out what's the contract being established, not which function it calls internally to satisfy that contract. Someone reading this only needs to know what to expect from this one function.

overlord/ifacestate/hotplug.go
```diff
@@ -30,3 +34,115 @@ func (m *InterfaceManager) HotplugDeviceAdded(devinfo *hotplug.HotplugDeviceInfo
...
+	if suffixIndex == end {
+		pname = append(pname, '-')
+		suffixIndex++
+	} else {
+		var err error
+		suffixNumValue, err = strconv.ParseUint(string(pname[suffixIndex+1:]), 10, 32)
+		if err != nil {
+			suffixIndex = end
+		}
+	}
```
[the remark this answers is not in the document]
stolowski: As for the naming, not hard opinion, I'm fine with no dash before numbers, done.

stolowski:
Feedback addressed, this is ready for re-review.

overlord/ifacestate/hotplug.go
```diff
@@ -30,3 +35,98 @@ func (m *InterfaceManager) HotplugDeviceAdded(devinfo *hotplug.HotplugDeviceInfo
...
+
+// makeSlotName sanitizes a string to make it a valid slot name that
+// passes validation rules implemented by ValidateSlotName (see snap/validate.go):
+// - only lowercase letter, digits and dashes are allowed
+// - must start with a letter
+// - no double dashes, cannot end with a dash.
+// In addition names are truncated not to exceed maxGenerateSlotNameLen characters.
+func makeSlotName(s string) string {
+	var out []rune
+	var charCount int
```
niemeyer: How's charCount different from len(out)?

overlord/ifacestate/hotplug.go
```diff
@@ -30,3 +35,98 @@ func (m *InterfaceManager) HotplugDeviceAdded(devinfo *hotplug.HotplugDeviceInfo
...
+// name from. The name created from attributes is sanitized to ensure it's a
+// valid slot name. The fallbackName is typically the name of the interface.
+func suggestedSlotName(devinfo *hotplug.HotplugDeviceInfo, fallbackName string) string {
+	var candidates []string
+	for _, attr := range nameAttrs {
+		name, ok := devinfo.Attribute(attr)
+		if ok {
+			name = makeSlotName(name)
+			if name != "" {
+				candidates = append(candidates, name)
```
niemeyer: Why is this building a list of names to then compute the shortest one, instead of simply comparing len(name) with len(current candidate) here?

stolowski: Good point, thanks, simplified.

overlord/ifacestate/hotplug.go
stolowski: Good point, the difference was subtle and not worth keeping I think - the dashes would not affect character count, so you could actually get a string longer than max length if there were dashes there. Simplified.

PR #5783, opened 2018-09-06
tests: add new core16-base test

mvo5:
This test ensures that the core16 base will work for basic
tasks.

This is the result of
https://forum.snapcraft.io/t/using-base-snaps-with-account-control-doesnt-work-as-intended/7217


sergiocazzolato:
Thanks for adding this test :), just a question
Perhaps we could just create a test-snapd-sh-core16 snap instead which allows just to make sh. It is gonna be simpler. What do you think?

PR #5786, opened 2018-09-06
tests: update prepare/restore for nightly suite

sergiocazzolato:
Fix the nightly test suite which is currently failing

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -582,28 +582,13 @@ suites:
         # drop this blacklist.
         systems: [-fedora-*, -opensuse-*, -arch-*, -amazon-*]
         prepare: |
-            #shellcheck source=tests/lib/prepare.sh
-            . "$TESTSLIB"/prepare.sh
-            #shellcheck source=tests/lib/systems.sh
-            . "$TESTSLIB"/systems.sh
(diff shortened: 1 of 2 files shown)


spread.yaml
```diff
@@ -582,28 +582,13 @@ suites:
         # drop this blacklist.
         systems: [-fedora-*, -opensuse-*, -arch-*, -amazon-*]
         prepare: |
-            #shellcheck source=tests/lib/prepare.sh
```
zyga: Woot, thank you. This is much nicer

tests/nightly/unity/task.yaml
```diff
@@ -28,4 +32,6 @@ execute: |
 
     echo "Then the app window is created"
     expected=".*?\"qmlscene: clockMainView\": \(\"qmlscene\" \"com\.ubuntu\.clock\"\)"
-    while ! xwininfo -tree -root | grep -Pq "$expected"; do sleep 1; done
+    while ! xwininfo -tree -root | grep -Pq "$expected";
+        do sleep 1;
```
zyga: The location of `do` is unusual here. Could you please move it to the previous line?

PR #5790, opened 2018-09-07
overlord: skip testUpdateWithAutoconnectRetry temporarily

mvo5:
The testUpdateWithAutoconnectRetry test fails during the snapd
package PPA build in ~50% of the builds. What arches fail seems
random. It seems to be a race, the PPA builders tend to be slower
than the developer machines or GCE instances that this test runs
on usually.

Pawel is investigating the issue. Until this issue is resolved
this PR skips the test so that we get reliable snapd builds back.


mvo5:
No longer needed #5794 contains a real fix.

PR #5792, opened 2018-09-07
[RFC] {config,snap}state: add new refresh.metered=force option and flip default

mvo5:
This PR adds a new option to refresh.metered: "force". This option
will force automatic refreshes even on metered connections. The
default (when unset) is now to not do automatic refreshes.

However if the maximum time that refreshes can be defered
(60 days) is over the refresh is forced even if the network is
metered (this has always been the case).


overlord/snapstate/autorefresh.go
```diff
@@ -157,8 +157,14 @@ func canRefreshOnMeteredConnection(st *state.State) (bool, error) {
 	if err != nil && err != state.ErrNoState {
 		return false, err
 	}
-
-	return onMetered != "hold", nil
+	switch onMetered {
+	case "hold":
+		return false, nil
+	case "force":
```
bboozzoo: `"always"`?

mvo5: I like this, happy to change it. The original name was suggested by @niemeyer so I would like to get his input on this before changing anything.

overlord/configstate/configcore/refresh.go
```diff
@@ -64,7 +64,7 @@ func validateRefreshSchedule(tr Conf) error {
 		return err
 	}
 	switch refreshOnMeteredStr {
-	case "", "hold":
+	case "", "hold", "force":
```
zyga: +1 but this needs to be documented

overlord/snapstate/autorefresh.go
```diff
@@ -157,8 +157,14 @@ func canRefreshOnMeteredConnection(st *state.State) (bool, error) {
...
+		return true, nil
+	default:
```
zyga: Should the default case be an error? After all there are only three valid values, "", "force" and "hold"

mvo5:
Setting to blocked as this has a regression risk on deployed core devices that are connected via 3g - once this feature is released those devices will not get auto-updates anymore (even though they did get them before).

overlord/snapstate/autorefresh.go
```diff
@@ -157,8 +157,14 @@ func canRefreshOnMeteredConnection(st *state.State) (bool, error) {
...
+		return false, nil
```
bboozzoo: There's new feedback https://forum.snapcraft.io/t/snap-refresh-over-metered-connections/5001/30
It looks like we should do something like:
```
// holding refreshes on metered connections is an opt-in on Core devices,
// but a default on classic systems
return !release.IsOnClassic, nil
```

mvo5:
Closing until there is a clear concensus.

PR #5795, opened 2018-09-07
cmd/snap-update-ns: re-factor pair of helpers to call fstatfs once

zyga:
This patch changes IsReadOnly and IsSnapdCreatedPrivateTmpfs to accept
a syscall.Statfs_t buffer and delegate the call that populates the
buffer to the upcoming helper that combines the two. I did this after
noticing a pattern of repeated fstatfs, on the same file descriptor,
inside tests that would exercise this code path. This lets use save
one of the calls.

### DIFF
--- a/cmd/snap-update-ns/utils.go
+++ b/cmd/snap-update-ns/utils.go
@@ -73,21 +73,17 @@ var (
 //
 // Directories are read only when they reside on file systems mounted in read
 // only mode or when the underlying file system itself is inherently read only.
-func IsReadOnly(dirFd int, dirName string) (bool, error) {
-	var fsData syscall.Statfs_t
-	if err := sysFstatfs(dirFd, &fsData); err != nil {
-		return false, fmt.Errorf("cannot fstatfs %q: %s", dirName, err)
-	}
+func IsReadOnly(dirFd int, dirName string, fsData *syscall.Statfs_t) bool {
 	// If something is mounted with f_flags & ST_RDONLY then is read-only.
 	if fsData.Flags&StReadOnly == StReadOnly {
-		return true, nil
+		return true
 	}
 	// If something is a known read-only file-system then it is safe.
 	// Older copies of snapd were not mounting squashfs as read only.
 	if fsData.Type == SquashfsMagic {
-		return true, nil
+		return true
 	}
-	return false, nil
+	return false
 }
 
 // IsSnapdCreatedPrivateTmpfs returns true if a directory is a tmpfs mounted by snapd.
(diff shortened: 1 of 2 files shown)


cmd/snap-update-ns/utils.go
```diff
@@ -73,21 +73,17 @@ var (
 //
 // Directories are read only when they reside on file systems mounted in read
 // only mode or when the underlying file system itself is inherently read only.
-func IsReadOnly(dirFd int, dirName string) (bool, error) {
-	var fsData syscall.Statfs_t
-	if err := sysFstatfs(dirFd, &fsData); err != nil {
-		return false, fmt.Errorf("cannot fstatfs %q: %s", dirName, err)
-	}
+func IsReadOnly(dirFd int, dirName string, fsData *syscall.Statfs_t) bool {
```
bboozzoo: You don't need `dirFd` and `dirName` anymore.

cmd/snap-update-ns/utils.go
```diff
@@ -97,14 +93,10 @@ func IsReadOnly(dirFd int, dirName string) (bool, error) {
 // the mount namespace. A directory is trusted if it is a tmpfs that was
 // mounted by snap-confine or snapd-update-ns. Note that sub-directories of a
 // trusted tmpfs are not considered trusted by this function.
-func IsSnapdCreatedPrivateTmpfs(dirFd int, dirName string, changes []*Change) (bool, error) {
-	var fsData syscall.Statfs_t
-	if err := sysFstatfs(dirFd, &fsData); err != nil {
-		return false, fmt.Errorf("cannot fstatfs %q: %s", dirName, err)
-	}
+func IsSnapdCreatedPrivateTmpfs(dirFd int, dirName string, fsData *syscall.Statfs_t, changes []*Change) bool {
```
bboozzoo: `dirFd` is unused

cmd/snap-update-ns/utils.go
```diff
@@ -97,14 +93,10 @@ func IsReadOnly(dirFd int, dirName string) (bool, error) {
 // the mount namespace. A directory is trusted if it is a tmpfs that was
```
bboozzoo: The comment needs an update too.

cmd/snap-update-ns/utils.go
zyga: Ohh, that's right. Thank you, removed.

cmd/snap-update-ns/utils.go
zyga: Same :)

cmd/snap-update-ns/utils.go
zyga: Updated

cmd/snap-update-ns/utils_test.go
```diff
@@ -759,151 +759,87 @@ func (s *utilsSuite) TestSecureOpenPathRoot(c *C) {
...
 func (s *utilsSuite) TestIsReadOnlySquashfsMountedRw(c *C) {
-	statfs := syscall.Statfs_t{Type: update.SquashfsMagic}
 	path := "/some/path"
-	s.sys.InsertFstatfsResult("fstatfs 3 <ptr>", statfs)
-	fd, err := s.sys.Open(path, syscall.O_DIRECTORY, 0)
-	c.Assert(err, IsNil)
-	defer s.sys.Close(fd)
-	result, err := update.IsReadOnly(fd, path)
-	c.Assert(err, IsNil)
+	statfs := &syscall.Statfs_t{Type: update.SquashfsMagic}
```
stolowski: The tests become much shorter and easier to read/write with this change, nice.

PR #5797, opened 2018-09-07
osutil, o/snapshotstate, o/sss/backend: quick fixes

chipaca:
osutil, o/snapshotstate, o/sss/backend: quick fixes

This change addresses a few racy errors that started appearing in
spread once snapshotstate had landed, and some issues that were found
during the debugging of same.

In particular:

* `o/snapshotstate/backend`

  the runuser wrapper had a bug (or runuser has a bug?)  where the -- is
  needed because runuser's option parsing does not stop at he first
  non-option.

* `o/snapshotstate`

  In TestRestoreIntegrationFails sometimes one of the restores actually
  succeeds, and is undone, instead of erroring out.  Also, if running as
  root, don't run the integration tests, as the faking of the user does not
  amuse runuser

* `osutil`

  have a lock so that if one goroutine tries to mkdirallchown /foo/bar, and
  another tries to mkdirallchown /foo/baz, they can't both decide they need
  to make /foo and then have one fail.

  This is hackish, and needs a proper fix, but contention for the lock is
  going to be low and fixing it properly will be tricky. We'll do that
  later.

### DIFF
--- a/osutil/mkdirallchown.go
+++ b/osutil/mkdirallchown.go
@@ -22,14 +22,22 @@ package osutil
 import (
 	"os"
 	"path/filepath"
+	"sync"
 	"syscall"
 
 	"github.com/snapcore/snapd/osutil/sys"
 )
 
+// XXX: we need to come back and fix this; this is a hack to unblock us.
+//      As part of the fixing we should unify with the similar code in
+//      cmd/snap-update-ns/utils.(*Secure).MkdirAll
+var mu sync.Mutex
+
 // MkdirAllChown is like os.MkdirAll but it calls os.Chown on any
 // directories it creates.
 func MkdirAllChown(path string, perm os.FileMode, uid sys.UserID, gid sys.GroupID) error {
+	mu.Lock()
+	defer mu.Unlock()
 	return mkdirAllChown(filepath.Clean(path), perm, uid, gid)
 }
 
--- a/overlord/snapshotstate/backend/backend_test.go
+++ b/overlord/snapshotstate/backend/backend_test.go
@@ -436,6 +436,9 @@ func (s *snapshotSuite) TestAddDirToZip(c *check.C) {
 }
 
 func (s *snapshotSuite) TestHappyRoundtrip(c *check.C) {
+	if os.Geteuid() == 0 {
+		c.Skip("this test cannot run as root")
+	}
 	logger.SimpleSetup()
 
 	info := &snap.Info{SideInfo: snap.SideInfo{RealName: "hello-snap", Revision: snap.R(42)}, Version: "v1.33"}
--- a/overlord/snapshotstate/backend/backend_test.go
+++ b/overlord/snapshotstate/backend/backend_test.go
@@ -437,7 +437,7 @@ func (s *snapshotSuite) TestAddDirToZip(c *check.C) {
 
 func (s *snapshotSuite) TestHappyRoundtrip(c *check.C) {
 	if os.Geteuid() == 0 {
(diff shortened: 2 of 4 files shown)


osutil/mkdirallchown.go
```diff
@@ -22,14 +22,22 @@ package osutil
...
 	"path/filepath"
+	"sync"
 	"syscall"
 
 	"github.com/snapcore/snapd/osutil/sys"
 )
 
+// XXX: we need to come back and fix this; this is a hack to unblock us.
+//      As part of the fixing we should unify with the similar code in
+//      cmd/snap-update-ns/utils.(*Secure).MkdirAll
```
zyga: Thank you for this comment.

overlord/snapshotstate/backend/backend_test.go
```diff
@@ -436,6 +436,9 @@ func (s *snapshotSuite) TestAddDirToZip(c *check.C) {
 }
 
 func (s *snapshotSuite) TestHappyRoundtrip(c *check.C) {
+	if os.Geteuid() == 0 {
+		c.Skip("this test cannot run as root")
```
zyga: Not being familiar with this code it would be more useful to say why the test cannot run as root.

overlord/snapshotstate/backend/helpers.go
```diff
@@ -165,10 +165,11 @@ func maybeRunuserCommand(username string, args ...string) *exec.Cmd {
 		// runuser only works for euid 0, and doesn't make sense for root
 		return exec.Command(args[0], args[1:]...)
 	}
-	sudoArgs := make([]string, len(args)+2)
-	copy(sudoArgs[2:], args)
+	sudoArgs := make([]string, len(args)+3)
+	copy(sudoArgs[3:], args)
 	sudoArgs[0] = "-u"
 	sudoArgs[1] = username
+	sudoArgs[2] = "--"
```
zyga: Hah, eventually all software evolves to parse `--`

overlord/snapshotstate/snapshotstate_test.go
```diff
@@ -488,6 +488,9 @@ func (snapshotSuite) TestSaveOneSnap(c *check.C) {
 }
 
 func (snapshotSuite) TestSaveIntegration(c *check.C) {
+	if os.Geteuid() == 0 {
+		c.Skip("this test cannot run as root")
```
zyga: Same comment as above

overlord/snapshotstate/snapshotstate_test.go
```diff
@@ -946,11 +958,15 @@ func (snapshotSuite) TestRestoreIntegrationFails(c *check.C) {
 	tasks := change.Tasks()
 	c.Check(tasks, check.HasLen, 3)
 	for _, task := range tasks {
-		c.Check(task.Status(), check.Equals, state.ErrorStatus)
 		if strings.Contains(task.Summary(), `"too-snap"`) {
+			c.Check(task.Status(), check.Equals, state.ErrorStatus)
 			c.Check(strings.Join(task.Log(), "\n"), check.Matches, `\S+ ERROR mkdir \S+: permission denied`)
 		} else {
-			c.Check(strings.Join(task.Log(), "\n"), check.Matches, `\S+ ERROR context canceled`)
+			if task.Status() == state.ErrorStatus {
```
mvo5: A comment would be nice why sometimes we go into "ErrorStatus" and sometimes into "UndoStatus". I.e. why we handle both cases in the test.

PR #5802, opened 2018-09-10
cmd/snap-update-ns: introduce trespassing state tracking

zyga:
This patch adds a pair of types that are used together as a remedy for
trespassing bug where snap-update-ns writes to places outside of
$SNAP_DATA or $SNAP_COMMON. The types are used as follows:

Assumptions track the global state of the filesystem: directories that
have no write restrictions (like /tmp or $SNAP_DATA) and past mount
operations so that we can detect if a tmpfs was mounted by snapd or not.

Restrictions track state of ongoing operation that may write outside of
$SNAP_DATA or other allowed place. If restriction checking fails the
full path kept by the restrictions object is used to construct a
TrespassingError.

The patch also moves the existing trespassing helpers (isReadOnly and
IsSnapdCreatedPrivateTmpfs) to the same file as the rest of trespassing
code and make both private.

This is a separate pull request so that we can discuss the implementation
of this aspect separately from how the state is passed around the call
chain, which exists in #5805 if you want to look at the end-to-end fix.


cmd/snap-update-ns/trespassing.go
```diff
@@ -0,0 +1,234 @@
...
+// even if it would hit the real host filesystem (or transit through the host
+// filesystem). This is intended to be used with certain well-known locations
+// such as /tmp, $SNAP_DATA and $SNAP.
+func (as *Assumptions) AddUnrestrictedPaths(paths ...string) {
+	as.unrestrictedPaths = append(as.unrestrictedPaths, paths...)
+}
+
+// isRestricted returns true if a path follows restricted writing scheme.
+//
+// Writing to a restricted path results in step-by-step validation of each
```
bboozzoo: It feels this paragraph would be more useful if added at the call site.

cmd/snap-update-ns/trespassing.go
```diff
@@ -0,0 +1,234 @@
...
+func (as *Assumptions) RestrictionsFor(desiredPath string) *Restrictions {
+	if as.isRestricted(desiredPath) {
+		return &Restrictions{assumptions: as, desiredPath: desiredPath, restricted: true}
+	}
+	return nil
+}
+
+// Restrictions contains meta-data of a compound write operation.
+//
+// This structure helps various functions that write to the filesystem to keep
```
bboozzoo: maybe drop `various`

cmd/snap-update-ns/trespassing.go
```diff
@@ -0,0 +1,234 @@
...
+// is based on the full path of the desired object being constructed. This allows
+// various write helpers to avoid trespassing on host filesystem in places that
+// are not expected to be written to by snapd (e.g. outside of $SNAP_DATA).
+type Restrictions struct {
+	assumptions *Assumptions
+	desiredPath string
+	restricted  bool
+}
+
+// Check verifies if writing to a directory would trespass on the host.
```
bboozzoo: s/if/whether/

cmd/snap-update-ns/trespassing.go
```diff
@@ -0,0 +1,234 @@
...
+	ViolatedPath string
+	DesiredPath  string
+}
+
+// Error returns a formatted error message.
+func (e *TrespassingError) Error() string {
+	return fmt.Sprintf("cannot write to %q because it would affect the host in %q", e.DesiredPath, e.ViolatedPath)
+}
+
+// isReadOnly returns true if a directory is ready only.
```
bboozzoo: returns true for read-only directories

cmd/snap-update-ns/trespassing.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	// If something is a known read-only file-system then it is safe.
+	// Older copies of snapd were not mounting squashfs as read only.
+	if fsData.Type == SquashfsMagic {
+		return true
+	}
+	return false
+}
+
+// isSnapdCreatedPrivateTmpfs returns true if a directory is a tmpfs mounted by snapd.
```
bboozzoo: `..checks whether a directory resides on a tmpfs mounted by snapd`

cmd/snap-update-ns/trespassing.go
zyga: I tweaked it more but yeah, I kept the spirit.

cmd/snap-update-ns/trespassing.go
```diff
@@ -0,0 +1,231 @@
...
+	return false
+}
+
+// isSnapdCreatedPrivateTmpfs checks whether a directory resides on a tmpfs mounted by snapd
+//
+// The function inspects the directory and a list of changes that were applied
+// to the mount namespace. A directory is trusted if it is a tmpfs that was
+// mounted by snap-confine or snapd-update-ns. Note that sub-directories of a
+// trusted tmpfs are not considered trusted by this function.
+func isSnapdCreatedPrivateTmpfs(dirName string, fsData *syscall.Statfs_t, changes []*Change) bool {
```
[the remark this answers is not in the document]
zyga: We certainly can :) I'll make that change.

PR #5805, opened 2018-09-10
cmd/snap-update-ns: enforce trespassing checks

zyga:
This patch enables and enforces the trespassing checks. The rules as as
follows. Writes to /tmp (which is managed by snap-confine)
/var/lib/snap, and /snap/$SNAP_NAME are unrestricted. That is, they may
potentially write to the host filesystem. All other writes are
restricted and are internally replaced with a writable mimic, as if we
had encountered a read-only directory where we wish to create a new file
or directory for content bind mount.

The main function constructs the Assumptions object an pass it down all
the way to the function that performs mount changes. There is quite a
lot of argument passing to eventually pass the restrictions object to
filesystem operations such as MkDirAll, MkFileAll or MkSymlinkAll.

Performed mount changes are tracked so that we can keep track of the
exact set of tmpfs filesystems that were mounted by snapd. This requires
no new outside state so we are compatible with existing mount namespaces
and are safe to upgrade to.

There are ample new unit tests and, most importantly, the layout
integration test no longer needs to correct for the bug!

This branch is built on top of https://github.com/snapcore/snapd/pull/5802


cmd/snap-update-ns/trespassing.go
```diff
@@ -0,0 +1,234 @@
...
+	unrestrictedPaths []string
+	pastChanges       []*Change
+}
+
+// AddUnrestrictedPaths adds a list of directories where writing is allowed
+// even if it would hit the real host filesystem (or transit through the host
+// filesystem). This is intended to be used with certain well-known locations
+// such as /tmp, $SNAP_DATA and $SNAP.
+func (as *Assumptions) AddUnrestrictedPaths(paths ...string) {
+	as.unrestrictedPaths = append(as.unrestrictedPaths, paths...)
```
stolowski: Should we check that added paths are absolute?

cmd/snap-update-ns/trespassing.go
```diff
@@ -0,0 +1,234 @@
...
+//
+// Writing to a restricted path results in step-by-step validation of each
+// directory, starting from the root of the file system. Unless writing is
+// allowed a mimic must be constructed to ensure that writes are not visible in
+// undesired locations of the host filesystem.
+//
+// Provided path is the full, absolute path of the entity that needs to be
+// created (directory, file or symbolic link).
+func (as *Assumptions) isRestricted(path string) bool {
+	// Anything rooted at one of the unrestricted paths is not restricted.
```
stolowski: Also, would it make sense to throw away paths whose prefix we already have right in the `AddUnrestrictedPaths`? e.g. AddUnrestrictedPaths("/foo", "/foo/bar") would only remember "/foo"? Or is it not worth optimizing?

cmd/snap-update-ns/trespassing.go
```diff
@@ -0,0 +1,234 @@
...
+package main
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+	"syscall"
+)
+
+// Assumptions track the assumptions about the state of the filesystem.
```
stolowski: Nice concept.

cmd/snap-update-ns/trespassing.go
```diff
@@ -0,0 +1,234 @@
...
+func (as *Assumptions) canWriteToDirectory(dirFd int, dirName string) (bool, error) {
+	if !as.isRestricted(dirName) {
+		return true, nil
+	}
+	var fsData syscall.Statfs_t
+	if err := sysFstatfs(dirFd, &fsData); err != nil {
+		return false, fmt.Errorf("cannot fstatfs %q: %s", dirName, err)
+	}
+	// Writing to read only directories is allowed because EROFS is handled
+	// by each of the writing helpers already.
```
stolowski: This looks a little peculiar. Can you elaborate a little more (perhaps in the code, if it makes sens) what are the general semantics, i.e. why do we say 'true' if we know it's false?

cmd/snap-update-ns/trespassing.go
zyga: Right now it's not optimising. We are only adding three paths (hard-coded in main.go) that we allow.

cmd/snap-update-ns/trespassing.go
zyga: We could but see below.

cmd/snap-update-ns/trespassing.go
zyga: We are allowing _attempts_ to write because we know the code already handles EROFS correctly. We are going to say "yeah, go ahead, you cannot do any harm here". Note that we _attempt_ to write but we also handle EEXIST in the code so that when we are looking at creating `/usr/share/foo` we _attempt_ to create `/usr` but it exists so there is no error, then we _attempt_ to create `share` but it also exists then we attempt to create `foo` but it doesn't exist so we get EROFS which is handled already. If we said no here you would get a mimic even _before_ you considered if `usr` exists (because it is racy not to attempt that) and the whole thing would fall apart.

stolowski: Makes sense, thanks for explanation.

tests/main/layout/task.yaml
```diff
@@ -61,6 +60,15 @@ execute: |
...
         [ "$(test-snapd-layout.sh -c "stat /var/spool/rsyslog -c '%g'")" = "4" ]
 
+        echo "layout declarations didn't leak to the host"
+        test ! -e /etc/demo
+        test ! -e /etc/demo.conf
+        test ! -e /etc/demo.cfg
+        test ! -e /usr/share/demo
+        test ! -e /var/lib/demo
+        test ! -e /var/cache/demo
+        test ! -e /opt/demo
```
stolowski: Nice.

PR #5807, opened 2018-09-10
interfaces: extra argument for static attrs in NewConnectedPlug/NewConnectedSlot

stolowski:
Support passing static plug / slot attributes to NewConnectedPlug / NewConnectedSlot constructors - if not nil, they take precedence over static Attrs from plugInfo/slotInfo. 
This PR is purely mechanical, except for the change in NewConnectedPlug/Slot methods and their unit test.
This is an offshot of #5451 review and will become of prerequisite of that PR.


interfaces/builtin/location_control_test.go
```diff
@@ -87,7 +87,9 @@ func (s *LocationControlInterfaceSuite) TestConnectedPlugSnippetUsesSlotLabelAll
 		Name:      "location",
 		Interface: "location",
 		Apps:      map[string]*snap.AppInfo{"app1": app1, "app2": app2},
-	}, nil)
+	}, nil,
+
```
zyga: Spacing is odd here. Can you remove the newline and see if this looks more like before please?

interfaces/connection.go
```diff
@@ -103,19 +103,31 @@ func getAttribute(snapName string, ifaceName string, staticAttrs map[string]inte
 }
 
 // NewConnectedSlot creates an object representing a connected slot.
-func NewConnectedSlot(slot *snap.SlotInfo, dynamicAttrs map[string]interface{}) *ConnectedSlot {
+func NewConnectedSlot(slot *snap.SlotInfo, staticAttrs, dynamicAttrs map[string]interface{}) *ConnectedSlot {
+	var static map[string]interface{}
+	if staticAttrs != nil {
```
zyga: Will `staticAttrs` and `slot.Attrs` be, at the same time, non-nil and different?

interfaces/connection.go
```diff
@@ -103,19 +103,31 @@ func getAttribute(snapName string, ifaceName string, staticAttrs map[string]inte
 }
 
 // NewConnectedSlot creates an object representing a connected slot.
-func NewConnectedSlot(slot *snap.SlotInfo, dynamicAttrs map[string]interface{}) *ConnectedSlot {
+func NewConnectedSlot(slot *snap.SlotInfo, staticAttrs, dynamicAttrs map[string]interface{}) *ConnectedSlot {
```
bboozzoo: Wondering if this could be:
```
type EndpointAttributes {
    Static map[string]interface{}
    Dynamic map[string]interface{}
}
func NewConnectedSlot(slot *snap.SlotInfo attrs *EndpointAttributes) *ConnectedSlot {
   ...
}
func NewConnectedPlug(slot *snap.PlugInfo attrs *EndpointAttributes) *ConnectedPlug {
   ...
}
```

interfaces/builtin/location_control_test.go
stolowski: Fixed. Indeed, it came from gofmt...

interfaces/connection.go
stolowski: In theory they shouldn't be different. However since we decided to keep both dynamic and static attributes in conn state (for consistency, and to have full picture of all attributes at the time connection was made), this is setting the stage for the other PR where reloadConnection uses static attributes from conn state.

interfaces/connection.go
stolowski: Yes, this is a nice idea indeed, thanks! But I think I'd rather keep it for later if we want to pursue it, the changes would be even bigger then they are now.

overlord/ifacestate/ifacestate_test.go
```diff
@@ -835,12 +835,15 @@ func (s *interfaceManagerSuite) TestDisconnectTask(c *C) {
 		Plug: interfaces.NewConnectedPlug(&snap.PlugInfo{
 			Snap:  &snap.Info{SuggestedName: "consumer"},
 			Name:  "plug",
-			Attrs: map[string]interface{}{"attr1": "value1"}},
+			Attrs: map[string]interface{}{"attr1": "value1"}}, nil,
+
```
zyga: Unexpected newline

PR #5808, opened 2018-09-10
snapd-env-generator: fix when PATH is empty or unset

mvo5:
When the PATH is empty or unset the generator will generate the
wrong output (PATH=:/snap/bin). This causes havoc. This PR fixes
it by outputing only a PATH=/snap/bin segment that systemd will
correctly piece together.

### DIFF
--- a/cmd/snapd-env-generator/main.c
+++ b/cmd/snapd-env-generator/main.c
@@ -32,7 +32,7 @@ int main(int argc, char **argv)
 
 	char *path = getenv("PATH");
 	if (path == NULL || sc_streq(path, "")) {
-               printf("PATH=%s\n", snap_bin_dir);
+	       // do nothing, until systemd is fixed, see LP#1791691
                return 0;
         }
 	char buf[PATH_MAX + 1] = { 0 };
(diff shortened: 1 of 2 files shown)


cmd/snapd-env-generator/main.c
```diff
@@ -31,8 +31,10 @@ int main(int argc, char **argv)
 	const char *snap_bin_dir = SNAP_MOUNT_DIR "/bin";
 
 	char *path = getenv("PATH");
-	if (path == NULL)
-		path = "";
+	if (path == NULL || sc_streq(path, "")) {
```
zyga: +1, shall I add `sc_strempty` for cases like this?

mvo5:
@smoser Thanks Scott. I just talked on irc with @xnox and he plans to fix this issue in systemd. Until then his advice is that if the PATH is unset our generator should do nothing. Picking a default PATH seems to be problematic - there are just too many too chose from: https://bugs.launchpad.net/ubuntu/+source/dpkg/+bug/1792004

powersj:
@mvo5 I want to confirm that you were going to land and upload your latest change of doing nothing hopefully today/tomorrow?

mvo5:
@powersj Yes, this is scheduled to be uploaded ASAP (hopefully today). We would have done it yesterday already but github was having too much trouble.

PR #5811, opened 2018-09-11
tests: add test that runs snapctl with a core18 snap

mvo5:
We currently test snapctl from inside a snap. This PR extends the
test to also check if snapctl works from inside a core18 based
snap.


chipaca:
The failures are now relevant :-)

PR #5814, opened 2018-09-11
daemon: fix snap list --all with parallel snap instances

bboozzoo:
`snap list --all` did not work when there were multiple instances of the same snap installed.

### DIFF
--- a/daemon/snap.go
+++ b/daemon/snap.go
@@ -128,7 +128,7 @@ func allLocalSnapInfos(st *state.State, all bool, wanted map[string]bool) ([]abo
 		var err error
 		if all {
 			for _, seq := range snapst.Sequence {
-				info, err = snap.ReadInfo(seq.RealName, seq)
+				info, err = snap.ReadInfo(name, seq)
 				if err != nil {
 					break
 				}
--- a/tests/main/listing/task.yaml
+++ b/tests/main/listing/task.yaml
@@ -5,6 +5,12 @@ prepare: |
     . "$TESTSLIB"/snaps.sh
     install_local test-snapd-tools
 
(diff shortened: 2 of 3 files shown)


daemon/snap.go
```diff
@@ -128,7 +128,7 @@ func allLocalSnapInfos(st *state.State, all bool, wanted map[string]bool) ([]abo
 		var err error
 		if all {
 			for _, seq := range snapst.Sequence {
-				info, err = snap.ReadInfo(seq.RealName, seq)
```
zyga: Is real name the same as name in the snap state?

bboozzoo: Not always. `RealName` is from `snap.SideInfo` and comes directly from snap YAML file. In snapstate we're using `snap.Info.InstanceName()` (i.e. `name` at this location), which is `RealName` + optionally `InstanceKey`. That's what we should be using here.

tests/main/listing/task.yaml
```diff
@@ -50,3 +58,10 @@ execute: |
         echo "$output"
         exit 1
     fi
+
+    output=$(snap list --all |grep 'test-snapd-tools_foo ')
```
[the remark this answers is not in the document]
bboozzoo: There should be exactly one, but I guess MATCH will be ok to use here too.

daemon/snap.go
chipaca: `RealName` as in the name blessed by the store, as opposed to the one in the snap YAML file that was merely requested.

bboozzoo: Missed how loading from the file only sets `SuggestedName`. Anyways, AFAICT `name` is the one we want to use here.

PR #5817, opened 2018-09-11
snapstate/tests: serialize all appends in fake backend

stolowski:
Serialize all append ops in fake test backend - this should fix the occasional failure observed on travis (and also locally when run the test in a loop) that manifested itself with a missing fake op (while all state tasks indicated all operations were done). Thanks to @mvo5 for steering me into this direction!


overlord/snapstate/backend_test.go
```diff
@@ -144,6 +145,9 @@ func (f *fakeStore) pokeStateLock() {
 func (f *fakeStore) SnapInfo(spec store.SnapSpec, user *auth.UserState) (*snap.Info, error) {
 	f.pokeStateLock()
 
+	f.fakeBackend.mu.Lock()
```
bboozzoo: If it's just about appends, would it be possible to limit the locking just the append() part? Eg:
```
func (f *fakeBackend) appendOp(op *fakeOp) {
    f.mu.Lock()
    defer f.mu.Unlock()
    f.ops = append(...)
}
```

stolowski: Neat idea, done, thanks!

overlord/snapstate/backend_test.go
```diff
@@ -881,9 +885,15 @@ func (f *fakeSnappyBackend) UpdateAliases(add []*backend.Alias, remove []*backen
...
 func (f *fakeSnappyBackend) RemoveSnapAliases(snapName string) error {
-	f.ops = append(f.ops, fakeOp{
+	f.AppendOp(&fakeOp{
 		op:   "remove-snap-aliases",
 		name: snapName,
 	})
 	return nil
 }
+
+func (f *fakeSnappyBackend) AppendOp(op *fakeOp) {
```
bboozzoo: Does it need to be public?

PR #5818, opened 2018-09-11
cmd/snap: handle "snap interfaces core" better

zyga:
The legacy "snap interfaces" command does client side filtering and thus
doesn't benefit from the server-side nickname/remapping created when the
core18 snap was introduced. With the introduction of the re-mapping
logic implicit slots are provided by various snaps that are always
exposed, on the API layer, under the "system" nickname.

The re-mapping is done server side so various commands like "connect"
and "disconnect" work perfectly fine with both old and new names. The
"snap interfaes" (plural) command is the exception as it was doing all
of the filtering locally.

This patch adds a tiny special case that allows matching "system"
nickname when the user has requested the "core" snap.

### DIFF
--- a/tests/main/interfaces-cli/task.yaml
+++ b/tests/main/interfaces-cli/task.yaml
@@ -23,3 +23,8 @@ execute: |
     echo "When the interfaces list is restricted by slot and snap"
     echo "Then only the requested slots are shown"
     snap interfaces -i "$PLUG" "$SNAP_NAME" | grep -Pzq "$expected"
+
+    echo "Implicit slots are exposed by a snap holding the nickname 'system'"
+    echo "but for compatibility they can also be listed when asking for 'core'"
+    snap interfaces -i network system | MATCH '^:network .*'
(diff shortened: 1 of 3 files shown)


zyga:
Sure, will do

tests/main/interfaces-cli/task.yaml
```diff
@@ -23,3 +23,8 @@ execute: |
     echo "When the interfaces list is restricted by slot and snap"
     echo "Then only the requested slots are shown"
     snap interfaces -i "$PLUG" "$SNAP_NAME" | grep -Pzq "$expected"
+
+    echo "Implicit slots are exposed by a snap holding the nickname 'system'"
+    echo "but for compatibility they can also be listed when asking for 'core'"
+    snap interfaces -i network system | MATCH '^:network .*'
+    snap interfaces -i network core | MATCH '^:network .*'
```
bboozzoo: Whould adding this be useful too?
```
diff -up <(snap interfaces system) <(snap interfaces core)
```

PR #5825, opened 2018-09-12
tests: add snap install hook with base: core18

mvo5:
This is to reproduce/verify the issue that was reported on:
https://forum.snapcraft.io/t/7297


tests/lib/snaps/test-snapd-snapctl-core18/meta/hooks/install
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -e
```
zyga: Is bash relevant to the issue?

mvo5: Its not, I moved to /bin/sh

sergiocazzolato:
@mvo5 tests seem to be failing for real with this change.

PR #5826, opened 2018-09-13
tests: refactor for nested suite and tests fixed

sergiocazzolato:
This is the first part of the refactor, then there is a change to make those tests run on core 18 as well.

### DIFF
--- a/tests/nested/core-revert/task.yaml
+++ b/tests/nested/core-revert/task.yaml
@@ -34,8 +34,6 @@ execute: |
 
     cd "$SPREAD_PATH"
     execute_remote "sudo snap install network-manager"
-    execute_remote "sudo snap install bluez"
-    execute_remote "sudo bluez.bluetoothctl -a"
     execute_remote "snap aliases" | MATCH "nmcli"
 
     # configure network-manager to take over the connection control on next reboot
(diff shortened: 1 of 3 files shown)


tests/nested/core-revert/task.yaml
```diff
@@ -45,28 +43,36 @@ execute: |
...
+        echo "ssh not stablished, exiting..."
+        exit 1
+    fi
 
     while ! execute_remote "snap changes" | MATCH "Done.*Revert \"core\" snap"; do sleep 1 ; done
     execute_remote "snap aliases" | MATCH "nmcli"
     execute_remote "snap info core" | MATCH "tracking: +${CORE_REFRESH_CHANNEL}"
     execute_remote "ifconfig" | MATCH eth0
-    execute_remote "sudo bluez.bluetoothctl -a"
+    #execute_remote "sudo bluez.bluetoothctl -a"
```
mvo5: Given that similar bluez installs were removed previously, why is this kept (and commented out) instead of removed?

tests/nested/extra-snaps-assertions/task.yaml
```diff
@@ -3,53 +3,22 @@ summary: create ubuntu-core image and execute the suite in a nested qemu instanc
 systems: [ubuntu-16.04-64, ubuntu-16.04-32]
 
 prepare: |
-    # FIXME: until https://github.com/snapcore/snapd/pull/3263 is available from
```
mvo5: Nice removal of no longer needed code!

PR #5831, opened 2018-09-14
client, cmd/snap: on !linux, exit when the client tries to Do something

chipaca:
Before this change snap client on non-linux would exit with a warning
when trying to instantiate a client.Client object inside a command. As
we need to move towards instantiating client before running the
commands parser and passing that client in to the command, this
approach proves problematic. This change makes it so that only when
trying to talk to the network does the non-linux client bail out.

### DIFF
--- a/client/client_other.go
+++ b/client/client_other.go
@@ -1,0 +1,26 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+// +build !linux
+
+/*
+ * Copyright (C) 2015 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


client/client_other.go
```diff
@@ -0,0 +1,26 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+// +build !linux
+
+/*
+ * Copyright (C) 2015 Canonical Ltd
```
zyga: Nitpick, 2015-2018?

cmd/snap/main_other.go
```diff
@@ -0,0 +1,48 @@
...
+	"fmt"
+	"net/http"
+	"os"
+	"runtime"
+
+	"github.com/snapcore/snapd/client"
+	"github.com/snapcore/snapd/i18n"
+)
+
+type otherDoer struct{}
```
stolowski: That's probably a matter of personal taste, but I'd consider renaming it to `nonLinuxDoer` or some such, and the file to `nonlinux_main.go` to make it more apparent what's going on - it's easy to miss the `// +build !linux` directive.

PR #5834, opened 2018-09-14
cmd/snap: commands no longer build their own client

chipaca:
Before this change, any command needing to talk to snapd would
instantiate a *client.Client and use that. With this, the client is
instantiated by run(), and passed to the commands (in a slightly
convoluted way, because flags is the one invoking the commands).


chipaca:
I answered @mvo5 face-to-face (wee), but for the record: we're doing this because we're getting information in client that needs addressing in the same way across all commands, which means we either have a single global client (ugly), handle it in every single command (painful), or pass the client in (ugh). This one is "ugh".

chipaca:
@mvo5 further to that, [here's the diff to having "snap warnings" and etc](https://github.com/snapcore/snapd/compare/master...chipaca:snap-warnings)

PR #5835, opened 2018-09-17
tests: find snaps just for edge and beta channels

sergiocazzolato:
Changes:
. Using test-snapd-just-{edge,beta} snaps
. Snap test-snapd-only-in-edge replaced (tests udpated)
. searching test updated to search snaps on edge and beta channels

### DIFF
--- a/tests/lib/snaps/test-snapd-just-beta/snapcraft.yaml
+++ b/tests/lib/snaps/test-snapd-just-beta/snapcraft.yaml
@@ -1,0 +1,15 @@
+name: test-snapd-just-beta
+version: 1.0
+summary: Basic snap published just on beta channel
+description: A basic snap published just on beta channel
+confinement: strict
+grade: stable
+
+apps:
+    snap-name:
+        command: snap-name
+
+parts:
+    copy:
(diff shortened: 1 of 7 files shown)


bboozzoo:
Was the snap released for i386?
```
+ MATCH 'error: snap "test-snapd-just-edge" is not available on stable but'
error: snap "test-snapd-just-edge" is not available on this architecture (i386)
       but exists on other architectures (amd64, armhf, ppc64el, s390x).
```

sergiocazzolato:
@bboozzoo  ready now

tests/lib/snaps/test-snapd-just-beta/snapcraft.yaml
```diff
@@ -0,0 +1,15 @@
+name: test-snapd-just-beta
+version: 1.0
+summary: Basic snap published just on beta channel
+description: A basic snap published just on beta channel
+confinement: strict
+grade: stable
```
mvo5: It would be much simpler to make this an: `architecture: [all]` snap but its fine.

PR #5840, opened 2018-09-18
interfaces/apparmor, interfaces/builtin: tweaks for parallel snap installs

bboozzoo:
Tweak the default AppArmor profile and builtin interfaces for parallel snap installation. 
I've tried to leave comments where appropriate so that future maintenance is possible.

As agreed in #5713 snaps supporting parallel installation should use `$SNAP_INSTANCE_NAME` environment variable instead of `$SNAP_NAME` to be safe. Although we now do the bind mounts of `/snap/foo_bar -> /snap/foo` and `/var/snap/foo_bar -> /var/snap/foo`, I've opened the default profile to cover access in the instance specific location too.

Note that some interfaces, namely fuse-support, greengrass, and cifs-mount use `@{SNAP_NAME}`. I am bit on the fence on whether those should be updated to have a corresponding set of rules using `@{SNAP_INSTANCE_NAME}` too.


interfaces/builtin/home.go
```diff
@@ -65,9 +65,9 @@ owner @{HOME}/snap[^/]**          rwklix,
 owner @{HOME}/{s,sn,sna}{,/} rwklix,
 
 # Allow access to @{HOME}/snap/ to allow directory traversals from
-# @{HOME}/snap/@{SNAP_NAME} through @{HOME}/snap to @{HOME}. While this leaks
-# snap names, it fixes usability issues for snaps that require this
-# transitional interface.
+# @{HOME}/snap/@{SNAP_INSTANCE_NAME} through @{HOME}/snap to @{HOME}.
+# While this leaks # snap names, it fixes usability issues for snaps
```
zyga: Spurious `#` after `leaks`

tests/main/parallel-install-basic/task.yaml
```diff
@@ -11,15 +11,8 @@ execute: |
     #shellcheck source=tests/lib/snaps.sh
     . "$TESTSLIB"/snaps.sh
 
-    devmode=
```
zyga: Whee :-)

interfaces/apparmor/backend_test.go
```diff
@@ -527,6 +528,51 @@ func (s *backendSuite) TestCombineSnippets(c *C) {
...
+		"###PROFILEATTACH### (attach_disconnected,mediate_deleted) {\n" +
+		"###SNIPPETS###\n" +
+		"}\n")
+	defer restoreClassicTemplate()
+	s.Iface.AppArmorPermanentSlotCallback = func(spec *apparmor.Specification, slot *snap.SlotInfo) error {
+		return nil
+	}
+	expected := `
+@{SNAP_NAME}="samba"
+@{SNAP_INSTANCE_NAME}="samba_foo"
```
jdstrand: Within the context of the apparmor profile, SNAP_INSTANCE_NAME is a little weird, but I understand why you did it. For someone auditing the profile, they will wonder "what is the difference between SNAP_NAME and SNAP_INSTANCE_NAME?".

It makes sense to me for the end result to be:
```
@{SNAP_NAME}="samba_foo"
# the snap 'basename' is used to support environment settings in parallel installs
@{SNAP_BASENAME}="samba"
```

(I'm not sure if 'SNAP_BASENAME' is the best name, but something like that)

I think I prefer this because it feels like SNAP_NAME should tend to match the security label of the snap (and therefore PROFILE_DBUS). The PR would also be smaller. Granted, SNAP_NAME doesn't match the snap's environment, but in this case I don't think that matters-- it is more important that one reading or authoring the profile understands how it works and that SNAP_NAME is the normal thing we use, with SNAP_BASENAME as an exceptional case. As it stands, we are adding a new apparmor variable that is actually the normal case, with the existing SNAP_NAME as the exception, which feels wrong.

interfaces/apparmor/template.go
```diff
@@ -363,26 +372,40 @@ var defaultTemplate = `
   @{INSTALL_DIR}/@{SNAP_NAME}/                   r,
```
jdstrand: I'd like to see a comment here: `# For parallel installs, see above`

interfaces/apparmor/template.go
```diff
@@ -363,26 +372,40 @@ var defaultTemplate = `
...
   @{INSTALL_DIR}/@{SNAP_NAME}/@{SNAP_REVISION}/    r,
   @{INSTALL_DIR}/@{SNAP_NAME}/@{SNAP_REVISION}/**  mrklix,
+  # parallel-installs: should snap use $SNAP_INSTANCE_NAME for these paths, make
+  # sure it has access too
+  @{INSTALL_DIR}/@{SNAP_INSTANCE_NAME}/                   r,
+  @{INSTALL_DIR}/@{SNAP_INSTANCE_NAME}/@{SNAP_REVISION}/    r,
+  @{INSTALL_DIR}/@{SNAP_INSTANCE_NAME}/@{SNAP_REVISION}/**  mrklix,
 
   # Read-only install directory for other revisions to help with bugs like
   # LP: #1616650 and LP: #1655992
```
jdstrand: Also add: `(support parallel installs, see above)`.

interfaces/apparmor/template.go
```diff
@@ -363,26 +372,40 @@ var defaultTemplate = `
...
   @{INSTALL_DIR}/@{SNAP_NAME}/**  mrkix,
+  @{INSTALL_DIR}/@{SNAP_INSTANCE_NAME}/**  mrkix,
 
   # Read-only home area for other versions
```
jdstrand: ` and parallel installs`.

interfaces/apparmor/template.go
```diff
@@ -392,25 +415,25 @@ var defaultTemplate = `
 
   # App-specific access to files and directories in /dev/shm. We allow file
   # access in /dev/shm for shm_open() and files in subdirectories for open()
```
jdstrand: Please add:
```
# Because snapd does not touch /dev/shm for parallel installs, only
# allow INSTANCE_NAME.
```

interfaces/apparmor/template.go
```diff
@@ -392,25 +415,25 @@ var defaultTemplate = `
...
-  /{dev,run}/shm/snap.@{SNAP_NAME}.** mrwlkix,
+  /{dev,run}/shm/snap.@{SNAP_INSTANCE_NAME}.** mrwlkix,
   # Also allow app-specific access for sem_open()
-  /{dev,run}/shm/sem.snap.@{SNAP_NAME}.* mrwk,
+  /{dev,run}/shm/sem.snap.@{SNAP_INSTANCE_NAME}.* mrwk,
 
   # Snap-specific XDG_RUNTIME_DIR that is based on the UID of the user
```
jdstrand: Please add:
```
# Because snapd does not touch /run/user for parallel installs, only
# allow INSTANCE_NAME.
```

interfaces/apparmor/template.go
```diff
@@ -392,25 +415,25 @@ var defaultTemplate = `
...
-  owner /run/user/[0-9]*/snap.@{SNAP_NAME}/   rw,
-  owner /run/user/[0-9]*/snap.@{SNAP_NAME}/** mrwklix,
+  owner /run/user/[0-9]*/snap.@{SNAP_INSTANCE_NAME}/   rw,
+  owner /run/user/[0-9]*/snap.@{SNAP_INSTANCE_NAME}/** mrwklix,
 
   # Allow apps from the same package to communicate with each other via an
   # abstract or anonymous socket
```
jdstrand: Please add:
```
# Because snapd does not touch abstract sockets for parallel installs, only
# allow INSTANCE_NAME.
```

interfaces/apparmor/template.go
```diff
@@ -392,25 +415,25 @@ var defaultTemplate = `
...
-  unix peer=(label=snap.@{SNAP_NAME}.*),
+  unix peer=(label=snap.@{SNAP_INSTANCE_NAME}.*),
 
   # Allow apps from the same package to communicate with each other via DBus.
   # Note: this does not grant access to the DBus sockets of well known buses
   # (will still need to use an appropriate interface for that).
```
jdstrand: Please add:
```
# Because snapd does not touch dbus for parallel installs, only allow
# INSTANCE_NAME.
```

interfaces/apparmor/template.go
```diff
@@ -392,25 +415,25 @@ var defaultTemplate = `
...
-  dbus (receive, send) peer=(label=snap.@{SNAP_NAME}.*),
+  dbus (receive, send) peer=(label=snap.@{SNAP_INSTANCE_NAME}.*),
 
   # Allow apps from the same package to signal each other via signals
```
jdstrand: Please add:
```
# Because snapd does not touch signals for parallel installs, only allow
# INSTANCE_NAME.
```

interfaces/apparmor/backend_test.go
jdstrand: In retrospect and talking through everything with @bboozzoo, I'm taking back this comment. Let's use SNAP_NAME and be sure to document it clearly in the profile.

interfaces/builtin/thumbnailer_service.go
```diff
@@ -127,7 +127,8 @@ func (iface *thumbnailerServiceInterface) AppArmorPermanentSlot(spec *apparmor.S
 func (iface *thumbnailerServiceInterface) AppArmorConnectedSlot(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
 	snippet := thumbnailerServiceConnectedSlotAppArmor
 	old := "###PLUG_SNAP_NAME###"
-	// TODO parallel-install: use of proper instance/store name
+	// parallel-installs: PLUG_SNAP_NAME is used in the context of dbus
+	// mediation rules, need to use the actual instance name
```
jdstrand: This is correct.

interfaces/builtin/unity7.go
```diff
@@ -639,7 +641,10 @@ func (iface *unity7Interface) AppArmorConnectedPlug(spec *apparmor.Specification
 	// but we don't care about that here because the rule above already
 	// does that) to '_'. Since we know that the desktop filename starts
 	// with the snap name, perform this conversion on the snap name.
-	// TODO parallel-install: use of proper instance/store name
+	//
+	// parallel-installs: UNITY_SNAP_NAME is used in the context of dbus
+	// mediation, this unintentionally opens access to dbus paths of keyed
+	// instances of @{SNAP_NAME} to @{SNAP_NAME} snap
```
jdstrand: Nice catch. Please add an AppArmor policy comment above the affected rules. Eg:
```
# NOTE: when @{SNAP_NAME} == @{SNAP_INSTANCE_NAME}, this rule
# allows the snap to access parallel installs of this snap.
```

(we get policy audits from 3rd parties all the time and we need to note the deficiencies there).

interfaces/apparmor/template.go
```diff
@@ -363,26 +372,40 @@ var defaultTemplate = `
...
+  @{INSTALL_DIR}/@{SNAP_INSTANCE_NAME}/@{SNAP_REVISION}/**  mrklix,
 
   # Read-only install directory for other revisions to help with bugs like
   # LP: #1616650 and LP: #1655992
   @{INSTALL_DIR}/@{SNAP_NAME}/**  mrkix,
+  @{INSTALL_DIR}/@{SNAP_INSTANCE_NAME}/**  mrkix,
 
   # Read-only home area for other versions
-  owner @{HOME}/snap/@{SNAP_NAME}/                  r,
-  owner @{HOME}/snap/@{SNAP_NAME}/**                mrkix,
```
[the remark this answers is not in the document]
bboozzoo: Added here and another one under `# Writable home area..`

interfaces/builtin/greengrass_support.go
```diff
@@ -100,6 +101,7 @@ mount options=(rw, nosuid, noexec) fstype=devpts devpts -> /var/snap/@{SNAP_NAME
 mount options=(rw, nosuid, nodev, noexec) fstype=tmpfs shm -> /var/snap/@{SNAP_NAME}/*/rootfs/dev/shm/,
 
 # add a few common devices
+# parallel-installs: SNAP_{DATA,COMMON} are remapped, need to use SNAP_NAME
 mount options=(rw, bind) /dev/full -> /var/snap/@{SNAP_NAME}/*/rootfs/dev/full,
 mount options=(rw, bind) /dev/null -> /var/snap/@{SNAP_NAME}/*/rootfs/dev/null,
 mount options=(rw, bind) /dev/random -> /var/snap/@{SNAP_NAME}/*/rootfs/dev/random,
```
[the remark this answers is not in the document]
bboozzoo: Wondering if the same thing should be done for fuse_support and cifs_mount. What do you think?

interfaces/builtin/unity7.go
```diff
@@ -482,6 +482,8 @@ dbus (receive)
 # this leaks the names of snaps with desktop files
 /var/lib/snapd/desktop/applications/ r,
 /var/lib/snapd/desktop/applications/mimeinfo.cache r,
+# parallel-installs: this leaks read access to desktop files owned by keyed
+# instances of @{SNAP_NAME} to @{SNAP_NAME} snap
 /var/lib/snapd/desktop/applications/@{SNAP_NAME}_*.desktop r,
```
jdstrand: This should be adjusted to be:
```
/var/lib/snapd/desktop/applications/@{SNAP_INSTANCE_NAME}_*.desktop r,
```

Let's make the commend slightly more clear with:
```
# parallel-installs: when @{SNAP_INSTANCE_NAME} == @{SNAP_NAME},
# this leaks read access to desktop files of parallel installs of the snap
```

bboozzoo:
> The same goes for various slot implementations where the policy allows them to put files in well-known locations such as /run (eg, docker-support, openvswitch, network-manager, mir, lxd and potentially other future interfaces). Maybe it makes sense to document something about slot implementations and well-known locations generally?

I think this falls under the general parallel installation awareness. I think we all agree that some snaps will not be usable to the usual extent (if all) when parallel installed. I'll try to document it in the forum topic. Other than that, I think we can revisit this and check what we can pull off realistically if people come asking about such functionality.

> Can you describe any considerations surrounding the content interface (eg, where either the providing or the consuming snap is a parallel install)?

There is a spread test exercising the content interface in the context of parallel installations. Only the plug part of the interface is used by parallel installed snaps (the idea was to mimic a use case when you have shared libraries/themes/data). I can add another test that would use parallel installed snaps in the slot in a follow-up.

PR #5841, opened 2018-09-18
interfaces/hotplug: hotplug spec takes one slot definition

stolowski:
Change hotplug spec to allow one slot at a time for given udev event and interface.

Background: a new instance of hotplug spec is passed to every hotplug interface to allow creation of hotplug slots in response to udev add event. The initial implementation (incorrectly) assumed one udev event could lead to multiple slots of same interface type getting created via the spec, but this doesn't make sense. If a device has multiple ports (or partitions or some other distinctive features), there will actually be separate udev events for them, they will have distinct device keys so each event will lead to a new slot getting created.

Note: due to this simplification the SlotSpec struct could potentially be dropped, but I think having Specification and SlotSpec makes it more future-proof and extensible, e.g. Specification can grow extra methods/attributes.


interfaces/hotplug/spec.go
```diff
@@ -34,31 +34,30 @@ type Definer interface {
...
-// AddSlot adds a specification of a slot.
-func (h *Specification) AddSlot(slotSpec *SlotSpec) error {
-	if _, ok := h.slots[slotSpec.Name]; ok {
-		return fmt.Errorf("slot %q already exists", slotSpec.Name)
+// SetSlot adds a specification of a slot.
+func (h *Specification) SetSlot(slotSpec *SlotSpec) error {
+	if h.slot != nil {
+		return fmt.Errorf("slot specification already created")
 	}
 	if err := snap.ValidateSlotName(slotSpec.Name); err != nil {
```
bboozzoo: The comment around SlotSpec suggest that Name can be empty and hotplug will assign something to it in such case. Will this error out if slotSpec.Name is empty?

interfaces/hotplug/spec_test.go
```diff
@@ -42,27 +42,18 @@ func (s *hotplugSpecSuite) TearDownTest(c *C) {
...
-	c.Assert(spec.AddSlot(&SlotSpec{Name: "slot2", Label: "A slot", Attrs: map[string]interface{}{"baz": "booze"}}), IsNil)
-
-	c.Assert(spec.Slots(), DeepEquals, []*SlotSpec{
-		{Name: "slot1", Label: "A slot", Attrs: map[string]interface{}{"foo": "bar"}},
-		{Name: "slot2", Label: "A slot", Attrs: map[string]interface{}{"baz": "booze"}},
-	})
-
-	c.Assert(spec.AddSlot(&SlotSpec{Name: "---", Label: "A slot", Attrs: map[string]interface{}{}}), ErrorMatches, `invalid slot name: "---"`)
+	c.Assert(spec.SetSlot(&SlotSpec{Name: "slot1", Label: "A slot", Attrs: map[string]interface{}{"foo": "bar"}}), IsNil)
+	c.Assert(spec.Slot(), DeepEquals, &SlotSpec{Name: "slot1", Label: "A slot", Attrs: map[string]interface{}{"foo": "bar"}})
```
bboozzoo: Probably mising a test with `&SlotSpec{Name: "", ..}`

interfaces/hotplug/spec.go
stolowski: Good catch, thanks! Fixed.

interfaces/hotplug/spec_test.go
stolowski: Indeed, added.

interfaces/hotplug/spec.go
```diff
@@ -34,60 +34,52 @@ type Definer interface {
 
 // SlotSpec is a definition of the slot to create in response to udev event.
 type SlotSpec struct {
-	// XXX: Name is the name the interface wants to give to the slot; we
-	// might want to mediate this though (e.g. generate automatically), so this
-	// may change/go away.
+	// Name is the name the interface wants to give to the slot. It can be
+	// left empty, in such case hotplug machinery will auto-generate it.
+	// Hotplug machinery make sure the proposed name is unique by appending
```
bboozzoo: s/make/makes/

interfaces/hotplug/spec.go
```diff
@@ -34,60 +34,52 @@ type Definer interface {
...
+	// numeric suffix if needed.
 	Name  string
 	Label string
 	Attrs map[string]interface{}
 }
 
-// Specification contains data about all slots that a hotplug interface wants to have created in response to uevent.
+// Specification contains a slot definition to create in response to uevent.
```
bboozzoo: Maybe consider using `hotplug event` instead of uevent. It's probably to ok to keep `uevent` somewhere closer to where you process actual udev events.

interfaces/hotplug/spec.go
```diff
@@ -34,60 +34,52 @@ type Definer interface {
 
 // SlotSpec is a definition of the slot to create in response to udev event.
 type SlotSpec struct {
-	// XXX: Name is the name the interface wants to give to the slot; we
-	// might want to mediate this though (e.g. generate automatically), so this
-	// may change/go away.
+	// Name is the name the interface wants to give to the slot. It can be
```
bboozzoo: Maybe something like this?
```
// Name is how the interface wants to name the slot. When
// left empty, one will be generated on demand. The hotplug
// machinery appends a suffix to ensure uniqueness of the name.
```

interfaces/hotplug/spec.go
```diff
@@ -34,60 +34,52 @@ type Definer interface {
 
 // SlotSpec is a definition of the slot to create in response to udev event.
 type SlotSpec struct {
```
bboozzoo: Why not call it `RequestedSlotSpec` or `DesiredSlotSpec`?

stolowski: I like `RequestedSlotSpec`, thanks, changed.

interfaces/hotplug/spec.go
```diff
@@ -32,62 +32,53 @@ type Definer interface {
...
-		keys = append(keys, k)
-	}
-	sort.Strings(keys)
-
-	slots := make([]*SlotSpec, 0, len(h.slots))
-	for _, k := range keys {
-		slots = append(slots, h.slots[k])
-	}
-	return slots
+// Slot returns specification of the slot created by given interface.
```
bboozzoo: `s/created/requested/`

PR #5843, opened 2018-09-19
client, cmd/snap: expose warnings to the world

chipaca:
This introduces two new commands, 'snap warnings', that lists the
things snapd is currently worried about, and 'snap okay', that tells
snapd to chill about the ones it's told you about.

### DIFF
--- a/client/warnings.go
+++ b/client/warnings.go
@@ -1,0 +1,65 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2018 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package client
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/url"
+	"time"
+)
+
+type Warning struct {
+	Message     string     `json:"message"`
+	FirstAdded  time.Time  `json:"first-added"`
+	LastAdded   time.Time  `json:"last-added"`
+	LastShown   *time.Time `json:"last-shown,omitempty"`
+	ExpireAfter string     `json:"expire-after,omitempty"`
+	RepeatAfter string     `json:"repeat-after,omitempty"`
+}
+
+type WarningsOptions struct {
+	All bool
+}
+
+// Warnings are short messages that are meant to alert the user of system events
+func (client *Client) Warnings(opts WarningsOptions) ([]*Warning, error) {
+	var warnings []*Warning
+	q := make(url.Values)
+	if opts.All {
+		q.Add("select", "all")
+	}
+	_, err := client.doSync("GET", "/v2/warnings", q, nil, nil, &warnings)
+	return warnings, err
+}
+
+// Okay asks snapd to chill about the warnings
+func (client *Client) Okay(t time.Time) error {
+	var body bytes.Buffer
+	var op = struct {
+		Action    string    `json:"action"`
+		Timestamp time.Time `json:"timestamp"`
+	}{Action: "okay", Timestamp: t}
+	if err := json.NewEncoder(&body).Encode(op); err != nil {
+		return err
+	}
+	_, err := client.doSync("POST", "/v2/warnings", nil, nil, &body, nil)
+	return err
+}
--- a/cmd/snap/cmd_warnings.go
+++ b/cmd/snap/cmd_warnings.go
@@ -1,0 +1,210 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2018 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package main
+
+import (
+	"encoding/json"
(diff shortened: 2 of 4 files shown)


client/warnings.go
```diff
@@ -0,0 +1,65 @@
...
+		q.Add("select", "all")
+	}
+	_, err := client.doSync("GET", "/v2/warnings", q, nil, nil, &warnings)
+	return warnings, err
+}
+
+// Okay asks snapd to chill about the warnings
+func (client *Client) Okay(t time.Time) error {
+	var body bytes.Buffer
+	var op = struct {
```
mvo5: (nitpick^2) Why not a `type warningsAction struct` or similar? Looks slightly unusual to embed it into the function.

cmd/snap/cmd_warnings.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+const warnFileEnvKey = "SNAPD_LAST_WARNING_TIMESTAMP_FILENAME"
+
+func warnFilename(homedir string) string {
+	if fn := os.Getenv(warnFileEnvKey); fn != "" {
+		return fn
+	}
+
+	return filepath.Join(homedir, ".snap", "warning")
```
mvo5: (nitpick) Should we use plural `warnings` (warning**s**) here?

client/warnings.go
```diff
@@ -0,0 +1,65 @@
...
+	"encoding/json"
+	"net/url"
+	"time"
+)
+
+type Warning struct {
+	Message     string     `json:"message"`
+	FirstAdded  time.Time  `json:"first-added"`
+	LastAdded   time.Time  `json:"last-added"`
+	LastShown   *time.Time `json:"last-shown,omitempty"`
```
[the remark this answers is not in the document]
chipaca: From the sending side if it's not a pointer it'll send `"0001-01-01T00:00:00Z"` which I felt looked weird. But this isn't the sending side, and from the receiving side this isn't needed. I should make it a regular time.Time.

client/warnings.go
```diff
@@ -0,0 +1,65 @@
...
+	ExpireAfter string     `json:"expire-after,omitempty"`
```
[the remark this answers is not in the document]
chipaca: it is a duration. I could make it more structured...

client/warnings.go
```diff
@@ -0,0 +1,65 @@
...
+	q := make(url.Values)
+	if opts.All {
+		q.Add("select", "all")
+	}
+	_, err := client.doSync("GET", "/v2/warnings", q, nil, nil, &warnings)
+	return warnings, err
+}
+
+// Okay asks snapd to chill about the warnings
+func (client *Client) Okay(t time.Time) error {
```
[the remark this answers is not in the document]
chipaca: YES

no.

cmd/snap/cmd_warnings.go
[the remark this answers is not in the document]
chipaca: when we move `~/.snap/`, we need to move this as well. Keeping it together seemed better.

cmd/snap/cmd_warnings.go
```diff
@@ -0,0 +1,210 @@
...
+	uid, gid, err := osutil.UidGid(user)
+	if err != nil {
+		return err
+	}
+
+	aw, err := osutil.NewAtomicFile(warnFilename(user.HomeDir), 0600, 0, uid, gid)
+	if err != nil {
+		return err
+	}
+	// Cancel once Committed is a NOP :-)
```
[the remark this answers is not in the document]
chipaca: I think we agreed to make it `Close`, but I haven't gotten around to it yet.

cmd/snap/cmd_warnings.go
```diff
@@ -0,0 +1,210 @@
...
+	if err := dec.Decode(&d); err != nil {
+		return time.Time{}, fmt.Errorf("cannot decode timestamp file: %v", err)
+	}
+	if dec.More() {
+		return time.Time{}, fmt.Errorf("spurious extra data in timestamp file")
+	}
+	return d.Timestamp, nil
+}
+
+func checkWarnings(count int, timestamp time.Time) {
```
mvo5: (nitpick^2) I wonder if this is the best function name. But I struggle to find a better name: `maybePresentWarningSummary` or something - but its a mouthful. I guess my quibble with checkW... is that on a check function I would expect a bool return. But anyway, nitpick^2 :)

cmd/snap/main.go
```diff
@@ -473,5 +473,7 @@ fixed-width fonts, so it can be hard to tell.
 		fmt.Fprintln(Stderr, msg)
 	}
 
+	checkWarnings(cli.WarningsSummary())
```
mvo5: Would it make sense to simply pass `cli` in here? Or even to make a `client.WarningSummary(w io.Writer)` or similar?

client/warnings.go
```diff
@@ -0,0 +1,65 @@
...
+	var warnings []*Warning
+	q := make(url.Values)
+	if opts.All {
+		q.Add("select", "all")
+	}
+	_, err := client.doSync("GET", "/v2/warnings", q, nil, nil, &warnings)
+	return warnings, err
+}
+
+// Okay asks snapd to chill about the warnings
```
mvo5: I couldn't help noticing that there are no tests for this AFAICT - should be easy to add, right :) ?

cmd/snap/cmd_warnings.go
```diff
@@ -0,0 +1,210 @@
...
+
+func init() {
+	addCommand("warnings", shortWarningsHelp, longWarningsHelp, func() flags.Commander { return &cmdWarnings{} }, timeDescs.also(map[string]string{
+		"all":     i18n.G("Show all warnings"),
+		"verbose": i18n.G("Show more information"),
+	}), nil)
+	addCommand("okay", shortOkayHelp, longOkayHelp, func() flags.Commander { return &cmdOkay{} }, nil, nil)
+}
+
+func (cmd *cmdWarnings) Execute(args []string) error {
```
mvo5: This is also not tested, adding a test would be great. Maybe it will result in some small refactor to extract bits from execute (but maybe not).

client/warnings.go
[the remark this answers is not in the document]
zyga: 🥇 what?

cmd/snap/cmd_warnings.go
```diff
@@ -0,0 +1,210 @@
...
+
+	return aw.Commit()
+}
+
+func lastWarningTimestamp() (time.Time, error) {
+	user, err := osutil.RealUser()
+	if err != nil {
+		return time.Time{}, fmt.Errorf("cannot determine real user: %v", err)
+	}
+	f, err := os.Open(warnFilename(user.HomeDir))
```
bboozzoo: A minimalistic solution would be to use an empty file and just look at the modification timestamp like systemd does with timer files.

cmd/snap/cmd_warnings.go
stolowski: Nitpick #2: we already keep `auth.json` in this directory, would it make sense to use .json extension here as well for consistency?

cmd/snap/cmd_warnings.go
chipaca: it's covered now

PR #5845, opened 2018-09-19
interface: add new `{personal,system}-files`  interface

mvo5:
This adds two new interfaces called "{personal,system}-files". It allows to get access to specific files or dirs in either the system location (starting with "/") or the personal home directory (starting with $HOME). Note that these interfaces require manual approval because they are very powerful.

An application might for example use:
```
name: fluxctl
version: 1.0
plugs:
 system-files:
  read: /etc/kude/config
 personal-files:
  write: [$HOME/.kube/config]
apps:
 app:
  command: fluxctl
  plugs: [system-files,personal-files]
```
Which will allow the fluxctl snap to control the .kube config.

This was requested in:
https://forum.snapcraft.io/t/6948
https://forum.snapcraft.io/t/7062

Note that due to limitations on how things are mapped into the snap namespace
random directories like /random cannot get mapped yet via the system-files interface.


interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,92 @@
...
+# Note, @{HOME} is the user's $HOME, not the snap's $HOME
+`
+
+type dotfilesInterface struct {
+	commonInterface
+}
+
+func (iface *dotfilesInterface) BeforePreparePlug(plug *snap.PlugInfo) error {
+	hasValidAttr := false
+	var sl []interface{}
```
zyga: I think we should validate that the value is an array of strings, otherwise it will accept an array of any kind of object.

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,92 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package builtin
+
+import (
+	"fmt"
+	"github.com/snapcore/snapd/interfaces"
```
bboozzoo: Can you group stdlib imports and our imports separately?

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,92 @@
...
+type dotfilesInterface struct {
+	commonInterface
+}
+
+func (iface *dotfilesInterface) BeforePreparePlug(plug *snap.PlugInfo) error {
+	hasValidAttr := false
+	var sl []interface{}
+	for _, s := range []string{"files", "dirs"} {
+		if err := plug.Attr(s, &sl); err == nil {
+			hasValidAttr = true
```
zyga: Please ensure that there are no dot files in here. Ideally call `filepath.Clean` and refuse unclean paths?

Note that we don't want to de-reference symbolic links here as it would allow an attacker to craft a symlink $HOME/.evil that points to /root/ and would allow us to create an apparmor rule that allows write to /root (for example)

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,92 @@
...
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
+`
+
+const dotfilesConnectedPlugAppArmor = `
+# Description: Can access hidden files in user's $HOME. This is restricted
+# because it gives file access to some of the user's $HOME.
+
+# Note, @{HOME} is the user's $HOME, not the snap's $HOME
```
bboozzoo: I think you can drop `Note` prefix

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (iface *dotfilesInterface) AppArmorConnectedPlug(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	var files, dirs []interface{}
+	_ = plug.Attr("files", &files)
+	_ = plug.Attr("dirs", &dirs)
+
+	spec.AddSnippet(dotfilesConnectedPlugAppArmor)
+	for _, file := range files {
+		s := fmt.Sprintf("owner @${HOME}/%s rwklix,", file)
```
bboozzoo: @zyga will probably leave a comment here

zyga: I would expect `rwkl` alone, not sure what `ix' is needed for.

CC @jdstrand - do you suppose we should allow executing programs in dot directories?

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,92 @@
...
+
+type dotfilesInterface struct {
+	commonInterface
+}
+
+func (iface *dotfilesInterface) BeforePreparePlug(plug *snap.PlugInfo) error {
+	hasValidAttr := false
+	var sl []interface{}
+	for _, s := range []string{"files", "dirs"} {
+		if err := plug.Attr(s, &sl); err == nil {
```
bboozzoo: It's missing some path validation. The path should be relative, without `..` or such. Also it should be AARE safe. I recall there's some code for this in wrapers/desktop.go, probably worth checking out.

bboozzoo: AARE checks are here: https://github.com/snapcore/snapd/blob/master/interfaces/builtin/daemon_notify.go#L67..L72

interfaces/builtin/daemon_notify.go
```diff
@@ -64,7 +64,7 @@ func (iface *daemoNotifyInterface) AppArmorConnectedPlug(spec *apparmor.Specific
 		// must be an absolute path or an abstract socket path
 		return fmt.Errorf("cannot use %q as notify socket path: not absolute", notifySocket)
 	}
-	illegalChars := `?*[]{}^"`
+	illegalChars := apparmorAARE
```
zyga: Thank you for doing this, how about moving those to `apparmor.AARE` (as a const), I believe we have those in one other place as well.

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,125 @@
...
+		np, ok := npp.(string)
+		if !ok {
+			return fmt.Errorf("%q must be a list of strings", attrName)
+		}
+		p := filepath.Clean(np)
+		if p != np {
+			return fmt.Errorf("%q must be clean", np)
+		}
+
+		if strings.Contains(p, "..") {
```
zyga: I believe you can discard this verification, Clean handles that automatically.

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,125 @@
...
+}
+
+func validatePaths(attrName string, paths []interface{}) error {
+	for _, npp := range paths {
+		np, ok := npp.(string)
+		if !ok {
+			return fmt.Errorf("%q must be a list of strings", attrName)
+		}
+		p := filepath.Clean(np)
+		if p != np {
```
zyga: Ha, I knew p != np ;-)

Joking aside, I think that Clean has one subtle effect, it will reject trailing slash. The trailing slash is required to express directory (as opposed to a file) in apparmor. This is not a problem in the way you use the values in the apparmor specification function below but it may be confusing for users that they cannot say:
```
dirs:
 - .foo/
```

interfaces/builtin/dotfiles_test.go
```diff
@@ -0,0 +1,203 @@
...
+	c.Assert(interfaces.BeforePreparePlug(s.iface, s.plugInfo), IsNil)
+}
+
+func (s *dotfilesInterfaceSuite) TestSanitizePlugHappy(c *C) {
+	const mockSnapYaml = `name: dotfiles-plug-snap
+version: 1.0
+plugs:
+ dotfiles:
+  files: [".file1"]
+  dirs: [".dir1"]
```
zyga: Can you please add a test that ensures that the path indeed starts with the dot character (and that it is not a trick like ./foo? I believe this is broken today.

interfaces/builtin/utils.go
```diff
@@ -31,6 +31,8 @@ import (
 // The maximum number of Usb bInterfaceNumber.
 const UsbMaxInterfaces = 32
 
+const apparmorAARE = `?*[]{}^"`
```
zyga: Those could go to apparmor/apparmor.go

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,125 @@
...
+		if strings.Contains(p, "..") {
+			return fmt.Errorf(`%q contains invalid ".."`, p)
+		}
+		illegalChars := apparmorAARE
+		if strings.ContainsAny(p, illegalChars) {
+			// must not contain any AppArmor regular expression (AARE)
+			// characters or double quotes
+			return fmt.Errorf("%q contains one of %s", p, illegalChars)
+		}
+
```
zyga: Please add a check here that ensures that the first component of the path contains at least two characters and that the first of those is a dot.

interfaces/builtin/daemon_notify.go
mvo5: I moved this into apparmor as a helper now.

interfaces/builtin/dotfiles.go
mvo5: It does to some extend: https://play.golang.org/p/DTs3K45CzlK - i.e. there may still be leading ".."

interfaces/builtin/dotfiles.go
mvo5: I wonder if we can re-purpose this interface as a general "access-stuff-in-home" interface. Given that we will only give this out after a manual review it seems like its ok to allow something like "access my ~/public_html" dir for a hypothetical homepage generator.

interfaces/builtin/dotfiles.go
jdstrand: I think we can do something like this (see below about discussion of '$HOME'):
```
if !strings.HasPrefix(np, "/") && !strings.HasPrefix(np, "$HOME/") {
    return fmt.Errof("%q must start with '/' or '$HOME'", np)
}

p := filepath.Clean(np)
if p != np {
    return fmt.Errorf("%q must be clean", np)
}
```

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,132 @@
...
+		if p != np {
+			return fmt.Errorf("%q must be clean", np)
+		}
+
+		if strings.Contains(p, "..") {
+			return fmt.Errorf(`%q contains invalid ".."`, p)
+		}
+		if strings.Contains(p, "~") {
+			return fmt.Errorf(`%q contains invalid "~"`, p)
+		}
```
jdstrand: Were you concerned about tilde expansion here? If so, I suggest dropping this since my suggested change would handle this ok.

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func (iface *dotfilesInterface) AppArmorConnectedPlug(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	var files, dirs []interface{}
+	_ = plug.Attr("files", &files)
+	_ = plug.Attr("dirs", &dirs)
+
+	spec.AddSnippet(dotfilesConnectedPlugAppArmor)
+	for _, file := range files {
+		s := fmt.Sprintf("owner @${HOME}/%s rwklix,", filepath.Clean(file.(string)))
```
jdstrand: I think we want either 'rkl' or or 'rwkl', but not not 'ix'. Perhaps at some future date we will want to add 'ix', but 'ix' could cause issues with 'conflicting x modifiers' that might impact combinations of interface connections (I'm going through this now with 'ix' in the home interface and a change I want to make to the classic template and docker-support interface, and it is a headache).

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,132 @@
...
+
+	"github.com/snapcore/snapd/interfaces"
+	"github.com/snapcore/snapd/interfaces/apparmor"
+	"github.com/snapcore/snapd/snap"
+)
+
+const dotfilesSummary = `allows access to specific hidden files in the home directory`
+
+const dotfilesBaseDeclarationSlots = `
+  dotfiles:
```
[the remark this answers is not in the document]
zyga: I was thinking about `user-data-access`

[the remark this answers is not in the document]
jdstrand: @zyga - "user-data-access" doesn't quite fit the "/etc" case (deb transition).

interfaces/builtin/dotfiles.go
jdstrand: Note that my suggestion of not exposing this AppArmor subtlety. Just have them express a path and we normalize it and create a rule that allows it as both a file and a directory.

zyga: I think a simplification where we allow both is a good idea.

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,132 @@
...
+		return fmt.Errorf(`cannot add dotfiles plug without valid "files" or "dirs" attribute`)
+	}
+
+	return nil
+}
+
+func (iface *dotfilesInterface) AppArmorConnectedPlug(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	var files, dirs []interface{}
+	_ = plug.Attr("files", &files)
+	_ = plug.Attr("dirs", &dirs)
```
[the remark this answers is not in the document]
mvo5: I implemented this using the trailing "/" to separate between files and dirs for now. But happy to change if you think its too magic.

interfaces/builtin/dotfiles.go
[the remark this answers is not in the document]
diddledan: just throwing ideas about, feel free to disregard as a looney blowing a gasket :-)

something with the word "config" or "configuration" in it mayhaps?
"config-control"?
"config-observe"?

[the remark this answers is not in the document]
jdstrand: @diddledan - it is potentially more than config though. It could be anything (to support deb migration).

mvo5:
Adding @niemeyer mostly to get a good name :)

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func formatPath(ip interface{}) (string, error) {
+	p, ok := ip.(string)
+	if !ok {
+		return "", fmt.Errorf("%[1]v (%[1]T) is not a string", ip)
+	}
+	prefix := ""
+	if strings.Count(p, "$HOME") > 0 {
+		p = strings.Replace(p, "$HOME", "@${HOME}", 1)
```
jdstrand: It should be: `p = strings.Replace(p, "$HOME", "@{HOME}", 1)`

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+func (iface *dotfilesInterface) AppArmorConnectedPlug(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	var reads, writes []interface{}
+	_ = plug.Attr("read", &reads)
+	_ = plug.Attr("write", &writes)
+
+	errPrefix := fmt.Sprintf(`cannot connect plug %s: `, plug.Name())
+	spec.AddSnippet(dotfilesConnectedPlugAppArmor)
+	if err := addSnippet(spec, "rkl,", reads); err != nil {
```
jdstrand: Not 'l' with 'read'.

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,158 @@
...
+	p, ok := ip.(string)
+	if !ok {
+		return "", fmt.Errorf("%[1]v (%[1]T) is not a string", ip)
+	}
+	prefix := ""
+	if strings.Count(p, "$HOME") > 0 {
+		p = strings.Replace(p, "$HOME", "@${HOME}", 1)
+		prefix = "owner "
+	}
+	return prefix + filepath.Clean(p), nil
```
jdstrand: We are going to want to end up with: `owner "@{HOME}/foo" rk,`, but AFAICS, this will end up with `owner @{HOME}/foo rk,` (ie, the quotes are dropped).

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,158 @@
...
+	for _, rp := range paths {
+		p, err := formatPath(rp)
+		if err != nil {
+			return err
+		}
+		if strings.HasSuffix(p, "/") {
+			spec.AddSnippet(fmt.Sprintf("%s/** %s", p, perm))
+		} else {
+			spec.AddSnippet(fmt.Sprintf("%s %s", p, perm))
+		}
```
jdstrand: I continue to prefer *not* exposing the trailing-/ apparmorism to snap developers and instead recommend they simply specify a path (we can even warn/error if they try to use the apparmor-ism) and we allow access to the path as a file or directory. Ie, this becomes:
```
spec.AddSnippet(fmt.Sprintf("%s{,/,/**} %s", p, perm))
```

interfaces/builtin/dotfiles.go
mvo5: I added quotes around all user supplied paths now.

interfaces/builtin/dotfiles.go
mvo5: Thanks, I updated this now.

interfaces/apparmor/apparmor_test.go
```diff
@@ -219,3 +219,11 @@ func (s *appArmorSuite) TestLoadedApparmorProfilesHandlesParsingErrors(c *C) {
 	c.Assert(err, ErrorMatches, `syntax error, expected: name \(mode\)`)
 	c.Check(profiles, IsNil)
 }
+
+func (s *appArmorSuite) TestValidateFreeFromAAREUnhappy(c *C) {
+	var testCases = []string{"a?", "*b", "c[c", "dd]", "e{", "f}", "g^", `h"`, "f\000", "g\x00"}
+
+	for _, s := range testCases {
+		c.Check(apparmor.ValidateFreeFromAARE(s), ErrorMatches, ".* contains a reserved apparmor char from .*", Commentf("%q is not raising an error", s))
+	}
```
zyga: Nitpick, it would be great to show that "foo" doesn't contain any AARE and passes the validation.

interfaces/builtin/daemon_notify.go
```diff
@@ -64,11 +64,8 @@ func (iface *daemoNotifyInterface) AppArmorConnectedPlug(spec *apparmor.Specific
 		// must be an absolute path or an abstract socket path
 		return fmt.Errorf("cannot use %q as notify socket path: not absolute", notifySocket)
 	}
-	illegalChars := `?*[]{}^"`
-	if strings.ContainsAny(notifySocket, illegalChars) {
-		// must not contain any AppArmor regular expression (AARE)
-		// characters or double quotes
-		return fmt.Errorf("cannot use %q as notify socket path: contains one of %s", notifySocket, illegalChars)
+	if err := apparmor.ValidateFreeFromAARE(notifySocket); err != nil {
+		return fmt.Errorf("cannot use %q as notify socket path: %s", notifySocket, err)
```
zyga: This is fine

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+const dotfilesSummary = `allows access to specific hidden files in the home directory`
+
+const dotfilesBaseDeclarationSlots = `
+  dotfiles:
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
```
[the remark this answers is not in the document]
zyga: @jdstrand @mvo5 @niemeyer perhaps you can agree to this while physically together so this can move on. I agree with Jamie that we ought to be very careful here.

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,153 @@
...
+	return nil
+}
+
+func (iface *dotfilesInterface) AppArmorConnectedPlug(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	var reads, writes []interface{}
+	_ = plug.Attr("read", &reads)
+	_ = plug.Attr("write", &writes)
+
+	errPrefix := fmt.Sprintf(`cannot connect plug %s: `, plug.Name())
+	spec.AddSnippet(dotfilesConnectedPlugAppArmor)
```
zyga: Nitpick, can you please remove `dotfilesConnectedPlugAppArmor`, format the whole chunk into a buffer and then add it as one snippet. We sort snippets so without the change the comment will be detached from the corresponding permissions it grants.

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,153 @@
...
+	if err := addSnippet(spec, "rk,", reads); err != nil {
+		return fmt.Errorf(errPrefix+"%v", err)
```
zyga: Nitpick, errPrefix does not contain any meta-characters interpreted by the standard library formatting system but perhaps we ought to be double plus careful and instead do the more correct: `return fmt.Errorf("%s%v", errPrefix, err)`?

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,153 @@
...
+	if strings.Count(p, "$HOME") > 0 {
+		p = strings.Replace(p, "$HOME", "@{HOME}", 1)
+		prefix = "owner "
+	}
+	p += "{,/,/**}"
+
+	return fmt.Sprintf("%s%q", prefix, filepath.Clean(p)), nil
+}
+
+func addSnippet(spec *apparmor.Specification, perm string, paths []interface{}) error {
```
zyga: Perhaps rename this to `allowPathAccess`?

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,153 @@
...
+	for _, rp := range paths {
+		p, err := formatPath(rp)
+		if err != nil {
+			return err
+		}
+		spec.AddSnippet(fmt.Sprintf("%s %s", p, perm))
```
zyga: This is a bit tricky, see my other comment on the main apparmor specification method. There is nothing here that guarantees that perm ends with `,` and thus forms a correct apparmor syntax. I would love to see this refactored slightly so that it is obvious the added snippet is correct.

dholbach:
Is anything missing apart from a name now? If yes, it'd be great to land this soon!

@mvo5 mentioned `file-access` as an idea to me, which I feel is more generic than `dot-files`, but still clear enough. What developers ask for is pretty much access to files. So :+1: from me.

Oh and hello everyone! :grinning:

oSoMoN:
Hello @dholbach !
I'm also eagerly looking forward to having this interface available, to allow transitioning the chromium deb package to a snap (need access to `$HOME/.config/chromium`).
How can I test this end-to-end?
Regarding naming, "files" (suggested by @jdstrand and acked by @mvo5) sounds good to me. @niemeyer can you chime in?

mvo5:
I have some ideas about the name: `files`, `filesystem`, `file-access`, `arbitrary-file-access`, `controlled-file-access` - but open for ideas.

mvo5:
The opinion of @pedronis about the name would also be very welcome :)

jdstrand:
I prefer 'files'. IMHO, '-access' in 'file-access' is redundant. I don't care for the 'arbitrary-' and 'controlled-' prefixes since the accesses aren't arbitrary and controlled seems redundant.

@niemeyer - can you also weigh in?

alexmurray:
This looks really neat - re name for this interface - I like the simplicity of `files` but I wonder if instead it is better to name it such that it is clear this is about providing access to files from outside of the snap environment - hence how about about `host-files` / `core-files` / `external-files` or something?

jdstrand:
I'm not opposed to adding something extra to 'files', but IMHO `core-files` isn't quite right because it implies something about the core snap which isn't quite right. `external-files` and `host-files` do convey something about the intent of the interface-- ie, to pull something in from somewhere else, but for some reason don't feel quite right. `extra-files`? I think I still prefer `files`...

jdstrand:
I'm not opposed to adding something extra to 'files', but IMHO `core-files` isn't quite right because it implies something about the core snap which isn't quite right. `external-files` and `host-files` do convey something about the intent of the interface-- ie, to pull something in from somewhere else, but for some reason don't feel quite right. `extra-files`? I think I still prefer `files`...

alexmurray:
`system-files` ?

interfaces/apparmor/apparmor.go
```diff
@@ -35,6 +35,17 @@ import (
 	"github.com/snapcore/snapd/osutil"
 )
 
+// ValidateFreeFromAARE will check that the given string does not
+// contain AppArmor regular expressions (AARE) or double quotes
+func ValidateFreeFromAARE(s string) error {
```
niemeyer: I suggest _ValidateNoAppArmorRegexp_ instead, as something that won't require readers to come looking for the implementation to understand the code in the call site.

interfaces/builtin/dotfiles.go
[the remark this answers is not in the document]
pedronis: The issue I think is that a user will not always be able to  make a good decision whether connecting files is appropriate or not, given that is very generic and abstract, and we don't have mechanisms atm to see/inform easily what attribute values (paths here) are involved afaik.

This is an area where prompting could help, OTOH here paths are hard-coded so it seems appropriate for the snap-declaration to control this.

So in summary I think it would be appropriate to start very carefully and actually block installation of snap using "files". Which would then make our task to provide appropriate snap-declarations and rules on a case-by-case basis.

@jdstrand any further thoughts on this?

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,159 @@
...
+		return fmt.Errorf("%s%v", errPrefix, err)
+	}
+	spec.AddSnippet(buf.String())
+
+	return nil
+}
+
+func init() {
+	registerIface(&dotfilesInterface{commonInterface{
+		name:                 "dotfiles",
```
niemeyer: The suggested _files_ name feels a bit awkward in the sense the snap definitely has access to files.. many files. :)

I suggest naming the interface as _sensitive-files._ This remains general, while also giving the feeling that is restricted for good reasons.

interfaces/builtin/dotfiles.go
```diff
@@ -0,0 +1,159 @@
...
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/snapcore/snapd/interfaces"
+	"github.com/snapcore/snapd/interfaces/apparmor"
+	"github.com/snapcore/snapd/snap"
+)
+
+const dotfilesSummary = `allows access to files or directories`
```
niemeyer: files => sensitive files

interfaces/builtin/dotfiles.go
pedronis: But is unclear that the file are "sensitive", it depends,  they are apriori inaccessible because we start from a conservative position. I think the trouble we are having with naming this is that is an expedient interface but problematic, it's hard for the name to capture what it allows because in this case it really fully depends on the attributes on it. We need to convey something in the range of "specific" to "otherwise-inaccessible" to indeed "sensitive".

pedronis: So thinking some more:

* "raw-files" or "bare-files",  it undescores that this give *direct* access to files *without any mediation or predefined purpose* (and that would not be accessible otherwise)
* "fenced-files" or something similar (fenced seemed the shortest word with this possible meaning), it underscores that the files would *not* be apriori *accessible*
* "sensitive-files",  it underscores that the file might be *sensitive* (that might be misleading in general) and were not previously accessible
* "designated-files" or "specific-files",  it would underscore mainly that this lets specify precise files/dirs

jdstrand: Yeah, @pedronis is right. The interface will allow access to anything the developer chooses-- it's our policies around it and what the developer chooses that make it sensitive or not. Eg, ~/.config/foo might be sensitive, but /etc/foo might not. I don't hate "sensitive-files" but it doesn't seem quite right either.

I don't really like "raw-files" or "bare-files" for the same reason @niemeyer doesn't like "files"-- they are just files and no more raw or bare than files already allowed by the existing interface policy. "fenced-files" seems weird. "designated-files" and "specific-files" seems redundant and don't hit my ear quite right.

What about "extra-files"? It could be argued as redundant too, but at least it conveys that these are files separate from what would normally be allowed.

interfaces/builtin/dotfiles.go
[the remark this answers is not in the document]
jdstrand: @pedrons - +1 for installation constraint. When @niemeyer and I initially discussed it, we definitely wanted to carefully guard this. It isn't meant to be a get out of jail free card (&lt;/rimshot&gt;) but something where snap declarations carefully control what additional accesses are needed *and* humans have a conversation about the paths and if they should be writable. If we start more restricted, we can always open it up later if we want; it is much harder going the other way.

jdstrand:
@alexmurray - depending on how you think about it, "system-files" doesn't work quite right for the same reason "dotfiles" doesn't. /etc/foo is a "system file" to me, while ~/.foo is a dot file. This interface supports both. That said, "system-files" could be interpreted as "files from the system, not the mount namespace", but that seems like a weird thing to describe....

pedronis:
> @alexmurray - depending on how you think about it, "system-files" doesn't work quite right for the same reason "dotfiles" doesn't. /etc/foo is a "system file" to me, while ~/.foo is a dot file. This interface supports both. That said, "system-files" could be interpreted as "files from the system, not the mount namespace", but that seems like a weird thing to describe....

also afaiu this doesn't really dinstinguish or does anything special about namespace vs host files, it gives access to file however they are mounted (or not mounted)

interfaces/builtin/dotfiles.go
pedronis: Stepping back a bit, we have "content",  which we didn't call "shared-content" or "protected-content" or "private-content",

ATM I could live with "files" (tough generic/vacuous as it might be) or "misc-files"  (as in miscellaneous) which is similar to "extra-files"  but "extra" has a nuance of "(from) another place/set" somehow for me that seems confusing here.

mvo5:
We just discussed this and we will use the two names: `system-files` and `personal-files` for this.

mvo5:
I updated this PR now to use {system,personal}-files - however it does not validate based on that. When we discussed this in a hangout it was unclear if this is needed at this level because each of these interfaces need to be approved manually with a matching snap declaration so we can perform the checks at this level. I can also write validation code if desired.

mvo5:
We should probably also add a spread test but I will need a matching snap declaration (maybe @jdstrand and I can work on this on Monday)

interfaces/builtin/system_files.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+const systemFilesSummary = `allows access to system files or directories`
+
+const systemFilesBaseDeclarationSlots = `
+  system-files:
+    allow-installation:
```
pedronis: the plan is still to have allow-installation: false here, cc @jdstrand

interfaces/builtin/personal_files.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+const personalFilesSummary = `allows access to personal files or directories`
+
+const personalFilesBaseDeclarationSlots = `
+  personal-files:
+    allow-installation:
```
pedronis: the plan is still to have allow-installation: false here, cc @jdstrand

interfaces/builtin/system_files.go
```diff
@@ -0,0 +1,71 @@
...
+
+type systemFilesInterface struct {
+	commonFilesInterface
+}
+
+func validateSinglePathSystem(np string) error {
+	if !strings.HasPrefix(np, "/") {
+		return fmt.Errorf(`%q must start with "/"`, np)
+	}
+	if strings.Contains(np, "$HOME") {
```
sergiocazzolato: I think here we should check $HOME is not a prefix, right?

interfaces/builtin/system_files.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+	"strings"
+)
+
+const systemFilesSummary = `allows access to system files or directories`
+
+const systemFilesBaseDeclarationSlots = `
+  system-files:
+    allow-installation: false
+    deny-connection: true
```
sergiocazzolato: I see this error when running the spread tests.

+ snap install --edge core
- Mount snap "core" (5940) (installation not allowed by "system-files" slot rule of interface "system-files")

sergiocazzolato: Also, I see my test snap that it has not the system/personal files interface (I updated the interfaces just to be able to test them)
https://paste.ubuntu.com/p/WxJQY2Y7kB/

mvo5:
@sergiocazzolato Thanks for looking into this. I fixed the incorrect base declaration now, hopefully travis is more happy now.

interfaces/builtin/system_files.go
pedronis: ah, sorry that was meant for plugs

interfaces/builtin/personal_files.go
jdstrand: What we want is for personalFilesBaseDeclarationPlugs to have 'allow-installation: false' but personalFilesBaseDeclarationSlots is scoped to slot-snap-type: core. As such, this LGTM.

interfaces/builtin/personal_files.go
```diff
@@ -0,0 +1,79 @@
...
+    allow-installation: false
+    deny-auto-connection: true
+`
+
+const personalFilesBaseDeclarationSlots = `
+  personal-files:
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-connection: true
```
jdstrand: Our normal pattern for super-privileged interfaces (eg, docket-support, kernel-module-control, etc) doesn't use `deny-connection: true` on the slots part of the based declaration.

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,141 @@
...
+		return fmt.Errorf("%q must be clean", np)
+	}
+	if strings.Contains(p, "~") {
+		return fmt.Errorf(`%q contains invalid "~"`, p)
+	}
+	if err := apparmor.ValidateNoAppArmorRegexp(p); err != nil {
+		return err
+	}
+
+	if err := iface.extraPathValidate(np); err != nil {
```
jdstrand: Maybe add a comment that this must be implemented elsewhere.

interfaces/builtin/system_files.go
jdstrand: The typical pattern is to not have `deny-connection: true` (see above).

interfaces/builtin/system_files.go
jdstrand: Yes, for plugs, which has it so that is good.

interfaces/builtin/system_files.go
jdstrand: For system-files, we don't want $HOME at all, so the simple Contains check here is correct.

interfaces/builtin/system_files_test.go
```diff
@@ -0,0 +1,181 @@
...
+		{`read: [ 123 ]`, `"read" must be a list of strings`},
+		{`read: [ "/foo/./bar" ]`, `"/foo/./bar" must be clean`},
+		{`read: [ "../foo" ]`, `"../foo" must start with "/"`},
+		{`read: [ "/foo[" ]`, `"/foo\[" contains a reserved apparmor char from .*`},
+		{`write: ""`, `"write" must be a list of strings`},
+		{`write: bar`, `"write" must be a list of strings`},
+		{`read: [ "~/foo" ]`, `"~/foo" contains invalid "~"`},
+		{`read: [ "/foo/~/foo" ]`, `"/foo/~/foo" contains invalid "~"`},
+		{`read: [ "/foo/../foo" ]`, `"/foo/../foo" must be clean`},
+		{`read: [ "/home/$HOME/foo" ]`, `\$HOME cannot be used in "/home/\$HOME/foo"`},
```
jdstrand: This is an interesting test case. We could disallow '/home' as a prefix, but we would always be chasing configurations where the user home directories are somewhere else. I think you are right to not second guess the snap and let it be a matter of snap declaration store policies on what we allow.

tests/main/interfaces-personal-files/task.yaml
```diff
@@ -0,0 +1,96 @@
...
+    ensure_file_exists_backup_real /root/.testdir1/.testfilé2
+    ensure_file_exists_backup_real "/root/.testdir1/test file2"
+    ensure_dir_exists_backup_real /root/.testdir1/.testdir2
+
+    # Not accessible dirs and files
+    ensure_dir_exists_backup_real /tmp/.testdir1
+    ensure_file_exists_backup_real /tmp/.testfile1
+    ensure_dir_exists_backup_real /tmp/root/
+    ensure_file_exists_backup_real /tmp/root/testfile2
+    ensure_file_exists_backup_real /tmp/root/.testfile2
```
jdstrand: Why are these entries in /tmp dirs not accessible? The default policy should allow these due to the private /tmp for the snap.... Perhaps /var/tmp would be a better test?

tests/main/interfaces-personal-files/task.yaml
```diff
@@ -0,0 +1,96 @@
...
+        exit 1
+    fi
+    MATCH "Permission denied" < call.error
+
+    echo "Then the snap is not able to to access files and dirs outside $HOME"
+    test-snapd-sh.with-personal-files-plug -c "ls /tmp/.testdir1" && exit 1
+    test-snapd-sh.with-personal-files-plug -c "cat /tmp/.testfile1" && exit 1
+    test-snapd-sh.with-personal-files-plug -c "ls /tmp/root/" && exit 1
+    test-snapd-sh.with-personal-files-plug -c "cat /tmp/root/testfile2" && exit 1
+    test-snapd-sh.with-personal-files-plug -c "cat /tmp/root/.testfile2" && exit 1
```
jdstrand: These should be accessible. I suspect what is happening is that you are getting ENOENT errors because while you call ensure_file_exists_backup_real, that is within the context of the global mount namespace, and not the snap namespace. If you use /var/tmp instead, it would address this (or really, `/root/.not-allowed/blah`). There is a missing `MATCH "Permission denied" < call.error` to prove that it was a permission denied and not some other error.

tests/main/interfaces-system-files/task.yaml
```diff
@@ -0,0 +1,91 @@
...
+    # Fist layer of dirs and files
+    ensure_file_exists_backup_real /tmp/.testfile1
+    ensure_file_exists_backup_real /tmp/testfile1
+    ensure_dir_exists_backup_real /tmp/.testdir1
+    ensure_dir_exists_backup_real /tmp/testdir1
+
+    # Second layer of dirs and files
+    ensure_file_exists_backup_real /tmp/.testdir1/.testfilé2
+    ensure_file_exists_backup_real "/tmp/.testdir1/test file2"
+    ensure_dir_exists_backup_real /tmp/root
```
jdstrand: Similarly, these are not good test cases since /tmp should be allowed (see above).

tests/main/interfaces-system-files/task.yaml
```diff
@@ -0,0 +1,91 @@
...
+    test-snapd-sh.with-personal-files-plug -c "touch /tmp/.testfile1/testfilé2"
+    if test-snapd-sh.with-personal-files-plug -c "echo test >> /tmp/testfile1" 2> call.error; then
+        echo "Expected permission error writing the system file"
+        exit 1
+    fi
+    MATCH "Permission denied" < call.error
+
+    echo "Then the snap is not able to to access files and dirs in $HOME"
+    test-snapd-sh.with-system-files-plug -c "ls /root/.testdir1" && exit 1
+    test-snapd-sh.with-system-files-plug -c "cat /root/.testfile1" && exit 1
```
jdstrand: Shouldn't we have a MATCH here too?

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,147 @@
...
+	return nil
+}
+
+func (iface *commonFilesInterface) validateSinglePath(np string) error {
+	if strings.HasSuffix(np, "/") {
+		return fmt.Errorf(`%q cannot end with "/"`, np)
+	}
+	if strings.Contains(np, "@{") {
+		return fmt.Errorf(`%q should not use "@{"`, np)
+	}
```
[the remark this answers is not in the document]
mvo5: Indeed, I removed this particular check and updated the tests (the bits in the test are still invalid the error is just different).

tests/main/interfaces-personal-files/task.yaml
mvo5: I updated the test to use /var/tmp

tests/main/interfaces-system-files/task.yaml
mvo5: Good point, added that to the tests.

tests/main/interfaces-system-files/task.yaml
```diff
@@ -0,0 +1,94 @@
...
+
+    echo "Then the snap is able to write just $TESTDIR/.testdir1 and $TESTDIR/.testfile1"
+    test-snapd-sh.with-system-files-plug -c "echo test >> $TESTDIR/.testfile1"
+    test-snapd-sh.with-system-files-plug -c "touch $TESTDIR/.testdir1/testfilé2"
+
+    if [ "$(snap debug confinement)" = partial ] ; then
+        exit 0
+    fi
+
+    if test-snapd-sh.with-system-files-plug -c "echo test >> $TESTDIR/testfile1" 2> call.error; then
```
jdstrand: It made it a little hard to verify that this was correct with it named as 'testfile1'. Naming it 'deniedwrite1' or something might be easier. Not a blocker.

tests/main/interfaces-system-files/task.yaml
jdstrand: This is where I though the MATCH should be to notice EACCES vs ENOENT.

tests/main/interfaces-personal-files/task.yaml
```diff
@@ -0,0 +1,98 @@
...
+        exit 1
+    fi
+    MATCH "Permission denied" < call.error
+
+    echo "Then the snap is not able to to access files and dirs outside $HOME"
+    test-snapd-sh.with-personal-files-plug -c "ls /var/tmp/.testdir1" && exit 1
+    test-snapd-sh.with-personal-files-plug -c "cat /var/tmp/.testfile1" && exit 1
+    test-snapd-sh.with-personal-files-plug -c "ls /var/tmp/root/" && exit 1
+    test-snapd-sh.with-personal-files-plug -c "cat /var/tmp/root/testfile2" && exit 1
+    test-snapd-sh.with-personal-files-plug -c "cat /var/tmp/root/.testfile2" && exit 1
```
jdstrand: This is another place where I though the MATCH should be to notice EACCES vs ENOENT.

interfaces/apparmor/apparmor.go
```diff
@@ -35,6 +35,17 @@ import (
 	"github.com/snapcore/snapd/osutil"
 )
 
+// ValidateFreeFromAARE will check that the given string does not
+// contain AppArmor regular expressions (AARE) or double quotes
+func ValidateNoAppArmorRegexp(s string) error {
```
zyga: The name and the comment don't match.

interfaces/apparmor/apparmor.go
```diff
@@ -35,6 +35,17 @@ import (
...
+	const AARE = `?*[]{}^"` + "\x00"
+
+	if strings.ContainsAny(s, AARE) {
+		return fmt.Errorf("%q contains a reserved apparmor char from %s ", s, AARE)
```
zyga: Is the trailing space relevant?

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,162 @@
...
+
+func formatPath(ip interface{}) (string, error) {
+	p, ok := ip.(string)
+	if !ok {
+		return "", fmt.Errorf("%[1]v (%[1]T) is not a string", ip)
+	}
+	prefix := ""
+	if strings.HasPrefix(p, "$HOME") && strings.Count(p, "$HOME") == 1 {
+		p = strings.Replace(p, "$HOME", "@{HOME}", 1)
+		prefix = "owner "
```
zyga: We had some issues with `owner` in the past. I wonder if this works with encrypted `$HOME`. While a spread test for it would be great I think that giving it a casual quick test would be useful.

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,162 @@
...
+	panic(fmt.Sprintf("invalid perm: %d", a))
+}
+
+func formatPath(ip interface{}) (string, error) {
+	p, ok := ip.(string)
+	if !ok {
+		return "", fmt.Errorf("%[1]v (%[1]T) is not a string", ip)
+	}
+	prefix := ""
+	if strings.HasPrefix(p, "$HOME") && strings.Count(p, "$HOME") == 1 {
```
zyga: Is there a validation phase that checks that `p` has at most one `$HOME` and errors if it does not? I'm asking because for a string like `"$HOME/sweet/$HOME"` we will just pass that unexpanded to apparmor which is, at least, confusing.

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,162 @@
...
+
+const (
+	filesRead filesAAPerm = iota
+	filesWrite
+)
+
+func (a filesAAPerm) String() string {
+	switch a {
+	case filesRead:
+		return "rk,"
```
zyga: I would prefer if the comma was not a part of this value and instead be placed at the point of expansion.

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,162 @@
...
+		p, err := formatPath(rawPath)
+		if err != nil {
+			return err
+		}
+		fmt.Fprintf(buf, "%s %s\n", p, perm)
+	}
+	return nil
+}
+
+func (iface *commonFilesInterface) validatePaths(attrName string, paths []interface{}) error {
```
zyga: Not a strong opinion or something to change now but I think that having a set of things like validate one path, validate a list of paths, etc, would be useful as plain functions. I suspect that more than one interface needs to have code like this.

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,162 @@
...
+func (iface *commonFilesInterface) validateSinglePath(np string) error {
+	if strings.HasSuffix(np, "/") {
+		return fmt.Errorf(`%q cannot end with "/"`, np)
+	}
+	p := filepath.Clean(np)
+	if p != np {
+		return fmt.Errorf("%q must be clean", np)
+	}
+	if strings.Contains(p, "~") {
+		return fmt.Errorf(`%q contains invalid "~"`, p)
```
zyga: Nitpick but for consistency this could be `%q cannot contain "~"` like above.

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,162 @@
...
+	return nil
+}
+
+func (iface *commonFilesInterface) validateSinglePath(np string) error {
+	if strings.HasSuffix(np, "/") {
+		return fmt.Errorf(`%q cannot end with "/"`, np)
+	}
+	p := filepath.Clean(np)
+	if p != np {
+		return fmt.Errorf("%q must be clean", np)
```
zyga: The concept of path cleanness is not something I encountered before I started working with go. I don't think it will make a useful error message. How about `"%q cannot be used, try %q", where the 2nd %q is the cleaned value.

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,162 @@
...
+		return fmt.Errorf(`%q contains invalid "~"`, p)
+	}
+	if err := apparmor.ValidateNoAppArmorRegexp(p); err != nil {
+		return err
+	}
+
+	// extraPathValidation must be implemented by the interface
+	// that build on top of the abstract commonFilesInterface
+	if iface.extraPathValidate == nil {
+		panic("extraPathValidate must be set when using the commonFilesInterface")
```
zyga: Eh :) I wish we had a type system that could capture this.

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,162 @@
...
+	return nil
+}
+
+func (iface *commonFilesInterface) BeforePreparePlug(plug *snap.PlugInfo) error {
+	hasValidAttr := false
+	for _, att := range []string{"read", "write"} {
+		if _, ok := plug.Attrs[att]; !ok {
+			continue
+		}
+		il, ok := plug.Attrs[att].([]interface{})
```
zyga: I find `il` a bit unintelligible. Can this simply be called `paths`?

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,162 @@
...
+		if !ok {
+			return fmt.Errorf("cannot add %s plug: %q must be a list of strings", iface.name, att)
```
zyga: Should the string "cannot add foo plug:" be a part of an error message one level above this?

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+		if err := iface.validatePaths(att, il); err != nil {
+			return fmt.Errorf("cannot add %s plug: %s", iface.name, err)
+		}
+		hasValidAttr = true
+	}
+	if !hasValidAttr {
+		return fmt.Errorf(`cannot add %s plug: needs valid "read" or "write" attribute`, iface.name)
```
zyga: Same comment as above about the placement of "cannot add foo plug:" prefix.

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,162 @@
...
+
+	return nil
+}
+
+func (iface *commonFilesInterface) AppArmorConnectedPlug(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	var reads, writes []interface{}
+	_ = plug.Attr("read", &reads)
+	_ = plug.Attr("write", &writes)
+
+	errPrefix := fmt.Sprintf(`cannot connect plug %s: `, plug.Name())
```
zyga: We are not connecting plugs here, the prefix "cannot connect plug %s" should be elsewhere IMO.

interfaces/builtin/personal_files.go
```diff
@@ -0,0 +1,79 @@
...
+type personalFilesInterface struct {
+	commonFilesInterface
+}
+
+func validateSinglePathHome(np string) error {
+	if !strings.HasPrefix(np, "$HOME/") {
+		return fmt.Errorf(`%q must start with "$HOME"`, np)
+	}
+	if strings.Count(np, "$HOME") > 1 {
+		return fmt.Errorf(`$HOME must only be used at the start of the path of %q`, np)
```
zyga: The error message for the string `$HOME/sweet/$HOME` will be a little confusing. How about literally: `"$HOME must be used exactly once"`?

interfaces/builtin/personal_files.go
```diff
@@ -0,0 +1,79 @@
...
+# users's home directory.
+# This is restricted because it gives file access to arbitrary locations.
+`
+
+type personalFilesInterface struct {
+	commonFilesInterface
+}
+
+func validateSinglePathHome(np string) error {
+	if !strings.HasPrefix(np, "$HOME/") {
```
zyga: `$HOME` vs `${HOME}`?

I'm not sure how this looks for consistency with other variable-like syntax we have in interfaces.

interfaces/builtin/personal_files.go
```diff
@@ -0,0 +1,79 @@
...
+		return fmt.Errorf(`%q must start with "$HOME"`, np)
```
zyga: Ironically the string `$HOME` will fail validation with the message `"$HOME" must start with "$HOME"`

interfaces/builtin/system_files.go
```diff
@@ -0,0 +1,79 @@
...
+type systemFilesInterface struct {
+	commonFilesInterface
+}
+
+func validateSinglePathSystem(np string) error {
+	if !strings.HasPrefix(np, "/") {
+		return fmt.Errorf(`%q must start with "/"`, np)
+	}
+	if strings.Contains(np, "$HOME") {
+		return fmt.Errorf(`$HOME cannot be used in %q`, np)
```
zyga: How about something like `system-files interface cannot refer to $HOME`?

interfaces/builtin/personal_files.go
zyga: Update: after reading system-files I would change all those error messages to just: `personal-files interface must refer to $HOME`

tests/lib/files.sh
```diff
@@ -45,8 +45,8 @@ clean_dir() {
 ensure_file_exists() {
```
zyga: The name of the function is surprising because nothing mentions the extra `.fake` file being created.

tests/lib/snaps/test-snapd-sh/meta/snap.yaml
```diff
@@ -1,6 +1,15 @@
 name: test-snapd-sh
 summary: A no-strings-attached, no-fuss shell for writing tests
 version: 1.0
+
+plugs:
+    personal-files:
+        read: [$HOME/.testdir1, $HOME/.testfile1, $HOME/testfile1, $HOME/testdir1]
```
zyga: Is it guaranteed that `$HOME` expanded has no trailing slash? If it may have a trailing slash then the paths won't be clean. I wonder if we should handle that (cleaning $HOME ourselves)

interfaces/builtin/common_files.go
mvo5: I followed the example of the "HOME" interface here.

interfaces/builtin/common_files.go
mvo5: There is a validation part and I added some tests. However I think this code should simply replace all $HOME" with ${HOME} as it does not know about the other validations and should not need to know. I added a comment and changed the code slightly.

interfaces/builtin/personal_files.go
mvo5: Not quite sure I fully understand the suggestion but: `${HOME}` does not validate: `"\${HOME}/foo" contains a reserved apparmor char from .*`},` - I added a test for this.

interfaces/builtin/personal_files.go
mvo5: I fix this.

tests/lib/files.sh
mvo5: Yeah, I can look at this in a followup, I inherited this one.

interfaces/builtin/common_files.go
jdstrand: I would wait to break this out until there is a need to-- these are needed for quite specific reasons atm.

interfaces/apparmor/apparmor.go
```diff
@@ -35,6 +35,17 @@ import (
 	"github.com/snapcore/snapd/osutil"
 )
 
+// ValidateNoAppArmorRegexp will check that the given string does not
+// contain AppArmor regular expressions (AARE), double quotes or \0.
+func ValidateNoAppArmorRegexp(s string) error {
+	const AARE = `?*[]{}^"` + "\x00"
```
zyga: Curious about the peculiar syntax here, is there a reason for this concatenation?

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,165 @@
...
+
+const (
+	filesRead filesAAPerm = iota
+	filesWrite
+)
+
+func (a filesAAPerm) String() string {
+	switch a {
+	case filesRead:
+		return "rk"
```
zyga:
```suggestion
		return "rk" // [r]ead and loc[k]
```

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,165 @@
...
+	case filesWrite:
+		return "rwkl"
```
zyga:
```suggestion
		return "rwkl" // [r]ead, [w]rite, loc[k] and [l]ink
```

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,165 @@
...
+	// Note that the {personal,system}-files interface impose
+	// limitations on the $HOME usage - system-files forbids it,
+	// personal only allows starting with $HOME in the path.
+	if strings.Contains(p, "$HOME") {
+		p = strings.Replace(p, "$HOME", "@{HOME}", -1)
+		prefix = "owner "
+	}
+	p += "{,/,/**}"
+
+	return fmt.Sprintf("%s%q", prefix, filepath.Clean(p)), nil
```
zyga: Calling file path.Clean on a non-path expression may have unintended consequences (stripping the trailing slash). I think you want to call it on the part before you append `{,/,/**}`. Perhaps you want to remove the `p += ...` above and simply encode that in the format string here.

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func allowPathAccess(buf *bytes.Buffer, perm filesAAPerm, paths []interface{}) error {
+	for _, rawPath := range paths {
+		p, err := formatPath(rawPath)
+		if err != nil {
+			return err
+		}
+		fmt.Fprintf(buf, "%s %s,\n", p, perm)
```
zyga: Nice

interfaces/builtin/common_files.go
```diff
@@ -0,0 +1,165 @@
...
+	return nil
+}
+
+func (iface *commonFilesInterface) validateSinglePath(np string) error {
+	if strings.HasSuffix(np, "/") {
+		return fmt.Errorf(`%q cannot end with "/"`, np)
+	}
+	p := filepath.Clean(np)
+	if p != np {
+		return fmt.Errorf("cannot use %q: try %q", np, filepath.Clean(np))
```
zyga: That is nice, thank you

tests/main/interfaces-personal-files/task.yaml
```diff
@@ -0,0 +1,98 @@
...
+    . "$TESTSLIB/files.sh"
+
+    # Fist layer of dirs and files
+    ensure_file_exists_backup_real /root/.testfile1
+    ensure_file_exists_backup_real /root/testfile1
+    ensure_dir_exists_backup_real /root/.testdir1
+    ensure_dir_exists_backup_real /root/testdir1
+
+    # Second layer of dirs and files
+    ensure_file_exists_backup_real /root/.testdir1/.testfilé2
```
zyga: Touché ;-)

dholbach:
Will this get into the Christmas release of `snapd`? :-) :santa:

interfaces/apparmor/apparmor.go
mvo5: Because its a "`" type string I can't embedd the \0 in there easily. I could change it to:
```
	const AARE = "?*[]{}^\"\x00"
```
not sure if that is more readable though (I have no opinion either way, both look a bit hard to read to me).

PR #5846, opened 2018-09-20
snap: tweak commands

mvo5:
Tweak things.

### DIFF
--- a/tests/main/help/task.yaml
+++ b/tests/main/help/task.yaml
@@ -4,7 +4,7 @@ systems: [fedora-*]
 
 execute: |
     bad=""
-    for CMD in $( GO_FLAGS_COMPLETION=1 snap | grep -vFx help ); do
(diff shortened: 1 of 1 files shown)


cmd/snap/cmd_blame.go
```diff
@@ -0,0 +1,55 @@
...
+
+func (x *cmdBlame) Execute(args []string) error {
+	if len(args) > 0 {
+		return ErrExtraArgs
+	}
+	if len(authors) == 0 {
+		return nil
+	}
+
+	fmt.Fprintf(Stdout, "It's all %s's fault.\n", authors[rand.Intn(len(authors))])
```
zyga: I think we need something like this:
```
author := authors[rand.Intn(len(authors))]
if strings.HasSuffix(author, "s") {
fmt.Fprintf(Stdout, "It's all %s' fault.\n", author)
} else {
fmt.Fprintf(Stdout, "It's all %s's fault.\n", author)
}

PR #5847, opened 2018-09-20
overlord/snapstate: improve cleaup in mount-snap handler

bboozzoo:
When SetupSnap() fails, the cleanup path would not remove a shared snap
directory if one was created. Unify the cleanup paths in mount snap handler and
make sure to attempt to cleanup the shared directory in all error paths.

### DIFF
--- a/overlord/snapstate/backend_test.go
+++ b/overlord/snapstate/backend_test.go
@@ -621,6 +621,9 @@ func (f *fakeSnappyBackend) SetupSnap(snapFilePath, instanceName string, si *sna
 	case "gadget":
 		snapType = snap.TypeGadget
 	}
+	if instanceName == "borken-in-setup" {
+		return snapType, fmt.Errorf("cannot install snap %q", instanceName)
+	}
 	return snapType, nil
 }
 
--- a/overlord/snapstate/handlers.go
+++ b/overlord/snapstate/handlers.go
@@ -515,11 +515,29 @@ func (m *SnapManager) doMountSnap(t *state.Task, _ *tomb.Tomb) error {
 		return err
 	}
 
+	cleanup := func() {
+		st.Lock()
+		defer st.Unlock()
+
+		otherInstances, err := hasOtherInstances(st, snapsup.InstanceName())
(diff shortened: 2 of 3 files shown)


overlord/snapstate/handlers.go
```diff
@@ -515,11 +515,29 @@ func (m *SnapManager) doMountSnap(t *state.Task, _ *tomb.Tomb) error {
 		return err
 	}
 
+	cleanup := func() {
+		st.Lock()
+		defer st.Unlock()
+
+		otherInstances, err := hasOtherInstances(st, snapsup.InstanceName())
```
zyga: nitpick, `otherInstances` feels like there is a collection of other snaps here. Can we call this `ok` or something similar?

overlord/snapstate/handlers_mount_test.go
```diff
@@ -195,6 +195,63 @@ func (s *mountSnapSuite) TestDoMountSnapError(c *C) {
...
+	v1 := "name: borken\nversion: 1.0\n"
+	testSnap := snaptest.MakeTestSnapWithFiles(c, v1, nil)
+
+	s.state.Lock()
+	defer s.state.Unlock()
+	si1 := &snap.SideInfo{
+		RealName: "borken-in-setup",
+		Revision: snap.R(1),
+	}
+	si2 := &snap.SideInfo{
```
zyga: Having two revisions here seems unnecessary. Unless I'm missing something I'd just remove one to make the test easier to follow _or_ add a comment explaining how we use those revisions.

overlord/snapstate/backend_test.go
```diff
@@ -621,6 +621,9 @@ func (f *fakeSnappyBackend) SetupSnap(snapFilePath, instanceName string, si *sna
 	case "gadget":
 		snapType = snap.TypeGadget
 	}
+	if instanceName == "borken-in-setup" {
```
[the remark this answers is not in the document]
bboozzoo: That's planned, we already have `borken` that triggers certain error paths in fake backend.

PR #5851, opened 2018-09-21
tests: don't fail interfaces-bluez test if bluez is already installed

plars:
skip installation of the bluez snap in interfaces-bluez spread test if bluez is already installed

### DIFF
--- a/tests/main/interfaces-bluez/task.yaml
+++ b/tests/main/interfaces-bluez/task.yaml
@@ -11,9 +11,11 @@ environment:
     SNAP_NAME: bluez
 
 execute: |
-    echo "Installing bluez snap from the store ..."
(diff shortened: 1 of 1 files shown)


tests/main/interfaces-bluez/task.yaml
```diff
@@ -11,9 +11,11 @@ environment:
     SNAP_NAME: bluez
 
 execute: |
-    echo "Installing bluez snap from the store ..."
-    expected="(?s)$SNAP_NAME .* from Canonical✓ installed\\n"
-    snap install "$SNAP_NAME" | grep -Pzq "$expected"
+    if ! snap list |grep -Pq "^bluez +" ; then
```
chipaca: nah, do `snap list bluez`

PR #5859, opened 2018-09-25
many: introduce facts, export experimental features as facts

zyga:
Facts are pieces of information determined by snapd. Facts are useful in
avoiding duplication of C and Go implementation and where state access is
required to determine something.  Facts are written to a root-owned directory
`/var/lib/snapd/facts` are trusted just like the state or the installed snaps.

The first set of facts exported from snapd to snap-update-ns and snap-confine
are the enabled experimental flags. This will be used by the upcoming
experimental flag that controls persistence and updates of the per-user,
per-snap mount namespace.


overlord/configstate/configcore/corecfg.go
```diff
@@ -54,12 +54,7 @@ func coreCfg(tr Conf, key string) (result string, err error) {
 
 // supportedConfigurations will be filled in by the files (like proxy.go)
 // that handle this configuration.
-var supportedConfigurations = map[string]bool{
```
bboozzoo: IMO having this explicit was nice when grepping around the code.

overlord/configstate/configcore/experimental.go
```diff
@@ -0,0 +1,72 @@
...
+
+func handleExperimentalFlags(tr Conf) error {
+	var buf bytes.Buffer
+	for _, flag := range experimentalFlags {
+		value, err := coreCfg(tr, "experimental."+flag)
+		if err != nil {
+			return err
+		}
+		if value == "true" {
+			fmt.Fprintf(&buf, "%s=%s\n", strings.TrimPrefix(flag, "core.experimental."), value)
```
bboozzoo: Why not just use `flag` ?

overlord/configstate/configcore/experimental.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+	}
+
+	if !osutil.IsDirectory(dirs.FactsDir) {
+		if err := os.MkdirAll(dirs.FactsDir, 0755); err != nil {
+			return err
+		}
+	}
+
+	return osutil.AtomicWriteFile(filepath.Join(dirs.FactsDir, "experimental-features"), buf.Bytes(), 0644, 0)
```
bboozzoo: How about just `experimental` or `experiments` or `feature-flags` (for that's what it is) ?

overlord/configstate/configcore/corecfg.go
zyga: It's either one or the other way, other parts of the code have this defined programmatically as well.

overlord/configstate/configcore/experimental.go
zyga: Because the file contains the enabled experimental flags, not all of configuration. It looks more natural this way.

overlord/configstate/configcore/experimental.go
zyga: `feature-flags` seems nice but I wanted to be explicit. Still I'm happy to rename this any way we agree to.

niemeyer:
Closing for now based on our conversation.

PR #5860, opened 2018-09-25
interfaces/hotplug: helpers and struct updates

stolowski:
This is a preparatory PR for upcoming hotplug branches that will slightly reduce diff sizes. With this I'm introducing minor additions to existing structers (connState, slot Info), add some task helpers for hotplug-related tasks and define new struct (HotplugSlotDef) used to persist data of hotplug slots.

- connState gets "hotplugDeviceKey" and "hotplugRemoved" flag.
- SlotInfo gets optional "HotplugDeviceKey" value.

The goal of the above is to associate slots and connections with hotplug devices. The "hotplugRemoved" flag in connState is used to keep the state of a connection if the device was unplugged, and restore it when it's plugged again.

The "hotplug-slots" in the state is used to keep track of hotplug slots.


snap/info.go
```diff
@@ -648,12 +648,13 @@ func (slot *SlotInfo) String() string {
...
-	Label     string
-	Apps      map[string]*AppInfo
-	Hooks     map[string]*HookInfo
+	Name             string
+	Interface        string
+	Attrs            map[string]interface{}
+	Label            string
+	Apps             map[string]*AppInfo
+	Hooks            map[string]*HookInfo
+	HotplugDeviceKey string
```
zyga: Perhaps add a newline in case there will be more hot plug things here?

stolowski: Added, with a comment, thanks.

overlord/ifacestate/helpers.go
```diff
@@ -337,6 +337,13 @@ type connState struct {
 	DynamicPlugAttrs map[string]interface{} `json:"plug-dynamic,omitempty"`
 	StaticSlotAttrs  map[string]interface{} `json:"slot-static,omitempty"`
 	DynamicSlotAttrs map[string]interface{} `json:"slot-dynamic,omitempty"`
+	// Hotplug-related attributes: HotplugRemoved indicates a connection
+	// that disappeared because the device was removed, but may potentially
+	// be restored in the future if we see the device again.
+	// HotplugDeviceKey is the key of the associated device; it's empty for
+	// connections of regular slots.
+	HotplugRemoved   bool   `json:"hotplug-removed,omitempty"`
+	HotplugDeviceKey string `json:"hotplug-key,omitempty"`
```
niemeyer: I suggest naming the first one as _HotplugGone_ instead. The difference is subtle.. "gone" feels a little more like the device was lost somehow, instead of proactively removed via snapd or similar.

The second one would be better as _HotplugKey,_ which reflects the json name  already in place, and is short and sweet.

overlord/ifacestate/helpers.go
```diff
@@ -719,3 +726,41 @@ func ensureSystemSnapIsPresent(st *state.State) error {
...
 }
+
+func hotplugTaskSetAttrs(task *state.Task, deviceKey, ifaceName string) {
+	task.Set("device-key", deviceKey)
+	task.Set("interface", ifaceName)
+}
+
+func hotplugTaskGetAttrs(task *state.Task) (deviceKey, ifaceName string, err error) {
+	if err = task.Get("interface", &ifaceName); err != nil {
+		return "", "", fmt.Errorf("internal error: failed to get interface name: %s", err)
```
niemeyer: Lacks context..

"internal error: cannot get interface name from hotplug task: %s"

Note the s/failed to/cannot/

overlord/ifacestate/helpers.go
```diff
@@ -719,3 +726,41 @@ func ensureSystemSnapIsPresent(st *state.State) error {
...
+	}
+	if err = task.Get("device-key", &deviceKey); err != nil {
+		return "", "", fmt.Errorf("internal error: failed to get device key: %s", err)
```
niemeyer: Same as above.

overlord/ifacestate/helpers.go
```diff
@@ -719,3 +726,41 @@ func ensureSystemSnapIsPresent(st *state.State) error {
 	_, err := snapstate.CoreInfo(st)
 	return err
 }
+
+func hotplugTaskSetAttrs(task *state.Task, deviceKey, ifaceName string) {
+	task.Set("device-key", deviceKey)
+	task.Set("interface", ifaceName)
+}
+
+func hotplugTaskGetAttrs(task *state.Task) (deviceKey, ifaceName string, err error) {
```
niemeyer: The more natural order on the results here might be _ifaceName, deviceKey,_ since the device key is built by and dependent on the interface.

overlord/ifacestate/helpers.go
```diff
@@ -719,3 +726,41 @@ func ensureSystemSnapIsPresent(st *state.State) error {
...
+	if err = task.Get("interface", &ifaceName); err != nil {
+		return "", "", fmt.Errorf("internal error: failed to get interface name: %s", err)
+	}
+	if err = task.Get("device-key", &deviceKey); err != nil {
+		return "", "", fmt.Errorf("internal error: failed to get device key: %s", err)
+	}
+	return deviceKey, ifaceName, err
+}
+
+type HotplugSlotDef struct {
```
niemeyer: Why "Def"?  Doesn't sound like a convention we have in place, and in general every struct is a "*Def".

overlord/ifacestate/helpers.go
```diff
@@ -719,3 +726,41 @@ func ensureSystemSnapIsPresent(st *state.State) error {
...
+	Name             string                 `json:"name"`
+	Interface        string                 `json:"interface"`
+	StaticAttrs      map[string]interface{} `json:"static-attrs,omitempty"`
+	HotplugDeviceKey string                 `json:"device-key"`
+}
+
+func getHotplugSlots(st *state.State) (map[string]HotplugSlotDef, error) {
+	var slots map[string]HotplugSlotDef
```
niemeyer: These should probably use pointers.. it looks like an identity type instead of a value type, and there's no reason to have all the data inside the map itself.

snap/info.go
```diff
@@ -654,6 +654,10 @@ type SlotInfo struct {
 	Label     string
 	Apps      map[string]*AppInfo
 	Hooks     map[string]*HookInfo
+
+	// HotplugDeviceKey is the key of the associated device; it's empty for
```
niemeyer: The comment is stating the code. Any DeviceKey is _always_ the _Key_ of the _Device._ Such comments are worse than no comment because it forces the reader to parse an english sentence looking for useful data, but there's none.

A better comment might be:
```
// HotplugKey is a unique key built by the slot's interface using properties of a
// hotplugged so that the same slot may be made available if the device is reinserted.

overlord/ifacestate/helpers.go
stolowski: Renamed to HotplugSlotInfo, per your suggestion from the followup PR, thanks!

PR #5862, opened 2018-09-25
cmd: fix C formatting

zyga:


zyga:
It's applied automatically but let me double-check that.

zyga:
@bboozzoo good catch, the snap-update-ns directory was not included. This is fixed now.

mvo5:
We will loose `git blame`ability for a lot of code once this is applied. Maybe unavoidable if we really want to have `make fmt` for everything.

zyga:
We can look past the format change with git log, that code doesn't have that much churn going on.

PR #5865, opened 2018-09-26
cmd/snap: add a bunch of TRANSLATORS notes (and a little more i18n)

chipaca:
All option descriptions were missing a TRANSLATORS note about not
starting the description with lowercase; I've added that. I've also
removed the ambiguous "probably" from the existing note on argument
descriptions, and clarified it in the two cases where the probability
was relevant.

Also, and opportunistically (and only at first because this PR got too
big), I added some i18n.G's that were missing.

I recommend you read this diff with the "ignore whitespace" option turned on.

### DIFF
--- a/cmd/snap/cmd_prepare_image.go
+++ b/cmd/snap/cmd_prepare_image.go
@@ -49,9 +49,9 @@ core device images.
 			return &cmdPrepareImage{}
 		}, map[string]string{
 			// TRANSLATORS: This should not start with a lowercase letter.
-			"extra-snaps": "Extra snaps to be installed",
+			"extra-snaps": i18n.G("Extra snaps to be installed"),
 			// TRANSLATORS: This should not start with a lowercase letter.
-			"channel": "The channel to use",
+			"channel": i18n.G("The channel to use"),
 		}, []argDesc{
 			{
 				// TRANSLATORS: This needs to be wrapped in <>s.
--- a/cmd/snap/cmd_prepare_image.go
+++ b/cmd/snap/cmd_prepare_image.go
@@ -48,18 +48,20 @@ core device images.
 		func() flags.Commander {
 			return &cmdPrepareImage{}
 		}, map[string]string{
+			// TRANSLATORS: This should not start with a lowercase letter.
 			"extra-snaps": "Extra snaps to be installed",
-			"channel":     "The channel to use",
+			// TRANSLATORS: This should not start with a lowercase letter.
+			"channel": "The channel to use",
 		}, []argDesc{
 			{
 				// TRANSLATORS: This needs to be wrapped in <>s.
 				name: i18n.G("<model-assertion>"),
-				// TRANSLATORS: This should probably not start with a lowercase letter.
+				// TRANSLATORS: This should not start with a lowercase letter.
 				desc: i18n.G("The model assertion name"),
 			}, {
 				// TRANSLATORS: This needs to be wrapped in <>s.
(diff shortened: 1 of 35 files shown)


cmd/snap/cmd_prepare_image.go
```diff
@@ -48,18 +48,20 @@ core device images.
 		func() flags.Commander {
 			return &cmdPrepareImage{}
 		}, map[string]string{
+			// TRANSLATORS: This should not start with a lowercase letter.
 			"extra-snaps": "Extra snaps to be installed",
-			"channel":     "The channel to use",
+			// TRANSLATORS: This should not start with a lowercase letter.
+			"channel": "The channel to use",
```
diddledan: are these two strings passed through i18n elsewhere? if not, then we either need to remove the `TRANSLATORS:` comment or add the `i18n.G` function call

cmd/snap/cmd_snap_op.go
```diff
@@ -211,11 +211,16 @@ func (mxd mixinDescs) also(m map[string]string) mixinDescs {
...
+	"channel": i18n.G("Use this channel instead of stable"),
+	// TRANSLATORS: this should start with an upper case character (if they exist)
+	"beta": i18n.G("Install from the beta channel"),
+	// TRANSLATORS: this should start with an upper case character (if they exist)
+	"edge": i18n.G("Install from the edge channel"),
+	// TRANSLATORS: this should start with an upper case character (if they exist)
 	"candidate": i18n.G("Install from the candidate channel"),
-	"stable":    i18n.G("Install from the stable channel"),
+	// TRANSLATORS: this should start with an upper case character (if they exist)
+	"stable": i18n.G("Install from the stable channel"),
```
diddledan: These don't need the `(if they exist)` text for translators. If the channel doesn't contain anything the text just isn't shown (shouldn't be), so the translation will never need to indicate a dual nature.

cmd/snap/cmd_snap_op.go
```diff
@@ -301,8 +306,11 @@ type modeMixin struct {
 }
 
 var modeDescs = mixinDescs{
-	"classic":  i18n.G("Put snap in classic mode and disable security confinement"),
-	"devmode":  i18n.G("Put snap in development mode and disable security confinement"),
+	// TRANSLATORS: this should start with an upper case character (if they exist)
+	"classic": i18n.G("Put snap in classic mode and disable security confinement"),
+	// TRANSLATORS: this should start with an upper case character (if they exist)
+	"devmode": i18n.G("Put snap in development mode and disable security confinement"),
+	// TRANSLATORS: this should start with an upper case character (if they exist)
```
diddledan: Likewise, these don't need the `(if they exist)` caveat.

cmd/snap/cmd_snap_op.go
chipaca: ... if *upper case characters* exist.

diddledan: aha. I misunderstood the meaning. In that case then I'm fine with this.

cmd/snap/cmd_snap_op.go
diddledan: Per above comment I'm fine with this :-)

cmd/snap/cmd_snap_op.go
mvo5: This curious why for most it says "this should not start with a lower-case char" but for this one (and some more) it is different?

PR #5867, opened 2018-09-27
many: enable layouts by default

zyga:
This branch switches the default value of `experimental.layouts` to true while
retaining the ability to still disable it on a per-machine basis.


mvo5:
A bit of a meta question - why not simply remove the option? Or is the idea that people can still disable layouts if they are scared of them?

zyga:
I started with just removing the code but then I thought that for some things it may be better to still have an off switch. It is less sharp and risky change than removing the option altogether.

tests/main/layout/task.yaml
```diff
@@ -7,9 +7,6 @@ details: |
     hooks get permissions to access those areas.
 
 prepare: |
-    echo "Ensure feature flag is enabled"
-    snap set core experimental.layouts=true
```
stolowski: Can we extend this test after exisitng checks to remove the snap, set `experimental.layouts=false` and check that it fails to install when the feature is missing?
And a little bit of a generic question (and not strictly related to this change): what happens when you disable the feature while having a snap that depends on it in the system?

overlord/snapstate/snapstate.go
```diff
@@ -459,20 +459,20 @@ func defaultContentPlugProviders(st *state.State, info *snap.Info) []string {
 	return out
 }
 
-func getFeatureFlagBool(tr *config.Transaction, flag string) (bool, error) {
-	var v interface{} = false
+func getFeatureFlagBool(tr *config.Transaction, flag string, unset bool) (bool, error) {
```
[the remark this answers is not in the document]
zyga: Default is a keyword

tests/main/layout/task.yaml
zyga: The same as if you have a parallel-instance snap installed, something undesired. We only check during a refresh.

overlord/snapstate/snapstate.go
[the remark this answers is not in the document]
stolowski: How about defaultValue?

zyga:
I feel comfortable with the set of fixes that are piled up for 2.36 now so I'd like this in.
CC @mvo5

overlord/snapstate/snapstate.go
```diff
@@ -68,7 +68,7 @@ func doInstall(st *state.State, snapst *SnapState, snapsup *SnapSetup, flags int
 		}
 		if model == nil || model.Base() == "" {
 			tr := config.NewTransaction(st)
-			experimentalAllowSnapd, err := getFeatureFlagBool(tr, "experimental.snapd-snap")
+			experimentalAllowSnapd, err := getFeatureFlagBool(tr, "experimental.snapd-snap", false)
```
bboozzoo: How about just dropping the code that looks at the flag? It should still work if you revert snapd because the actual setting is not removed.

mvo5: I had the same feeling when reviewing the code, YAGNI and all that. But we had a quick irc discussion with Gustavo and he likes the idea of keeping the flag.

overlord/snapstate/snapstate.go
```diff
@@ -471,20 +471,20 @@ func defaultContentPlugProviders(st *state.State, info *snap.Info) []string {
 	return out
 }
 
-func GetFeatureFlagBool(tr *config.Transaction, flag string) (bool, error) {
-	var v interface{} = false
+func GetFeatureFlagBool(tr *config.Transaction, flag string, unset bool) (bool, error) {
```
mvo5: How do you feel about emulating default args like: http://paste.ubuntu.com/p/5rYxwwvx54 ? This way we can write "GetFeatureFlagBool(tr, "something") and only add "GetFeatureFlagBool(tr, "foo", true)" if the default is different?

zyga: Ah, that's nice. Yeah.

overlord/snapstate/snapstate_test.go
```diff
@@ -12080,8 +12093,8 @@ func (s *snapmgrTestSuite) TestUpdateLayoutsChecksFeatureFlag(c *C) {
 	_, err := snapstate.Update(s.state, "some-snap", "channel-for-layout", snap.R(0), s.user.ID, snapstate.Flags{})
 	c.Assert(err, ErrorMatches, "experimental feature disabled - test it by setting 'experimental.layouts' to true")
 
-	// enable layouts
-	tr := config.NewTransaction(s.state)
+	// When layouts are enabled we can refresh to a snap depending on the feature.
```
bboozzoo: I think we're missing a test for the update path but with layouts feature using the default. That would be the case of a snap that did not use layouts before but got them now. It's probably enough to just test the `Update()` as the checks are in `doUpdate()` which is called by `UpdateMany()` too.

PR #5868, opened 2018-09-27
tests/main/snap-env: extend to cover parallel installations of snaps

bboozzoo:
Now that we can run parallel installed snaps, add parallel installs variant of the test.

### DIFF
--- a/tests/main/snap-env/task.yaml
+++ b/tests/main/snap-env/task.yaml
@@ -1,63 +1,85 @@
 summary: inspect all the set environment variables prefixed with SNAP_ and XDG_
 
+environment:
+    NAME: test-snapd-tools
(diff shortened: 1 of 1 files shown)


tests/main/snap-env/task.yaml
```diff
@@ -1,63 +1,85 @@
 summary: inspect all the set environment variables prefixed with SNAP_ and XDG_
 
+environment:
+    NAME: test-snapd-tools
```
mvo5: This is slightly unusal AFAICT. In most tests that use environment we have a "/description" for all uses. Mixing "NAME" and "NAME/paralle" is slightly unusual. Maybe "NAME/normal" and "NAME/parallel" (same for insance_key)?

PR #5871, opened 2018-09-27
snapstate: only report errors if there is an actual error 

mvo5:
The code in the undo handler was too naive. It would report errors
even if the change was aborted and not actually errored. This PR
fixes this and only report real errors.

### DIFF
--- a/overlord/snapstate/handlers.go
+++ b/overlord/snapstate/handlers.go
@@ -383,7 +383,17 @@ func (m *SnapManager) undoPrepareSnap(t *state.Task, _ *tomb.Tomb) error {
 		extra["UbuntuCoreTransitionCount"] = strconv.Itoa(ubuntuCoreTransitionCount)
 	}
 
-	if !settings.ProblemReportsDisabled(st) {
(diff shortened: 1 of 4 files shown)


overlord/snapstate/handlers.go
```diff
@@ -383,7 +383,17 @@ func (m *SnapManager) undoPrepareSnap(t *state.Task, _ *tomb.Tomb) error {
 		extra["UbuntuCoreTransitionCount"] = strconv.Itoa(ubuntuCoreTransitionCount)
 	}
 
-	if !settings.ProblemReportsDisabled(st) {
+	// Only report and error if there is an actual error in the change,
+	// we could undo things because the user canceled the change.
+	isErr := func() bool {
```
niemeyer: Why is this an inline function?  If this is just the equivalent of the following, I'd keep it simple:
```
var isErr bool
for _, tt := range t.Change.Tasks() {
    if tt.Status() == state.ErrorStatus {
        isErr = true
        break
    }
}

PR #5872, opened 2018-09-27
tests/main/parallel-install-local: rename from *-sideload, extend to run snaps

bboozzoo:
Extend the test to run installed snaps. Remove use of 'sideload' and rename the test.


tests/main/parallel-install-local/task.yaml
```diff
@@ -1,10 +1,13 @@
-summary: Checks for parallel installation of sideloaded snaps
+summary: Checks parallel installation of snaps from local files
+
+restore:
```
zyga: I'm a bit sensitive to not using `restore: |`, it can be easily misused if someone adds another line. Would you mind changing that before it is merged?

PR #5873, opened 2018-09-27
tests/main/parallel-install-store: run installed snap

bboozzoo:
Extend the test to run the installed snap.


tests/main/parallel-install-store/task.yaml
```diff
@@ -24,5 +26,5 @@ execute: |
     #   more than once in the request.
     MATCH 'cannot install "test-snapd-tools"' < run.err
 
-restore:
```
mvo5: heh, nice catch

PR #5875, opened 2018-09-27
interfaces/builtin: avahi interface update

kubiko:
Avahi behaves differently on core and classic systems. Socket is created in different location on classic  system deployed by deb package and on core by avahi snap.

Additionally client, depending on way it talks to daemon, needs to sometimes connect socket dir interface. However there is no alternative to this on classic system, when system does not expose content interface for socket directory.
Instead give avahi-daemon permission to create socked in the same place it does it on classic system

### DIFF
--- a/interfaces/builtin/avahi_observe.go
+++ b/interfaces/builtin/avahi_observe.go
@@ -45,6 +45,10 @@ const avahiObserveBaseDeclarationSlots = `
 const avahiObservePermanentSlotAppArmor = `
 network netlink,
 
+# Allow access to deamon to create socket
+/{,var/}run/avahi-daemon/  w,
+/{,var/}run/avahi-daemon/{pid,socket} rw,
+
 # Description: Allow operating as the avahi service. This gives
 # privileged access to the system.
 #include <abstractions/dbus-strict>
@@ -231,6 +235,8 @@ const avahiObservePermanentSlotDBus = `
 const avahiObserveConnectedPlugAppArmor = `
 # Description: allows domain, record, service, and service type browsing
 # as well as address, host and service resolving
+
+/{,var/}run/avahi-daemon/socket rw,
 
 #include <abstractions/dbus-strict>
 dbus (send)


interfaces/builtin/avahi_observe.go
```diff
@@ -232,6 +236,8 @@ const avahiObserveConnectedPlugAppArmor = `
 # Description: allows domain, record, service, and service type browsing
 # as well as address, host and service resolving
 
+/{,var/}run/avahi-daemon/{pid,socket} rw,
```
bboozzoo: There's a change just like this one in the previous hunk.

bboozzoo: Pfff ignore this please.

interfaces/builtin/avahi_observe.go
```diff
@@ -45,6 +45,10 @@ const avahiObserveBaseDeclarationSlots = `
 const avahiObservePermanentSlotAppArmor = `
 network netlink,
 
+# Allow access to deamon to create socket
+/{,var/}run/avahi-daemon/  w,
+/{,var/}run/avahi-daemon/{pid,socket} rw,
```
jdstrand: Wouldn't this be better served in SNAP_COMMON or SNAP_DATA?

interfaces/builtin/avahi_observe.go
jdstrand: Why does a client of avahi need to be able to write avahi's pid file? Can you describe what the socket is used for (which might answer my question about SNAP_COMMON and SNAP_DATA)?

kubiko: indeed oversight. Client should not need to write pid file, removing it

interfaces/builtin/avahi_observe.go
kubiko: SNAP_COMMON is where socket is created at the moment, when avahi is installed as snap. Problem is that on classic system when avahi is deployed through deb packages, socket is created in /run/avahi-daemon which creates several problems.
Client needs to worry about two location, libnss_mdns.so has path hardcoded and compiled in, so one would need to handle two binary versions in one snap. Also on classic system there is no content interface so client will still need permission /run/avahi-daemon/socket rw
So cleanest way seems to be to unify between core and classic

PR #5877, opened 2018-09-27
cmd/snap-update-ns,osutil: move helpers to osutil

zyga:
This patch moves a large set of helper functions into the osuitl
package. The helpers are as follows: OpenPath, BindMount, MkFile,
MkDir, MkSymlink, MkPrefix, MkFileAll, MkDirAll, MkSymlinkAll.

In addition the Assumptions and Restrictions infrastructure is moved
with the constraint that Assumptions becomes an interface and
Restrcitions are moved as-is. This allows us to keep the implementation
of Assumptions tightly bound to the internals of snap-update-ns while
allowing for restriction-free operation or of an implementation of
custom logic.

The last change is that the system call mocking machinery is made public
so that certain system calls are exposes as public functions that are
automatically mocked with the MockSystemCalls test helper. The result is
that any package can now use all of the helpers and observe their
behavior at the same time without having to carry a local copy of the
mocking system.


zyga:
I'll close this for now, not sure if the cost is worth doing this.

PR #5879, opened 2018-09-27
vendor, cmd/snap: refactor to accommodate the new less buggy go-flags

chipaca:
go-flags on master has some fixes to manpage generation that are
interesting for us. In particular, the paragraphs following a command
with only hidden options were not dedented properly, and included a
spurious "Usage:" line, making it very messy and hard to find the
following command:

```
   help
       Show help about a command

       The help command displays information about snap commands.

       Usage: snap [OPTIONS] help [help-OPTIONS]

          info
              Show detailed information about snaps
```

would now render properly as

```
   help
       Show help about a command

       The help command displays information about snap commands.

   info
       Show detailed information about snaps
```

as well as some the alignment of options in help output no longer
leaving space for hidden options.

However, another bug fixed by this revision is that the
`PassAfterNonOption` flag now performs as advertised, meaning that
`snap install foo --classic` and `snap install --classic foo` are
different, if that flag is used. This is bad for us because we want
the fexibility it used to provide, but we also need this behaviour for
`snap run`.

That is, everywhere except in `snap run` we want options that come
after arguments to be treated as options, but in `snap run`,
anything that comes after the snap name argument we want to pass
through.

The `PassDoubleDash` flag would let us do that, by simply including a
`"--"` before the snap name. Unfortunately if we switch to requiring
this, it will break people who might be running `snap run` by hand and
dependend on the old behaviour.

~~This commit drops our use of `PassAfterNonOption`, and tweaks
`os.Args` to add a `"--"` before the snapname in `snap run` calls, if
the user hasn't added it themself.~~

This commit only adds `PassAfterNonOption` when we're trying to run `snap run`.

HTH, HAND.

### DIFF
--- a/cmd/snap/cmd_alias_test.go
+++ b/cmd/snap/cmd_alias_test.go
@@ -38,8 +38,8 @@ Once this manual alias is setup the respective application command can be
 invoked just using the alias.
 
 [alias command options]
-          --no-wait     Do not wait for the operation to finish but just print
-                        the change id.
+      --no-wait       Do not wait for the operation to finish but just print
+                      the change id.
 `
 	s.testSubCommandHelp(c, "alias", msg)
 }
--- a/cmd/snap/cmd_connect_test.go
+++ b/cmd/snap/cmd_connect_test.go
@@ -54,8 +54,8 @@ Connects the provided plug to the slot in the core snap with a name matching
 the plug name.
 
 [connect command options]
-          --no-wait        Do not wait for the operation to finish but just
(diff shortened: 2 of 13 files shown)


bboozzoo:
This means we'd have to update go-flags in Fedora now?

chipaca:
@bboozzoo we don't depend on new features (except that the tests that include the exact help output might break), we just no longer depend on old bugs.

chipaca:
Oooh, don't merge this one just yet

PR #5880, opened 2018-09-28
interfaces/repo: two helper methods for hotplug

stolowski:
Two helpers related to hotplug - getting a slot based on device key and interface type, and getting connections for given device key and interface type.

Based on #5860, the actual change is only in `repo.go` and `repo_test.go`.


interfaces/repo.go
```diff
@@ -764,6 +764,47 @@ func (r *Repository) connected(snapName, plugOrSlotName string) ([]*ConnRef, err
 	return conns, nil
 }
 
+// ConnectionsForDeviceKey returns all hotplug connections for given device key and interface name.
+func (r *Repository) ConnectionsForDeviceKey(deviceKey, ifaceName string) ([]*ConnRef, error) {
+	r.m.Lock()
+	defer r.m.Unlock()
+
+	snapName, err := r.guessSystemSnapName()
```
zyga: This is not something I'd like to change in this branch but I would like to remove this function and teach the repo about the snap that holds implicit and hot plug slots.

interfaces/repo.go
```diff
@@ -764,6 +764,47 @@ func (r *Repository) connected(snapName, plugOrSlotName string) ([]*ConnRef, err
 	return conns, nil
 }
 
+// ConnectionsForDeviceKey returns all hotplug connections for given device key and interface name.
+func (r *Repository) ConnectionsForDeviceKey(deviceKey, ifaceName string) ([]*ConnRef, error) {
```
niemeyer: Following up on points in base PRs: s/DeviceKey/HotplugKey/

Equivalent in docs above it and in parameter name.

Also, per reviews in base PRs, the order should generally be:

_(ifaceName, deviceKey)_

since the key is built by the interface itself, and so only makes inside the context of the given interface.

interfaces/repo.go
```diff
@@ -764,6 +764,47 @@ func (r *Repository) connected(snapName, plugOrSlotName string) ([]*ConnRef, err
...
+	r.m.Lock()
+	defer r.m.Unlock()
+
+	snapName, err := r.guessSystemSnapName()
+	if err != nil {
+		return nil, err
+	}
+	var conns []*ConnRef
+	for _, slotInfo := range r.slots[snapName] {
+		if slotInfo.HotplugDeviceKey == deviceKey && slotInfo.Interface == ifaceName {
```
niemeyer: This ends up as _slotInfo.HotplugKey == hotplugKey_, which is quite nice.

The order of comparisons should be inverted though, for the same reason mentioned above.

interfaces/repo.go
```diff
@@ -764,6 +764,47 @@ func (r *Repository) connected(snapName, plugOrSlotName string) ([]*ConnRef, err
...
+			}
+		}
+	}
+
+	return conns, nil
+}
+
+// HotplugSlotForDeviceKey returns a hotplug slot for given device key and interface name or nil
+// if there is no slot.
+func (r *Repository) HotplugSlotForDeviceKey(deviceKey, ifaceName string) (*snap.SlotInfo, error) {
```
niemeyer: The name should use a similar structure to the previous function, since it has exactly the same purpose and layout. So something like _SlotForHotplugKey_ instead.

The parameter also needs to be renamed and order switched, per previous note.

Doc above it needs to be changed accordingly.

interfaces/repo.go
```diff
@@ -764,6 +764,47 @@ func (r *Repository) connected(snapName, plugOrSlotName string) ([]*ConnRef, err
...
+	r.m.Lock()
+	defer r.m.Unlock()
+
+	snapName, err := r.guessSystemSnapName()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, slotInfo := range r.slots[snapName] {
+		if slotInfo.HotplugDeviceKey == deviceKey && slotInfo.Interface == ifaceName {
```
niemeyer: Same as above (order switch, renames).

overlord/ifacestate/helpers.go
```diff
@@ -719,3 +726,41 @@ func ensureSystemSnapIsPresent(st *state.State) error {
 	_, err := snapstate.CoreInfo(st)
 	return err
 }
+
+func hotplugTaskSetAttrs(task *state.Task, deviceKey, ifaceName string) {
+	task.Set("device-key", deviceKey)
```
niemeyer: I think I missed this in previous reviews: these should all be hotplugKey and "hotplug-key" as well, so it's consistent across the board.

It's also not yet clear how these will be used, so we may still need to revisit these attributes after the real task show up to account for context, etc.

overlord/ifacestate/helpers.go
```diff
@@ -719,3 +726,41 @@ func ensureSystemSnapIsPresent(st *state.State) error {
 	_, err := snapstate.CoreInfo(st)
 	return err
 }
+
+func hotplugTaskSetAttrs(task *state.Task, deviceKey, ifaceName string) {
```
niemeyer: These can be setHotplugAttrs and getHotplugAttrs, consistent with the names already in place for the other functions below (getHotplugSlots, and setHotplugSlots).

PR #5883, opened 2018-09-28
tests,cmd/snap-update-ns: add test showing mount update bug

zyga:
This patch adds a unit and integration test that shows how mount
namespace update process may misbehave due to presence of leftover empty
mount points or symlinks.


zyga:
I have a fix now but I'm trying to remove a need for a race condition there. I will publish it here soon.

zyga:
I'll close this for now, debugging locally.

PR #5884, opened 2018-09-28
overlord/snapshotstate: store the SnapID in snapshot, block restore if changed

chipaca:
This change makes it so that the snap ID is stored in the snapshot,
and if on restore that ID is non-empty, and there is a snap with the
snapshot's snap's name installed that has a non-empty snap ID, and
these two IDs are different, the restore is fails.

This feature was requested at the Brussels sprint, as a minimal
prerequisite to renames (and merges). This then needs to grow to
handle the rename, but at least for now it will do no damage.

### DIFF
--- a/overlord/snapshotstate/snapshotstate.go
+++ b/overlord/snapshotstate/snapshotstate.go
@@ -69,7 +69,7 @@ func allActiveSnapNames(st *state.State) ([]string, error) {
 	return names, nil
 }
 
-func snapNamesInSnapshotSet(setID uint64, requested []string) (snapsFound, filenames []string, err error) {
+func snapNamesInSnapshotSet(setID uint64, requested []string) (snapsFound, snapIDs, filenames []string, err error) {
 	sort.Strings(requested)
 	found := false
 	err = backendIter(context.TODO(), func(r *backend.Reader) error {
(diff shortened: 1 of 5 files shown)


overlord/snapshotstate/snapshotstate.go
```diff
@@ -197,6 +202,30 @@ func Restore(st *state.State, setID uint64, snapNames []string, users []string)
 
 	ts = state.NewTaskSet()
 
+	for i, snapID := range snapIDs {
+		if snapID == "" {
+			// snapshotted snap was sideloaded, never mind
```
zyga: `// The snap referring to the snapshot was locally installed without assertions`

overlord/snapshotstate/snapshotstate.go
```diff
@@ -197,6 +202,30 @@ func Restore(st *state.State, setID uint64, snapNames []string, users []string)
...
+			// snap not installed
+			continue
+		}
+		sideInfo := snapst.CurrentSideInfo()
+		if sideInfo == nil {
+			// ... shouldn't happen (but also not installed)
+			continue
+		}
+		if sideInfo.SnapID == "" {
+			// current snap is sideloaded, never mind then
```
zyga: Same re-wording suggestion as above.

overlord/snapshotstate/snapshotstate.go
```diff
@@ -197,6 +202,30 @@ func Restore(st *state.State, setID uint64, snapNames []string, users []string)
...
+			continue
+		}
+		if sideInfo.SnapID != snapID {
+			return nil, nil, fmt.Errorf("cannot restore snapshot over id change: %.7s… → %.7s…", snapID, sideInfo.SnapID)
```
zyga: Will the -> arrow render OK on simple terminals? Perhaps call a helper to translate it to "->"

chipaca: It works on the linux console with default settings, which was the good-enough-for-now pass mark we've used before.

overlord/snapshotstate/snapshotstate.go
chipaca: I'm going with s/sideloaded/unasserted/, thank you for pointing it out

PR #5885, opened 2018-09-28
Adding DPDK interface for DPDK Snap

wililupy:
Hello,
I created two new interfaces, dpdk-control and dpdk-observe. 
dpdk-control is used to allow creation of dpdk required paths for use with the dpdk snap I am currently working on. (github.com/wililupy/dpdk-snap). 
dpdk-observe just reads processes and configurations and kernel space so that dpdk can function properly.
If you have any questions, don't hesitate to contact me.
Luke.

### DIFF
--- a/interfaces/builtin/dpdk_observe_test.go
+++ b/interfaces/builtin/dpdk_observe_test.go
@@ -1,7 +1,7 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 
 /*
- * Copyright (C) 2017 Canonical Ltd
+ * Copyright (C) 2018 Canonical Ltd
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 3 as


interfaces/builtin/dpdk_control.go
```diff
@@ -0,0 +1,49 @@
...
+package builtin
+
+const dpdkControlSummary = `allows using dpdk`
+
+const dpdkControlBaseDeclarationSlots = `
+  dpdk-control:
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
```
zyga: This is okay

interfaces/builtin/dpdk_control.go
```diff
@@ -0,0 +1,49 @@
...
+`
+
+const dpdkControlConnectedPlugAppArmor = `
+# Description: Allow control to dpdk.
```
zyga: Is the interface designed to allow DPDK to do its thing or for other things to talk to a snap providing DPDK? If the answer is _the former_ then this interface should be called _dpkg-support_ and be uniquely granted to dpdk.

interfaces/builtin/dpdk_control_test.go
```diff
@@ -0,0 +1,108 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2017 Canonical Ltd
```
zyga: It is 2018

interfaces/builtin/dpdk_control.go
```diff
@@ -0,0 +1,49 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2017 Canonical Ltd
```
zyga: It is 2018

interfaces/builtin/dpdk_observe.go
```diff
@@ -0,0 +1,49 @@
...
+        - core
+    deny-auto-connection: true
+`
+
+const dpdkObserveConnectedPlugAppArmor = `
+# Description: Allow access to dpdk.
+/sys/kernel/mm/hugepages/{,**} r,
+/dev/hugepages/{,**} r,
+/run/dpdk/{,**} r,
+/proc/{,**} r,
```
zyga: Can you provide the rationale for reading _everything_ in proc? This feels too powerful to me.

interfaces/builtin/dpdk_control.go
wililupy: The plug allows the dpdk snap to do it's thing, but it also has to allow any application installed on top of the device to be able to make use of the DPDK modules.

interfaces/builtin/dpdk_observe.go
wililupy: During my testing, I was running into issues where dpdk could not read it's proc running and would cause a core dump. I was hoping that there was already an interface in place for this (process-control or system-observe) but I had to explicitly set it in the dpdk interface for it to function properly.

chipaca:
@wililupy note you have unit test failures in that last run

wililupy:
I am going to rewrite the interface. I am currently working on an interface for dpdk to work properly when fully confined, and then working on a dpdk-consume interface so that snaps/applications that rely on the dpdk libraries will be able to use them. We can close this PR as when I have finished the new one, I will create a new PR.

zyga:
@wililupy I'm closing this based on your feedback. Feel free to open a new PR with v2 patchset.

PR #5887, opened 2018-09-29
tests: moving core-snap-refresh-on-core test from main to nested suite

sergiocazzolato:
This test is moved because it is set as manual in the main suite and it is
failing when the core is refreshed. If we move the test to the nested
suite, it is easier to run it and control the reboots which are
currently failing. Currently this test is being executed in a spread-cron branch when a new core is pushed to edge channel.

In a next PR I'll add checks for the boot properties.

Currently it is failing because after the revert (from edge to stable),
the installed core is the stable one and the tracking is edge, which is
wrong and makes the test fail.

See the error log:
https://paste.ubuntu.com/p/vw7DsjxZhk/


tests/lib/nested.sh
```diff
@@ -187,3 +199,8 @@ del_device(){
     echo "device_del $DEVICE_ID" | nc -q 0 127.0.0.1 "$MON_PORT"
     echo "device deleted"
 }
+
+get_nested_core_revision(){
+    ID=$1
```
zyga:
```suggestion
    ID="$1"
```

sergiocazzolato:
@stolowski thanks for reviewing this

PR #5888, opened 2018-09-30
[RFC] use stages to run "cheaper" tests first

chipaca:
This reorganises our travis tests so that a number of cheaper (in
time, money, and environmental impact) tests are run before running
the full integration suite.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,9 @@ jobs:
       name: go 1.6/trusty unit test suite
       go: 1.6
       before_install:
-        - sudo apt update
+        - sudo apt --quiet update
       install:
-        - sudo apt build-dep snapd
+        - sudo apt --quiet build-dep snapd
         - ./get-deps.sh
       script:
         - ./run-checks --unit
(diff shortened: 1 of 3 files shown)


chipaca:
> Build dist: xenial
> Build id: 435306699
> Job id: 435306700
> Runtime kernel version: 4.14.12-041412-generic
> travis-build version: 1ed393ed4
> Build image provisioning date and time
> Tue Dec  5 20:11:19 UTC 2017
> Operating System Details
> Distributor ID:	Ubuntu
> Description:	Ubuntu 14.04.5 LTS

I suspect `dist: xenial` does not do what I thought it did (supported values are `precise`, `trusty`, and `xenial`, though, so ... was worth a shot)

chipaca:
stages are done sequentially (with everything in a stage done in ~parallel), so it shouldn't

chipaca:
To be clear: if the "cheap" tests pass every time, this approach will slow us down: we'll be using up a slot for about 15 minutes before spread kicks in (we can probably cut that down to ~5 minutes by picking tests that are 'short' and only running those from here). This only benefits us time-wise if we think unit / static / license checks are at least ~1:3 of all "real" failures --- if it's a transient failure, you can restart the spread suite independently of the rest.

chipaca:
That [last run](https://travis-ci.org/snapcore/snapd/builds/435647396), plus [travis' docs](https://docs.travis-ci.com/user/customizing-the-build/#build-timeouts) show that the 50 minute timeout is per job, counted from when the job itself starts running.

PR #5889, opened 2018-09-30
make run-checks --static pass again w/shellcheck installed

chipaca:
We'd slipped, here.

### DIFF
--- a/packaging/arch/PKGBUILD
+++ b/packaging/arch/PKGBUILD
@@ -36,13 +36,12 @@ prepare() {
   cd "$pkgname"
 
   export GOPATH="$srcdir/go"
-  mkdir -p "$GOPATH"
 
   # Have snapd checkout appear in a place suitable for subsequent GOPATH. This
   # way we don't have to go get it again and it is exactly what the tag/hash
   # above describes.
-  mkdir -p "$(dirname "$GOPATH/src/${_gourl}")"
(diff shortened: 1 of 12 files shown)


packaging/arch/PKGBUILD
```diff
@@ -36,13 +36,12 @@ prepare() {
...
 
   export GOPATH="$srcdir/go"
-  mkdir -p "$GOPATH"
 
   # Have snapd checkout appear in a place suitable for subsequent GOPATH. This
   # way we don't have to go get it again and it is exactly what the tag/hash
   # above describes.
-  mkdir -p "$(dirname "$GOPATH/src/${_gourl}")"
-  ln --no-target-directory -fs "$srcdir/$pkgname" "$GOPATH/src/${_gourl}"
+  mkdir -p "$(dirname "$srcdir/go/src/${_gourl}")"
```
mvo5: Why is this change needed? What is shellcheck complaining about? It seems like this is a bit less readable now (or maybe that is just me)?

chipaca: that one isn't shellcheck. It's right after it, inside the same guard as shellcheck, checking for uses of GOPATH that assume it's only ever got a single element in it.

I can also tell it to skip the PKGBUILD file.

mvo5: Aha, I get it - its the ":" check. I think its fine unless @bboozzoo dislikes it.

PR #5890, opened 2018-09-30
overlord/snapshotstate: small refactor of internal helpers

chipaca:
overlord/snapshotstate had an internal helper, `snapNamesInSnapshotSet`, that used to returned the snap names and their filenames of a given snap id with matching snaps; it was `func(uint64, []string) ([]string, []string, error)`, and was used by restore/check/forget to find the snapshots that were part of a set and that needed to be included in a given change.

In #5884, that function changed to also return the snap IDs, so that restore could error out if a snap's ID changed around a restore. it was, thus, `func(uint64, []string) ([]string, []string, []string, error)`

Next I want to block on Epoch changes... so rather than going with a `func(uint64, []string) ([]string, []string, []string, []snap.Epoch, error)` I decided to refactor it a little bit, introducing an internal, ad hoc, `snapshotSnapSummary`.

I hope you like it.


overlord/snapshotstate/export_test.go
```diff
@@ -32,19 +32,32 @@ import (
...
+	Filename                   = filename
+	DoSave                     = doSave
+	DoRestore                  = doRestore
+	UndoRestore                = undoRestore
+	CleanupRestore             = cleanupRestore
+	DoCheck                    = doCheck
+	DoForget                   = doForget
 )
 
+func (shsums snapshotSnapSummaries) Asplode() []map[string]string {
```
mvo5: Hm, maybe `func (sss snapshotSnapSummaries) Summaries() []map[string]string {` ? The `shsums` reads a lot like hashsums to me and Asplode is a hard to remember :)

overlord/snapshotstate/snapshotstate.go
```diff
@@ -197,14 +229,38 @@ func Restore(st *state.State, setID uint64, snapNames []string, users []string)
...
 	ts = state.NewTaskSet()
 
-	for i, name := range snapsFound {
-		desc := fmt.Sprintf("Restore data of snap %q from snapshot set #%d", name, setID)
+	for _, summary := range summaries {
+		if summary.snapID == "" {
+			// snapshotted snap was unasserted, never mind
+			continue
+		}
+		snapst, ok := all[summary.snap]
```
bboozzoo: I'm wondering if it's better to use individual `snapshot.Get()` or pay the price of unmarshalling all snaps upfront.

overlord/snapshotstate/export_test.go
chipaca: sure

bboozzoo: Slicify?

overlord/snapshotstate/snapshotstate.go
chipaca: assuming you meant `snapstate.Get()`, perhaps yes. Doing the `All` makes the worst case better, at the expense of the expected common case. Not sure.

I'm hoping to revisit all the `snapstate.All` calls when we have Go 1.7+, because [reasons](https://gist.github.com/chipaca/10ecdf364db962c9b5ee590f0f610b2a).

overlord/snapshotstate/export_test.go
chipaca: `PotatoChicken()`

PR #5891, opened 2018-09-30
snap: give Epoch a CanRead helper

chipaca:
The logic behind whether an epoch can read another epoch's data isn't
hard nor error prone, but we might as well have a helper for it.

### DIFF
--- a/snap/epoch.go
+++ b/snap/epoch.go
@@ -162,16 +162,8 @@ func (e *Epoch) Validate() error {
 		return &EpochError{Message: epochListNotSorted}
 	}
 
-	// O(𝑚𝑛) instead of O(𝑚log𝑛) for the binary search we could do, but
-	// 𝑚 and 𝑛 <10 per above, so the simple solution is good enough (and if
-	// that alone makes you nervous, know that it is ~2× faster in the
-	// worst case; bisect starts being faster at ~50 entries).
-	for _, r := range e.Read {
-		for _, w := range e.Write {
-			if r == w {
-				return nil
-			}
-		}
+	if intersect(e.Read, e.Write) {
+		return nil
 	}
 	return &EpochError{Message: noEpochIntersection}
 }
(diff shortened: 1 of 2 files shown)


snap/epoch.go
```diff
@@ -212,6 +212,37 @@ func (e *Epoch) String() string {
 	return string(buf)
 }
 
+// CanRead checks whether this epoch can read the data written by the
+// other one.
+func (e *Epoch) CanRead(other *Epoch) bool {
```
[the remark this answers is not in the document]
chipaca: or I could make an `intersects(rs, ws []uint32) bool` and call it from both

snap/epoch.go
```diff
@@ -212,6 +204,44 @@ func (e *Epoch) String() string {
...
+	if other != nil {
+		ws = other.Write
+	}
+	if len(rs) == 0 {
+		rs = []uint32{0}
+	}
+	if len(ws) == 0 {
+		ws = []uint32{0}
+	}
+
```
zyga: Is there something that ensures that `rs` and `ws` are indeed sorted?

snap/epoch.go
```diff
@@ -212,6 +204,44 @@ func (e *Epoch) String() string {
...
+	return intersect(rs, ws)
+}
+
+func intersect(rs, ws []uint32) bool {
+	// O(𝑚𝑛) instead of O(𝑚log𝑛) for the binary search we could do, but
+	// 𝑚 and 𝑛 < 10, so the simple solution is good enough (and if that
+	// alone makes you nervous, know that it is ~2× faster in the worst
+	// case; bisect starts being faster at ~50 entries).
+	for _, r := range rs {
```
zyga: Big nitpick about complexity.

snap/epoch.go
chipaca: Either by construction (when the epoch is a string), or from `snap.(*Epoch).Validate()`, when the epoch is more complex.
In tests, you can build an invalid epoch, but in the wild they all come from one of those two :-)

PR #5892, opened 2018-10-01
tests: shellchecks part 9

bboozzoo:
Part 9


tests/main/snap-service-stop-mode-sigkill/task.yaml
```diff
@@ -24,11 +25,11 @@ execute: |
 
     refresh_modes="sigterm sigterm-all"
     for s in $refresh_modes; do
-        systemctl show -p ActiveState snap.test-snapd-service.test-snapd-${s}-service | MATCH "ActiveState=active"
+        systemctl show -p ActiveState "snap.test-snapd-service.test-snapd-${s}-service" | MATCH "ActiveState=active"
     done
 
     echo "we expect two sleep processes (children) from the two sigterm services"
-    n=$(ps afx | grep [3]133731337 | grep -v grep | wc -l)
+    n=$(pgrep -c -f '[3]133731337')
```
sparkiegeek: You don't need the `[3]` character class trick when using `pgrep`, as per pgrep(1), pgrep "will never report itself as a match"

tests/main/snapd-snap/task.yaml
```diff
@@ -12,5 +12,5 @@ restore: |
 execute: |
     echo "Installing snapscraft"
     apt install -y snapcraft
-    cd $PROJECT_PATH
+    cd "$PROJECT_PATH" || exit
```
mvo5: This is not really needed because we run our spread with `set -e` so cd failing will result in an abort. But it seems like shellcheck does not know about us running with set -e :/ I wonder if we can generate it automatically in the inputs we feed to shellcheck?

tests/main/try-snap-is-optional/task.yaml
```diff
@@ -4,8 +4,8 @@ systems: [-ubuntu-core-*]
 
 execute: |
     echo "When try is executed inside a snap directory"
-    cd $TESTSLIB/snaps/test-snapd-tools
+    cd "$TESTSLIB"/snaps/test-snapd-tools || exit 1
```
mvo5: Same as above, I am not sure we really need these extra guards

tests/main/snap-service-stop-mode-sigkill/task.yaml
bboozzoo: Thanks for the tip, will push an update.

PR #5894, opened 2018-10-01
many: enable AppArmor on Arch

bboozzoo:
In the wake for upcoming changes in snapd AUR packages, enable use of AppArmor in arch tests.

So far, only one task is failing:
```
2018-10-01 14:12:59 Failed tasks: 1
    - google:arch-linux-64:tests/main/snap-confine
```
The reason being AppArmor is enabled in the kernel, but disabled at boot. We need to update the image to pass `apparmor=1 security=apparmor` in the kernel command line. (cc @sergiocazzolato)

Edit: The PR includes a #5893 which turned out to be useful when debugging this.


interfaces/apparmor/backend.go
```diff
@@ -458,14 +458,9 @@ func downgradeConfinement() bool {
...
 	case release.DistroLike("arch"):
-		if strings.HasSuffix(kver, "-hardened") {
-			if cmp, _ := strutil.VersionCompare(kver, "4.17.4"); cmp >= 0 {
-				// The linux-hardened 4.17.4+ package has
-				// apparmor enabled, do not downgrade the
-				// confinement template.
-				return false
-			}
-		}
+		// The default kernel has AppArmor enabled since 4.18.8, the
```
mvo5: I assume the checks are no longer needed because the new dependency on arch ensures that the right kernel gets pulled in, correct?

bboozzoo: Arch is on the latest mainline or LTS kernel. The mainline is 4.18 at the moment, so it's fresh enough to have all the features we need (IIRC we are fine AppArmor wise with anything 4.14+). The LTS kernel (4.14.72) does not have AppArmor enabled so the backend is effectively unused. On top of this people are expected to update whole distro when installing stuff because there's no way to ensure rpm or deb like binary compatibility so one is always running the latest and greatest (for better or worse).

packaging/arch/PKGBUILD
```diff
@@ -182,12 +183,10 @@ package() {
   # Remove snappy core specific units
   rm -fv "$pkgdir/usr/lib/systemd/system/snapd.system-shutdown.service"
   rm -fv "$pkgdir/usr/lib/systemd/system/snapd.autoimport.service"
-  rm -fv "$pkgdir/usr/lib/systemd/system/snapd.apparmor.service"
```
zyga: Thank you for ensuring that this is shipped!

PR #5895, opened 2018-10-01
client, daemon: support passing of 'unaliased' option when installing from local files

bboozzoo:
When doing `snap install --unaliased foo_123.snap` the 'unaliased' option was never passed to the daemon and the API code did not handle it.

### DIFF
--- a/client/snap_op.go
+++ b/client/snap_op.go
@@ -43,6 +43,13 @@ type SnapOptions struct {
 	Users []string `json:"users,omitempty"`
 }
 
+func writeFieldBool(mw *multipart.Writer, key string, val bool) error {
+	if !val {
+		return nil
+	}
+	return mw.WriteField(key, "true")
+}
+
 func (opts *SnapOptions) writeModeFields(mw *multipart.Writer) error {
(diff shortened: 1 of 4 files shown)


client/snap_op.go
```diff
@@ -54,18 +61,18 @@ func (opts *SnapOptions) writeModeFields(mw *multipart.Writer) error {
...
-		if err != nil {
+		if err := writeFieldBool(mw, o.f, o.b); err != nil {
 			return err
 		}
 	}
 
 	return nil
 }
 
+func (opts *SnapOptions) writeOptionFields(mw *multipart.Writer) error {
```
chipaca: You could rename `writeModeFields` to this, and just include Unaliased in that loop, right?

bboozzoo: Yes, although the code in API seems to split handling of mode flags from other options. I tried (maybe unnecessarily) to keep the same split here. I'm ok with rolling this into a single func if you feel strongly about it.

chipaca: I feel strongly about many things, but this is not one of them.

PR #5896, opened 2018-10-02
snapcraft.yaml: set grade to stable and add type: snapd

mvo5:
Trivial update to the snapcraft.yaml so that we can release 2.36 into candidate and stable. As a drive-by also enables the `type: snapd`.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -10,9 +10,8 @@ description: |
 version: set-by-version-script-thxbye
 version-script: |
   ./mkversion.sh --output-only
-grade: devel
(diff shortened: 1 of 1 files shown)


bboozzoo:
`snapcraft` is not aware of `snapd` type yet apparently:

```
Processing triggers for libc-bin (2.23-0ubuntu10) ...

+ snapcraft
Issues while validating snapcraft.yaml: The 'type' property does not match the required schema: 'snapd' is not one of ['app', 'base', 'gadget', 'kernel', 'os']
```

PR #5231, opened 2018-05-29
interfaces/joystick: force use of the device cgroup with joystick interface

jdstrand:
Because of the unconditional /dev/input/event[0-9]* AppArmor rule, we need
to ensure that the device cgroup is in effect even when there are no
joysticks present so that we don't give away all input to the snap. Use
/dev/full for this purpose.

The defect is not in any released version of snapd, but we need to either
include this or revert https://github.com/snapcore/snapd/pull/5203 from 2.33 and master.

PR #5897, opened 2018-10-02
interfaces/builtin: add gpio-keys interface for accessing events

bergotorino:
This new interface makes possible to read events coming from GPIO keys
such as some sort of buttons available on devices. The purpose is to
allow events to be captured by snaps that can trigger actions based on
them.

### DIFF
--- a/interfaces/builtin/gpio_keys.go
+++ b/interfaces/builtin/gpio_keys.go
@@ -1,0 +1,89 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2018 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package builtin
+
+import (
+	"github.com/snapcore/snapd/interfaces"
+	"github.com/snapcore/snapd/interfaces/udev"
+)
+
+const gpioKeysSummary = `allows access to gpio keys events`
+
+const gpioKeysBaseDeclarationSlots = `
+  gpio-keys:
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
+`
+
+const gpioKeysConnectedPlugAppArmor = `
+# Description: Allow reading and writing events to gpio keys devices
+
+#
+# evdev-based interface
+#
+
+# /dev/input/event* is unfortunately not namespaced and includes all input
+# devices, including keyboards and mice, which allows input sniffing and
+# injection. Until we have inode tagging of devices, we use a glob rule here
+# and rely on udev tagging to only add evdev devices to the snap's device
+# cgroup that are marked with ENV{ID_PATH}=="platform-gpio-keys". As such,
+# even though AppArmor allows all evdev, the device cgroup does not.
+/dev/input/event[0-9]* rw,
+
+# Allow reading for supported event reports for all input devices. See
+# https://www.kernel.org/doc/Documentation/input/event-codes.txt
+# FIXME: this is a very minor information leak and snapd should instead query
+# udev for the specific accesses associated with the above devices.
+/sys/devices/**/input[0-9]*/capabilities/* r,
+`
+
+// Add the gpio keys devices. They come up with ENV{ID_PATH} set to
+// "platform-gpio-keys" value.
+//
+// Because of the unconditional /dev/input/event[0-9]* AppArmor rule, we need
+// to ensure that the device cgroup is in effect even when there are no
+// gpio keys present so that we don't give away all input to the snap.
+var gpioKeysConnectedPlugUDev = []string{
+	`KERNEL=="event[0-9]*", SUBSYSTEM=="input", ENV{ID_PATH}=="platform-gpio-keys"`,
+}
+
+type gpioKeysInterface struct {
+	commonInterface
+}
+
+func (iface *gpioKeysInterface) UDevConnectedPlug(spec *udev.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	spec.TriggerSubsystem("input")
+	return iface.commonInterface.UDevConnectedPlug(spec, plug, slot)
+}
+
+func init() {
+	registerIface(&gpioKeysInterface{commonInterface{
+		name:                  "gpio-keys",
+		summary:               gpioKeysSummary,
+		implicitOnCore:        true,
+		implicitOnClassic:     true,
+		baseDeclarationSlots:  gpioKeysBaseDeclarationSlots,
+		connectedPlugAppArmor: gpioKeysConnectedPlugAppArmor,
+		connectedPlugUDev:     gpioKeysConnectedPlugUDev,
+		reservedForOS:         true,
+	}})
+}
--- a/interfaces/builtin/gpio_keys_test.go
+++ b/interfaces/builtin/gpio_keys_test.go
@@ -1,0 +1,106 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2018 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package builtin_test
+
+import (
+	. "gopkg.in/check.v1"
+
+	"github.com/snapcore/snapd/interfaces"
+	"github.com/snapcore/snapd/interfaces/apparmor"
+	"github.com/snapcore/snapd/interfaces/builtin"
+	"github.com/snapcore/snapd/interfaces/udev"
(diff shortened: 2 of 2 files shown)


interfaces/builtin/gpio_keys.go
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+package builtin
+
+import (
+	"github.com/snapcore/snapd/interfaces"
+	"github.com/snapcore/snapd/interfaces/udev"
+)
+
+const gpioKeysSummary = `allows access to gpio keys events`
```
zyga: Nitpick: use singular `keys` since the `events` are already plural.

interfaces/builtin/gpio_keys.go
```diff
@@ -0,0 +1,89 @@
...
+
+const gpioKeysBaseDeclarationSlots = `
```
zyga: The policy looks fine, you will need to add a new plug to a spread test that exercises all interface types (spread will fail in this so you will know what to do).

interfaces/builtin/gpio_keys.go
```diff
@@ -0,0 +1,89 @@
...
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
+`
+
+const gpioKeysConnectedPlugAppArmor = `
+# Description: Allow reading and writing events to gpio keys devices
+
+#
```
zyga: Perhaps discard the `evdev-based interface` fragment since the description below is more than adequate.

interfaces/builtin/gpio_keys.go
```diff
@@ -0,0 +1,89 @@
...
+`
+
+// Add the gpio keys devices. They come up with ENV{ID_PATH} set to
+// "platform-gpio-keys" value.
+//
+// Because of the unconditional /dev/input/event[0-9]* AppArmor rule, we need
+// to ensure that the device cgroup is in effect even when there are no
+// gpio keys present so that we don't give away all input to the snap.
+var gpioKeysConnectedPlugUDev = []string{
+	`KERNEL=="event[0-9]*", SUBSYSTEM=="input", ENV{ID_PATH}=="platform-gpio-keys"`,
```
zyga: This looks ok but please evaluate if we should use ID_PATH or ID_PATH_TAG, I don't know what the difference is from the top of my head.

interfaces/builtin/gpio_keys.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+type gpioKeysInterface struct {
+	commonInterface
+}
+
+func (iface *gpioKeysInterface) UDevConnectedPlug(spec *udev.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	spec.TriggerSubsystem("input")
```
zyga: This looks fine. CC @jdstrand

interfaces/builtin/gpio_keys.go
```diff
@@ -0,0 +1,89 @@
...
+	return iface.commonInterface.UDevConnectedPlug(spec, plug, slot)
+}
+
+func init() {
+	registerIface(&gpioKeysInterface{commonInterface{
+		name:                  "gpio-keys",
```
zyga: The name is perhaps most interesting. Those are not GPIOs as exposed in sysfs. Perhaps to avoid confusing with the GPIO interface we could call this interface `platform-keys` or even `platform-input`. CC @niemeyer

interfaces/builtin/gpio_keys.go
estutzenberger: FWIW, `udevadm info --query=property --name=/dev/input/by-path/platform-gpio-keys-event` gives the following on the system:
```
DEVLINKS=/dev/input/by-path/platform-gpio-keys-event
DEVNAME=/dev/input/event1
DEVPATH=/devices/platform/gpio-keys/input/input1/event1
ID_INPUT=1
ID_INPUT_KEY=1
ID_PATH=platform-gpio-keys
ID_PATH_TAG=platform-gpio-keys
MAJOR=13
MINOR=65
SUBSYSTEM=input
USEC_INITIALIZED=11131064
net.ifnames=0
```

bergotorino:
Updated the comments as asked. The tests will come next.

bergotorino:
Added tests to make sure interface works as expected and that keyboard devices cannot be accessed. This addresses points a) and b) (hopefully) from the @jdstrand's comment.

bergotorino:
Having addressed a few of the comments so far I think I need more guidance on how to complete all of them. @jdstrand could you review and provide suggestions? Thanks in advance!

interfaces/builtin/device_buttons.go
```diff
@@ -0,0 +1,92 @@
...
+var deviceButtonsConnectedPlugUDev = []string{
+	`KERNEL=="event[0-9]*", SUBSYSTEM=="input", ENV{ID_INPUT_KEY}=="1", ENV{ID_INPUT_KEYBOARD}!="1"`,
+}
+
+type deviceButtonsInterface struct {
+	commonInterface
+}
+
+func (iface *deviceButtonsInterface) UDevConnectedPlug(spec *udev.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	spec.TriggerSubsystem("input")
```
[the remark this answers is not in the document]
bergotorino: Thanks, this is a good catch. I have already made the requested change.

tests/main/security-dev-input-event-denied/task.yaml
```diff
@@ -63,3 +65,37 @@ execute: |
     # joystick interface is connected since a keyboard shouldn't be added to
     # the device cgroup.
     MATCH "Operation not permitted" < call.error
+
+    # 2. Device Buttons
```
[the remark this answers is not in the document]
bergotorino: Good catch, done!

interfaces/builtin/device_buttons_test.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func (s *DeviceButtonsInterfaceSuite) TestUDevSpec(c *C) {
+	spec := &udev.Specification{}
+	c.Assert(spec.AddConnectedPlug(s.iface, s.plug, s.slot), IsNil)
+	c.Assert(spec.Snippets(), HasLen, 1)
+	c.Assert(spec.Snippets(), testutil.Contains, `# device-buttons
+KERNEL=="event[0-9]*", SUBSYSTEM=="input", ENV{ID_INPUT_KEY}=="1", ENV{ID_INPUT_KEYBOARD}!="1", TAG+="snap_consumer_app"`)
+	c.Assert(spec.TriggeredSubsystems(), DeepEquals, []string{"input"})
+}
```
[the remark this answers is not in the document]
bergotorino: Done

interfaces/udev/udev.go
```diff
@@ -64,6 +64,15 @@ func ReloadRules(subsystemTriggers []string) error {
 				return fmt.Errorf("cannot run udev triggers for joysticks: %s\nudev output:\n%s", err, string(output))
 			}
 			inputJoystickTriggered = true
+		} else if subsystem == "input/key" {
+			// If one of the interfaces said it uses the input
+			// subsystem for input keys, then trigger the keys
+			// events in a way that is specific to input keys
+			// to not block other inputs.
+			output, err = exec.Command("udevadm", "trigger", "--property-match=ID_INPUT_KEY=1", "--property-match=ID_INPUT_KEYBOARD!=1").CombinedOutput()
```
zyga: This looks good, thanks!

zyga:
```

FAIL: device_buttons_test.go:100: DeviceButtonsInterfaceSuite.TestStaticInfo

device_buttons_test.go:104:
    c.Assert(si.Summary, Equals, `allows access to device buttons as input event`)
... obtained string = "allows access to device buttons as input events"
... expected string = "allows access to device buttons

```

bboozzoo:
@jdstrand the PR was updated, can you take another look?

tests/main/security-dev-input-event-denied/task.yaml
```diff
@@ -63,3 +73,30 @@ execute: |
...
+
+    echo "Then the snap is not able to access an evdev keyboard"
+    if test-snapd-event "-event-kbd" 2>"${PWD}"/call.error; then
+        echo "Expected permission error calling evtest with disconnected plug"
+        exit 1
+    fi
+    # AppArmor is 'Permission denied' which is expected with default policy
+    MATCH "Operation not permitted" < call.error
```
jdstrand: The comment here doesn't go with the MATCH statement and what this is showing is that you have a device cgroup denial. You should disconnect the joystick interface (with a comment) before ' # 2. Device Buttons', then use:
```
MATCH "Permission denied" < call.error
```

bboozzoo: I fixed that and ran into issues. With `device-buttons` connected, none of the devices matched, hence `snap-confine` did not put itself under device cgroup in the first place. As a result, the test snap had unconfined access to all input devices. I'll address that in a separate PR.

jdstrand: This was why I wanted the spread tests! :) Thanks for finding this. In this PR, please add "/dev/full" to the cgroup like we did in the joystick interface here: https://github.com/snapcore/snapd/pull/5231/files

If/when the unconditional cgroup is added (ie, your other PR), then this /dev/full here and in the joystick interface can be removed.

bboozzoo:
@jdstrand pushed the workaround you suggested. Please take another look.

interfaces/builtin/device_buttons.go
```diff
@@ -0,0 +1,93 @@
...
+	`KERNEL=="event[0-9]*", SUBSYSTEM=="input", ENV{ID_INPUT_KEY}=="1", ENV{ID_INPUT_KEYBOARD}!="1"`,
+	`KERNEL=="full", SUBSYSTEM=="mem"`,
+}
+
+type deviceButtonsInterface struct {
+	commonInterface
+}
+
+func (iface *deviceButtonsInterface) UDevConnectedPlug(spec *udev.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	spec.TriggerSubsystem("key")
```
jdstrand: This should be "input/key" since that is what the change to udev.go is looking for. Sorry for not noticing this sooner.

tests/main/security-dev-input-event-denied/task.yaml
```diff
@@ -63,3 +73,36 @@ execute: |
...
+    fi
+    # AppArmor is 'Permission denied' which is expected with default policy
+    MATCH "Permission denied" < call.error
+
+    echo "When the device-buttons plug is connected"
+    snap connect test-snapd-event:device-buttons
+    udevadm settle
+
+    # Note, '-event-kbd' devices aren't device buttons (those are
+    # -gpio-keys-even (evdev event*) and therefore shouldn't be added to
```
jdstrand: Typo: -gpio-keys-even**t**

PR #5899, opened 2018-10-02
tests: shellchecks, final round

bboozzoo:
The final batch of shellchecks.


tests/unit/c-unit-tests-clang/task.yaml
```diff
@@ -23,9 +23,9 @@ debug: |
     cat "$SPREAD_PATH/cmd/autogarbage/test-suite.log" || true
 
 execute: |
-    cd "$SPREAD_PATH/cmd/"
+    cd "$SPREAD_PATH/cmd/" || exit 1
```
mvo5: Same nitpick as in my other PR, this is not really needed because we set -e

mvo5: I think we should do something like this:
```
diff --git a/spread-shellcheck b/spread-shellcheck
index 97b7fcc7c..a985eb467 100755
--- a/spread-shellcheck
+++ b/spread-shellcheck
@@ -95,6 +95,7 @@ class ShellcheckFailures(Exception):
 
 
 def checksection(data):
+    data = "#!/bin/sh\nset -e\n" + data
     proc = subprocess.Popen("shellcheck -s {} -x -".format(SHELLCHECK_SHELL),
                             stdout=subprocess.PIPE,
                             stdin=subprocess.PIPE,
```
instead.

tests/main/ubuntu-core-writablepaths/task.yaml
```diff
@@ -4,7 +4,8 @@ systems: [ubuntu-core-16-*]
 
 execute: |
     echo "Ensure everything in writable-paths is actually writable"
-    while read -r line; do
+    #shellcheck disable=SC2002
```
mvo5: Thank you!

PR #5902, opened 2018-10-02
cmd/snap: tweak UX of snap refresh --list

chipaca:
`snap refresh --list` had a poor description, that misled @basak but
fortunately they told us (thanks!) so now here it is, fixed (I hope).

Also, `snap refresh --list some-snap` would not complain, and just
ignore the extra argument. This too was confusing. This one I 'fixed'
by having it error; I could instead filter, but it's a lot of work
right now.

### DIFF
--- a/cmd/snap/cmd_snap_op.go
+++ b/cmd/snap/cmd_snap_op.go
@@ -629,18 +629,14 @@ func (x *cmdRefresh) Execute([]string) error {
 
 	if x.Time {
 		if x.asksForMode() || x.asksForChannel() {
-			return errors.New(i18n.G("--time does not take mode nor channel flags"))
+			return errors.New(i18n.G("--time does not take mode or channel flags"))
 		}
 		return x.showRefreshTimes()
 	}
 
 	if x.List {
-		if len(x.Positional.Snaps) > 0 {
-			return errors.New(i18n.G("--list does not accept positional arguments"))
-		}
-
-		if x.asksForMode() || x.asksForChannel() {
-			return errors.New(i18n.G("--list does not take mode nor channel flags"))
+		if len(x.Positional.Snaps) > 0 || x.asksForMode() || x.asksForChannel() {
+			return errors.New(i18n.G("--list does not accept additional arguments"))
 		}
 
 		return x.listRefresh()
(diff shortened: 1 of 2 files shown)


cmd/snap/cmd_snap_op.go
```diff
@@ -932,7 +936,7 @@ func init() {
 			// TRANSLATORS: This should not start with a lowercase letter.
 			"revision": i18n.G("Refresh to the given revision, to which you must have developer access"),
 			// TRANSLATORS: This should not start with a lowercase letter.
-			"list": i18n.G("Show available snaps for refresh but do not perform a refresh"),
+			"list": i18n.G("Show the snaps that would be updated on refresh, but do not perform the refresh"),
```
degville: Might it be worth adding that the output lists the store version and revision, rather than  the version installed (which is unclear, I think)?
eg. `Shows the new versions of snaps that would be updated with the next refresh.`

cmd/snap/cmd_snap_op_test.go
```diff
@@ -905,6 +905,20 @@ func (s *SnapOpSuite) TestRevertMissingName(c *check.C) {
...
 }
 
+func (s *SnapSuite) TestRefreshListLessOptions(c *check.C) {
+	s.RedirectClientToTestServer(func(w http.ResponseWriter, r *http.Request) {
+		c.Fatal("expected to get 0 requests")
+	})
+
+	for _, flag := range []string{"--beta", "--channel=potato", "--classic"} {
+		_, err := snap.Parser(snap.Client()).ParseArgs([]string{"refresh", "--list", flag})
+		c.Assert(err, check.ErrorMatches, "--list does not take mode nor channel flags")
```
degville: I think it should be *or* rather than *nor*.

cmd/snap/cmd_snap_op.go
```diff
@@ -635,6 +635,10 @@ func (x *cmdRefresh) Execute([]string) error {
 	}
 
 	if x.List {
+		if len(x.Positional.Snaps) > 0 {
+			return errors.New(i18n.G("--list does not accept positional arguments"))
```
degville: I know this is correct, but would _additional_, rather than _positional_, be clearer?

PR #5903, opened 2018-10-02
overlord/snapshotstate: store epoch in snapshot, check on restore

chipaca:
We weren't storing the snap's epoch in its data's snapshot, and we
weren't checking the current revision could read the restored
data. This change fixes that.

### DIFF
--- a/overlord/snapshotstate/snapshotstate.go
+++ b/overlord/snapshotstate/snapshotstate.go
@@ -237,10 +237,12 @@ func Restore(st *state.State, setID uint64, snapNames []string, users []string) 
 				return nil, nil, fmt.Errorf("unexpected error while reading snap info: %v", err)
 			}
 			if !info.Epoch.CanRead(&summary.epoch) {
-				return nil, nil, fmt.Errorf("cannot restore snapshot to incompatible epoch: %s → %s", &summary.epoch, &info.Epoch)
+				const tpl = "cannot restore snapshot for %q: current snap (epoch %s) cannot read snapshot data (epoch %s)"
+				return nil, nil, fmt.Errorf(tpl, summary.snap, &info.Epoch, &summary.epoch)
 			}
 			if summary.snapID != "" && info.SnapID != "" && info.SnapID != summary.snapID {
-				return nil, nil, fmt.Errorf("cannot restore snapshot over id change: %.7s… → %.7s…", summary.snapID, info.SnapID)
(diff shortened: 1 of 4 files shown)


overlord/snapshotstate/snapshotstate.go
```diff
@@ -230,30 +230,20 @@ func Restore(st *state.State, setID uint64, snapNames []string, users []string)
...
-		if sideInfo.SnapID != summary.snapID {
-			return nil, nil, fmt.Errorf("cannot restore snapshot over id change: %.7s… → %.7s…", summary.snapID, sideInfo.SnapID)
+		if snapst, ok := all[summary.snap]; ok {
+			info, err := snapst.CurrentInfo()
+			if err != nil {
+				// how?
+				return nil, nil, fmt.Errorf("unexpected error while reading snap info: %v", err)
+			}
+			if !info.Epoch.CanRead(&summary.epoch) {
+				return nil, nil, fmt.Errorf("cannot restore snapshot to incompatible epoch: %s → %s", &summary.epoch, &info.Epoch)
```
bboozzoo: I feel like wording here may be a bit confusing, though I cannot come up with anything better that isn't much longer at the same time.

Maybe something like `cannot restore snapshot from epoch %s to incompatible epoch %s` (though it still feels a bit off).

chipaca: `"cannot restore snapshot for %s: current epoch (%s) cannot read snapshot epoch (%s)"`?

chipaca: I tweaked it some more: `"cannot restore snapshot for %q: current snap (epoch %s) cannot read snapshot data (epoch %s)"`

And I changed the ID one to match: `"cannot restore snapshot for %q: current snap (ID %.7s…) does not match snapshot (ID %.7s…)"`

overlord/snapshotstate/snapshotstate.go
```diff
@@ -230,30 +230,22 @@ func Restore(st *state.State, setID uint64, snapNames []string, users []string)
...
-		if sideInfo.SnapID == "" {
-			// current snap is unasserted, never mind then
-			continue
-		}
-		if sideInfo.SnapID != summary.snapID {
-			return nil, nil, fmt.Errorf("cannot restore snapshot over id change: %.7s… → %.7s…", summary.snapID, sideInfo.SnapID)
+		if snapst, ok := all[summary.snap]; ok {
+			info, err := snapst.CurrentInfo()
+			if err != nil {
+				// how?
```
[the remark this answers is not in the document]
chipaca: not in this PR :-D

PR #5908, opened 2018-10-03
tests,cmd/snap-update-ns: add test showing mount update bug

zyga:
This patch adds a unit and integration test that shows how mount
namespace update process may misbehave due to presence of leftover empty
mount points or symlinks.


zyga:
CC: @mvo5 I tagged this for 2.36 as it is a prereq of a real 2.36 bug fix branch.

PR #5912, opened 2018-10-03
interfaces/apparmor: handle overlayfs snippet for snap-update-ns

zyga:
When running from a live CD which is using overlayfs on the root
filesystem we would get denials from snap-update-ns that tries to
construct content interface connections as appropriate bind mounts.

This is because additional permissions required to support overlayfs
were added to individual application profiles but not to the per-snap
snap-update-ns profile. This is now fixed.

Fixes: https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1794953


zyga:
@niemeyer I noticed we don't have a spread test for this case and I'm working on conjuring up one. It's slightly more involved so it will be proposed separately.

PR #5913, opened 2018-10-03
 daemon: expose snapshots to the API

chipaca:
With this change, snapd gains a /v2/snapshots endpoint, that supports
GET (to list snapshots) and POST (to modify them), as well as a new
"snapshot" verb on /v2/snaps, to create snapshots.


daemon/api_test.go
```diff
@@ -7522,3 +7540,289 @@ func (s *appSuite) TestErrToResponse(c *check.C) {
...
+		c.Assert(err, check.IsNil, comm)
+
+		rsp := changeSnapshots(snapshotCmd, req, nil).(*resp)
+		c.Check(rsp.Type, check.Equals, ResponseTypeError, comm)
+		c.Check(rsp.Status, check.Equals, 400, comm)
+		c.Check(rsp.Result.(*errorResult).Message, check.Matches, test.error, comm)
+	}
+}
+
+func (s *apiSuite) TestChangeSnapshots404(c *check.C) {
```
mvo5: If I run this test in isolation it crashes:
```
go test -check.f TestChangeSnapshots404
AppArmor status: apparmor is enabled and all features are available

PANIC: api_test.go:7696: apiSuite.TestChangeSnapshots404

... Panic: runtime error: invalid memory address or nil pointer dereference (PC=0x459E2A)

/usr/lib/go-1.10/src/runtime/panic.go:502
  in gopanic
/usr/lib/go-1.10/src/runtime/panic.go:63
  in panicmem
/usr/lib/go-1.10/src/runtime/signal_unix.go:388
  in sigpanic
api.go:3027
  in changeSnapshots
api_test.go:7726
  in apiSuite.TestChangeSnapshots404
/usr/lib/go-1.10/src/reflect/value.go:308
  in Value.Call
/usr/lib/go-1.10/src/runtime/asm_amd64.s:2361
  in goexit
OOPS: 0 passed, 1 PANICKED
--- FAIL: Test (0.01s)
FAIL
exit status 1
```

overlord/snapshotstate/backend/backend_test.go
```diff
@@ -509,3 +509,53 @@ func (s *snapshotSuite) TestHappyRoundtrip(c *check.C) {
...
+		filepath.Join(dirs.SnapDataDir, "hello-snap"),
+	} {
+		c.Check(os.Rename(filepath.Join(dir, "42"), filepath.Join(dir, "17")), check.IsNil)
+	}
+
+	newroot := c.MkDir()
+	c.Assert(os.MkdirAll(filepath.Join(newroot, "home", "snapuser"), 0755), check.IsNil)
+	dirs.SetRootDir(newroot)
+
+	var diff = func() *exec.Cmd {
```
mvo5: (nitpick^2) you could fold the `diff().Run()` directly into diff. I.e.: `return cmd.Run()`.

daemon/api.go
```diff
@@ -2918,3 +2961,99 @@ func getWarnings(c *Command, r *http.Request, _ *auth.UserState) Response {
...
+
+	st := c.d.overlord.State()
+	st.Lock()
+	defer st.Unlock()
+
+	switch action.Action {
+	case "check":
+		affected, ts, err = snapshotCheck(st, action.SetID, action.Snaps, action.Users)
+	case "restore":
+		affected, ts, err = snapshotRestore(st, action.SetID, action.Snaps, action.Users)
```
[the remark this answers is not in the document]
chipaca: Like with AFAIK all the handlers, we hold the lock while we create the state changes. They don't necessarily run with the lock held.

In the particular case of restore, the `doRestore` handler in `overlord/snapshotstate/snapshotmgr.go` starts without the lock held, calls `prepareRestore` which grabs the lock, reads the info about what snap to restore, and reads that snap's config; it releases the lock on return, at which point `doRestore` runs the actual `overlord/snapshotstate/backend/reader.Restore` which calls `tar` et al. Once that succeeds, `doRestore` grabs the lock and applies the saved config (and sets things up to be able to roll back if something else fails)

daemon/api_test.go
chipaca: WAT

chipaca: figured it out 🤦

overlord/snapshotstate/backend/backend_test.go
chipaca: I think I'll leave it for now: this is in the other PR

daemon/api.go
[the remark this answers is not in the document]
mvo5: Aha, silly me - of course, this is just creating the task sets.

daemon/api_snapshots_test.go
```diff
@@ -0,0 +1,321 @@
...
+	c.Check(rsp.Status, check.Equals, 200)
+	c.Check(rsp.Result, check.DeepEquals, snapshots)
+}
+
+func (s *snapshotSuite) TestListSnapshotsFiltering(c *check.C) {
+	snapshots := []client.SnapshotSet{{ID: 1}, {ID: 42}}
+
+	defer daemon.MockSnapshotList(func(_ context.Context, setID uint64, _ []string) ([]client.SnapshotSet, error) {
+		c.Assert(setID, check.Equals, uint64(42))
+		return snapshots[1:], nil
```
mvo5: :heart:

PR #5916, opened 2018-10-04
data: run snapd.autoimport.service only after seeding

mvo5:
The snapd.autoimport.service will try to import assertions from
removable device that are cold-plugged into the device.

On first-boot there is a race here, if the autoimport service is
run before some key assertions are imported during the seeding
the autoimport (cold-plug) will fail.

This PR fixes this by adding an explicit
 `After=snapd.seeded.service`


zyga:
Please don't restart this test, the logs are precious to Chipaca to debug.

zyga:
The error is more widespread and Chipaca has added a follow up to collect more data so removing label.

PR #5271, opened 2018-06-06
cmd/snap: attempt to start the document portal if running with a session bus

jhenstridge:
This is an early prototype, and maybe not the right way to implement the feature.

When starting a desktop application, we want to make sure xdg-document-portal is running (if it is installed), since we want its FUSE mount to be visible before trying to set up the user mount namespace.

Some points on the implementation:

1. Check that the snap has a plug for the `desktop` interface.  This doesn't check that the plug is actually connected, since `snap run` doesn't communicate with `snapd`.   But this shouldn't be harmful in the rare cases where a snap has the plug but it is disconnected.

2. give up early if `$DBUS_SESSION_BUS_ADDRESS` isn't set.  This avoids any attempts to auto-launch a session bus from cron jobs or root services.

3. If we get back a `ServiceUnknown` error, we assume xdg-document-portal isn't installed and continue without printing a warning.

PR #5917, opened 2018-10-04
cmd/snap: attempt to start the document portal if running with a sess…

zyga:
…ion bus (#5271)

* cmd: attempt to start the document portal if running with a session bus

* tests: add a spread test to verify that document portal activation works

* tests: handle different package names on Fedora, SUSE, and Arch

* tests: changes to fix spread failures on Ubuntu 14.04 and OpenSUSE Leap 42.3

* cmd/snap: changes requested in review

* tests: fix shellcheck warnings

* tests: remove packages installed by test in restore

* tests: more shellcheck fixes

* packaging: conflict with old xdg-desktop-portal on Tumbleweed

* cmd/snap: check whether the app or hook specifically uses the desktop plug.

* cmd/snap: if $XDG_RUNTIME_DIR/doc is a mount point, assume
xdg-document-portal is running.

* tests: disable document-portal-activation spread test on Amazon Linux

Amazon Linux does not ship Python 3 versions of the DBus or GObject
bindings, which this test relies on for the fake document portal
implementation.  It isn't a particularly interesting target for this
feature, so I don't think it is worth spending too much time on.

* cmd/snap: record whether the portal services was available

This means that there should only be one attempt to start the document
portal per user session on systems where xdg-desktop-portal is not
installed.

* cmd/snap: use osutil.IsMounted implementation, and stub out on darwin

* cmd/snap: remove duplicate calculation of $XDG_RUNTIME_DIR

* cmd/snap: rename .snap-portals-unavailable -> .portals-unavailable

### DIFF
--- a/cmd/snap/cmd_run.go
+++ b/cmd/snap/cmd_run.go
@@ -23,6 +23,7 @@ import (
 	"bufio"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"os/user"
@@ -33,6 +34,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/godbus/dbus"
 	"github.com/jessevdk/go-flags"
 
 	"github.com/snapcore/snapd/client"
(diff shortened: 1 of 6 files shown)


jhenstridge:
Failing test is a prepare run for google:ubuntu-16.04-64:tests/completion/, failing on "systemctl restart snapd.socket" -- seems unrelated to the changes in this PR.

mvo5:
This is now merged to release/2.36 via master

PR #5921, opened 2018-10-04
spread-shellcheck: use threads to parallelise

chipaca:
Current spread-shellcheck does things serially. The task is of a class
called "embarrassingly parallel", so I did the obvious thing:

Before:

    $ time ./spread-shellcheck spread.yaml tests

    real    1m27.587s
    user    0m53.500s
    sys     0m15.368s

After:

    $ time ./spread-shellcheck spread.yaml tests

    real    0m13.977s
    user    0m56.736s
    sys     0m11.536s

One thing to improve would be to add a lock and hold it around any
call to logging (especially as we do more than one call to logging in
a row and expect them to be adjacent in the output). In practice, not
really a problem right now.


spread-shellcheck
```diff
@@ -171,6 +177,9 @@ def checkpath(loc):
                 logging.error("%s: section '%s':\n%s", entry, section, error)
             failed.append(entry)
```
bboozzoo: I think you need to wrap appending failed files with a lock:
```
lock  = threading.Lock()
def check1path(entry):
     try:
         checkfile(entry)
     except ShellcheckError as serr:
         ...
         with lock:
             failed.append(entry)
```
While at it you may just wrap the whole exception handling to cover the logging too.

How about this:
```
diff --git a/spread-shellcheck b/spread-shellcheck
index 317ad5c0f..aa8e2c55f 100755
--- a/spread-shellcheck
+++ b/spread-shellcheck
@@ -18,6 +18,7 @@ import logging
 import os
 import subprocess
 import argparse
+from concurrent.futures import ThreadPoolExecutor
 
 import yaml
 
@@ -60,9 +61,10 @@ class ShellcheckRunError(Exception):
 
 
 class ShellcheckError(Exception):
-    def __init__(self):
+    def __init__(self, path):
         super().__init__()
         self.sectionerrors = {}
+        self.path = path
 
     def addfailure(self, section, error):
         self.sectionerrors[section] = error
@@ -109,18 +111,17 @@ def checksection(data):
 
 
 def checkfile(path):
-    logging.debug("checking file %s", path)
     with open(path) as inf:
         data = yaml.load(inf)
 
-    errors = ShellcheckError()
+    errors = ShellcheckError(path)
 
     for section in SECTIONS:
         if section not in data:
             continue
 
         try:
-            logging.debug("checking section %s", section)
+            logging.debug("%s: checking section %s", path, section)
             checksection(data[section])
         except ShellcheckRunError as serr:
             logging.debug("%s: shellcheck failed in section '%s'",
@@ -130,12 +131,11 @@ def checkfile(path):
     if path.endswith('spread.yaml') and 'suites' in data:
         # check suites
         for suite in data['suites'].keys():
-            logging.debug('checking suite %s', suite)
             for section in SECTIONS:
                 if section not in data['suites'][suite]:
                     continue
                 try:
-                    logging.debug("checking section %s", section)
+                    logging.debug("%s (suite %s): checking section %s", path, suite, section)
                     checksection(data['suites'][suite][section])
                 except ShellcheckRunError as serr:
                     errors.addfailure('suites/' + suite + '/' + section,
@@ -160,16 +160,20 @@ def checkpath(loc):
         locations = [loc]
 
     failed = []
-    for entry in locations:
-        try:
-            checkfile(entry)
-        except ShellcheckError as serr:
-            logging.error(('shellcheck failed for file %s in sections: '
-                           '%s; error log follows'),
-                          entry, ', '.join(serr.sectionerrors.keys()))
-            for section, error in serr.sectionerrors.items():
-                logging.error("%s: section '%s':\n%s", entry, section, error)
-            failed.append(entry)
+    with ThreadPoolExecutor() as executor:
+        results = executor.map(checkfile, locations)
+        while True:
+            try:
+                next(results)
+            except StopIteration:
+                break
+            except ShellcheckError as serr:
+                logging.error(('shellcheck failed for file %s in sections: '
+                               '%s; error log follows'),
+                              serr.path, ', '.join(serr.sectionerrors.keys()))
+                for section, error in serr.sectionerrors.items():
+                    logging.error("%s: section '%s':\n%s", serr.path, section, error)
+                failed.append(serr.path)
 
     if failed:
         raise ShellcheckFailures(failures=failed)
```

chipaca: Feel free to push to the PR ... TOMORROW WHEN YOU'RE NOT ON HOLIDAY

... maniac

chipaca: for reference, you need to say that last bit like this
https://youtu.be/9TnGjq9mWSI?t=152

chipaca:
(p.s. at 15 seconds I could add it to the "quick" travis preflight checks)

PR #5928, opened 2018-10-04
cmd: put our manpages in section 8

chipaca:
I hadn't realised, but our manpages are a mess.

We currently ship four manpages: `snap(1)`, `snap-confine(1)`,
`snap-env-generator(7)`, and `snap-discard-ns(5)`.

* Section 5 is where file formats are described, like `fstab(5)` or
  `passwd(5)`. `snap-discard-ns` is a program that can only be run as
  root; it goes in section 8.

* Section 7 is where generators in the abstract are described, in
  `systemd.generator(7)`. The generators themselves go in section 8,
  like `systemd-fstab-generator(8)`. So `snap-env-generator` goes in
  section 8.

* I always thought we shipped `snap(8)`: in `debian/rules`, I added
  the rule to create `snap.8`, and then shipped that. I didn't notice
  that `go-flags` wrote it as `snap(1)`, and of course the content
  wins over the filename. My rationale here is that `snap` is like
  `apt(8)`, `ifconfig(8)`, `mount(8)`, ...: users can use them to
  query things, but to modify the system state they need to be admins.

* `snap-confine` is setuid root, and so users _can_ run it, but it's
  not meant for users to run it directly. Just thinking of a system
  administrator running `snap-confine` directly is a stretch. Even so,
  I'm not sure: is it more like `gtk-query-immodule(1)`, or more like
  `ssh-keysign(8)`? I'd put it in section 8, but I can see arguments
  for it to stay in section 1.

So: in this PR, I move all our manpages to section 8. Bring it.

### DIFF
--- a/cmd/Makefile.am
+++ b/cmd/Makefile.am
@@ -186,8 +186,8 @@ decode-mount-opts/decode-mount-opts$(EXEEXT): LIBS += -Wl,-Bstatic $(decode_moun
 
 libexec_PROGRAMS += snap-confine/snap-confine
 if HAVE_RST2MAN
-dist_man_MANS += snap-confine/snap-confine.1
-CLEANFILES += snap-confine/snap-confine.1
+dist_man_MANS += snap-confine/snap-confine.8
+CLEANFILES += snap-confine/snap-confine.8
 endif
 EXTRA_DIST += snap-confine/snap-confine.rst
 EXTRA_DIST += snap-confine/snap-confine.apparmor.in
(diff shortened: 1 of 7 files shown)


keszybz:
This is just going to confuse users. IMHO a much better solution is to remove the compat snap(1) alias from snap (the python package), and let snap own the "snap" name. `/usr/bin/snap` exists, hence there should be snap(1) man page that describes this tool.

Conan-Kudo:
@keszybz the `snap` tool doesn't work without admin privileges anyway, so it belongs in section 8 regardless.

keszybz:
a) it is found in user path, so it should have a man page. b) sure it works, e.g. `snap list`.

Conan-Kudo:
@keszybz And `dnf search` works without admin powers, but it's still in section 8 because the primary actions (i.e. installing and removing software) requires admin privileges.

keszybz:
Well, we're in the realm of opinions here. My take is that the 1/8 man page split stopped being useful when we added `/sbin` to user $PATH. The more immediate issue is that when users type `man snap` to figure out what this does, they'll get an unrelated page for a fairly obscure backup tool in python.

Conan-Kudo:
@keszybz That's true, but there's been a bug report over it for a while, and the maintainer never did anything about it. :(

keszybz:
Yeah, the project does not seem very alive. All the more reason to grab the man page name ;)

PR #5930, opened 2018-10-05
wrappers: do not depend on network.taget in socket units, tweak generated units

bboozzoo:
Depending on network.target (which is currently used as PrerequisiteTarget) may
lead to dependncy cycles, see https://forum.snapcraft.io/t/7666 for details.
Socket units should depend on mount unit only.

While at it, tweak the generated unit files to not leave empty lines and extend
tests to verify contents of generated socket unit files.

### DIFF
--- a/wrappers/services.go
+++ b/wrappers/services.go
@@ -492,18 +492,19 @@ WantedBy={{.ServicesTarget}}
 
 func genServiceSocketFile(appInfo *snap.AppInfo, socketName string) []byte {
 	socketTemplate := `[Unit]
-# Auto-generated, DO NO EDIT
+# Auto-generated, DO NOT EDIT
 Description=Socket {{.SocketName}} for snap application {{.App.Snap.InstanceName}}.{{.App.Name}}
 Requires={{.MountUnit}}
-Wants={{.PrerequisiteTarget}}
-After={{.MountUnit}} {{.PrerequisiteTarget}}
+After={{.MountUnit}}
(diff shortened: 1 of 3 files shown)


wrappers/services.go
```diff
@@ -492,18 +492,19 @@ WantedBy={{.ServicesTarget}}
 
 func genServiceSocketFile(appInfo *snap.AppInfo, socketName string) []byte {
 	socketTemplate := `[Unit]
-# Auto-generated, DO NO EDIT
+# Auto-generated, DO NOT EDIT
 Description=Socket {{.SocketName}} for snap application {{.App.Snap.InstanceName}}.{{.App.Name}}
 Requires={{.MountUnit}}
-Wants={{.PrerequisiteTarget}}
-After={{.MountUnit}} {{.PrerequisiteTarget}}
```
mvo5: With that removed, we can stop passing `		PrerequisiteTarget string` into the template in genServiceSocketFile, right?

wrappers/services.go
```diff
@@ -492,18 +492,19 @@ WantedBy={{.ServicesTarget}}
 
 func genServiceSocketFile(appInfo *snap.AppInfo, socketName string) []byte {
 	socketTemplate := `[Unit]
-# Auto-generated, DO NO EDIT
+# Auto-generated, DO NOT EDIT
```
chipaca: `Y U EDIT`

PR #5933, opened 2018-10-05
osutil, interfaces/apparmor: add and use of osutil.UnlinkMany

chipaca:
This adds `osutil.UnlinkMany`, which is mostly just a convenience
function to remove multiple files form a single directory. It also
uses it from interfaces/apparmor.

### DIFF
--- a/interfaces/apparmor/apparmor.go
+++ b/interfaces/apparmor/apparmor.go
@@ -30,7 +30,6 @@ import (
 	"io"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"strings"
 
 	"github.com/snapcore/snapd/osutil"
@@ -88,12 +87,8 @@ func unloadProfiles(names []string, cacheDir string) error {
 	if err != nil {
 		return fmt.Errorf("cannot unload apparmor profile: %s\napparmor_parser output:\n%s", err, string(output))
 	}
-	for _, name := range names {
-		err = os.Remove(filepath.Join(cacheDir, name))
-		// It is not an error if the cache file wasn't there to remove.
-		if err != nil && !os.IsNotExist(err) {
-			return fmt.Errorf("cannot remove apparmor profile cache: %s", err)
-		}
+	if err := osutil.UnlinkMany(cacheDir, names); err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("cannot remove apparmor profile cache: %s", err)
 	}
 	return nil
 }
--- a/osutil/unlink.go
+++ b/osutil/unlink.go
@@ -1,0 +1,59 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2018 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package osutil
+
+import (
+	"os"
+	"syscall"
+)
+
+// UnlinkMany removes multiple files from a single directory.
+//
+// If dirname is not a directory, this will fail.
+//
+// This will abort at the first removal error (but ENOENT is ignored).
+//
+// Filenames must refer to files. They don't necessarily have to be
+// relative paths to the given dirname, but if they aren't why are you
+// using this function?
+//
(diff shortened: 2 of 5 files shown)


osutil/unlink.go
```diff
@@ -0,0 +1,59 @@
...
+//
+// This will abort at the first removal error (but ENOENT is ignored).
+//
+// Filenames must refer to files. They don't necessarily have to be
+// relative paths to the given dirname, but if they aren't why are you
+// using this function?
+//
+// Errors are *os.PathError, for convenience
+func UnlinkMany(dirname string, filenames []string) error {
+	dirfd, err := syscall.Open(dirname, syscall.O_RDONLY|syscall.O_CLOEXEC|syscall.O_DIRECTORY|0x200000, 0)
```
zyga: Can we dispense with 0x20000 as a numerical constant please? This is double-plus confusing since O_PATH is defined with an octal value so simple grep won't find it.

osutil/unlink.go
```diff
@@ -0,0 +1,59 @@
...
+//
+// If dirname is not a directory, this will fail.
+//
+// This will abort at the first removal error (but ENOENT is ignored).
+//
+// Filenames must refer to files. They don't necessarily have to be
+// relative paths to the given dirname, but if they aren't why are you
+// using this function?
+//
+// Errors are *os.PathError, for convenience
```
zyga: I would explicitly state that the first encountered error stops the iteration and is returned.

osutil/unlink.go
chipaca: Whoops, I forgot to remove that. I added it while wondering whether I should call fsync on dirfd, and whether fsync would work with O_PATH.

osutil/unlink.go
chipaca: Doesn't `This will abort at the first removal error` explicitly state exactly that?

interfaces/apparmor/apparmor.go
```diff
@@ -88,12 +87,8 @@ func unloadProfiles(names []string, cacheDir string) error {
 	if err != nil {
 		return fmt.Errorf("cannot unload apparmor profile: %s\napparmor_parser output:\n%s", err, string(output))
 	}
-	for _, name := range names {
-		err = os.Remove(filepath.Join(cacheDir, name))
-		// It is not an error if the cache file wasn't there to remove.
-		if err != nil && !os.IsNotExist(err) {
-			return fmt.Errorf("cannot remove apparmor profile cache: %s", err)
-		}
+	if err := osutil.UnlinkMany(cacheDir, names); err != nil && !os.IsNotExist(err) {
```
mvo5: Do we need the check for os.IsNotExist() here? The docstring for osutil.UnlinMany says that ENOENT is ignored?

osutil/unlink.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/snapcore/snapd/osutil/sys"
+)
+
+// UnlinkMany removes multiple files from a single directory.
+//
+// If dirname is not a directory, this will fail.
+//
+// This will abort at the first removal error (but ENOENT is ignored).
+//
+// Filenames must refer to files. They don't necessarily have to be
```
mvo5: Hm, so I could do: osutil.UnlinkMany("/etc/", ["/var/lib/foo"]) ? If so thats a slightly odd interface, I mean: it seems to me that either it should enforce that all removes are in this basedir (so no path in the filenames) or we don't need a basedir. Or am I missing something?

interfaces/apparmor/apparmor.go
chipaca: The cacheDir itself might not exist (and there's a test for that situation)

osutil/unlink.go
chipaca: yeah, it's weird, but that's `unlinkat` interface. As any checks we did would be racy, and the function is a convenience thing more than anything else, I didn't think it was worth it

mvo5: I don't strongly mind either way, but something like: `for _, name := range filesnames { if filepath.Base() != name { return fmt.Errorf("can only use base names") }` shouldn't be racy. Or am I misunderstanding the race?

osutil/unlink_test.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (s *unlinkSuite) SetUpTest(c *check.C) {
+	s.d = c.MkDir()
+	s.mkFixture(c)
+}
+
+func (s *unlinkSuite) mkFixture(c *check.C) {
+	for _, fname := range []string{"foo", "bar", "baz", "quux"} {
+		f, err := os.Create(filepath.Join(s.d, fname))
```
mvo5: (nitpick^99) ioutil.WriteFile() would save the f.Close() line ;)

osutil/unlink.go
chipaca: Yes, I could do that. In fact I was going to use strings.IndexByte to much the same effect. But then I thought: what does this gain us? I didn't have an answer, so I didn't do it. I still don't have an answer :-) but if there is something to be gained, I'd be glad to add the check.

chipaca: (i missed your comment before merging, fwiw, but i'll gladly push a PR that fixes this if wanted)

PR #5940, opened 2018-10-06
store: speedup unit tests

chipaca:
Before this change,

    $ go test -count 1 ./store
    ok  	github.com/snapcore/snapd/store	11.921s

and after,

    $ go test -count 1 ./store
    ok  	github.com/snapcore/snapd/store	0.633s

### DIFF
--- a/store/download_test.go
+++ b/store/download_test.go
@@ -35,6 +35,7 @@ import (
 	"github.com/juju/ratelimit"
 	"golang.org/x/net/context"
 	. "gopkg.in/check.v1"
+	"gopkg.in/retry.v1"
 
 	"github.com/snapcore/snapd/dirs"
 	"github.com/snapcore/snapd/osutil"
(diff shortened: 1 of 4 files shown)


store/download_test.go
```diff
@@ -149,7 +155,7 @@ func (s *downloadSuite) TestDownloadCancellation(c *C) {
 		io.WriteString(w, "foo")
 		syncCh <- struct{}{}
 		io.WriteString(w, "bar")
-		time.Sleep(time.Duration(1) * time.Second)
+		time.Sleep(10 * time.Millisecond)
```
bboozzoo: Won't this fail occasionally on a slow machine? Maybe we could make this test conditional on `testing.Short()`?

chipaca: it shouldn't, but it might. If it does we can add the conditional and drop the timeouts back, ok?

bboozzoo: Yes, that would be fair. @mvo5 wdyt?

PR #5944, opened 2018-10-06
cmd/snap: speed up unit tests

chipaca:
cmd/snap started at nearly 13s, but most of that is because ssh2 is
slow to start up. I was able to shave a couple of seconds elsewhere
though. And I tagged the ssh2 tests as not-short.

Before:

    $ go test -count 1 ./cmd/snap
    ok  	github.com/snapcore/snapd/cmd/snap	13.868s

and then,

    $ go test -count 1 ./cmd/snap
    ok  	github.com/snapcore/snapd/cmd/snap	11.470s
    $ go test -short -count 1 ./cmd/snap
    ok  	github.com/snapcore/snapd/cmd/snap	1.251s

### DIFF
--- a/cmd/snap/cmd_userd_test.go
+++ b/cmd/snap/cmd_userd_test.go
@@ -69,16 +69,15 @@ func (s *userdSuite) TestUserd(c *C) {
 		}()
 
 		needle := "io.snapcraft.Launcher"
-		for i := 0; i < 10; i++ {
+		for i := 0; i < 100; i++ {
 			for _, objName := range s.SessionBus.Names() {
 				if objName == needle {
 					return
 				}
-				time.Sleep(1 * time.Second)
 			}
-
+			time.Sleep(10 * time.Millisecond)
 		}
-		c.Fatalf("%s does not appeared on the bus", needle)
+		c.Fatalf("%s has not appeared on the bus", needle)
 	}()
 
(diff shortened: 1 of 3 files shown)


cmd/snap/cmd_userd_test.go
```diff
@@ -69,16 +69,15 @@ func (s *userdSuite) TestUserd(c *C) {
 		}()
 
 		needle := "io.snapcraft.Launcher"
-		for i := 0; i < 10; i++ {
+		for i := 0; i < 100; i++ {
```
mvo5: The total wait of this loop is now one second - this may be an issue on slow VMs (think autopkgtest). Not sure if it is, just saysing that this changes the total wait of this loop.

bboozzoo: Could we make the end value conditional on `testing.Short()`?

chipaca: 1 second seemed plenty to me, but I can kick it to 10 seconds.

chipaca: problem is, if you're using `-short` because you're on a slow machine, and when `testing.Short()` we wait less, ...

chipaca:
I'll bump the loop to 1000, tonight  probably (don't want this silly thing spamming travis during the day)

PR #5947, opened 2018-10-08
many: cleanup remaining parallel installs TODOs

bboozzoo:
Cleanup TODOs where usage of InstanceName() vs. SnapName() is clear now.

### DIFF
--- a/snap/validate.go
+++ b/snap/validate.go
@@ -84,8 +84,6 @@ func ValidateInstanceName(instanceName string) error {
 		return err
 	}
 	if !validInstanceKey.MatchString(instanceKey) {
-		// TODO parallel-install: extend the error message once snap
-		// install help has been updated
 		return fmt.Errorf("invalid instance key: %q", instanceKey)
 	}
 	return nil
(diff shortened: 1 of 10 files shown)


snap/validate.go
```diff
@@ -282,8 +280,7 @@ func validateSocketAddrPath(socket *SocketInfo, fieldName string, path string) e
 }
 
 func validateSocketAddrAbstract(socket *SocketInfo, fieldName string, path string) error {
-	// TODO parallel-install: use of proper instance/store name, discuss socket activation in parallel install world
-	prefix := fmt.Sprintf("@snap.%s.", socket.App.Snap.InstanceName())
+	prefix := fmt.Sprintf("@snap.%s.", socket.App.Snap.SnapName())
```
stolowski: Why this change?

bboozzoo: I should probably add a comment here. The `path` comes from `snap.yaml`, so it can only match snap name (`foo`), but not the actual instance name (`foo_bar`). There will be an obvious conflict in the system if there is more than one snap with the same abstract socket address, but we expect the adminstrator to update the `snap.*.socket` files appropriately.

stolowski: Right, I see. Yes, I think a comment here makes sense.

PR #5949, opened 2018-10-08
osutil,asserts,daemon: support force password change in system-user assertion

mvo5:
This PR adds a new optional `force-password-change: {true,false}` field to the system-user assertion. When set to `true` it will force the user to change the password on first login.

This addresses https://bugs.launchpad.net/snappy/+bug/1774509

### DIFF
--- a/osutil/user_test.go
+++ b/osutil/user_test.go
@@ -169,6 +169,9 @@ func (s *createUserSuite) TestAddUserWithPasswordForceChange(c *check.C) {
 	c.Check(s.mockAddUser.Calls(), check.DeepEquals, [][]string{
 		{"adduser", "--force-badname", "--gecos", "my gecos", "--disabled-password", "karl.popper"},
 	})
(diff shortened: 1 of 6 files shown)


osutil/user_test.go
```diff
@@ -149,7 +152,38 @@ func (s *createUserSuite) TestAddUserWithPassword(c *check.C) {
...
+		Gecos:               "my gecos",
+		Password:            "$6$salt$hash",
+		ForcePasswordChange: true,
+	})
+	c.Assert(err, check.IsNil)
+
+	c.Check(s.mockAddUser.Calls(), check.DeepEquals, [][]string{
+		{"adduser", "--force-badname", "--gecos", "my gecos", "--disabled-password", "karl.popper"},
+	})
+	c.Check(s.mockPasswd.Calls(), check.DeepEquals, [][]string{
```
bboozzoo: For completeness you can check whether `usermod` was called too.

PR #5951, opened 2018-10-09
spread-shellcheck: fix interleaved error messages, tweaks

bboozzoo:
When shellcheck identifies erorrs in many files, the error log will be
interleaved as the logging is happening from inside of the helper ran by a
thread pool executor. Tweak the code so that the helpers return the error
instead of doing the logging directly.

Minor logging cleanup.

### DIFF
--- a/spread-shellcheck
+++ b/spread-shellcheck
@@ -66,9 +66,10 @@ class ShellcheckRunError(Exception):
 
 
 class ShellcheckError(Exception):
-    def __init__(self):
+    def __init__(self, path):
         super().__init__()
         self.sectionerrors = {}
+        self.path = path
 
     def addfailure(self, section, error):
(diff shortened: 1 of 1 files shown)


spread-shellcheck
```diff
@@ -166,19 +163,24 @@ def checkpath(loc, max_workers):
...
-            checkfile(entry)
-        except ShellcheckError as serr:
+            checkfile(path)
+        except ShellcheckError as err:
+            return err
+        return None
+
+    with ThreadPoolExecutor(max_workers=max_workers) as executor:
+        for serr in executor.map(check1path, locations):
+            if not isinstance(serr, ShellcheckError):
```
chipaca: wouldn't this work better with `if serr is None`?

PR #5955, opened 2018-10-09
cmd/snap, tests: snapshots for all

chipaca:
This introduces a way for users to create, list, verify, restore, and
remove snapshots of their snap's data and configuration.

Includes also #5913 so ignore that part of it (easy: they're in separate files).


chipaca:
The spread failure is real: 14.04 doesn't have a `runuser`. We'd originally gone with that because it was in `util-linux`, which is more likely to be installed already. I'll push a separate PR that falls back to `sudo` if `runuser` is missing.

cmd/snap/cmd_snapshot.go
```diff
@@ -0,0 +1,297 @@
...
+	Users      string `long:"users"`
+	Positional struct {
+		ID    snapshotID          `positional-arg-name:"<id>"`
+		Snaps []installedSnapName `positional-arg-name:"<snap>"`
+	} `positional-args:"yes" required:"yes"`
+}
+
+var shortCheckHelp = i18n.G("Check a snapshot")
+var longCheckHelp = i18n.G(`
+The check-snapshot command checks a snapshot against its hashsums.
```
bboozzoo: Maybe something more user friendly, `The check snapshot command verifies correctness of given snapshot`, or `.. consistency..` although that feels a bit technical.

cmd/snap/cmd_snapshot.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+	_, err = x.wait(changeID)
+	if err == noWait {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	// TODO: also mention the home archives that were actually restoreed
```
bboozzoo: s/restoreed/restored/

Sounds like something that could be mentioned in the help message?

cmd/snap/cmd_snapshot.go
```diff
@@ -0,0 +1,297 @@
...
+		fmt.Fprintf(Stdout, i18n.G("Snapshot #%d of %s restored.\n"),
+			x.Positional.ID, strutil.Quoted(snaps))
+	} else {
+		fmt.Fprintf(Stdout, i18n.G("Snapshot #%d restored.\n"), x.Positional.ID)
+	}
+	return nil
+}
+
+func init() {
+	addCommand("saved",
```
bboozzoo: How about `list`? Might be just me, but list feels more natural. (i.e. like we have `snap list`)

bboozzoo: Apparently `saved` was decided upon after some discussion. Don't wan't want to start the bikeshed again.

cmd/snap/cmd_snapshot.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+	if len(list) == 0 {
+		fmt.Fprintln(Stdout, i18n.G("No snapshots found."))
+		return nil
+	}
+	w := tabWriter()
+	defer w.Flush()
+
+	fmt.Fprintf(w, "%s\t%s\t%s\t%s\t%s\t%s\t%s\n",
+		// TRANSLATORS: 'Set' as in group or bag of things
```
zyga: I really appreciate the attention to detail here!

cmd/snap/cmd_snapshot.go
```diff
@@ -0,0 +1,297 @@
...
+		i18n.G("Set"),
+		"Snap",
+		i18n.G("Age"),
```
zyga: `# Translators: as in how old something is, not as in the stone age`

cmd/snap/cmd_snapshot.go
```diff
@@ -0,0 +1,297 @@
...
+		fmt.Fprintf(Stdout, i18n.NG("Snapshot #%d of snap %s forgotten.\n", "Snapshot #%d of snaps %s forgotten.\n", len(snaps)), x.Positional.ID, strutil.Quoted(snaps))
+	} else {
+		fmt.Fprintf(Stdout, i18n.G("Snapshot #%d forgotten.\n"), x.Positional.ID)
+	}
+	return nil
+}
+
+type checkSnapshotCmd struct {
+	waitMixin
+	Users      string `long:"users"`
```
zyga: There's a small inconsistency in enumerable arguments, for users we need to say `--users foo,bar` but for snap names we just list them `foo bar`. I don't think this is easy to improve but I just wanted to raise this.

cmd/snap/cmd_snapshot.go
```diff
@@ -0,0 +1,297 @@
...
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	// TODO: also mention the home archives that were actually checked
+	if len(snaps) > 0 {
+		// TRANSLATORS: the %s is a comma-separated list of quoted snap names
+		fmt.Fprintf(Stdout, i18n.G("Snapshot #%d of snaps %s check passed.\n"),
```
zyga: Perhaps `.... passed checks` or `... verified successfully` or `... is correct` or something similar. I think the sentence reads backwards a little as-is.

cmd/snap/cmd_snapshot.go
```diff
@@ -0,0 +1,297 @@
...
+			x.Positional.ID, strutil.Quoted(snaps))
+	} else {
+		fmt.Fprintf(Stdout, i18n.G("Snapshot #%d check passed.\n"), x.Positional.ID)
```
zyga: Same remark as above.

cmd/snap/cmd_snapshot.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+var shortRestoreHelp = i18n.G("Restore a snapshot")
+var longRestoreHelp = i18n.G(`
+The restore command restores a snapshot.
+`)
+
+func (x *restoreCmd) Execute([]string) error {
+	snaps := installedSnapNames(x.Positional.Snaps)
+	var users []string
```
zyga: Perhaps a small helper for handling the users field across many commands:
```
func splitUsers(users string) []string {
   if len(users) > 0 {
     return strings.Split(users, ",")
   }
   return nil
}
```

cmd/snap/cmd_snapshot.go
zyga: Perhaps that there is some logic that indicates what data is restored but I don't know this myself so I cannot specify.

cmd/snap/cmd_snapshot.go
```diff
@@ -0,0 +1,297 @@
...
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	// TODO: also mention the home archives that were actually restoreed
+	if len(snaps) > 0 {
+		// TRANSLATORS: the %s is a comma-separated list of quoted snap names
+		fmt.Fprintf(Stdout, i18n.G("Snapshot #%d of %s restored.\n"),
```
zyga: Nitpicking mode: `Restored snapshot #%d of snaps %s\n`

cmd/snap/cmd_snapshot.go
```diff
@@ -0,0 +1,297 @@
...
+func init() {
+	addCommand("saved",
+		shortSavedHelp,
+		longSavedHelp,
+		func() flags.Commander {
+			return &savedCmd{}
+		},
+		durationDescs.also(map[string]string{
+			// TRANSLATORS: This should not start with a lowercase letter.
+			"id": i18n.G("Only list this snapshot."),
```
zyga: Suggestion: `Show only a specific snapshot`

cmd/snap/cmd_snapshot.go
```diff
@@ -0,0 +1,297 @@
...
+		nil)
+
+	addCommand("save",
+		shortSaveHelp,
+		longSaveHelp,
+		func() flags.Commander {
+			return &saveCmd{}
+		}, durationDescs.also(waitDescs).also(map[string]string{
+			// TRANSLATORS: This should not start with a lowercase letter.
+			"users": i18n.G("Only snapshot these users' files (a comma-separated list)."),
```
zyga: Suggestion: `Snapshot data of only specific users (comma separated).`

cmd/snap/cmd_snapshot.go
```diff
@@ -0,0 +1,297 @@
...
+		}), nil)
+
+	addCommand("restore",
+		shortRestoreHelp,
+		longRestoreHelp,
+		func() flags.Commander {
+			return &restoreCmd{}
+		}, waitDescs.also(map[string]string{
+			// TRANSLATORS: This should not start with a lowercase letter.
+			"users": i18n.G("Only restore these users' files (a comma-separated list)."),
```
zyga: Suggestion: `Restore data of only specific users (comma separated).`

cmd/snap/cmd_snapshot.go
```diff
@@ -0,0 +1,297 @@
...
+		}, waitDescs, nil)
+
+	addCommand("check-snapshot",
+		shortCheckHelp,
+		longCheckHelp,
+		func() flags.Commander {
+			return &checkSnapshotCmd{}
+		}, waitDescs.also(map[string]string{
+			// TRANSLATORS: This should not start with a lowercase letter.
+			"users": i18n.G("Only check these users' files (a comma-separated list)."),
```
zyga: Suggestion: `Check data of only specific users (comma separated).`

cmd/snap/cmd_snapshot.go
chipaca: good catch

cmd/snap/cmd_snapshot.go
bboozzoo: As sugested on IRC, maybe we should extend the long help of `restore` and `save` a bit. Include something like this:
```
...
When no users are provided, restoring a snapshot restores data of all users.
```

cmd/snap/cmd_snapshot.go
chipaca: I'll move `daemon.splitQS` to `strutil`

chipaca: (... in a followup)

chipaca:
@degville I took your suggestions, implemented them, and then iterated a few more times. Then I reorganised the code so I could see all the help together, and iterated a bit more.

I think you'll like the result. But it might still have room for improvement, so I'll leave this open just in case.

cmd/snap/cmd_snapshot.go
```diff
@@ -43,10 +43,67 @@ type savedCmd struct {
...
+data of the snaps included in the specified snapshot.
+
+The check operation runs the same data integrity verification that is
+performed when a snapshot is restored.
+
+You can choose to check all the data in a snapshot, which is the
+default, or you can specify the data of which snaps to check, or for
+which users, or a combination of these.
+
+If a snap is included in a check-snapshot operation, excluding its
```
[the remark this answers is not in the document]
degville: It is not currently possible to exclude a specific snap's system and configuration data in a restore operation?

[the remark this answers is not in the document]
chipaca: doesn't that read like we're saying that if the snapshot is for foo, bar and baz snaps, you can't exclude system and config data from foo and not not exclude it from the others?

cmd/snap/cmd_snapshot.go
```diff
@@ -43,10 +43,67 @@ type savedCmd struct {
...
+	shortSaveHelp    = i18n.G("Save a snapshot of the current data")
+	shortForgetHelp  = i18n.G("Delete a snapshot")
+	shortCheckHelp   = i18n.G("Check a snapshot")
+	shortRestoreHelp = i18n.G("Restore a snapshot")
+)
+
 var longSavedHelp = i18n.G(`
-The saved command lists the snapshots that have been created previously with
-the 'save' command.
+The saved command lists the snapshots that have been created
```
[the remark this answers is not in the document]
zyga: @chipaca could we do some ANSI art magic here to apply bold or italics to the actual command being mentioned. It would show for a good chunk of people and would immediately look better IMO (though not in this PR)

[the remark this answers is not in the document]
chipaca: I've been wanting to find a way to do this for a while. What I want to do is have a formatter that takes e.g. `%Q` to mean "special-quoted" or something, that uses italics and typographic quotes if available, falling back to simple quotes.

It's fiddly work, and would mess up widths if done without care. I know how to do it, but I haven't gotten to it yet.

cmd/snap/cmd_snapshot.go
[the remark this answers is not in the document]
degville: yeah, you're right. It's the use of included and excluded in the original text that confuses me, but I can't really think of anything better and it does work.

PR #5957, opened 2018-10-10
overlord/snapshotstate/backend: fall back on sudo when no runuser

chipaca:
Without this change, snapshots don't work on 14.04.

### DIFF
--- a/overlord/snapshotstate/backend/helpers.go
+++ b/overlord/snapshotstate/backend/helpers.go
@@ -171,35 +171,39 @@ func pickUserWrapper() string {
 	return ""
 }
 
-var sudo = pickUserWrapper()
+var userWrapper = pickUserWrapper()
 
-// maybeRunuserCommand returns an exec.Cmd that will, if the current
-// effective user id is 0 and username is not "root", call runuser(1)
-// (or sudo(8) if runuser(1) isn't available) to change to the given
-// username before running the given command.
+// tarAsUser returns an exec.Cmd that will, if the current effective user id is
+// 0 and username is not "root", and if either runuser(1) or sudo(8) are found
+// on the PATH, run tar as the given user.
 //
-// If username is "root", or the effective user id is 0, the given
-// command is passed directly to exec.Command.
+// If the effective user id is not 0, or username is "root", exec.Command is
+// used directly; changing the user id would fail (in the first case) or be a
+// no-op (in the second).
 //
-// TODO: maybe move this to osutil
-func maybeRunuserCommand(username string, args ...string) *exec.Cmd {
-	if sysGeteuid() == 0 && username != "root" && sudo != "" {
-		sudoArgs := make([]string, len(args)+4)
-		copy(sudoArgs[4:], args)
-		sudoArgs[0] = sudo
-		sudoArgs[1] = "-u"
-		sudoArgs[2] = username
-		sudoArgs[3] = "--"
-		return &exec.Cmd{
-			Path: sudo,
-			Args: sudoArgs,
+// If neither runuser nor sudo are found on the path, exec.Command is also used
+// directly. This will result in tar running as root in this situation (so it
+// will fail if on NFS; I don't think there's an attack vector though).
+func tarAsUser(username string, args ...string) *exec.Cmd {
+	if sysGeteuid() == 0 && username != "root" {
+		if userWrapper != "" {
+			uwArgs := make([]string, len(args)+5)
+			copy(uwArgs[5:], args)
+			uwArgs[0] = userWrapper
+			uwArgs[1] = "-u"
+			uwArgs[2] = username
+			uwArgs[3] = "--"
+			uwArgs[4] = "tar"
+			return &exec.Cmd{
+				Path: userWrapper,
+				Args: uwArgs,
+			}
 		}
+		// TODO: use warnings instead
+		logger.Noticef("No user wrapper found; running tar for user data as root. Please make sure 'sudo' or 'runuser' (from util-linux) is on $PATH to avoid this.")
 	}
 
-	// TODO: use warnings instead
-	logger.Noticef("No user wrapper found; running tar for user data as root. Please make sure 'sudo' or 'runuser' (from util-linux) is on $PATH to avoid this.")
-
-	return exec.Command(args[0], args[1:]...)
+	return exec.Command("tar", args...)
 }
 
 func MockUserLookup(newLookup func(string) (*user.User, error)) func() {
--- a/overlord/snapshotstate/backend/helpers.go
+++ b/overlord/snapshotstate/backend/helpers.go
@@ -33,6 +33,7 @@ import (
 
 	"github.com/snapcore/snapd/client"
 	"github.com/snapcore/snapd/dirs"
+	"github.com/snapcore/snapd/logger"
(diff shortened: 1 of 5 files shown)


overlord/snapshotstate/backend/helpers.go
```diff
@@ -152,26 +152,43 @@ func allUsers() ([]*user.User, error) {
...
 // TODO: maybe move this to osutil
 func maybeRunuserCommand(username string, args ...string) *exec.Cmd {
-	if username == "root" || sys.Geteuid() != 0 {
-		// runuser only works for euid 0, and doesn't make sense for root
-		return exec.Command(args[0], args[1:]...)
+	if sysGeteuid() == 0 && username != "root" {
+		// runuser and sudo happen to work the same way in this case.
+		// The main reason to prefer runuser over sudo is that runuser
+		// is part of util-linux, which is considered essential,
+		// whereas sudo is an addon which could be removed.
```
bboozzoo: Thanks for adding a note in case we forget the reasoning again.

overlord/snapshotstate/backend/helpers.go
```diff
@@ -152,26 +152,43 @@ func allUsers() ([]*user.User, error) {
...
+		for _, cmd := range []string{"runuser", "sudo"} {
+			if lp, err := execLookPath(cmd); err == nil {
```
bboozzoo: Nitpick, looking for the actual command could be done in `sync.Once()`, eg.
```
var cmdSeekedFor sync.Once
var sudoOrRunuser string
func pickUserWrapper() {
    for _, cmd := range []string{"runuser", "sudo"} {
        if lp, err := execLookPath(cmd); err != nil {
            sudoOrRunuser = lp
            break
        }
    }
}

func maybeRunuserCommand(.....) {
    cmdSeekedFor.Do(pickUserWrapper)
    if sysGeteuid() == 0 && username != "root" && sudoOnRunuser != "" {
        ....
    }
    return exec.Command(args[0], args[1:]...)
}
```

overlord/snapshotstate/backend/helpers.go
```diff
@@ -152,26 +152,43 @@ func allUsers() ([]*user.User, error) {
...
 func maybeRunuserCommand(username string, args ...string) *exec.Cmd {
-	if username == "root" || sys.Geteuid() != 0 {
-		// runuser only works for euid 0, and doesn't make sense for root
-		return exec.Command(args[0], args[1:]...)
+	if sysGeteuid() == 0 && username != "root" {
+		// runuser and sudo happen to work the same way in this case.
+		// The main reason to prefer runuser over sudo is that runuser
+		// is part of util-linux, which is considered essential,
+		// whereas sudo is an addon which could be removed.
+		for _, cmd := range []string{"runuser", "sudo"} {
```
bboozzoo: Should we log an error/warning when neither is found?

chipaca: yes

overlord/snapshotstate/backend/helpers.go
chipaca: hm, might as well do it in `init`, n'eh?

overlord/snapshotstate/backend/helpers.go
```diff
@@ -152,26 +153,53 @@ func allUsers() ([]*user.User, error) {
...
+var (
+	sysGeteuid   = sys.Geteuid
+	execLookPath = exec.LookPath
+)
+
+func pickUserWrapper() string {
+	// runuser and sudo happen to work the same way in this case.
+	// The main reason to prefer runuser over sudo is that runuser
+	// is part of util-linux, which is considered essential,
+	// whereas sudo is an addon which could be removed.
```
zyga: Hold on, but the rationale for adding this branch is that we ... don't have runuser everywhere, right?

overlord/snapshotstate/backend/helpers.go
```diff
@@ -152,26 +153,53 @@ func allUsers() ([]*user.User, error) {
...
+	for _, cmd := range []string{"runuser", "sudo"} {
+		if lp, err := execLookPath(cmd); err == nil {
+			return lp
+		}
+	}
+	return ""
+}
+
+var sudo = pickUserWrapper()
```
zyga: I think the variable name is no longer appropriate. Perhaps `launcher`?

overlord/snapshotstate/backend/helpers.go
```diff
@@ -152,26 +153,53 @@ func allUsers() ([]*user.User, error) {
...
 // If username is "root", or the effective user id is 0, the given
 // command is passed directly to exec.Command.
 //
 // TODO: maybe move this to osutil
 func maybeRunuserCommand(username string, args ...string) *exec.Cmd {
-	if username == "root" || sys.Geteuid() != 0 {
-		// runuser only works for euid 0, and doesn't make sense for root
-		return exec.Command(args[0], args[1:]...)
+	if sysGeteuid() == 0 && username != "root" && sudo != "" {
+		sudoArgs := make([]string, len(args)+4)
```
zyga: While perhaps a little bit YAGNI for now, since I assume the code works, I would consider having the launcher be an interface with a function RunWithArgs() that encapsulates this kind of knowledge. Right now there's a hidden assumption that "sudo" behaves like sudo, down the the command line handling.

overlord/snapshotstate/backend/helpers.go
```diff
@@ -152,26 +153,53 @@ func allUsers() ([]*user.User, error) {
...
 	}
-	sudoArgs := make([]string, len(args)+3)
-	copy(sudoArgs[3:], args)
-	sudoArgs[0] = "-u"
-	sudoArgs[1] = username
-	sudoArgs[2] = "--"
 
-	return exec.Command("runuser", sudoArgs...)
+	// TODO: use warnings instead
+	logger.Noticef("No user wrapper found; running tar for user data as root. Please make sure 'sudo' or 'runuser' (from util-linux) is on $PATH to avoid this.")
```
zyga: Hmm, when this is the case what would happen? Should things just fail loudly?

chipaca: I'm not sure. It'll break on NFS and you'll get an error. Other than that it should be fine.

overlord/snapshotstate/backend/helpers.go
chipaca: I could unify things:
* `pickUserWrapper` is alright, i think
* so `sudo` -> `userWrapper`
* so `maybeRunuserCommand` -> `userWrapCommand` ?

overlord/snapshotstate/backend/helpers.go
zyga: Ah, that's a good thing to know. We can already check if home is using NFS so perhaps all of this can be optional eventually?

overlord/snapshotstate/backend/helpers.go
zyga: `runAsUser`

overlord/snapshotstate/backend/helpers.go
bboozzoo: May be worth mentioning in the comment that runuser is avaialble since util-linux 2.23 and some old systems that we support are on an even older version.

PR #5959, opened 2018-10-10
systemd: extend Status() to work for socket and timer units

bboozzoo:
In wake of extending reporting of service statuses in `snap services` output we
will need to ask for the status of timer and socket units. Those do not have the
Type property. Extend systemd.Status() to work for timers and sockets. While at
it, tweak the field names of returned struct to closer reflect that those are
units (and not necessarily services).


systemd/systemd.go
```diff
@@ -216,66 +216,76 @@ func (*systemd) LogReader(serviceNames []string, n int, follow bool) (io.ReadClo
...
-	cur := &ServiceStatus{}
+	sts := make([]*UnitStatus, 0, len(unitNames))
+	cur := &UnitStatus{}
 	seen := map[string]bool{}
 
 	for _, bs := range statusregex.FindAllSubmatch(bs, -1) {
 		if len(bs[0]) == 0 {
 			// systemctl separates data pertaining to particular services by an empty line
+			unitType := filepath.Ext(cur.UnitName)
+			expected := expectedAll
```
zyga: The handling of expected attributes is a little bit confusing. Could we just have expectedSocket, expectedTimer, expectedMount, expectedService? I think it is fine to repeat something explicitly and pick the right _expected..._ variable based on the extension.

systemd/systemd.go
```diff
@@ -216,66 +216,76 @@ func (*systemd) LogReader(serviceNames []string, n int, follow bool) (io.ReadClo
...
 var statusregex = regexp.MustCompile(`(?m)^(?:(.+?)=(.*)|(.*))?$`)
 
-type ServiceStatus struct {
-	Daemon          string
-	ServiceFileName string
-	Enabled         bool
-	Active          bool
+type UnitStatus struct {
+	Daemon   string
+	UnitName string
```
chipaca: d'oh, so much better

PR #5960, opened 2018-10-11
snap, wrappers: support restart-delay, generate RestartSec=<value> in service units

bboozzoo:
This was asked for in the forum https://forum.snapcraft.io/t/expose-a-more-consistent-subset-of-systemds-service-directives/2268/52

It is possible to control the delay systemd applies between service restarts. The snap YAML looks like this:
```
name: foo
version: 1.0
apps:
  foo:
    daemon: simple
    restart-delay: 12s
```

(cc @kyrofa )

### DIFF
--- a/snap/validate.go
+++ b/snap/validate.go
@@ -623,7 +623,7 @@ func validateAppRestart(app *AppInfo) error {
 
 	if app.RestartDelay != 0 {
 		if !app.IsService() {
-			return fmt.Errorf("cannot define restart-delay for application %q as it's not a service", app.Name)
+			return fmt.Errorf("application %q must be a service to define restart-delay", app.Name)
 		}
 
 		if app.RestartDelay < 0 {
@@ -633,7 +633,7 @@ func validateAppRestart(app *AppInfo) error {
 
 	if app.RestartCond != "" {
 		if !app.IsService() {
-			return fmt.Errorf("cannot define restart-condition for application %q as it's not a service", app.Name)
+			return fmt.Errorf("application %q must be a service to define restart-condition", app.Name)
 		}
 	}
 	return nil
--- a/snap/validate.go
+++ b/snap/validate.go
@@ -614,6 +614,31 @@ func validateAppTimer(app *AppInfo) error {
 	return nil
 }
 
+func validateAppRestart(app *AppInfo) error {
+	// app.RestartCond value is validated when unmarshalling
+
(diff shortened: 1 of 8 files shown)


snap/validate.go
```diff
@@ -614,6 +614,31 @@ func validateAppTimer(app *AppInfo) error {
...
+func validateAppRestart(app *AppInfo) error {
+	// app.RestartCond value is validated when unmarshalling
+
+	if app.RestartDelay == 0 && app.RestartCond == "" {
+		return nil
+	}
+
+	if app.RestartDelay != 0 {
+		if !app.IsService() {
+			return fmt.Errorf("cannot define restart-delay for application %q as it's not a service", app.Name)
```
zyga: Nitpicking on this: `restart-delay is only applicable to services`

snap/validate.go
```diff
@@ -614,6 +614,31 @@ func validateAppTimer(app *AppInfo) error {
...
+		}
+
+		if app.RestartDelay < 0 {
+			return fmt.Errorf("cannot use a negative restart-delay for application %q", app.Name)
```
zyga: `restart-delay cannot be negative`

snap/validate.go
```diff
@@ -614,6 +614,31 @@ func validateAppTimer(app *AppInfo) error {
...
+		}
+	}
+
+	if app.RestartCond != "" {
+		if !app.IsService() {
+			return fmt.Errorf("cannot define restart-condition for application %q as it's not a service", app.Name)
```
zyga: `restart-condition is only applicable to services`

Now with all three of those it would be great if we had a way to scope the error in the context of an application. This would require reworking some of the validation to be done per app and encapsulate the error with the app name. Something like `in application "foo": restart-delay cannot be negative, restart-condition is only applicable to services", etc.

Feel free to ignore all three suggestions though, this may require a rework of the error handling logic to implement.

wrappers/services.go
```diff
@@ -389,6 +389,9 @@ X-Snappy=yes
 ExecStart={{.App.LauncherCommand}}
 SyslogIdentifier={{.App.Snap.InstanceName}}.{{.App.Name}}
 Restart={{.Restart}}
+{{- if .App.RestartDelay}}
+RestartSec={{.App.RestartDelay.Seconds}}
```
zyga: The snapd version of the variable name is so much better than the systemd version!

PR #5961, opened 2018-10-11
snap/pack, snap/squashfs: use type to determine mksquashfs args

chipaca:
We should be using different args for 'snap pack' than we are doing:

* cores and bases should be built with xattrs
* everything else should be squashed to root
* we should use mksquashfs from core if available

this fixes all that. Also, splits squashfs tests into their own
package (most of the lines in the diff are probably about this tbh).

### DIFF
--- a/snap/squashfs/export_test.go
+++ b/snap/squashfs/export_test.go
@@ -1,0 +1,75 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2018 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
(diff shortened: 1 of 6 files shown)


snap/squashfs/export_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+func MockFromCore(newFromCore func(string, ...string) (*exec.Cmd, error)) (restore func()) {
+	oldFromCore := osutilCommandFromCore
+	osutilCommandFromCore = newFromCore
+	return func() {
+		osutilCommandFromCore = oldFromCore
+	}
+}
+
+func Alike(a, b os.FileInfo, c *check.C, comment check.CommentInterface) {
```
stolowski: I think this helper would benefit from a short comment.

PR #5966, opened 2018-10-11
snap: overhaul validation error messages

bboozzoo:
Overhaul error messages from snap validation in the hope of making them more useful by adding context.

### DIFF
--- a/snap/validate.go
+++ b/snap/validate.go
@@ -571,11 +571,11 @@ func validateAppOrderNames(app *AppInfo, dependencies []string) error {
 		// dependency is not defined
 		other, ok := app.Snap.Apps[dep]
 		if !ok {
-			return fmt.Errorf("refers to missing application %q in before/after", dep)
+			return fmt.Errorf("before/after references a missing application %q", dep)
 		}
 
 		if !other.IsService() {
-			return fmt.Errorf("refers to non-service application %q in before/after", dep)
+			return fmt.Errorf("before/after references a non-service application %q", dep)
 		}
 	}
 	return nil
--- a/snap/validate.go
+++ b/snap/validate.go
@@ -276,7 +276,7 @@ func validateSocketAddrPath(socket *SocketInfo, fieldName string, path string) e
 
 	if !(strings.HasPrefix(path, "$SNAP_DATA/") || strings.HasPrefix(path, "$SNAP_COMMON/")) {
 		return fmt.Errorf(
-			"invalid %q: only $SNAP_DATA and $SNAP_COMMON prefixes are allowed", fieldName)
+			"invalid %q: must have a prefix of $SNAP_DATA or $SNAP_COMMON", fieldName)
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


snap/validate.go
```diff
@@ -270,12 +271,12 @@ func validateSocketAddr(socket *SocketInfo, fieldName string, address string) er
...
 func validateSocketAddrPath(socket *SocketInfo, fieldName string, path string) error {
 	if clean := filepath.Clean(path); clean != path {
-		return fmt.Errorf("socket %q has invalid %q: %q should be written as %q", socket.Name, fieldName, path, clean)
+		return fmt.Errorf("invalid %q: %q should be written as %q", fieldName, path, clean)
 	}
 
 	if !(strings.HasPrefix(path, "$SNAP_DATA/") || strings.HasPrefix(path, "$SNAP_COMMON/")) {
 		return fmt.Errorf(
-			"socket %q has invalid %q: only $SNAP_DATA and $SNAP_COMMON prefixes are allowed", socket.Name, fieldName)
+			"invalid %q: only $SNAP_DATA and $SNAP_COMMON prefixes are allowed", fieldName)
```
chipaca: I've got a problem with this one: `only FOO and BAR prefixes are allowed` does not mean quite the same as `must have a prefix of FOO or BAR`, which is the case here. Granted this one was wrong before :-)

snap/validate.go
```diff
@@ -310,13 +311,13 @@ func validateSocketAddrNetHost(socket *SocketInfo, fieldName string, address str
 		}
 	}
 
-	return fmt.Errorf("socket %q has invalid %q address %q", socket.Name, fieldName, address)
+	return fmt.Errorf("invalid %q address %q", fieldName, address)
```
chipaca: maybe add `(only addresses referring to localhost are valid)`? or is that explained elsewhere?

snap/validate.go
```diff
@@ -304,19 +305,20 @@ func validateSocketAddrNet(socket *SocketInfo, fieldName string, address string)
...
-		if address == validAddress {
+	validAddresses := []string{"127.0.0.1", "[::1]", "[::]"}
+	for _, valid := range validAddresses {
+		if address == valid {
 			return nil
 		}
 	}
 
-	return fmt.Errorf("socket %q has invalid %q address %q", socket.Name, fieldName, address)
+	return fmt.Errorf("invalid %q address %q, must be one of: %s", fieldName, address, strings.Join(validAddresses, ", "))
```
stolowski: Nice.

snap/validate.go
```diff
@@ -562,20 +564,18 @@ func validateAppOrderCycles(apps map[string]*AppInfo) error {
...
+		return errors.New("must be a service to define before/after ordering")
 	}
 
 	for _, dep := range dependencies {
 		// dependency is not defined
 		other, ok := app.Snap.Apps[dep]
 		if !ok {
-			return fmt.Errorf("application %q refers to missing application %q in before/after",
-				app.Name, dep)
+			return fmt.Errorf("refers to missing application %q in before/after", dep)
```
stolowski: +1 for changing the original message, but the new version feels like it's missing a subject. How about: `reference to a missing application ....`? Perhaps @degville  can advise again.

snap/validate.go
```diff
@@ -562,20 +564,18 @@ func validateAppOrderCycles(apps map[string]*AppInfo) error {
...
 		}
 
 		if !other.IsService() {
-			return fmt.Errorf("application %q refers to non-service application %q in before/after",
-				app.Name, dep)
+			return fmt.Errorf("refers to non-service application %q in before/after", dep)
```
stolowski: Same.

snap/validate.go
degville: +1 on @stolowski's suggestion, or how about `before/after references a missing application`?

snap/validate.go
degville: `before/after references a non-service application, %q` ... ?

PR #5968, opened 2018-10-11
interfaces: updates for default, screen-inhibit-control, tpm, {hardware,system,network}-observe

jdstrand:
* interfaces/default: allow read of /etc/sysconfig/clock
* interfaces/screen-inhibit-control: also allow interface="org.gnome.ScreenSaver"
* interfaces/{hardware,system,network}-observe: allow reading /proc/modules
* snap-confine.apparmor.in: add bug reference for pivot_root
* interfaces/password-manager-service: fix typo in comment
* interfaces/tpm: fix apparmor rules to match what is allowed by udev
* interfaces/default: allow snaps to use/create DBus private buses (see commit message for more detail

References:
* https://forum.snapcraft.io/t/hiri-wont-start/7075
* https://forum.snapcraft.io/t/what-interface-i-need-to-run-lsmod/7441/3
* https://forum.snapcraft.io/t/how-to-use-dbus-run-session-on-ubuntu-core/7077/2

### DIFF
--- a/interfaces/apparmor/template.go
+++ b/interfaces/apparmor/template.go
@@ -55,6 +55,7 @@ var defaultTemplate = `
   /etc/ld.so.preload r,
 
   # The base abstraction doesn't yet have this
+  /etc/sysconfig/clock r,
   /lib/terminfo/** rk,
   /usr/share/terminfo/** k,
   /usr/share/zoneinfo/** k,
--- a/interfaces/apparmor/template.go
+++ b/interfaces/apparmor/template.go
@@ -136,6 +136,8 @@ var defaultTemplate = `
   /{,usr/}bin/cpio ixr,
   /{,usr/}bin/cut ixr,
   /{,usr/}bin/date ixr,
+  /{,usr/}bin/dbus-daemon ixr,
+  /{,usr/}bin/dbus-run-session ixr,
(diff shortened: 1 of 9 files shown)


interfaces/apparmor/template.go
```diff
@@ -135,6 +136,8 @@ var defaultTemplate = `
   /{,usr/}bin/cpio ixr,
   /{,usr/}bin/cut ixr,
   /{,usr/}bin/date ixr,
+  /{,usr/}bin/dbus-daemon ixr,
+  /{,usr/}bin/dbus-run-session ixr,
```
niemeyer: When do we want snaps to run those?

jdstrand: Please see the individual commit message for this: https://github.com/snapcore/snapd/pull/5968/commits/af9edc528d94ce5fd33eb2eff3713b029c083303

They can run them any time they want to create an dbus private bus that is private to the snap. This isn't conceptually different from them setting up an abstract socket without dbus or simply shipping their own dbus-daemon and dbus-run-session. We could yank these two rules (but leave the others) so they must ship these, but I didn't see a benefit to that.

jdstrand:
Thanks!

PR #5975, opened 2018-10-12
ifacestate/hooks: only create interface hook tasks if hooks exist

stolowski:
Only create interface hooks tasks if relevant hook exists (i.e. hook file is present, or hook is explicitely listed in snap yaml - the Hooks map has both and is populated by ReadInfo). 

This should slightly reduce execution time as we avoid unneccessary no-op hook tasks which would be ignored by runHook anyway.

### DIFF
--- a/overlord/ifacestate/ifacestate.go
+++ b/overlord/ifacestate/ifacestate.go
@@ -101,7 +101,8 @@ func connect(st *state.State, plugSnap, plugName, slotSnap, slotName string, fla
 	//  - connect task
 	//  - connect-slot-<slot> hook
 	//  - connect-plug-<plug> hook
-	// The tasks run in sequence (are serialized by WaitFor).
+	// The tasks run in sequence (are serialized by WaitFor). The hooks are optional
+	// and their tasks are not created if not provided by snap.
 	// The prepare- hooks collect attributes via snapctl set.
 	// 'snapctl set' can only modify own attributes (plug's attributes in the *-plug-* hook and
 	// slot's attributes in the *-slot-* hook).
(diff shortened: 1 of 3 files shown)


overlord/ifacestate/ifacestate.go
```diff
@@ -176,56 +201,73 @@ func connect(st *state.State, plugSnap, plugName, slotSnap, slotName string, fla
 	connectInterface.Set("slot-static", slotStatic)
 	connectInterface.Set("plug-dynamic", emptyDynamicAttrs)
 	connectInterface.Set("slot-dynamic", emptyDynamicAttrs)
+	if prepareSlotConnection != nil {
+		connectInterface.WaitFor(prepareSlotConnection)
+	} else {
+		if preparePlugConnection != nil {
```
zyga: What is the point of the else {  } clause here? I would suspect that it would just be okay to have the else part equally unconditional as the if clause. Is the purpose to untangle a dependency loop if the slot prepare waits for plug prepare? If so, can we have a comment stating as much here? In case I'm wrong a comment that describes the purpose of this would be lovely as well.

overlord/ifacestate/ifacestate.go
```diff
@@ -176,56 +201,73 @@ func connect(st *state.State, plugSnap, plugName, slotSnap, slotName string, fla
...
+		}
 	}
 
-	summary = fmt.Sprintf(i18n.G("Run hook %s of snap %q"), connectPlugHookSetup.Hook, connectPlugHookSetup.Snap)
-	connectPlugConnection := hookstate.HookTaskWithUndo(st, summary, connectPlugHookSetup, undoConnectPlugHookSetup, initialContext)
-	connectPlugConnection.WaitFor(connectSlotConnection)
-
-	return state.NewTaskSet(preparePlugConnection, prepareSlotConnection, connectInterface, connectSlotConnection, connectPlugConnection), nil
+	ts := &state.TaskSet{}
+	for _, t := range []*state.Task{preparePlugConnection, prepareSlotConnection, connectInterface, connectSlotConnection, connectPlugConnection} {
```
zyga: Nice :)

overlord/ifacestate/ifacestate.go
stolowski: The only reason is to have simple dependency chain here (and make tests simpler thanks to that); there would be no harm in waiting on both tasks. 
FWTW the old code had simple 1-1 wait dependency as well.

overlord/ifacestate/ifacestate.go
```diff
@@ -101,7 +101,8 @@ func connect(st *state.State, plugSnap, plugName, slotSnap, slotName string, fla
 	//  - connect task
 	//  - connect-slot-<slot> hook
 	//  - connect-plug-<plug> hook
-	// The tasks run in sequence (are serialized by WaitFor).
+	// The tasks run in sequence (are serialized by WaitFor). The hooks are optional
+	// and their tasks are not created if not provided by snap.
```
bboozzoo: A positive scenario may be easier to follow than a double-negative one:
`... tasks are created when hook exists or is declared in the snap`

overlord/ifacestate/ifacestate.go
bboozzoo: I recall seeing a pattern (probably in `snapstate.go:doInstall()`?) where an extra variable, `var last *task.Task` or somesuch was assigned to when adding tasks and used for `WaitFor()`, though this sounds like a material for a follow-up.

stolowski: Right, I know what you mean. I'll see how it fits here in a potential followup PR, thanks.

PR #5978, opened 2018-10-12
interfaces/home: don't allow snaps to write to $HOME/bin

zyga:
The $HOME/bin directory is added to default PATH on some systems. To
prevent sandbox escape in this specific case deny writing to this
directory.

Bug-Link: https://bugs.launchpad.net/snapd/+bug/1797556

### DIFF
--- a/interfaces/builtin/home.go
+++ b/interfaces/builtin/home.go
@@ -75,10 +75,9 @@ owner @{HOME}/snap/ r,
 owner /run/user/[0-9]*/gvfs/{,**} r,
 owner /run/user/[0-9]*/gvfs/*/**  w,
 
-# On some systems HOME/bin is in the default path in ahead of system
-# directories. To prevent snaps from writing arbitrary executables waiting to
-# be launched by the user, deny writes to the directory altogether.
-deny @{HOME}/bin/{,**} w,
+# Disallow writes to the well-known directory included in
+# the user's PATH on several distributions
+deny @{HOME}/bin/{,**} wl,
 `
 
 const homeConnectedPlugAppArmorWithAllRead = `
--- a/interfaces/builtin/home.go
+++ b/interfaces/builtin/home.go
@@ -77,7 +77,7 @@ owner /run/user/[0-9]*/gvfs/*/**  w,
 
 # Disallow writes to the well-known directory included in
 # the user's PATH on several distributions
-deny @{HOME}/bin/{,**} wl,
+audit deny @{HOME}/bin/{,**} wl,
 `
 
 const homeConnectedPlugAppArmorWithAllRead = `
--- a/interfaces/builtin/home.go
+++ b/interfaces/builtin/home.go
@@ -74,6 +74,11 @@ owner @{HOME}/snap/ r,
 # files; only allow writes to files, not the mount point).
 owner /run/user/[0-9]*/gvfs/{,**} r,
 owner /run/user/[0-9]*/gvfs/*/**  w,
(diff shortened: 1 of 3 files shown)


tests/regression/lp-1797556/task.yaml
```diff
@@ -0,0 +1,19 @@
...
+    su -l -c 'mkdir -p /home/test/bin/' test
+restore: |
+    su -l -c 'rmdir /home/test/bin/ || true' test
+execute: |
+    test -d /home/test/bin
+    test ! -e /home/test/bin/evil
+    if snap debug sandbox-features | grep 'apparmor:.* kernel:file'; then
+        ! su -l -c 'test-snapd-sh.with-home-plug -c touch /home/test/bin/evil' test
+    fi
+    test ! -e /home/test/bin/evil
```
stolowski: Since this is used 3 times in the script, I'd make a variable for it, it always helps avoid nasty typos that go unnoticed in such negative checks.

zyga: I prefer to avoid such variables because they make the test a little bit harder to understand, in addition, in some cases the variable is hard to actually expand given the number of quoting levels.

interfaces/builtin/home.go
```diff
@@ -74,6 +74,11 @@ owner @{HOME}/snap/ r,
 # files; only allow writes to files, not the mount point).
 owner /run/user/[0-9]*/gvfs/{,**} r,
 owner /run/user/[0-9]*/gvfs/*/**  w,
+
+# On some systems HOME/bin is in the default path in ahead of system
+# directories. To prevent snaps from writing arbitrary executables waiting to
+# be launched by the user, deny writes to the directory altogether.
+deny @{HOME}/bin/{,**} w,
```
jdstrand: Please add 'l' otherwise they can create hardlinks in here to a file they control.

interfaces/builtin/home.go
```diff
@@ -74,6 +74,11 @@ owner @{HOME}/snap/ r,
 # files; only allow writes to files, not the mount point).
 owner /run/user/[0-9]*/gvfs/{,**} r,
 owner /run/user/[0-9]*/gvfs/*/**  w,
+
+# On some systems HOME/bin is in the default path in ahead of system
+# directories. To prevent snaps from writing arbitrary executables waiting to
+# be launched by the user, deny writes to the directory altogether.
```
jdstrand: The fact that $HOME/bin is ahead of system paths makes it particularly bad, but isn't the only reason for the change. I suggest:
```
  # Disallow writes to the well-known directory included in
  # the user's PATH on several distributions
```

jdstrand:
@mvo5 - when approved, please include in 2.36.

interfaces/builtin/home.go
```diff
@@ -74,6 +74,10 @@ owner @{HOME}/snap/ r,
 # files; only allow writes to files, not the mount point).
 owner /run/user/[0-9]*/gvfs/{,**} r,
 owner /run/user/[0-9]*/gvfs/*/**  w,
+
+# Disallow writes to the well-known directory included in
+# the user's PATH on several distributions
+deny @{HOME}/bin/{,**} wl,
```
jdstrand: I still like 'audit deny' since I'd like to know when an application tried to do this, but that is not a blocker. +1

zyga: +1 I will add the audit, I forgot that deny stopped auditing

PR #5979, opened 2018-10-12
install: don't start disabled services

anonymouse64:
Note that this PR intends to only fix a specific subset of the issue with service startup during install; that is services that are disabled in the install hook, etc. and do *not* have sockets or timers. 
How to handle sockets or timers is still not yet clear, so we would like to just fix this specific bug, which is clearly wrong. There is a forum topic open to discuss the larger situation with sockets and timers here: https://forum.snapcraft.io/t/how-to-manage-services-with-sockets-timers/7904

Also note that there is basically no regression potential here, because we don't currently check services when installing at all, and as such all services are unconditionally installed. 

Ping @bboozzoo

### DIFF
--- a/systemd/systemd.go
+++ b/systemd/systemd.go
@@ -308,9 +308,8 @@ func (s *systemd) IsEnabled(serviceName string) (bool, error) {
 	if err == nil {
 		return true, nil
 	}
-	// note that "systemctl is-enabled <name>" returns exit code 1 for disabled services
-	// as well as outputting "disabled\n", so we need to cast the error, and
-	// also trim the whitespace from the output to check the command output
+	// "systemctl is-enabled <name>" prints `disabled\n` to stderr and returns exit code 1
+	// for disabled services
 	sysdErr, ok := err.(*Error)
 	if ok && sysdErr.exitCode == 1 && strings.TrimSpace(string(sysdErr.msg)) == "disabled" {
 		return false, nil
--- a/systemd/systemd.go
+++ b/systemd/systemd.go
@@ -130,6 +130,7 @@ type Systemd interface {
 	Kill(service, signal, who string) error
 	Restart(service string, timeout time.Duration) error
 	Status(services ...string) ([]*ServiceStatus, error)
+	IsEnabled(service string) (bool, error)
 	LogReader(services []string, n int, follow bool) (io.ReadCloser, error)
 	WriteMountUnitFile(name, revision, what, where, fstype string) (string, error)
 	Mask(service string) error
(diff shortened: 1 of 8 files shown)


systemd/systemd.go
```diff
@@ -301,6 +302,22 @@ func (s *systemd) Status(serviceNames ...string) ([]*ServiceStatus, error) {
 	return sts, nil
 }
 
+// IsEnabled checkes whether the given service is enabled
+func (s *systemd) IsEnabled(serviceName string) (bool, error) {
+	_, err := systemctlCmd("--root", s.rootDir, "is-enabled", serviceName)
+	if err == nil {
+		return true, nil
+	}
+	// note that "systemctl is-enabled <name>" returns exit code 1 for disabled services
```
bboozzoo: IMO it's enough to describe why we're doing things, what is done is seen in the code. I'd rephrase the comment to something like:

```
// "systemctl is-enabled <name>" prints `disabled\n` to stderr and returns exit code 1 
// for disabled services
```

anonymouse64: This is done.

anonymouse64:
I think the test failure is independent of this PR, see:
```
2018-10-15 13:25:59 Error executing google:ubuntu-16.04-64:tests/main/interfaces-accounts-service : 
-----
+ echo 'Ensure things run'
Ensure things run
++ dbus-launch --sh-syntax
+ eval 'DBUS_SESSION_BUS_ADDRESS='\''unix:abstract=/tmp/dbus-4t9ZOR0RiP,guid=2c27a5b364ad4dad8e4369195bc49564'\'';
export DBUS_SESSION_BUS_ADDRESS;
DBUS_SESSION_BUS_PID=29168;'
++ DBUS_SESSION_BUS_ADDRESS=unix:abstract=/tmp/dbus-4t9ZOR0RiP,guid=2c27a5b364ad4dad8e4369195bc49564
++ export DBUS_SESSION_BUS_ADDRESS
++ DBUS_SESSION_BUS_PID=29168
+ echo 29168
++ gnome-keyring-daemon --login
++ printf password
+ eval ''
++ gnome-keyring-daemon --start
+ eval SSH_AUTH_SOCK=/run/user/0/keyring/ssh
++ SSH_AUTH_SOCK=/run/user/0/keyring/ssh
+ echo 29178
+ echo 'Creating account'
Creating account
+ gdbus call --session --timeout 300 --dest=org.gnome.OnlineAccounts --object-path=/org/gnome/OnlineAccounts/Manager --method=org.gnome.OnlineAccounts.Manager.AddAccount imap_smtp test@example.com 'Display Name' '{}' '{'\''Enabled'\'': '\''true'\'', '\''EmailAddress'\'': '\''test@example.com'\'', '\''Name'\'': '\''Test User'\'', '\''ImapHost'\'': '\''imap.example.com'\'', '\''SmtpHost'\'': '\''mail.example.com'\''}'
+ gdbus monitor --session --dest org.gnome.OnlineAccounts
Error: GDBus.Error:org.freedesktop.DBus.Error.InvalidArgs: Type of message, '(ssssa{ss})', does not match expected type '(sssa{sv}a{ss})'
```

Additionally, the test added here passed on all platforms it ran on via spread.

chipaca:
Yes, the accounts-service was giving us trouble

mvo5:
I also cherry-picked this into 2.36

PR #5980, opened 2018-10-12
interfaces/apparmor: conditionally add explicit deny rules for ptrace

jdstrand:
AppArmor deny rules cannot be undone by allow rules which makes deny
rules difficult to work with arbitrary combinations of interfaces.
Sometimes it is useful to suppress noisy denials and because that can
currently only be done with explicit deny rules, adding explicit deny
rules unconditionally makes it difficult for interfaces to be used in
combination. Define suppressPtraceTrace to allow an interface to request
suppression and define usesPtraceTrace to omit the explict deny rules
such that:
```
  if suppressPtraceTrace && !usesPtraceTrace {
      add 'deny ptrace (trace),'
  }
```
Add UsesPtraceTrace() and SuppressPtraceTrace() that can be called in
the interfaces to set the above and adjust addContent() to use the above
logic to conditionally add a single explicit deny ptrace (trace) rule.

Adjust system-observe and network-observe to suppress via the common
interface and browser-support to use the new SuppressPtraceTrace().

With the above, the policy is equivalent to what was generated before
(except we only have a single explicit denial rather than several when
the above 3 interfaces are all connected at once). Future PRs will
adjust other interfaces to use UsesPtraceTrace() as needed so these
interfaces can still ptrace but continue to use the other interfaces.

### DIFF
--- a/interfaces/apparmor/backend.go
+++ b/interfaces/apparmor/backend.go
@@ -552,7 +552,9 @@ func addContent(securityTag string, snapInfo *snap.Info, opts interfaces.Confine
 					snippet := strings.Replace(overlayRootSnippet, "###UPPERDIR###", overlayRoot, -1)
 					tagSnippets += snippet
 				}
+			}
 
+			if !ignoreSnippets {
 				// For policy with snippets that request
 				// suppression of 'ptrace (trace)' denials, add
 				// the suppression rule unless another
--- a/interfaces/apparmor/backend.go
+++ b/interfaces/apparmor/backend.go
@@ -419,12 +419,12 @@ func (b *Backend) deriveContent(spec *Specification, snapInfo *snap.Info, opts i
 	// Add profile for each app.
 	for _, appInfo := range snapInfo.Apps {
 		securityTag := appInfo.SecurityTag()
-		addContent(securityTag, snapInfo, opts, spec.SnippetForTag(securityTag), content)
+		addContent(securityTag, snapInfo, opts, spec.SnippetForTag(securityTag), content, spec)
 	}
 	// Add profile for each hook.
 	for _, hookInfo := range snapInfo.Hooks {
 		securityTag := hookInfo.SecurityTag()
-		addContent(securityTag, snapInfo, opts, spec.SnippetForTag(securityTag), content)
+		addContent(securityTag, snapInfo, opts, spec.SnippetForTag(securityTag), content, spec)
 	}
 	// Add profile for snap-update-ns if we have any apps or hooks.
 	// If we have neither then we don't have any need to create an executing environment.
@@ -482,7 +482,7 @@ func downgradeConfinement() bool {
 	return true
 }
 
-func addContent(securityTag string, snapInfo *snap.Info, opts interfaces.ConfinementOptions, snippetForTag string, content map[string]*osutil.FileState) {
+func addContent(securityTag string, snapInfo *snap.Info, opts interfaces.ConfinementOptions, snippetForTag string, content map[string]*osutil.FileState, spec *Specification) {
 	// Normally we use a specific apparmor template for all snap programs.
 	policy := defaultTemplate
 	ignoreSnippets := false
@@ -552,7 +552,16 @@ func addContent(securityTag string, snapInfo *snap.Info, opts interfaces.Confine
 					snippet := strings.Replace(overlayRootSnippet, "###UPPERDIR###", overlayRoot, -1)
 					tagSnippets += snippet
 				}
+
+				// For policy with snippets that request
+				// suppression of 'ptrace (trace)' denials, add
+				// the suppression rule unless another
+				// interface said it uses them.
+				if spec.suppressPtraceTrace && !spec.usesPtraceTrace {
+					tagSnippets += ptraceTraceDenySnippet
+				}
 			}
+
 			return tagSnippets
 		}
 		return ""
--- a/interfaces/apparmor/backend.go
+++ b/interfaces/apparmor/backend.go
@@ -559,7 +559,7 @@ func addContent(securityTag string, snapInfo *snap.Info, opts interfaces.Confine
 				// suppression of 'ptrace (trace)' denials, add
 				// the suppression rule unless another
 				// interface said it uses them.
-				if spec.suppressPtraceTrace && !spec.usesPtraceTrace {
+				if spec.SuppressPtraceTrace() && !spec.UsesPtraceTrace() {
 					tagSnippets += ptraceTraceDenySnippet
 				}
 			}
--- a/interfaces/apparmor/backend_test.go
+++ b/interfaces/apparmor/backend_test.go
@@ -1634,10 +1634,10 @@ func (s *backendSuite) TestPtraceTraceRule(c *C) {
 	} {
 		s.Iface.AppArmorPermanentSlotCallback = func(spec *apparmor.Specification, slot *snap.SlotInfo) error {
 			if tc.uses {
-				spec.UsesPtraceTrace()
+				spec.SetUsesPtraceTrace()
 			}
 			if tc.suppress {
-				spec.SuppressPtraceTrace()
+				spec.SetSuppressPtraceTrace()
 			}
 			return nil
 		}
--- a/interfaces/apparmor/backend_test.go
+++ b/interfaces/apparmor/backend_test.go
@@ -1606,7 +1606,7 @@ func (s *backendSuite) TestPtraceTraceRule(c *C) {
 			suppress: true,
 			expected: false,
 		},
-		// classic with jail, never suppress
+		// classic with jail, only suppress if suppress == true and uses == false
 		{
 			opts:     interfaces.ConfinementOptions{Classic: true, JailMode: true},
 			uses:     false,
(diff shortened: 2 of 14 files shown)


chipaca:
[`interfaces/apparmor/spec.go:54:32: "explict" is a misspelling of "explicit"`](https://travis-ci.org/snapcore/snapd/jobs/440845633#L1083)

interfaces/apparmor/backend.go
```diff
@@ -419,12 +419,12 @@ func (b *Backend) deriveContent(spec *Specification, snapInfo *snap.Info, opts i
 	// Add profile for each app.
 	for _, appInfo := range snapInfo.Apps {
 		securityTag := appInfo.SecurityTag()
-		addContent(securityTag, snapInfo, opts, spec.SnippetForTag(securityTag), content)
+		addContent(securityTag, snapInfo, opts, spec.SnippetForTag(securityTag), content, spec)
```
zyga: Since we now pass both spec and securityTag into addContent we could make it call SnippetForTag internally. I can do this as a follow up, not to block this.

interfaces/apparmor/backend_test.go
```diff
@@ -1515,3 +1515,148 @@ func (s *backendSuite) TestDowngradeConfinement(c *C) {
...
+			suppress: false,
+			expected: false,
+		},
+		{
+			opts:     interfaces.ConfinementOptions{Classic: true},
+			uses:     true,
+			suppress: true,
+			expected: false,
+		},
+		// classic with jail, never suppress
```
zyga: Hmm, wasn't classic with jail supposed to be the same as strict?

interfaces/apparmor/spec.go
```diff
@@ -477,3 +491,21 @@ func (spec *Specification) AddPermanentSlot(iface interfaces.Interface, slot *sn
 	}
 	return nil
 }
+
+// UsesPtraceTrace records when to omit explicit ptrace deny rules
+func (spec *Specification) UsesPtraceTrace() {
```
zyga: I would have expected `UsesPtraceTrace` to return `bool` and not set any internal state. How about calling the set `UsePtraceTrace`, `SupressPtraceTrace`. Since this is all a part of the package we could drop the getter functions entirely and just either expose a single public function that computes `uses && !suppress` or just access the fields directly.

interfaces/apparmor/template.go
```diff
@@ -566,6 +566,19 @@ var overlayRootSnippet = `
   "###UPPERDIR###/{,**/}" r,
 `
 
+var ptraceTraceDenySnippet = `
```
zyga: The comment is very nice and informational. Thank you! I love that you included the commit ID for the relevant mainline bug fix.

interfaces/builtin/common.go
```diff
@@ -55,6 +55,9 @@ type commonInterface struct {
 	connectedSlotKModModules []string
 	permanentPlugKModModules []string
 	permanentSlotKModModules []string
+
+	usesPtraceTrace     bool
```
zyga: If you follow my naming suggestion above this could become
```
usePtraceTrace bool
suppressPtraceTrace bool
```

tests/main/interfaces-browser-support/task.yaml
```diff
@@ -106,6 +106,14 @@ execute: |
     fi
 
     if [ "$(snap debug confinement)" = strict ] ; then
+        if [ "$ALLOW_SANDBOX" = "false" ]; then
```
zyga: Thank you for updating the spread tests to capture the improvement!

interfaces/apparmor/spec.go
mvo5: I agree with zyga, this is a bit unusual in terms of go style. Most getters of the form "GetFoo() bool" are simply called "Foo() bool". Maybe something along the lines of what zyga suggested, e.g. http://paste.ubuntu.com/p/chSKC2jSBp/ ? This diff would add an explicit Set{Uses,Suppresses}PtraceTrace which simplifies some if/else. I kept the Get* to keep the diff small but I would like to rename them as zyga suggested. Happy to work on this fwiw.

interfaces/builtin/common_test.go
```diff
@@ -82,3 +83,71 @@ func MockEvalSymlinks(test *testutil.BaseTest, fn func(string) (string, error))
...
+	}
+	spec = &apparmor.Specification{}
+	c.Assert(spec.GetUsesPtraceTrace(), Equals, false)
+	c.Assert(spec.GetSuppressPtraceTrace(), Equals, false)
+	c.Assert(spec.AddConnectedPlug(iface, plug, slot), IsNil)
+	c.Assert(spec.GetUsesPtraceTrace(), Equals, false)
+	c.Assert(spec.GetSuppressPtraceTrace(), Equals, true)
+
+	// setting both, only uses is set
+	iface = &commonInterface{
```
mvo5: See above, not sure we need this special case.

interfaces/builtin/common.go
```diff
@@ -86,6 +89,11 @@ func (iface *commonInterface) BeforePrepareSlot(slot *snap.SlotInfo) error {
 }
 
 func (iface *commonInterface) AppArmorConnectedPlug(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	if iface.usesPtraceTrace {
+		spec.UsesPtraceTrace()
+	} else if iface.suppressPtraceTrace {
```
mvo5: I'm not sure why we need this? It seems like the apparmor/backend.go is doing the right thing if both are set. Not sure what this extra special case buys us. Or am I missing something?

interfaces/apparmor/backend.go
jdstrand: That sounds fine in a followup PR.

interfaces/apparmor/spec.go
jdstrand: I'm fine with reorganizing as requested. I'd prefer not to have Set* allow setting both ways because the interfaces will only need to call it with `true`. If they call it with anything else, then it could introduce difficult to diagnose bugs where sometimes the rule is added and sometimes not based on on interface connection ordering. Eg, interface foo uses `SetSuppressPtraceTrace(true)`, then interface bar does `SetSuppressPtraceTrace(false)`. We'll get different behavior depending on which one snapd decides to connect first. Unless you strongly feel otherwise, I'll leave the one-way behavior (but still name it Set*).

interfaces/builtin/common.go
jdstrand: I did it this way because there is no point setting `uses` with `suppress` since `uses` always trumps `suppress` and figured the common interface should reflect best practices. Happy to change if you prefer.

interfaces/apparmor/backend_test.go
jdstrand: Done

interfaces/apparmor/spec.go
jdstrand: Done

interfaces/builtin/common.go
jdstrand: I chose 'usesPtraceTrace' since the interface is declaring it uses it. I chose 'suppressPtraceTrace' because the interface is saying "please supress the ptrace trace denials if you can". I can rename things if you still want.

jdstrand:
@mvo5 and @zyga - I updated the PR and responded to feedback. I think the is ready for another review.

interfaces/apparmor/spec.go
jdstrand: @zyga - regarding dropping getter, I use them in common_test.go and felt this was cleaner.

zyga: I noticed but for such cases we can use `export_test.go`, though at this stage it's not a strong feeling.

jdstrand: @zyga - I recall having some trouble with mocking certain things in the PRs though I can't recall if this was one of them. If you don't have a strong feeling that it is a blocker, I suggest letting it in as is (it might be useful down the line). If it is strong enough for you wanting me to do a followup PR, I can.

zyga:
I think this is fine now but let me look quickly again.

interfaces/apparmor/backend.go
```diff
@@ -553,6 +553,17 @@ func addContent(securityTag string, snapInfo *snap.Info, opts interfaces.Confine
 					tagSnippets += snippet
 				}
 			}
+
+			if !ignoreSnippets {
```
zyga: This is fine, it is done after we've seen all the snippets for a given snap.

interfaces/apparmor/backend_test.go
```diff
@@ -1515,3 +1515,148 @@ func (s *backendSuite) TestDowngradeConfinement(c *C) {
...
+			suppress: false,
+			expected: false,
+		},
+		{
+			opts:     interfaces.ConfinementOptions{DevMode: true},
+			uses:     true,
+			suppress: true,
+			expected: false,
+		},
+		// classic, never suppress
```
zyga: Nice :-)

interfaces/builtin/common.go
zyga: Should the interface complain that both are set? I think this is orthogonal to the problem at hand, just about internal API usability. I think I agree with mvo, it's probably better as a "bunt" pusher of options than extra logic.

I can reduce this in the follow-up PR.

jdstrand:

PR #5981, opened 2018-10-13
interfaces/many: updates to support k8s worker nodes

jdstrand:
* interfaces/kubernetes-support: update to support flavors for worker nodes.
 Refactor the kubernetes-interface to support different flavors of access control via the 'flavor' attribute:
 * kubelet worker node
   * manipulating mounts, etc
   * ptracing containers
   * adjusting panic and panic_on_oops
   * managing kube-pods cgroups
  * kubeproxy workder node
   * managing kube-proxy cgroups
  * default (unspecified): all of the above
 In this manner a kubernetes snap containing many services can more finely tailor accessing for the service (eg, kubeproxy doesn't need to mount)
* interfaces/docker-support: misc access for k8s driving docker
* interfaces/system-observe: allow reading panic, panic_on_oom and panic_on_oops
* add interfaces/kernel-module-observe
* interfaces/kernel-module-control: whitespace changes
* interfaces/kernel-module-control: also allow reading of modprobe.d dirs
* interfaces/firewall-control: various reads/writes for nf. listen on @xtables

Note: this is branched from https://github.com/snapcore/snapd/pull/5980

### DIFF
--- a/interfaces/builtin/docker_support.go
+++ b/interfaces/builtin/docker_support.go
@@ -81,6 +81,10 @@ unix (bind,listen) type=stream addr="@/containerd-shim/moby/*/shim.sock\x00",
 /sys/fs/cgroup/*/docker/   rw,
 /sys/fs/cgroup/*/docker/** rw,
 
+# Also allow cgroup writes to kubernetes pods
+/sys/fs/cgroup/*/kubepods/ rw,
+/sys/fs/cgroup/*/kubepods/** rw,
+
 # Allow tracing ourself (especially the "runc" process we create)
 ptrace (trace) peer=@{profile_name},
 
@@ -111,7 +115,7 @@ pivot_root,
 /sbin/apparmor_parser ixr,
 /etc/apparmor.d/cache/ r,
 /etc/apparmor.d/cache/.features r,
-/etc/apparmor.d/cache/docker* rw,
+/etc/apparmor.d/{,cache/}docker* rw,
 /etc/apparmor.d/tunables/{,**} r,
 /etc/apparmor.d/abstractions/{,**} r,
 /etc/apparmor/parser.conf r,
@@ -532,6 +536,10 @@ ptrace (read, trace) peer=unconfined,
 # This grants raw access to device files and thus device ownership
 /dev/** mrwkl,
 @{PROC}/** mrwkl,
+
+# When kubernetes drives docker, it creates files in the container at arbitrary
+# locations.
+/** wl,
 `
 
 const dockerSupportPrivilegedSecComp = `
@@ -567,6 +575,7 @@ func (iface *dockerSupportInterface) AppArmorConnectedPlug(spec *apparmor.Specif
 	if privileged {
 		spec.AddSnippet(dockerSupportPrivilegedAppArmor)
 	}
+	spec.UsesPtraceTrace()
 	return nil
 }
 
--- a/interfaces/builtin/kernel_module_control.go
+++ b/interfaces/builtin/kernel_module_control.go
@@ -40,6 +40,7 @@ const kernelModuleControlConnectedPlugAppArmor = `
 
 capability sys_module,
 @{PROC}/modules r,
+/{,usr/}bin/kmod ixr,
 
 # FIXME: moved to physical-memory-observe (remove this in series 18)
 /dev/mem r,
@@ -49,8 +50,10 @@ capability sys_module,
 # required to verify kernel modules that are loaded.
 capability syslog,
 
-# Allow plug side to read information about loaded kernel modules
+# Allow reading information about loaded kernel modules
 /sys/module/{,**} r,
+/etc/modprobe.d/{,**} r,
+/lib/modprobe.d/{,**} r,
 `
 
 const kernelModuleControlConnectedPlugSecComp = `
--- a/interfaces/builtin/kernel_module_control.go
+++ b/interfaces/builtin/kernel_module_control.go
@@ -38,19 +38,19 @@ const kernelModuleControlBaseDeclarationSlots = `
 const kernelModuleControlConnectedPlugAppArmor = `
 # Description: Allow insertion, removal and querying of modules.
 
-  capability sys_module,
-  @{PROC}/modules r,
+capability sys_module,
+@{PROC}/modules r,
 
-  # FIXME: moved to physical-memory-observe (remove this in series 18)
-  /dev/mem r,
+# FIXME: moved to physical-memory-observe (remove this in series 18)
+/dev/mem r,
 
-  # Required to use SYSLOG_ACTION_READ_ALL and SYSLOG_ACTION_SIZE_BUFFER when
-  # /proc/sys/kernel/dmesg_restrict is '1' (syslog(2)). These operations are
-  # required to verify kernel modules that are loaded.
-  capability syslog,
+# Required to use SYSLOG_ACTION_READ_ALL and SYSLOG_ACTION_SIZE_BUFFER when
+# /proc/sys/kernel/dmesg_restrict is '1' (syslog(2)). These operations are
+# required to verify kernel modules that are loaded.
+capability syslog,
 
-  # Allow plug side to read information about loaded kernel modules
-  /sys/module/{,**} r,
+# Allow plug side to read information about loaded kernel modules
+/sys/module/{,**} r,
 `
 
 const kernelModuleControlConnectedPlugSecComp = `
--- a/interfaces/builtin/kernel_module_observe.go
+++ b/interfaces/builtin/kernel_module_observe.go
@@ -1,0 +1,58 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2018 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package builtin
+
+const kernelModuleObserveSummary = `allows querying of kernel modules`
+
+const kernelModuleObserveBaseDeclarationSlots = `
+  kernel-module-observe:
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
+`
+
+const kernelModuleObserveConnectedPlugAppArmor = `
+# Description: Allow querying of kernel modules.
+/{,usr/}bin/kmod ixr, # seccomp and no SYS_MODULE prevents loading/removing
+@{PROC}/modules r,
+
+# Required to use SYSLOG_ACTION_READ_ALL and SYSLOG_ACTION_SIZE_BUFFER when
+# /proc/sys/kernel/dmesg_restrict is '1' (syslog(2)). These operations are
+# required to verify kernel modules that are loaded.
+capability syslog,
+
+# Allow reading information about loaded kernel modules
+/sys/module/{,**} r,
+/etc/modprobe.d/{,**} r,
+/lib/modprobe.d/{,**} r,
+`
+
+func init() {
+	registerIface(&commonInterface{
+		name:                  "kernel-module-observe",
+		summary:               kernelModuleObserveSummary,
+		implicitOnCore:        true,
+		implicitOnClassic:     true,
+		baseDeclarationSlots:  kernelModuleObserveBaseDeclarationSlots,
+		connectedPlugAppArmor: kernelModuleObserveConnectedPlugAppArmor,
+		reservedForOS:         true,
+	})
+}
--- a/interfaces/builtin/kubernetes_support.go
+++ b/interfaces/builtin/kubernetes_support.go
@@ -1,7 +1,7 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 
 /*
- * Copyright (C) 2017 Canonical Ltd
+ * Copyright (C) 2017-2018 Canonical Ltd
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 3 as
@@ -19,6 +19,17 @@
 
 package builtin
 
+import (
+	"fmt"
+
+	"github.com/snapcore/snapd/interfaces"
+	"github.com/snapcore/snapd/interfaces/apparmor"
+	"github.com/snapcore/snapd/interfaces/kmod"
+	"github.com/snapcore/snapd/interfaces/seccomp"
+	"github.com/snapcore/snapd/interfaces/udev"
+	"github.com/snapcore/snapd/snap"
+)
+
 const kubernetesSupportSummary = `allows operating as the Kubernetes service`
 
 const kubernetesSupportBaseDeclarationPlugs = `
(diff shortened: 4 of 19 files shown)


interfaces/builtin/kernel_module_observe.go
```diff
@@ -0,0 +1,58 @@
...
+package builtin
+
+const kernelModuleObserveSummary = `allows querying of kernel modules`
+
+const kernelModuleObserveBaseDeclarationSlots = `
+  kernel-module-observe:
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
```
zyga: This is fine

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,227 @@ const kubernetesSupportBaseDeclarationSlots = `
...
+const kubernetesSupportConnectedPlugAppArmorCommon = `
+# Common rules for running as a kubernetes node
 
-# what is this for?
-#include <abstractions/dbus-strict>
-
-# Allow reading all information regarding cgroups for all processes
+# reading cgroups
 capability sys_resource,
-@{PROC}/@{pid}/cgroup r,
```
zyga: Is this removal intentional?

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,227 @@ const kubernetesSupportBaseDeclarationSlots = `
...
+# all processes, not just containers.
 @{PROC}/@{pid}/oom_score_adj rw,
-/sys/kernel/mm/hugepages/ r,
+@{PROC}/sys/vm/overcommit_memory rw,
+/sys/kernel/mm/hugepages/{,**} r,
+
+capability dac_override,
+
+/usr/bin/systemd-run Cxr -> systemd_run,
+profile systemd_run (attach_disconnected,mediate_deleted) {
```
zyga: Since we are not otherwise using `mediate_deleted` would it make sense to explain why it is used here? After some googling the best I could find was https://gitlab.com/apparmor/apparmor/wikis/FAQ

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,227 @@ const kubernetesSupportBaseDeclarationSlots = `
...
-@{PROC}/sys/vm/panic_on_oom r,
 
-@{PROC}/diskstats r,
-@{PROC}/@{pid}/cmdline r,
+/dev/kmsg r,
+
+# kubelet calls out to systemd-run for some mounts, but not all of them and not
+# unmounts...
+capability sys_admin,
+mount /var/snap/@{SNAP_NAME}/common/{,**} -> /var/snap/@{SNAP_NAME}/common/{,**},
```
zyga: This is okay, it stays within the single snap.

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,227 @@ const kubernetesSupportBaseDeclarationSlots = `
...
+`
 
-# Allow listing kernel modules. Note, seccomp blocks module loading syscalls
-/sys/module/apparmor/parameters/enabled r,
-/bin/kmod ixr,
-/etc/modprobe.d/{,**} r,
+var kubernetesSupportConnectedPlugKmodKubeProxy = []string{
+	`ip_vs_rr`,
+	`ip_vs_sh`,
+	`ip_vs_wrr`,
```
zyga: Do you need to add the `ip_vs` module here? It is referenced above in apparmor rules.

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,227 @@ const kubernetesSupportBaseDeclarationSlots = `
...
+	`libcrc32c`,
+	`llc`,
+	`stp`,
+}
 
-# Allow ptracing Docker containers
-ptrace (read, trace) peer=docker-default,
-ptrace (read, trace) peer=snap.docker.dockerd,
+type kubernetesSupportInterface struct{}
```
zyga: Note that embedding commonInterface would let you avoid some of the boilerplate. Did you consider doing that but decided not to or would you be open to simplifying this?

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,227 @@ const kubernetesSupportBaseDeclarationSlots = `
...
+}
+
+func (iface *kubernetesSupportInterface) AppArmorConnectedPlug(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	snippet := kubernetesSupportConnectedPlugAppArmorCommon
 
-var kubernetesSupportConnectedPlugKmod = []string{`llc`, `stp`}
+	systemd_run_extra := ""
+	var flavor string
+	_ = plug.Attr("flavor", &flavor)
+	if flavor == "kubelet" {
```
zyga: IMO this would look more natural as a switch statement.

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,227 @@ const kubernetesSupportBaseDeclarationSlots = `
...
+	old := "###KUBERNETES_SUPPORT_SYSTEMD_RUN###"
+	spec.AddSnippet(strings.Replace(snippet, old, systemd_run_extra, -1))
+	return nil
+}
+
+func (iface *kubernetesSupportInterface) SecCompConnectedPlug(spec *seccomp.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	var flavor string
+	_ = plug.Attr("flavor", &flavor)
+
+	if flavor != "kubeproxy" {
```
zyga: This feels odd, why not just `if flavour == "kubelet" || flavour == "kubeproxy" { ... }`

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,227 @@ const kubernetesSupportBaseDeclarationSlots = `
...
+		spec.AddSnippet(snippet)
+	}
+	return nil
+}
+
+func (iface *kubernetesSupportInterface) UDevConnectedPlug(spec *udev.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	var flavor string
+	_ = plug.Attr("flavor", &flavor)
+
+	if flavor != "kubeproxy" {
```
zyga: Same remark as above.

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,227 @@ const kubernetesSupportBaseDeclarationSlots = `
...
+		for _, rule := range kubernetesSupportConnectedPlugUDevKubelet {
+			spec.TagDevice(rule)
+		}
+	}
+	return nil
+}
+
+func (iface *kubernetesSupportInterface) KModConnectedPlug(spec *kmod.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	var flavor string
```
zyga: Since this pattern repeats often it might make sense to add a small helper:

```
func k8sFlavour(plug *interfaces.ConnectedPlug) string {
    var flavour string
    _ = plug.Attr("flavor", &flavour)
    return flavour
}
```

PS: please forgive the British English spelling above.

interfaces/builtin/docker_support.go
```diff
@@ -532,6 +536,10 @@ ptrace (read, trace) peer=unconfined,
 # This grants raw access to device files and thus device ownership
 /dev/** mrwkl,
 @{PROC}/** mrwkl,
+
+# When kubernetes drives docker, it creates files in the container at arbitrary
```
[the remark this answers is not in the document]
anonymouse64: I think this is because default docker containers use `pivot_root`, and as such AppArmor doesn't see the "real" path.

Note that what the docker snap does is different than "default" docker, because we specifically only let it do a `chroot` and not a `pivot_root`: https://git.launchpad.net/~docker/+git/snap/tree/patches/snappy-real-chroot.patch
@jdstrand would applying that patch to whatever part of microk8s that uses docker remove the need for this arbitrary file locations?

interfaces/builtin/docker_support.go
```diff
@@ -81,6 +81,10 @@ unix (bind,listen) type=stream addr="@/containerd-shim/moby/*/shim.sock\x00",
 /sys/fs/cgroup/*/docker/   rw,
 /sys/fs/cgroup/*/docker/** rw,
 
+# Also allow cgroup writes to kubernetes pods
+/sys/fs/cgroup/*/kubepods/ rw,
+/sys/fs/cgroup/*/kubepods/** rw,
```
anonymouse64: Is there a reason why these kubernetes specific accesses are in the docker specific interface and not in the kubernetes specific interface? I'm not familiar with the whole picture of how kubernetes is used with strict confinement, but this bit looks like it belongs in something kubernetes specific and not docker specific (as I think kubernetes is a superset of docker, interface wise).

interfaces/builtin/docker_support.go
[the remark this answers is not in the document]
jdstrand: @anonymouse64 - that is a great observation and I've advised the k8s team to adjust their microk8s packaging to use your docker (I used a modified microk8s snap to test this interface), but they haven't yet (it's more than just this that is a little wonky). I don't know their plans for microk8s, but they did say they wanted to have their k8s worker nodes *not* ship docker, but instead use your snap. Until these questions are answered, I think we need to keep the rule here. However, I'm all for removing the rule down the line (since docker-support is a super-privileged interface, no one can deliver it via the store without us knowing).

interfaces/builtin/kernel_module_control.go
```diff
@@ -38,19 +38,22 @@ const kernelModuleControlBaseDeclarationSlots = `
...
 
-  capability sys_module,
-  @{PROC}/modules r,
+capability sys_module,
+@{PROC}/modules r,
+/{,usr/}bin/kmod ixr,
 
-  # FIXME: moved to physical-memory-observe (remove this in series 18)
-  /dev/mem r,
+# FIXME: moved to physical-memory-observe (remove this in series 18)
```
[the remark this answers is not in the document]
jdstrand: I think it is too late for this. If others don't think so, sure.

interfaces/builtin/kubernetes_support.go
jdstrand: No, it is autoloaded via the others.

interfaces/builtin/kubernetes_support.go
jdstrand: I converted from commonInterface so I could more readily do the attribute handling. If there is an easier way, sure, I'm up for it being simplified.

interfaces/builtin/kubernetes_support.go
jdstrand: I did it this way because if you specify neither kubelet or kubeproxy, then you get both and this was an easy way to do that.

interfaces/builtin/docker_support.go
jdstrand: These access are in the kubernetes-interface already, but they have to be added to the docker-support interface because while kubelet drives docker (docker is one of the container runtime backends), docker then needs to work with the cgroups kubelet setup.

interfaces/builtin/kubernetes_support.go
jdstrand: It's easy enough to change so I'll do so.

interfaces/builtin/kubernetes_support.go
jdstrand: Done

interfaces/builtin/kubernetes_support.go
jdstrand: Done

interfaces/builtin/kubernetes_support.go
jdstrand: Done

interfaces/builtin/kubernetes_support.go
jdstrand: Done

jdstrand:
@zyga - I answered/addressed your comments. This should be ready for another review.

interfaces/builtin/docker_support.go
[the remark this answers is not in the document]
anonymouse64: I see, it sounds then like the ideal thing for the microk8s snap is to set it up to use docker over the `docker-executables` content interface (and then also plug `docker-daemon` `docker` interface) that the docker snap slot's so that the k8s nodes end up talking to the `dockerd` confined in the docker snap. But yes, as pointed out docker-support is already super privileged and it can always be refined back later as this is implemented if there's a urgent need for microk8s (it seems like there is).

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,226 @@ const kubernetesSupportBaseDeclarationSlots = `
...
+# ptracing unconfined.
+ptrace (trace),
+deny ptrace (trace) peer=unconfined,
+
+@{PROC}/[0-9]*/attr/ r,
+@{PROC}/[0-9]*/fdinfo/ r,
+@{PROC}/[0-9]*/map_files/ r,
+@{PROC}/[0-9]*/ns/ r,
+
+@{PROC}/sys/kernel/panic w,
```
zyga: This allows causing a kernel panic, why do we want to have this permission?

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,226 @@ const kubernetesSupportBaseDeclarationSlots = `
...
+	}
+
+	old := "###KUBERNETES_SUPPORT_SYSTEMD_RUN###"
+	spec.AddSnippet(strings.Replace(snippet, old, systemd_run_extra, -1))
+	return nil
+}
+
+func (iface *kubernetesSupportInterface) SecCompConnectedPlug(spec *seccomp.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	flavor := k8sFlavor(plug)
+	if flavor == "kubelet" || flavor == "" {
```
zyga: Is `flavor. == ""` a default that really means `kubelet`? If so could that knowledge be folded into `k8sFlavor`?

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,226 @@ const kubernetesSupportBaseDeclarationSlots = `
...
+		for _, m := range kubernetesSupportConnectedPlugKmodKubeProxy {
+			if err := spec.AddModule(m); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func (iface *kubernetesSupportInterface) BeforePrepareSlot(slot *snap.SlotInfo) error {
```
zyga: This could go away if you embed `commonInterface`.

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,226 @@ const kubernetesSupportBaseDeclarationSlots = `
...
-# docker-support? Right now kubernetes needs this which allows it to control
-# all processes on the system and on <4.8 kernels, escape confinement.
-ptrace (read, trace) peer=unconfined,
-`
+func (iface *kubernetesSupportInterface) Name() string {
+	return "kubernetes-support"
+}
 
-var kubernetesSupportConnectedPlugKmod = []string{`llc`, `stp`}
+func (iface *kubernetesSupportInterface) StaticInfo() interfaces.StaticInfo {
```
zyga: This could go away if you embed `commonInterface`.

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,226 @@ const kubernetesSupportBaseDeclarationSlots = `
...
-ptrace (read, trace) peer=docker-default,
-ptrace (read, trace) peer=snap.docker.dockerd,
+type kubernetesSupportInterface struct{}
 
-# Should we have a 'privileged' mode for kubernetes like we do with
-# docker-support? Right now kubernetes needs this which allows it to control
-# all processes on the system and on <4.8 kernels, escape confinement.
-ptrace (read, trace) peer=unconfined,
-`
+func (iface *kubernetesSupportInterface) Name() string {
```
zyga: This could go away if you embed `commonInterface`.

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,226 @@ const kubernetesSupportBaseDeclarationSlots = `
...
+	// It's fine if flavor isn't specified, but if it is, it needs to be
+	// either "kubelet" or "kubeproxy"
+	if t, ok := plug.Attrs["flavor"]; ok && t != "kubelet" && t != "kubeproxy" {
+		return fmt.Errorf(`kubernetes-support plug requires "flavor" to be either "kubelet" or "kubeproxy"`)
+	}
+
+	return nil
+}
+
+func (iface *kubernetesSupportInterface) AutoConnect(*snap.PlugInfo, *snap.SlotInfo) bool {
```
zyga: This could go away if you embed `commonInterface`.

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,226 @@ const kubernetesSupportBaseDeclarationSlots = `
...
+/bin/umount ixr,
+deny /run/mount/utab rw,
+umount /var/snap/@{SNAP_INSTANCE_NAME}/common/**,
+`
+
+const kubernetesSupportConnectedPlugAppArmorKubeletSystemdRun = `
+  # kubelet mount rules
+  capability sys_admin,
+  /bin/mount ixr,
+  mount -> /var/snap/@{SNAP_INSTANCE_NAME}/common/**,
```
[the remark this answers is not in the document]
jdstrand: @zyga - note this is a so called 'super-privileged' interface that requires an installation constraint to install and the policy can best be described as advisory (like docker-support). Also note that k8s is a general purpose container orchestration tool that is used in part to drive docker (or containerd or other backends) and the docker-support interface has a bare `mount,` rule. I observed many different mounts into the common directory and feel the above rule is as fine-grained as we want to get as I believe it will allow k8s to function, and going finer offers no security benefit since the the snap will have access to the docker (or other container runtime) socket and can simply ask it to make arbitrary mounts.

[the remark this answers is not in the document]
jdstrand: That said, if you feel strongly about this point, I can investigate further.

interfaces/builtin/kubernetes_support.go
jdstrand: Default means "all policy", ie, *both* kubelet and kubeproxy and any future policy. The kubelet and kubeproxy flavors are for when you want to limit kubelet and (particularly) kubeproxy to not have all of the policy of the other.

interfaces/builtin/kubernetes_support.go
jdstrand: It means this: "The value in this file represents the number of seconds the kernel waits before rebooting on a panic. When you use the software watchdog, the recommended setting is 60."

The reason is here: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/container_manager_linux.go#L346

jdstrand: I added a comment for this.

interfaces/builtin/kubernetes_support.go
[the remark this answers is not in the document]
jdstrand: Ok, I looked at this and in very simple tests, it seems like we can use this rule:
```
mount tmpfs -> /var/snap/@{SNAP_INSTANCE_NAME}/common/**,
```

The directories in common are all random strings so the dst can't be mediated further. I'm not confident this rule won't need to be expanded in the future, but we can do that in a future iteration. I'm updating the rule now.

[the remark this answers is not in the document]
jdstrand: This is Done.

interfaces/builtin/kubernetes_support.go
```diff
@@ -35,65 +47,226 @@ const kubernetesSupportBaseDeclarationSlots = `
...
+	var flavor string
+	_ = plug.Attr("flavor", &flavor)
+	return flavor
+}
+
+func (iface *kubernetesSupportInterface) AppArmorConnectedPlug(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	snippet := kubernetesSupportConnectedPlugAppArmorCommon
+	systemd_run_extra := ""
+
+	switch flavor := k8sFlavor(plug); flavor {
```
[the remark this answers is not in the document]
jdstrand: Done

interfaces/builtin/kubernetes_support.go
jdstrand: Done

jdstrand:
@zyga - I believe I addressed/answered all your comments.

interfaces/builtin/kubernetes_support.go
```diff
@@ -174,20 +174,12 @@ var kubernetesSupportConnectedPlugKmodKubeProxy = []string{
...
-func (iface *kubernetesSupportInterface) StaticInfo() interfaces.StaticInfo {
-	return interfaces.StaticInfo{
-		Summary:              kubernetesSupportSummary,
-		BaseDeclarationPlugs: kubernetesSupportBaseDeclarationPlugs,
-		BaseDeclarationSlots: kubernetesSupportBaseDeclarationSlots,
-		ImplicitOnCore:       true,
-		ImplicitOnClassic:    true,
-	}
+func (iface *kubernetesSupportInterface) BeforePrepareSlot(slot *snap.SlotInfo) error {
+	return sanitizeSlotReservedForOS(iface, slot)
```
zyga: You may want to run "super" version of the method as well:`iface.commonInterface.BeforePrepareSlot` though I down't think we do much validation there now.

interfaces/builtin/kubernetes_support.go
```diff
@@ -131,7 +131,7 @@ const kubernetesSupportConnectedPlugAppArmorKubeletSystemdRun = `
   # kubelet mount rules
   capability sys_admin,
   /bin/mount ixr,
-  mount -> /var/snap/@{SNAP_INSTANCE_NAME}/common/**,
+  mount tmpfs -> /var/snap/@{SNAP_INSTANCE_NAME}/common/**,
```
zyga: Do we want (type=tmpfs) as well?

jdstrand:

PR #5984, opened 2018-10-15
ifacestate/hooks: only create interface hook tasks if hooks exist (2.36)

stolowski:
This is cherry-pick of #5975 for 2.36.


stolowski:
Marking block for a moment, currently investigating an issue in master which might (low chance, but no non-zero) related to these changes.

stolowski:
> Marking blocked for a moment, currently investigating an issue in master which might (low chance, but no non-zero) related to these changes.

Ok, false alarm, the problem was in the branch and test I'm working on. Label removed.

PR #5987, opened 2018-10-15
cmd: refactor IPC and lifecycle of the helper process

zyga:
This branch makes the helper process used by snap confine for the capture of
mount namespace more flexible. It can now be started and commanded separately
from the rest of the process, whereas in the past is was fused with a call that
also switched or created namespaces. It can now execute numbered commands,
namely exit and capture, not just capture and exit together.

From the system point of view the only observable change is that we are now
using a pipe instead of en eventfd to interact with the helper. There is no
breaking or risky change induced by this branch.

Those changes allow the helper to be later re-used to capture the mount
namespace created for each user running snap applications on the system. Which
will be done in subsequent patches that also start using the temporary feature
flag file for the experimental per-user mount namespaces.

As a reviewer you may find this post handy:
https://forum.snapcraft.io/t/snapd-2-36-snap-confine-logic-walkthrough/7843

This now includes revised messages from https://github.com/snapcore/snapd/pull/6004

### DIFF
--- a/cmd/snap-confine/ns-support.c
+++ b/cmd/snap-confine/ns-support.c
@@ -185,8 +185,6 @@ struct sc_mount_ns {
 	// Identifier of the child process that is used during the one-time (per
 	// group) initialization and capture process.
 	pid_t child;
-	// Flag set when this process created a fresh namespace should populate it.
-	bool should_populate;
 };
 
 static struct sc_mount_ns *sc_alloc_mount_ns(void)
@@ -458,10 +456,9 @@ static void helper_main(struct sc_mount_ns *group, struct sc_apparmor *apparmor,
 static void helper_capture_ns(struct sc_mount_ns *group, pid_t parent);
 static void helper_exit(void);
 
-int sc_create_or_join_mount_ns(struct sc_mount_ns *group,
-			       struct sc_apparmor *apparmor,
-			       const char *base_snap_name,
-			       const char *snap_name)
+int sc_join_preserved_ns(struct sc_mount_ns *group, struct sc_apparmor
+			 *apparmor, const char *base_snap_name,
+			 const char *snap_name)
 {
 	// Open the mount namespace file.
 	char mnt_fname[PATH_MAX] = { 0 };
@@ -506,7 +503,7 @@ int sc_create_or_join_mount_ns(struct sc_mount_ns *group,
 		// Inspect and perhaps discard the preserved mount namespace.
 		if (sc_inspect_and_maybe_discard_stale_ns
 		    (mnt_fd, snap_name, base_snap_name) == EAGAIN) {
-			return EAGAIN;
+			return ESRCH;
 		}
 		// Remember the vanilla working directory so that we may attempt to restore it later.
 		char *vanilla_cwd SC_CLEANUP(sc_cleanup_string) = NULL;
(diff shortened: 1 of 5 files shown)


cmd/snap-confine/ns-support.c
```diff
@@ -549,137 +565,180 @@ int sc_create_or_join_mount_ns(struct sc_mount_ns *group,
...
+	debug("waiting for the helper process to exit");
 	int status = 0;
 	errno = 0;
 	if (waitpid(group->child, &status, 0) < 0) {
-		die("cannot wait for the support process for mount namespace capture");
+		die("cannot wait for the helper process");
 	}
 	if (!WIFEXITED(status) || WEXITSTATUS(status) != 0) {
-		die("support process for mount namespace capture exited abnormally");
+		die("helper process exited abnormally");
```
stolowski: Would it make sense to include exit status in the message here?

cmd/snap-confine/ns-support.c
```diff
@@ -549,137 +565,180 @@ int sc_create_or_join_mount_ns(struct sc_mount_ns *group,
...
-		     group->name);
-		exit(0);
-	} else {
-		group->child = pid;
-		// Unshare the mount namespace and set a flag instructing the caller that 
-		// the namespace is pristine and needs to be populated now.
-		debug("unsharing the mount namespace");
-		if (unshare(CLONE_NEWNS) < 0) {
-			die("cannot unshare the mount namespace");
+		if (write(group->pipe_helper[1], &command, sizeof command) < 0) {
```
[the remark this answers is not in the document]
zyga: If we are signalled then it's fine to just quit (we quit in that case). In general we either work successfully or fail immediately. We are not expecting any signals here.

cmd/snap-confine/ns-support.c
zyga: The raw exit status can encode a number of cases (exit code, kill by signal) and I didn't want to make the distinction here originally. I can include it in a follow-up.

cmd/snap-confine/ns-support.c
```diff
@@ -73,6 +73,11 @@ static const char *sc_ns_dir = SC_NS_DIR;
  **/
 #define SC_NS_MNT_FILE ".mnt"
 
+enum {
+	HELPER_CMD_EXIT,
```
bboozzoo:
```
HELPER_CMD_EXIT = 0,
```

cmd/snap-confine/snap-confine.c
```diff
@@ -231,32 +232,35 @@ int main(int argc, char **argv)
...
-				if (sc_create_or_join_mount_ns(group, &apparmor,
-							       base_snap_name,
-							       snap_instance) ==
-				    EAGAIN) {
-					die("unexpectedly the namespace needs to be discarded again");
+			int retval = sc_join_preserved_ns(group, &apparmor,
+							  base_snap_name,
+							  snap_instance);
+			if (retval == ESRCH) {
+				/* Stale mount namespace discarded or no mount namespace to
```
bboozzoo: Please use `// ..` comment style to be consistent with the rest of the code.

bboozzoo: So maybe we could hold off this change and apply it everywhere along with clang format? 

If you really want to use `/*` then it should probably be:
```
/*
 * ...
 */
```
Although, I see more `// ..` multiline comments than `/* ..  */`

zyga: I can look at clang-format (I'd love to move away from indent) but I feat that will just be a bit painful on history (I would be OK but I don't think people will let me land that change). For now I think we should just ... write comments :)

cmd/snap-confine/ns-support.c
zyga: I would have done something like that if there was a public protocol but since this is all inside one binary we can reorder or shuffle those around without consequences.

PR #5989, opened 2018-10-15
interfaces/system-key: add parser mtime and only discover features on write

jdstrand:
Because snap run will generate a system-key to compare with what is on
the system, the previous commit to add parser features support caused
apparmor_parser --preprocess to be called on every snap run invocation,
which is not allowed by the policy, not performant and more importantly,
not intended or needed.

The patch adds apparmor-parser-mtime to the system-key, adds it to
generateKey(), moves discovering the parser features to when the
system-key file is written to disk and omits verifying that the parser
features are the same in SystemKeyMismatch. In this manner, only a
stat() call to obtain the mtime is needed by snap run (fast and no
policy changes), snapd will regenerate the system-key when it discovers
the mtime changes and the parser features are always on disk for profile
generation.


release/apparmor.go
```diff
@@ -150,20 +144,30 @@ type apparmorParserFeature struct {
...
+	// system-key and that could be called by different users on the
+	// system, use a predictable search path for finding the parser.
+	parserSearchPath = "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin"
+)
+
+// lookupParser will look for apparmor_parser in a predictable search path
+func lookupParser() (string, error) {
+	for _, dir := range filepath.SplitList(parserSearchPath) {
+		path := filepath.Join(dir, "apparmor_parser")
+		if _, err := os.Stat(path); err == nil {
```
mvo5: You could use `osutil.FileExists()` here

mvo5: Sorry! You are right of course!

mvo5:
@jdstrand I resolved the conflict and updated the tests - please double check but should be all ok

jdstrand:
I have pressed restart so many times for this unrelated failure:
```
goroutine 1923 [chan receive, 9 minutes]:
github.com/snapcore/snapd/vendor/github.com/godbus/dbus.(*Conn).outWorker(0xc4200f2100)
	/tmp/static-unit-tests/src/github.com/snapcore/snapd/vendor/github.com/godbus/dbus/conn.go:427 +0x63
created by github.com/snapcore/snapd/vendor/github.com/godbus/dbus.(*Conn).Auth
	/tmp/static-unit-tests/src/github.com/snapcore/snapd/vendor/github.com/godbus/dbus/auth.go:119 +0x6ee
```

jdstrand:

PR #5995, opened 2018-10-16
tests/main/interfaces-accounts-service: switch to manual while we debug the test

bboozzoo:


bboozzoo:
Will reopen when needed.

PR #5997, opened 2018-10-16
ifacestate: simplify task chaining in ifacestate.Connect

stolowski:
Simplify task chaining in the Connect method, following the pattern we use in other methods that create multiple tasks. This is a followup to the previous PR which opttimized interface hooks.


mvo5:
Nice one!

PR #6002, opened 2018-10-16
interfaces/system-key: add parser mtime and only discover features on write - 2.36

jdstrand:
https://github.com/snapcore/snapd/pull/5989 for 2.36


jdstrand:

PR #6006, opened 2018-10-16
tests: move spread-shellcheck to run-checks --static

chipaca:
spread-shellcheck is fast enough we can run it as part of the static
suite on travis.


run-checks
```diff
@@ -185,6 +185,8 @@ if [ "$STATIC" = 1 ]; then
             exit 1
         fi
         unset regexp
+        echo Checking shell scripts inside spread yamls...
+        ./spread-shellcheck spread.yaml tests
```
bboozzoo: The spread task one used the latest version of shellcheck from snaps. Any chance we could use something more recent(-ish) here too?

mvo5: Maybe we could keep the shellcheck spread in addition to the fast check?

bboozzoo: If Travis supports snaps (does it?) we could `snap install` before :)

chipaca:
@bboozzoo @mvo5 with this test adding a whole minute to the 'quick' run, I'm not sure we want it in any case

chipaca:
@mvo @bboozzoo I changed the approach a little bit. With the current rearrangement, the 'quick' stage could be about a minute faster when things are good, or two minutes slower when we're hammering travis. It's also that little bit cleaner imho.
@bboozzoo note I tweaked `spread-shellcheck` a bit more.

spread-shellcheck
```diff
@@ -237,8 +234,19 @@ def main(opts):
 
 
 if __name__ == '__main__':
+    # fiddle with PATH so we pick up a snapped shellcheck first
+    if "/snap/bin" in os.environ["PATH"]:
```
bboozzoo: It's probably much easier to drop the piece that mangles the PATH and have this instead:
```
SHELLCHECK = os.getenv('SHELLCHECK', 'shellcheck)
...
def checksection(data):
    ...
    subprocess.Popen([SHELLCHECK, "-s", "bash", "-x", "-"]),
                     stdout=subprocess.PIPE,
                     stdin=subprocess.PIPE)
    ...
```

Then `SHELLCHECK=/snap/bin/shellcheck ./spread-shellcheck ...` should call `shellcheck` from snap.

spread-shellcheck
```diff
@@ -237,8 +234,19 @@ def main(opts):
...
+        for e in os.environ["PATH"].split(":"):
+            if "/snap/bin" in e:
+                path.insert(0, e)
+            else:
+                path.append(e)
+        os.environ["PATH"] = ":".join(path)
+
     opts = parse_arguments()
     if opts.verbose or D or V:
+        subprocess.call(["shellcheck", "--version"])
```
bboozzoo: This one then becomes `subprocess.call([SHELLCHECK, "--version"])`

PR #6007, opened 2018-10-17
tests/lib: add an @mozilla.com exception to the CLA checker.

chipaca:


chipaca:
It is my current understanding that people with @mozilla.com addresses should pass the CLA check. I'm waiting for confirmation about this, so please don't merge just yet.

PR #6008, opened 2018-10-17
client, daemon, cmd/snap: indicate that services are socket/timer activated

bboozzoo:
Update the output of `snap services` command to indicated whether a service is activated by a socket or a timer.

Given the current output, it's hard to tell why given app is not active:
```
$ snap services
Service                                 Startup   Current
lxd.activate                            enabled   inactive
lxd.daemon                              enabled   active
socket-activation.sleep-daemon          enabled   inactive
test-snapd-timer-service.random-timer   disabled  inactive
test-snapd-timer-service.regular-timer  disabled  inactive
```

Now, with the status of activators being passed from the daemon to the client side, the proposed output looks like this:

```
$ snap services
Service                                 Startup   Current   Notes
lxd.activate                            enabled   inactive  -
lxd.daemon                              enabled   active    socket-activated
socket-activation.sleep-daemon          enabled   inactive  socket-activated
test-snapd-timer-service.random-timer   disabled  inactive  timer-activated
test-snapd-timer-service.regular-timer  disabled  inactive  timer-activated
```

The PR does not yet show any information when some of the activators are not up. eg, a service that has 2 sockets, one can be disabled/stopped, the other is running.


chipaca:
I'm not sure the Startup/Current columns show what we want (I'd need to look), but that aspect isn't a regression.

zyga:
As a quick remark, `snap services` shows just the table headers when there is nothing to show. Perhaps we could have a crafted message like `no services found`?

daemon/snap.go
```diff
@@ -282,29 +282,69 @@ func clientAppInfosFromSnapAppInfos(apps []*snap.AppInfo) []client.AppInfo {
...
+			continue
+		}
+		if len(sts) != len(serviceNames) {
+			logger.Noticef("cannot get status of services of app %q: expected %v results, got %v", app.Name, len(serviceNames), len(sts))
+			continue
+		}
+		appInfo.Enabled = sts[0].Enabled
+		appInfo.Active = sts[0].Active
+
+		for _, st := range sts[1 : 1+len(app.Sockets)] {
```
zyga: The slice manipulation here feels fragile. I understand this is done to avoid separate systemctl queries but perhaps we could split it back into separate status slices just after `sysd.Status` call above?

cmd/snap/cmd_services.go
```diff
@@ -114,6 +115,38 @@ func svcNames(s []serviceName) []string {
 	return svcNames
 }
 
+func notesForSvc(app *client.AppInfo) string {
+	if !app.IsService() {
+		return "-"
+	}
+
+	var notes []string
```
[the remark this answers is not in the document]
bboozzoo: That would break the `append` & `if len(notes) == 0`

daemon/snap.go
bboozzoo: Hm, I'll see if I can iterate over all statuses instead.

cmd/snap/cmd_services.go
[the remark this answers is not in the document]
zyga: Ah right, I wish there was a simple version (as in, sans make), of `make([]string, 0, 2)`

cmd/snap/cmd_services_test.go
```diff
@@ -169,3 +170,130 @@ func (s *appOpSuite) TestAppOps(c *check.C) {
...
+	rest, err := snap.Parser(snap.Client()).ParseArgs([]string{"services"})
+	c.Assert(err, check.IsNil)
+	c.Assert(rest, check.HasLen, 0)
+	c.Check(s.Stderr(), check.Equals, "")
+	c.Check(s.Stdout(), check.Equals, "Service  Startup  Current  Notes\n")
+	// ensure that the fake server api was actually hit
+	c.Check(n, check.Equals, 1)
+}
+
+func (s *appOpSuite) TestAppStatusNotes(c *check.C) {
```
zyga: Thank you for this!

PR #6010, opened 2018-10-17
cmd/snap-discard-ns: add support for per-user mount namespaces

zyga:
This patch rewrites snap-discard-ns to support user mounts. This
involves scanning for the new per-user mount namespaces and mount
profiles.

The program was simplified to not use the "abstraction" of provided by
"ns-support.h". The abstraction was poor and with the upcoming changes
to snap-confine to accommodate per-user mount namespaces it is easier to
audit and understand the code directly. This change makes "ns-support.h"
private to snap-confine so it can evolve separately.

The manual page was updated to reflect the new functionality and refresh
some cruft like project bug report URL.


cmd/snap-discard-ns/snap-discard-ns.c
```diff
@@ -15,41 +15,187 @@
...
+	sc_must_snprintf(sys_fstab_pattern, sizeof sys_fstab_pattern,
+			 "snap\\.%s\\.fstab", snap_instance_name);
+	sc_must_snprintf(usr_fstab_pattern, sizeof usr_fstab_pattern,
+			 "snap\\.%s\\.*\\.fstab", snap_instance_name);
+	sc_must_snprintf(sys_mnt_pattern, sizeof sys_mnt_pattern,
+			 "%s\\.mnt", snap_instance_name);
+	sc_must_snprintf(usr_mnt_pattern, sizeof usr_mnt_pattern,
+			 "%s\\.*\\.mnt", snap_instance_name);
+
+	DIR *ns_dir = fdopendir(ns_dir_fd);
```
bboozzoo: Do we need to check for errors here?

tests/main/snap-discard-ns/mount-py2.py
```diff
@@ -0,0 +1,60 @@
...
+    """MountOpts contain various flags for the mount system call."""
+
+    Bind = 4096
+
+
+def mount(source, target, fstype, flags=0, data=""):
+    # type: (unicode, unicode, unicode, int, unicode) -> None
+    """mount is a thin wrapper around the mount library function."""
+    libc_name = find_library(b"c")
+    if libc_name is None:
```
bboozzoo: `if not libc_name:`?

tests/main/snap-discard-ns/mount.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+if [ "$(command -v python3)" != "" ]; then
+	exec ./mount-py3.py "$@";
```
bboozzoo: Can we have just one version of the script? The way you've written py2 version, it  should work with py3 just fine.

tests/main/snap-discard-ns/mount-py2.py
```diff
@@ -0,0 +1,60 @@
...
+        mount(opts.source, opts.target, b"", MountOpts.Bind)
+    else:
+        mount(opts.source, opts.target, opts.fstype)
+
+
+if __name__ == "__main__":
+    try:
+        main()
+    except Exception as err:
+        print(err, file=stderr)
```
bboozzoo: Did you forget to `raise`? This will mask any errors, even programming ones.

cmd/snap-discard-ns/snap-discard-ns.c
zyga: Hmmm, I think so, this is a `malloc` after all. I will add a check for this. Thank you!

tests/main/snap-discard-ns/mount-py2.py
zyga: I didn't intend to show the traceback but I think I did forget to return the exit status. I will fix this.

tests/main/snap-discard-ns/mount.sh
zyga: I tried to initially but in the end I decided no to. Using one version of the script puts most pressure on the unicode vs string vs bytes across the two versions. Right now both versions are correct but this could be subtly broken with a single script.

tests/main/snap-discard-ns/mount-py2.py
zyga: PEP8 suggests testing None the way I did to avoid masking the distinction between None and other things that evaluate to false in boolean context.

tests/main/snap-discard-ns/mount-py2.py
bboozzoo: I think you can then drop the whole try/except block and just call `main()` or catch say `OSError` and let the others fall through. Right now you'll catch stuff like `NameError` which is probably not very useful without the actual backtrace.

tests/main/snap-discard-ns/mount-py2.py
```diff
@@ -0,0 +1,61 @@
...
+
+    Bind = 4096
+
+
+def mount(source, target, fstype, flags=0, data=""):
+    # type: (unicode, unicode, unicode, int, unicode) -> None
+    """mount is a thin wrapper around the mount library function."""
+    libc_name = find_library(b"c")
+    if libc_name is None:
+        raise Exception("cannot find the C library")
```
bboozzoo: Maybe make that more specific, say `RuntimeError`

tests/main/snap-discard-ns/mount-py2.py
bboozzoo: Then you could
```
try (OSError, RuntimeError) as err:
   print(err, file=stderr)
   raise SystemExit(1)
```

tests/main/snap-discard-ns/mount-py2.py
```diff
@@ -0,0 +1,61 @@
...
+    parser = ArgumentParser()
+    parser.add_argument("source", help="path of the device or bind source")
+    parser.add_argument("target", help="path of the new mount point")
+    group = parser.add_mutually_exclusive_group(required=True)
+    group.add_argument("--bind", action="store_const", const="bind",
+                       help="create a new mount point of existing path")
+    group.add_argument("-t", "--type", help="filesystem type to mount")
+    opts = parser.parse_args()
+    if opts.bind is not None:
+        mount(opts.source, opts.target, b"", MountOpts.Bind)
```
bboozzoo: `mount(opts.source, opts.target, "", MountOpts.Bind)`? If you run this under Py3, `b"".encode('utf-8')` when calling mount will fail.

tests/main/snap-discard-ns/mount.sh
bboozzoo: I think one script is enough. With the exception of `b""`, I don't see anything wrong with py2 version that would blow up under py3.

tests/main/snap-discard-ns/mount-py2.py
```diff
@@ -0,0 +1,61 @@
...
+
+def mount(source, target, fstype, flags=0, data=""):
+    # type: (unicode, unicode, unicode, int, unicode) -> None
+    """mount is a thin wrapper around the mount library function."""
+    libc_name = find_library(b"c")
+    if libc_name is None:
+        raise Exception("cannot find the C library")
+    libc = CDLL(libc_name, use_errno=True)
+    retval = libc.mount(
+            c_char_p(source.encode("UTF-8")),
```
bboozzoo: `utf-8` is the default, so you may drop that.

tests/main/snap-discard-ns/mount-py2.py
zyga: What is the problem with b"".encode('utf-8')?

tests/main/snap-discard-ns/mount-py2.py
zyga: UTF-8 is not the default.

tests/main/snap-discard-ns/mount-py2.py
bboozzoo: `bytes` has no `encode()`, only `decode()`

tests/main/snap-discard-ns/mount-py2.py
bboozzoo: I think it is in Py3. In Py2 it was derived from your locale or sth. OTOH, I'm not sure we should be forcing utf-8 here which may not be the same as input encoding so the input path provided to the script will be a different byte sequence from one encoded to utf-8.

tests/main/snap-discard-ns/mount-py2.py
zyga: In python 2 bytes is just a string. It does have the encode method.
The b prefix makes it a plain string, not an unicode string. Note that this is the py2 version of the code.

Having said that I will look at unifying this even though I would much rather not have to use python 2 and all of its quirks at all.

bboozzoo: What I meant is that if you use the Py2 version of the script in Py3, `b""` is `bytes` which does not have `encode()`. I think it's the only line here that can cause compatibility problems.

IIRC `u""` is the unicode string in Py2, `b""` and `""` are the same (str). On the other hand, in Py3 `u""` is the same as `""` (str), but not `b""` (which is bytes).

tests/main/snap-discard-ns/mount-py2.py
zyga: I switched error handling, it should work fine as-is now.

tests/main/snap-discard-ns/mount-py2.py
zyga: Yeah but this was for Python 2 specifically. The Python 3 version did not need that. 

In Python 3 u"" is invalid syntax. Look at what I did now, it works and is Mypy-clean on both versions.

tests/main/snap-discard-ns/mount.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+if [ "$(command -v python3)" != "" ]; then
+	exec python3 ./mount.py "$@";
+elif [ "$(command -v python3)" != "" ]; then
```
bboozzoo: `s/3/2/` ?

tests/main/snap-discard-ns/mount.sh
```diff
@@ -0,0 +1,9 @@
...
+	exec python2 ./mount-py2.py "$@";
```
bboozzoo: `s/mount-py2/mount/`

tests/main/snap-discard-ns/mount.sh
zyga: Fixed

tests/main/snap-discard-ns/mount.sh
zyga: Fixed

cmd/snap-discard-ns/snap-discard-ns.c
```diff
@@ -15,41 +15,190 @@
...
+		/* We use dnet->d_name a lot so let's shorten it. */
+		const char *dname = dent->d_name;
+
+		/* Check the four patterns that we have against the name and set the
+		 * two should flags to decide further actions. Note that we always
+		 * unlink matching files so that is not reflected in the structure. */
+		bool should_unmount = false;
+		bool should_unlink = false;
+		struct variant {
+			const char *p;
```
stolowski: Nitpick, I would find `pattern` more descriptive.

cmd/snap-discard-ns/snap-discard-ns.c
```diff
@@ -15,41 +15,190 @@
...
+	 * - "$SNAP_INSTANCE_NAME.mnt"
+	 * - "$SNAP_INSTANCE_NAME.[0-9]+.mnt"
+	 *
+	 * Applied mount profiles to unlink:
+	 * - "snap.$SNAP_INSTANCE_NAME.fstab"
+	 * - "snap.$SNAP_INSTANCE_NAME.[0-9]+.fstab" */
+	char sys_fstab_pattern[PATH_MAX];
+	char usr_fstab_pattern[PATH_MAX];
+	char sys_mnt_pattern[PATH_MAX];
+	char usr_mnt_pattern[PATH_MAX];
```
stolowski: Not a real problem in this case since the resulting buffer will be large enough for our short patterns, but the use of PATH_MAX for patterns is slightly confusing here.

cmd/snap-discard-ns/snap-discard-ns.c
```diff
@@ -15,41 +15,190 @@
...
-	char profile_path[PATH_MAX] = { 0 };
-	sc_must_snprintf(profile_path, sizeof(profile_path),
-			 "/run/snapd/ns/snap.%s.fstab", snap_name);
-	if (unlink(profile_path) < 0) {
-		// Silently ignore ENOENT as the profile doens't have to be there.
-		if (errno != ENOENT) {
-			die("cannot remove current mount profile: %s",
-			    profile_path);
+
+	const char *snap_instance_name = argv[1];
```
stolowski: Should we make sure snap instance name is sane?

cmd/snap-discard-ns/snap-discard-ns.c
```diff
@@ -15,41 +15,190 @@
...
+		struct variant variants[4] = {
+			{.p = sys_mnt_pattern,.unmount = true},
+			{.p = usr_mnt_pattern,.unmount = true},
+			{.p = sys_fstab_pattern},
+			{.p = usr_fstab_pattern},
+		};
+		for (size_t i = 0; i < sizeof variants / sizeof *variants; ++i) {
+			struct variant *v = &variants[i];
+			debug("checking if %s matches %s", dname, v->p);
+			int match_result = fnmatch(v->p, dname, 0);
```
stolowski: The man of fnmatch says "FNM_NOMATCH if there is no match or another nonzero value if there is an error", so I think you should check against FNM_NOMATCH first before treating <0 as fatal.

cmd/snap-discard-ns/snap-discard-ns.c
```diff
@@ -15,41 +15,190 @@
...
+		struct stat file_info;
+		if (fstatat(ns_dir_fd, dname, &file_info,
+			    AT_SYMLINK_NOFOLLOW) < 0) {
+			die("cannot inspect file %s", dname);
+		}
+
+		/* We are only interested in regular files. The .mnt files, even if
+		 * bind-mounted, appear as regular files and not as symbolic links due
+		 * to the peculiarities of the Linux kernel. */
+		if (!S_ISREG(file_info.st_mode)) {
```
stolowski: Would it make sense to move this up & check early before trying to match?

cmd/snap-discard-ns/snap-discard-ns.c
```diff
@@ -15,41 +15,190 @@
...
+			struct statfs fs_info;
+			if (fstatfs(path_fd, &fs_info) < 0) {
+				die("cannot inspect file-system at %s", dname);
+			}
+			close(path_fd);
+			if (fs_info.f_type == NSFS_MAGIC
+			    || fs_info.f_type == PROC_SUPER_MAGIC) {
+				debug("unmounting %s", dname);
+				if (umount2(dname, MNT_DETACH | UMOUNT_NOFOLLOW)
+				    < 0) {
```
stolowski: This would read better if `< 0) {` was on the same line where the `if` starts, but I presume it was formatted automatically like this? Slightly odd given that the comment block above is much wider.

tests/main/snap-discard-ns/mount.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+if [ "$(command -v python3)" != "" ]; then
+	exec python3 ./mount.py "$@";
+elif [ "$(command -v python2)" != "" ]; then
+	exec python2 ./mount.py "$@";
```
stolowski: Why this logic if it works with both versions of python, why not simply check for `/usr/bin/python` (and have `#!/usr/bin/env python` shebang?

cmd/snap-discard-ns/snap-discard-ns.c
zyga: I stated with pattern but I wanted to make it fit in the formatting limitations and avoid the lines wrapping ugly. I will check if some abbreviation, perhaps `patrn` or similar would fit.

cmd/snap-discard-ns/snap-discard-ns.c
zyga: I will add a comment explaining that since those are path-related globs we use PATH_MAX to be on the safe side.

cmd/snap-discard-ns/snap-discard-ns.c
zyga: Good point, we probably should. I will make that adjustment.

cmd/snap-discard-ns/snap-discard-ns.c
zyga: I wrote this watching strace logs and realised we are stat-ing things that we don't care about. The pattern match is much faster than a system call so I chose to order it this way around.

cmd/snap-discard-ns/snap-discard-ns.c
zyga: This is automatically formatted, the mystery of `indent`. Perhaps given that this a near-full rewrite I could use clang-format for this file and actually switch to a sane format profile?

tests/main/snap-discard-ns/mount.sh
zyga: `python` is usually `python2` but I wanted to be explicit. In practice we will only pick up the other case (python2) on the Amazon Linux 2 systems.

cmd/snap-discard-ns/snap-discard-ns.c
zyga: `FNM_NOMATCH` is 1 but I agree that this works by implementation detail, not by correctness. I'll changed that now.

cmd/snap-discard-ns/snap-discard-ns.c
```diff
@@ -15,41 +15,199 @@
...
-	if (unlink(profile_path) < 0) {
-		// Silently ignore ENOENT as the profile doens't have to be there.
-		if (errno != ENOENT) {
-			die("cannot remove current mount profile: %s",
-			    profile_path);
+
+	const char *snap_instance_name = argv[1];
+	struct sc_error *err = NULL;
+	sc_instance_name_validate(snap_instance_name, &err);
+	sc_die_on_error(err);
```
stolowski: Nice, thank you.

cmd/snap-discard-ns/snap-discard-ns.c
stolowski: That might be a good idea... Feel free to give it a try before landing and apply if it produces sensible output.

tests/main/snap-discard-ns/mount.sh
stolowski: Fair enough, not a problem at all, just looks a little elaborate considering the fact that we don't actually care which python version it is.

PR #6011, opened 2018-10-17
Fix import as module by adding go.mod files

ryanjyoder:
When importing snapd as a module with go 1.11, following files cause errors:
* cmd/snap-confine/spread-tests/distros/debian.
* tests/completion/data/twisted/this isn't.innit

I've excluded these files by adding empty go.mod files at their top level directories. I believe this is the minimal set of mod files to import successful with go1.11.

Have you signed the [license agreement](https://www.ubuntu.com/legal/contributors) and read the [contribution guide](https://github.com/snapcore/snapd/blob/master/CONTRIBUTING.md)? yes and yes. :)


chipaca:
It looks like your signing of the CLA hasn't yet been processed.  When did you do it?

ryanjyoder:
I signed it right before I created the PR.

chipaca:
If it hasn't been processed in a few hours, I'll prod somebody about it.

ryanjyoder:
It looks like the CLA has been processed. I would welcome any feedback on this. I intentionally tried to create the minimum set of mod files to make this work, because I wasn't sure on the plan for migrating (or not migrating) to go modules. I'm personally most interested in consuming the `client` library.

chipaca:
I think this change breaks the way tests/unit/go-darwin works

PR #6012, opened 2018-10-17
interfaces: fix NormalizeInterfaceAttributes, add tests

zyga:
The NormalizeInterfaceAttributes was untested so I fixed that. While
working on tests I noticed that it was unexpectedly mutating the
argument it was given (while returning a return value) so I fixed and
tested that as well.


zyga:
@stolowski if it was supposed to do in-place normalisation then I think the return type is confusing and the documentation was not clear about that. It should do one or the other but not both IMO.

PR #6013, opened 2018-10-17
Revert "snap, client, daemon, store: use and expose "media" more (#5906)"

chipaca:
This reverts commit 08bbeb7289fb943ab6c78432579eaceae555ac71 (#5906).

### DIFF
--- a/client/packages.go
+++ b/client/packages.go
@@ -64,9 +64,8 @@ type Snap struct {
 	CommonIDs        []string      `json:"common-ids,omitempty"`
 	MountedFrom      string        `json:"mounted-from,omitempty"`
 
-	Prices      map[string]float64    `json:"prices,omitempty"`
-	Screenshots []snap.ScreenshotInfo `json:"screenshots,omitempty"`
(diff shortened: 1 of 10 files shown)


chipaca:
Heads up @robert-ancell :-(

robert-ancell:
I don't see an issue offering both screenshots and media. The screenshots field is deprecated and clients should migrate to using media. We expose that in the snapd-glib API by using deprecated flags that tell the developer at compile time to migrate.

Please don't revert this already released change - we already consume this API and this would be an API break. If the intention is to filter the media field to not include screenshots to avoid the duplication then let's just do that change instead. IMO that will be more complex for clients than the current solution but it is better than starting again.

niemeyer:
> Please don't revert this already released change - we already consume this API and this would be an API break.

Wait, what? Is that already in a production release? The affecting change was merged 13 days ago.

Per the note above, we do plan to offer you the data you need. What's at stake here is how this will be designed into the API.

chipaca:
We've come to an agreement that makes this revert unnecessary. More news soon.

PR #6017, opened 2018-10-18
interfaces: fix decoding of json numbers for static/dynamic attributes

stolowski:
Based on PR from @mvo5 - decode json with numbers in getConns, and normalize numbers. NormalizeAttributes understands json.Number and tries to get Int from it in the first place, then falls back to float. The only reasonable type for numbers in interface attributes is Int, but we don't prevent floats from being used at the moment - if we want Ints only, we would need to block them on other levels in the first place (e.g. yaml).

### DIFF
--- a/interfaces/utils/utils.go
+++ b/interfaces/utils/utils.go
@@ -47,12 +47,11 @@ func NormalizeInterfaceAttributes(value interface{}) interface{} {
 		return vc
 	case json.Number:
 		jsonval := value.(json.Number)
-		asInt, err := jsonval.Int64()
-		if err != nil {
-			asFloat, _ := jsonval.Float64()
-			return asFloat
+		if asInt, err := jsonval.Int64(); err == nil {
+			return asInt
 		}
-		return asInt
+		asFloat, _ := jsonval.Float64()
+		return asFloat
 	}
 	return value
 }
--- a/interfaces/utils/utils.go
+++ b/interfaces/utils/utils.go
@@ -19,6 +19,10 @@
 
 package utils
 
+import (
+	"encoding/json"
+)
+
 // NormalizeInterfaceAttributes normalises types of an attribute values.
 // The following transformations are applied: int -> int64, float32 -> float64.
 // The normalisation proceeds recursively through maps and slices.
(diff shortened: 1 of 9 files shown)


interfaces/utils/utils.go
```diff
@@ -41,6 +45,14 @@ func NormalizeInterfaceAttributes(value interface{}) interface{} {
 			vc[key] = NormalizeInterfaceAttributes(item)
 		}
 		return vc
+	case json.Number:
```
zyga: Can you please add a test for this case. I built the basics for it here: https://github.com/snapcore/snapd/pull/6012

stolowski: Added.

interfaces/utils/utils_test.go
```diff
@@ -44,10 +45,12 @@ func (s *utilsSuite) TestNormalizeInterfaceAttributes(c *C) {
 	// Funny that, I noticed it only because of missing test coverage.
 	c.Assert(normalize(float32(3.14)), Equals, float64(3.140000104904175))
 	c.Assert(normalize("banana"), Equals, "banana")
-	c.Assert(normalize([]interface{}{42, 3.14, "banana"}), DeepEquals,
-		[]interface{}{int64(42), float64(3.14), "banana"})
+	c.Assert(normalize([]interface{}{42, 3.14, "banana", json.Number("21"), json.Number("0.5")}), DeepEquals,
+		[]interface{}{int64(42), float64(3.14), "banana", int64(21), float64(0.5)})
 	c.Assert(normalize(map[string]interface{}{"i": 42, "f": 3.14, "s": "banana"}),
 		DeepEquals, map[string]interface{}{"i": int64(42), "f": float64(3.14), "s": "banana"})
+	c.Assert(normalize(json.Number("1")), Equals, int64(1))
```
zyga: +1, thanks!

overlord/ifacestate/helpers.go
```diff
@@ -522,9 +526,16 @@ func getPlugAndSlotRefs(task *state.Task) (interfaces.PlugRef, interfaces.SlotRe
...
+	var raw *json.RawMessage
+	err = st.Get("conns", &raw)
 	if err != nil && err != state.ErrNoState {
-		return nil, fmt.Errorf("cannot obtain data about existing connections: %s", err)
+		return nil, fmt.Errorf("cannot obtain raw data about existing connections: %s", err)
+	}
+	if raw != nil {
+		err = jsonutil.DecodeWithNumber(bytes.NewReader(*raw), &conns)
+		if err != nil {
+			return nil, fmt.Errorf("cannot obtain data about existing connections: %s", err)
```
zyga:
```suggestion
			return nil, fmt.Errorf("cannot decode data about existing connections: %s", err)
```

PR #6019, opened 2018-10-19
ifacestate: optimize disconnect hooks

stolowski:
Only run disconnect hooks is they exist - followup to optimization of connect hooks, not as critical, nonetheless always a small improvement. Since the test shares a lot of logic similiar to connect tests from previous PR, I refactored both to reduce duplicated code.


overlord/ifacestate/ifacestate.go
```diff
@@ -332,58 +332,73 @@ func disconnectTasks(st *state.State, conn *interfaces.Connection, flags disconn
...
-		summary := fmt.Sprintf(i18n.G("Run hook %s of snap %q"), disconnectPlugHookSetup.Hook, disconnectPlugHookSetup.Snap)
-		disconnectPlug := hookstate.HookTaskWithUndo(st, summary, disconnectPlugHookSetup, undoDisconnectPlugHookSetup, initialContext)
-		disconnectPlug.WaitAll(ts)
+			summary := fmt.Sprintf(i18n.G("Run hook %s of snap %q"), disconnectPlugHookSetup.Hook, disconnectPlugHookSetup.Snap)
+			disconnectPlug := hookstate.HookTaskWithUndo(st, summary, disconnectPlugHookSetup, undoDisconnectPlugHookSetup, initialContext)
 
-		if disconnectSlot != nil {
-			disconnectPlug.WaitFor(disconnectSlot)
+			addTask(disconnectPlug)
+			prev = disconnectPlug
```
zyga: Should `addTask` automatically set `prev`? It seems to be used this way each time.

overlord/ifacestate/ifacestate.go
```diff
@@ -332,58 +332,73 @@ func disconnectTasks(st *state.State, conn *interfaces.Connection, flags disconn
 	}
 
 	ts := state.NewTaskSet()
+	var prev *state.Task
+	addTask := func(t *state.Task) {
+		if prev != nil {
+			t.WaitFor(prev)
+		}
+		ts.AddTask(t)
```
zyga: Small suggestion, this would allow you not to remember to update `prev` after each of the calls to `addTask` below.

```suggestion
		ts.AddTask(t)
		prev = t
```

overlord/ifacestate/ifacestate.go
stolowski: Good idea, done.

overlord/ifacestate/ifacestate.go
stolowski: Good idea. Done.

PR #6023, opened 2018-10-19
overlord/snapstate, snap, wrappers: start services in the right order during install

bboozzoo:
We got a bug report that when installing a snap with services, the after before ordering is not preserved while services are started. See https://bugs.launchpad.net/snapd/+bug/1796125 for details.

Apparently, the order is not ensured by systemd when `systemctl <foo>` is called with multiple units. The upstream bug is https://github.com/systemd/systemd/issues/8102

Since the solution will not be provided soon, try to workaround the problem by doing the sort ourselves and starting the services one by one.

(cc @anonymouse64)


bboozzoo:
I'll be converting the reference test case from https://github.com/bboozzoo/snapd-services-ordering-test to a spread test in a followup.

wrappers/services_test.go
```diff
@@ -454,11 +454,9 @@ func (s *servicesTestSuite) TestStartSnapMultiServicesFailStartCleanup(c *C) {
 		sysdLog = append(sysdLog, cmd)
 		if len(cmd) >= 2 && cmd[0] == "start" {
 			name := cmd[len(cmd)-1]
-			if name == svc1Name {
-				// the services are being iterated in the "wrong" order
-				svc1Name, svc2Name = svc2Name, svc1Name
+			if name == svc2Name {
+				return nil, fmt.Errorf("failed")
```
chipaca: nice

snap/info.go
```diff
@@ -1138,3 +1138,68 @@ func (r ByType) Swap(i, j int) { r[i], r[j] = r[j], r[i] }
...
+
+	// list of apps without predecessors (no incoming edges)
+	queue := make([]*AppInfo, 0, len(apps))
+	for _, app := range apps {
+		if predecessors[app.Name] == 0 {
+			queue = append(queue, app)
+		}
+	}
+
+	// Kahn:
```
zyga: Khan?

bboozzoo: Nope, Kahn: https://en.wikipedia.org/wiki/Topological_sorting#CITEREFKahn1962

mvo5: Could you please add something (the URL or a short decription) after Kahn: ?

snap/validate.go
```diff
@@ -501,61 +500,9 @@ func validateAppSocket(socket *SocketInfo) error {
...
-		unsatisifed := bytes.Buffer{}
-		for name := range predecessors {
-			if unsatisifed.Len() > 0 {
-				unsatisifed.WriteString(", ")
-			}
-			unsatisifed.WriteString(name)
-		}
-		return fmt.Errorf("applications are part of a before/after cycle: %s", unsatisifed.String())
+func validateAppOrderCycles(apps []*AppInfo) error {
+	if _, err := SortServices(apps); err != nil {
```
mvo5: Nice!

snap/info.go
bboozzoo: Sure

chipaca:
`Before` and `after` are about ordering, not about requirements. And indeed this is what systemd does.

anonymouse64:
@chipaca there's an implicit difference between ordering and requirements in systemd that's not intuitive to me at least and as such I'm hesitant to import that difference into snapd without understanding it. 
While in many situations, doing the same thing as systemd makes sense and is probably the right thing for snapd to do, I have noticed that there are exceptions and I'm worndering if this should be one of those exceptions.

chipaca:
@anonymouse64 this PR does not "import the difference", though; it's making what's already there work in a more predictable way. Note the behaviour we're adding is also missing from systemd, so we're not averse to tweaking things.

However, the place to discuss the sort of thing you're wanting to discuss here is the forum, either under its own topic or building on [Service ordering](https://forum.snapcraft.io/t/service-ordering/1597?u=chipaca) or [Expose a more consistent subset of systemd’s service directives](https://forum.snapcraft.io/t/expose-a-more-consistent-subset-of-systemds-service-directives/2268?u=chipaca).

bboozzoo:
> I haven't had a chance to see what systemd does normally when booting a system with this kind of state installed unfortunately, but I think that probably snap install should do the same thing that systemd does when booting up the system.

@anonymouse64 I've checked what's the behavior of system during boot when there's a service (say B) depending on another service (say A) through `After=A` relation, but A is disabled. Then during boot, service B is started nonetheless. AFAICT, this is the correct behavior and follows the documentation. The docs mention that if you have a service that required another service to function and must be started once that other is active you need to use `After=` & `Requires=`. To be fair we should be able to express `Requires=` relation in the snap YAML too, but that's a topic for another PR.

snap/info.go
```diff
@@ -1138,3 +1138,70 @@ func (r ByType) Swap(i, j int) { r[i], r[j] = r[j], r[i] }
...
+	}
+
+	// list of successors of given app
+	successors := make(map[string][]*AppInfo, len(apps))
+	// count of predecessors (i.e. incoming edges) of given app
+	predecessors := make(map[string]int, len(apps))
+
+	for _, app := range apps {
+		for _, other := range app.After {
+			predecessors[app.Name]++
```
zyga: This is a loop invariant, you could just say `predecessors[app.Name] += len(app.After)` but I understand this is done for parity with the 2nd loop below.

snap/info.go
```diff
@@ -1138,3 +1138,70 @@ func (r ByType) Swap(i, j int) { r[i], r[j] = r[j], r[i] }
...
 }
+
+func SortServices(apps []*AppInfo) (sorted []*AppInfo, err error) {
+	nameToApp := make(map[string]*AppInfo, len(apps))
+	for _, app := range apps {
+		nameToApp[app.Name] = app
+	}
+
+	// list of successors of given app
+	successors := make(map[string][]*AppInfo, len(apps))
```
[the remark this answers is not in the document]
bboozzoo: Hm, I actually moved the code we previously had in `snap/validate.go`, which at the time of the merge already went through a number of renames. IIRC I started with something similar to what you suggest, but we ended up with the current naming anyway.

snap/info.go
```diff
@@ -1138,3 +1138,70 @@ func (r ByType) Swap(i, j int) { r[i], r[j] = r[j], r[i] }
...
+				queue = append(queue, successor)
+			}
+		}
+		sorted = append(sorted, app)
+	}
+
+	if len(predecessors) != 0 {
+		// apps with predecessors unaccounted for are a part of
+		// dependency cycle
+		unsatisifed := bytes.Buffer{}
```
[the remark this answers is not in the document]
bboozzoo: No can do, predecessors is a `map[string]int`.

overlord/snapstate/backend_test.go
```diff
@@ -668,12 +670,19 @@ func (f *fakeSnappyBackend) ReadInfo(name string, si *snap.SideInfo) (*snap.Info
...
 apps:
   svc1:
     daemon: simple
+    before: [svc3]
   svc2:
     daemon: simple
+    after: [svc1]
+  svc3:
+    daemon: simple
+    before: [svc2]
```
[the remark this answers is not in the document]
bboozzoo: Yeah, the ordering is deliberate. I'll leave a comment.

wrappers/services_test.go
```diff
@@ -470,13 +468,19 @@ func (s *servicesTestSuite) TestStartSnapMultiServicesFailStartCleanup(c *C) {
   daemon: simple
 `, &snap.SideInfo{Revision: snap.R(12)})
 
-	err := wrappers.StartServices(info.Services(), nil)
+	svcs := info.Services()
+	c.Assert(svcs, HasLen, 2)
+	if svcs[0].Name == "svc2" {
```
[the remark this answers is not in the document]
bboozzoo: It's not sorted, `info.Services()` iterates over `info.Apps`, which is a map so the order is random-ish.

PR #6025, opened 2018-10-20
Add go.mod files

ryanjyoder:
This adds minimal support for go modules by excluding irregular files which cause errors while importing this repo as a go module. Specifically these files were causing issues:
*  cmd/snap-confine/spread-tests/distros/debian.
*  tests/completion/data/twisted/this isn't.innit

Have you signed the [license agreement](https://www.ubuntu.com/legal/contributors) and read the [contribution guide](https://github.com/snapcore/snapd/blob/master/CONTRIBUTING.md)? yes and yes

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,1 @@
+module github.com/snapcore/snapd


Conan-Kudo:
Is this is a good idea to do right now?

chipaca:
This PR (or, rather, its predecessor which we then messed up and @ryanjyoder was kind enough to resubmit) has triggered quite a bit of internal dialogue, which is still ongoing, about what exactly is the snapd API _as a library_. We've been paying very loose attention to this, as we had no clients (that we knew of!) that were using it as a library directly --- and indeed in general we discourage it because we change things quicker than we can reasonably expect a consumer of the API to keep up, and we've made no stability promises about this API.

Nevertheless, we think you probably should be able to use the client package as a library without it pulling in everything else, and [we do a poor job of that today](http://r.chipaca.com/client.png), so more work is needed.

@ryanjyoder could you help us a little, by telling us what you'd be using the client for?

Thank you.

ryanjyoder:
My use case is pretty straight forward. I'm reading from the snapd REST API to discover which snaps are installed and configure them remotely. I did try placing a single go.mod file in the client directory, but I believe because it references other packages in snapd the entire thing gets pulled in. I can (as you suggested) pull in the client package if I move my project to my gopath, but then of course modules get ignored completely.

I mean the REST api is pretty simple. There's no reason i couldn't roll my own wrapper around it. It's just nice to know I'm using the "official" library, which I assume always stays up-to-date with the API.

chipaca:
That seems reasonable to me, if you're OK working on a library that doesn't yet promise API stability.
We'll be making it more library-like as time allows, which should stop it from pulling in everything under the sun.
I'll +1 based on this.

chipaca:
@Conan-Kudo why would it be a bad idea?

Conan-Kudo:
@chipaca Mainly because we have no API guarantees right now, and as far as I knew, we had no short-term goal to work toward resolving that.

zyga:
@niemeyer please give this a final +1 and merge if you agree.

chipaca:
@Conan-Kudo I think there are separate concerns: one is about API stability, and the other is about how library-like the client package is. We've always thought of client (loosely) as a library, but have been sloppy about it because nobody was using it as such. We'll improve on this. On the other hand, I don't think we're ready to offer stability guarantees. As long as we're clear on that one, it should be fine.

chipaca:
FWIW, FYI, with #6237 now on master, the client dependency graph now is a lot more palatable:
dependency graph after the refactor

PR #6026, opened 2018-10-20
cmd/snap: try not to panic on error from "snap try"

chipaca:
Without this change, when a user tries a "devmode" snap without saying
`--devmode`, the following happens:

    $ snap try
    panic: runtime error: invalid memory address or nil pointer dereference [recovered]
            panic: runtime error: invalid memory address or nil pointer dereference
    [signal 0xb code=0x1 addr=0x2b pc=0x55df7eb53bab]

    goroutine 1 [running]:
    panic(0x55df7f54b140, 0xc82000e140)
            /usr/lib/go-1.6/src/runtime/panic.go:481 +0x3ea
    main.main.func1()
            /build/snapd-zo0q60/snapd-2.36~pre2+git966.ef04a35~ubuntu16.04.1/_build/src/github.com/snapcore/snapd/cmd/snap/main.go:414 +0x7b
    panic(0x55df7f54b140, 0xc82000e140)
            /usr/lib/go-1.6/src/runtime/panic.go:443 +0x4ed
    main.errorToCmdMessage(0x0, 0x0, 0x7fbfc1736150, 0xc820286140, 0x0, 0x0, 0x0, 0x0, 0x0)
            /build/snapd-zo0q60/snapd-2.36~pre2+git966.ef04a35~ubuntu16.04.1/_build/src/github.com/snapcore/snapd/cmd/snap/error.go:158 +0xd2b
    main.run(0x0, 0x0)
            /build/snapd-zo0q60/snapd-2.36~pre2+git966.ef04a35~ubuntu16.04.1/_build/src/github.com/snapcore/snapd/cmd/snap/main.go:464 +0x2c3
    main.main()
            /build/snapd-zo0q60/snapd-2.36~pre2+git966.ef04a35~ubuntu16.04.1/_build/src/github.com/snapcore/snapd/cmd/snap/main.go:419 +0xab7

This affects 2.36 only. The root cause was older, but the panic is
caused by #5791 which is not yet released.

This change addresses the problem in two separate ways: it catches the
nil `opts` (which is always the case when `errorToCmdMessage` is
called from main), but also it calls `errorToCmdMessage` from
`(cmdTry).Execute` directly, which is where `opts` is known.

### DIFF
--- a/cmd/snap/error.go
+++ b/cmd/snap/error.go
@@ -101,6 +101,16 @@ func errorToCmdMessage(snapName string, e error, opts *client.SnapOptions) (stri
 	usesSnapName := true
 	var msg string
 	switch err.Kind {
+	case client.ErrorKindNotSnap:
+		msg = i18n.G(`%q does not contain an unpacked snap.
+
+Try 'snapcraft prime' in your project directory, then 'snap try' again.`)
+		if snapName == "" || snapName == "./" {
+			errValStr, ok := err.Value.(string)
+			if ok && errValStr != "" {
+				snapName = errValStr
+			}
+		}
 	case client.ErrorKindSnapNotFound:
 		msg = i18n.G("snap %q not found")
 		if snapName == "" {
@@ -155,7 +165,7 @@ func errorToCmdMessage(snapName string, e error, opts *client.SnapOptions) (stri
 		isError = false
 		msg = i18n.G(`snap %q is already installed, see 'snap help refresh'`)
 	case client.ErrorKindSnapNeedsDevMode:
-		if opts.Dangerous {
+		if opts != nil && opts.Dangerous {
 			msg = i18n.G("snap %q requires devmode or confinement override")
 			break
 		}
--- a/cmd/snap/main_test.go
+++ b/cmd/snap/main_test.go
@@ -28,6 +28,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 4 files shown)


cmd/snap/main_test.go
```diff
@@ -328,3 +329,13 @@ func (s *SnapSuite) TestFirstNonOptionIsRun(c *C) {
 		c.Check(snap.FirstNonOptionIsRun(), Equals, true)
 	}
 }
+
+// rewrappingMatcher takes a string that's expected to be in the output of
+// a command, and turns it into a regexp that survives rewraps
+func rewrappingMatcher(expected string) string {
```
bboozzoo: This looks quite useful, maybe it could go to `testutil`?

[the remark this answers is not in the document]
chipaca: I'll move it there in a followup. I was sure we were doing something like this elsewhere in cmd/snap but can't find it atm.

cmd/snap/error.go
```diff
@@ -155,7 +165,7 @@ func errorToCmdMessage(snapName string, e error, opts *client.SnapOptions) (stri
 		isError = false
 		msg = i18n.G(`snap %q is already installed, see 'snap help refresh'`)
 	case client.ErrorKindSnapNeedsDevMode:
-		if opts.Dangerous {
+		if opts != nil && opts.Dangerous {
```
zyga: I presume this is the crux of the issue.

zyga: I presume this is the crux of the issue.

PR #6027, opened 2018-10-22
overlord/ifacestate: don't conflict on own discard-snap tasks when refreshing & doing garbage collection

stolowski:
The change made ~1 month ago made auto-connect conflict with retry on "discard-snap" tasks, I missed however that "discard-snap" tasks are created also as part of refresh to remove inactive revisions (normal garbage collection). The trigger to this issue is to have at least 2 revisions of a snap installed, and making a refresh into a new revision. This was recently seen with 2.36pre2.

This PR fixes that by not conflicting on discard-snap if it affects different revision of a snap. I've also enhanced Retry error type to carry optional reason, and enhanced auto-connect error messages on Retry to give precise reason, making figuring these kind of problems out much easier. The existing debug message we had was not precise and was in fact misleading about true cause; also it required snapd executed with SNAPD_DEBUG=1.

### DIFF
--- a/overlord/ifacestate/handlers.go
+++ b/overlord/ifacestate/handlers.go
@@ -635,7 +635,7 @@ func obsoleteCorePhase2SetupProfiles(kind string, task *state.Task) (bool, error
 	return corePhase2, nil
 }
 
-func checkAutoconnectConflicts(st *state.State, plugSnap, slotSnap string) error {
+func checkAutoconnectConflicts(st *state.State, autoconnectTask *state.Task, plugSnap, slotSnap string) error {
 	for _, task := range st.Tasks() {
 		if task.Status().Ready() {
 			continue
@@ -650,8 +650,11 @@ func checkAutoconnectConflicts(st *state.State, plugSnap, slotSnap string) error
 			if err != nil {
 				return err
 			}
-			if plugRef.Snap == plugSnap || slotRef.Snap == slotSnap {
-				return &state.Retry{After: connectRetryTimeout}
+			if plugRef.Snap == plugSnap {
+				return &state.Retry{After: connectRetryTimeout, Reason: fmt.Sprintf("conflicting snap %s, task %q", plugSnap, k)}
+			}
+			if slotRef.Snap == slotSnap {
+				return &state.Retry{After: connectRetryTimeout, Reason: fmt.Sprintf("conflicting snap %s, task %q", slotSnap, k)}
 			}
 			continue
 		}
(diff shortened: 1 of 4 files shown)


overlord/ifacestate/handlers.go
```diff
@@ -650,8 +650,11 @@ func checkAutoconnectConflicts(st *state.State, plugSnap, slotSnap string) error
 			if err != nil {
 				return err
 			}
-			if plugRef.Snap == plugSnap || slotRef.Snap == slotSnap {
-				return &state.Retry{After: connectRetryTimeout}
+			if plugRef.Snap == plugSnap {
+				return &state.Retry{After: connectRetryTimeout, Reason: fmt.Sprintf("conflicting snap %s, task %q", plugSnap, k)}
```
mvo5: Should we include info that its the plug or slot side in the reason? Or is this always obvious from the context?

mvo5:
@stolowski unit tests are failing with:
```
FAIL: ifacestate_test.go:4382: interfaceManagerSuite.TestGadgetConnectConflictRetry

ifacestate_test.go:4409:
    c.Assert(chg.Err(), IsNil)
... value *state.changeError = &state.changeError{errors:[]state.taskError{state.taskError{task:"gadget connections", error:"gadget connect conflict check failed: no state entry for key"}}} ("cannot perform the following tasks:\n- gadget connections (gadget connect conflict check failed: no state entry for key)")
```

overlord/ifacestate/handlers.go
stolowski: Good idea, done.

overlord/ifacestate/handlers.go
```diff
@@ -682,9 +685,19 @@ func checkAutoconnectConflicts(st *state.State, plugSnap, slotSnap string) error
...
 		// other snap that affects us because of plug or slot
 		if k == "unlink-snap" || k == "link-snap" || k == "setup-profiles" || k == "discard-snap" {
+			autoconnectsup, err := snapstate.TaskSnapSetup(autoconnectTask)
+			if err != nil {
+				return err
+			}
+
+			// discard-snap is scheduled as part of garbage collection during refresh, if multiple revsions are already installed.
+			// this revision check avoids conflict with own discard tasks created as part of install/refresh.
+			if k == "discard-snap" && autoconnectsup.Revision().N != snapsup.Revision().N {
```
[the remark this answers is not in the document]
stolowski: Indeed, that's cheaper and more simple. Done.

PR #6028, opened 2018-10-22
testutil, cmd/snap: introduce and use testutil.EqualsWrapped and fly

chipaca:
This introduces a family of checkers to `testutil`, which are useful
for when you want to check for a string in output that can be wrapped
to fit the terminal, or aligned as part of a table. E.g. if you want
to check for the third line in

```
Name   Version                      Rev   Tracking  Publisher   Notes
bare   1.0                          5     edge      canonical✓  base
black  18.6b4                       1     edge      sergiusens  devmode
core   16-2.36~pre2+git967.54466bd  5780  edge      canonical✓  core
```

then you can use

```go
Check(theLine, EqualsPadded, "black 18.6b4 1 edge sergiusens devmode")
```

Similarily, to check that the output of installing a `devmode` snap
without `--devmode`,

```
error: The publisher of snap "eggg" has indicated that they do not consider this
       revision to be of production quality and that it is only meant for
       development or testing at this point. As a consequence this snap will not
       refresh automatically and may perform arbitrary system changes outside of
       the security sandbox snaps are generally confined to, which may put your
       system at risk.

       If you understand and want to proceed repeat the command including
       --devmode; if instead you want to install the snap forcing it into strict
       confinement repeat the command including --jailmode.
```

if you wanted to check that it says "only meant for development",
which might be on the same line or not depending on terminal width,
you'd simply say

```go
Check(output, ContainsWrapped, "only meant for development")
```

HTH, HAND!

### DIFF
--- a/testutil/checkers.go
+++ b/testutil/checkers.go
@@ -239,6 +239,125 @@ func (c *deepContainsChecker) Check(params []interface{}, names []string) (resul
 	}
 }
 
+type paddedChecker struct {
+	*check.CheckerInfo
+	isRegexp  bool
+	isPartial bool
+	isLine    bool
+}
+
+// EqualsPadded is a Checker that looks for an expected string to
+// be equal to another except that the other might have been padded
+// out to align with something else (so arbitrary amounts of
+// horizontal whitespace is ok at the ends, and between non-whitespace
+// bits).
+var EqualsPadded = &paddedChecker{
+	CheckerInfo: &check.CheckerInfo{Name: "EqualsPadded", Params: []string{"padded", "expected"}},
+	isLine:      true,
+}
+
+// MatchesPadded is a Checker that looks for an expected regexp in
+// a string that might have been padded out to align with something
+// else (so whitespace in the regexp is changed to [ \t]+, and ^[ \t]*
+// is added to the beginning, and [ \t]*$ to the end of it).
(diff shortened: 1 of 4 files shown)


testutil/checkers.go
```diff
@@ -239,6 +239,125 @@ func (c *deepContainsChecker) Check(params []interface{}, names []string) (resul
...
+	if !c.isLine {
+		expected = "(?s)" + expected
+	}
+
+	matches, err := regexp.MatchString(expected, other)
+	if err != nil {
+		// can't happen (really)
+		panic(err)
+	}
+	return matches, ""
```
bboozzoo: Wonder if we could produce a useful error message here. Maybe something with FindAllString()?

```
re, _ := regexp.Compile(expected)
matches, err := re.MatchString(other)
errMsg := ""
if !matches {
    somethingMatched := re.FindAllString(other, -1)
    if len(somethingMatched) > 0 {
        errMsg = fmt.Sprintf("this is what matched instead %q", somethingMatched)
    }
}
return matches, errMsg
```

chipaca: I'm not sure I understand correctly: AFAIU if `MatchString` returns `false`,  then `FindAllString` isn't going to find anything...?

bboozzoo: Uhh, right. Ignore that.

PR #6029, opened 2018-10-22
snapstate: add command-chain to supported featureset

kyrofa:
snapd supports the `assumes` property in a snap.yaml, allowing a snap to depend upon a specific snapd release, or specific snapd feature. Add command-chain as one of those features, which allows snaps to specify that they require command-chain in order to work properly.

It would be great to get this into 2.36 since that's the release actually containing command-chain. Of course, even if that doesn't happen, this should still be useful.


kyrofa:
@mvo5 what do you think about getting this into 2.36?

PR #6030, opened 2018-10-23
cmd/snap: tweak `snap services` output when there is no services

bboozzoo:
This is the current output when there are no snaps with services:

```
$ snap services
Service  Startup  Current  Notes
```

Try to fix it by printing a more user-friendly message.

Some bits were pulled from #6008

### DIFF
--- a/cmd/snap/cmd_services.go
+++ b/cmd/snap/cmd_services.go
@@ -125,7 +125,7 @@ func (s *svcStatus) Execute(args []string) error {
 	}
 
 	if len(services) == 0 {
-		fmt.Fprintln(Stderr, i18n.G("There are no snaps installed that provide a service."))
+		fmt.Fprintln(Stderr, i18n.G("There are no services provided by installed snaps."))
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/snap/cmd_services.go
```diff
@@ -124,6 +124,11 @@ func (s *svcStatus) Execute(args []string) error {
 		return err
 	}
 
+	if len(services) == 0 {
+		fmt.Fprintln(Stderr, i18n.G("No snaps providing services are installed."))
```
chipaca: that's a little clunky, indeed. My suggestion isn't much better, so I'll defer to @degville unless he comes up empty.

bboozzoo: Hm, a thought, since the command is `snap services`, maybe something like this could work `There are no services provided by snaps` (or `... installed snaps`). wdyt?

degville: @bboozzoo I think you're right - yes, that's better (I'd go with *installed snaps*).

cmd/snap/cmd_services.go
```diff
@@ -124,6 +124,11 @@ func (s *svcStatus) Execute(args []string) error {
 		return err
 	}
 
+	if len(services) == 0 {
+		fmt.Fprintln(Stderr, i18n.G("There are no services provided by installed snaps."))
```
zyga: Really nice, +1

PR #6040, opened 2018-10-24
data/apt: close stderr when calling snap in the apt install hook.

chipaca:
See [the forum] for what could happen otherwise. Targeting to 2.36 in case there's a .1 happening.

[the forum]: https://forum.snapcraft.io/t/weird-error-message-whenever-i-run-the-snap-command/6418/14?u=chipaca


mvo5:
Thanks for this !

PR #6044, opened 2018-10-25
cmd/snap-confine: remove stale mount profile along stale namespace

zyga:
When snap-confine attempts to start up and notices that the mount
namespace is stale and can be discarded it does so correctly but it
doesn't remove the .fstab file that has become stale because the mount
namespace was discarded. When snap-update-ns is subsequently used it
thinks that the mount namespace has all the changes applied to it and
then does nothing.

Thanks to Alan Pope for reporting the issue and providing helpful data.


zyga:
I will follow this up with a more proper fix, one that uses `snap-discard-ns` (so that there is really just one logic) but this will be more involved. This is a hot fix to back port and release.

I was thinking about how to write a spread test for this but it's somewhat tricky.

PR #6048, opened 2018-10-25
cmd/snap-exec: don't fail on some try mode snaps

zyga:
Snaps installed via "snap try" may have a dangling symlink from
/var/lib/snapd/snaps/snap_rev since the source location is unconstrained.  In
the execution environment that path may not exist or may point to unrelated
data, in a malicious case. The snap-exec command used snap.ReadInfo to obtain
information about the application or hook to exec. That function tries to stat
the snap file to compute the size of the snap. Since snap-exec does not need
the size and since this cannot be achieved in the execution environment we
should not attempt that.

This branch adds and uses a new helper that is exactly like ReadInfo but
doesn't compute the size of the snap.

Fixes: https://bugs.launchpad.net/snapd/+bug/1800004

### DIFF
--- a/snap/info.go
+++ b/snap/info.go
@@ -985,8 +985,10 @@ var SanitizePlugsSlots = func(snapInfo *Info) {
 	panic("SanitizePlugsSlots function not set")
 }
 
-// ReadInfo reads the snap information for the installed snap with the given name and given side-info.
-func ReadInfo(name string, si *SideInfo) (*Info, error) {
+// ReadInfoExceptSize reads the snap information for the installed snap with the given name and given side-info.
+//
+// The Size field of snap.Info is not set.
+func ReadInfoExceptSize(name string, si *SideInfo) (*Info, error) {
 	snapYamlFn := filepath.Join(MountDir(name, si.Revision), "meta", "snap.yaml")
 	meta, err := ioutil.ReadFile(snapYamlFn)
 	if os.IsNotExist(err) {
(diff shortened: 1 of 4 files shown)


snap/info.go
```diff
@@ -1001,6 +1003,24 @@ func ReadInfo(name string, si *SideInfo) (*Info, error) {
...
+// ReadInfo reads the snap information for the installed snap with the given name and given side-info.
+func ReadInfo(name string, si *SideInfo) (*Info, error) {
+	info, err := ReadInfoExceptSize(name, si)
+	if err != nil {
+		return nil, err
+	}
+
 	mountFile := MountFile(name, si.Revision)
 	st, err := os.Stat(mountFile)
 	if os.IsNotExist(err) {
```
kyrofa: Am I being silly, or is the only reason this works today because apparmor leaks stats? Consider: if the snap I'm `try`ing doesn't use the `home` plug, but the source code for the snap is in `~/src/snaps/` then the symlink that is `mountFile` is pointing to a location inaccessible by snap-exec. However, because appamor leaks stats, we happen to have the ability to determine that the symlink is pointing to something that exists.

zyga: Apparmor doesn't check stat, that's correct. As for Lstat - perhaps, cc @chipaca for opinion.

In any case inside the mount namespace we cannot even attempt to follow the symlink as someone may craft it to be malicious.

snap/info_test.go
```diff
@@ -334,6 +334,19 @@ func (s *infoSuite) TestReadInfoUnfindable(c *C) {
...
+	si := &snap.SideInfo{Revision: snap.R(42), EditedSummary: "esummary"}
+	p := filepath.Join(snap.MinimalPlaceInfo("sample", si.Revision).MountDir(), "meta", "snap.yaml")
+	c.Assert(os.MkdirAll(filepath.Dir(p), 0755), IsNil)
+	c.Assert(ioutil.WriteFile(p, []byte(`name: test`), 0644), IsNil)
+
+	info, err := snap.ReadInfoExceptSize("sample", si)
+	c.Check(err, IsNil)
+	c.Check(info.SnapName(), Equals, "test")
+	c.Check(info.Revision, Equals, snap.R(42))
+	c.Check(info.Summary(), Equals, "esummary")
```
stolowski: I'd add a check for .Size == 0.

chipaca:
Yeah, that size is wrong for 'try' snaps, info should also not show it.

PR #6051, opened 2018-10-26
cmd/snap-update-ns: parse the -u <uid> command line option

zyga:
This patch extends the C command line parser to support the -u <uid>
option. The option implies we are being called by snapd to update the
mount namespace of a given user. It implies we should jump to the
appropriate namespace ourselves, use per-user mount profile and perform
our task.

### DIFF
--- a/cmd/snap-update-ns/bootstrap.c
+++ b/cmd/snap-update-ns/bootstrap.c
@@ -325,7 +325,7 @@ int validate_instance_name(const char *instance_name)
 }
 
 // parse the -u argument, returns -1 on failure or 0 on success.
-static int parse_arg_u(int argc, char * const *argv, int *optind, int *uid_out)
+static int parse_arg_u(int argc, char * const *argv, int *optind, unsigned long *uid_out)
 {
 	if (*optind + 1 == argc || argv[*optind + 1] == NULL) {
 		bootstrap_msg = "-u requires an argument";
@@ -335,16 +335,22 @@ static int parse_arg_u(int argc, char * const *argv, int *optind, int *uid_out)
 	const char *uid_text = argv[*optind + 1];
 	errno = 0;
 	char *uid_text_end = NULL;
-	int parsed_uid = (int)strtol(uid_text, &uid_text_end, 10);
-	if ((parsed_uid == 0 && errno != 0)
+	unsigned long parsed_uid = strtoul(uid_text, &uid_text_end, 10);
+	if (
+			/* Reject overflow in parsed representation */
+			(parsed_uid == ULONG_MAX && errno != 0)
+			/* Reject leading whitespace allowed by strtoul. */
+			|| (isspace(*uid_text))
+			/* Reject empty string. */
 			|| (*uid_text == '\0')
+			/* Reject partially parsed strings. */
 			|| (*uid_text != '\0' && uid_text_end != NULL
 				&& *uid_text_end != '\0')) {
 		bootstrap_msg = "cannot parse user id";
 		bootstrap_errno = errno;
 		return -1;
 	}
-	if (parsed_uid < 0) {
+	if ((long)parsed_uid < 0) {
 		bootstrap_msg = "user id cannot be negative";
 		bootstrap_errno = 0;
 		return -1;
@@ -359,7 +365,7 @@ static int parse_arg_u(int argc, char * const *argv, int *optind, int *uid_out)
 // process_arguments parses given a command line
 // argc and argv are defined as for the main() function
 void process_arguments(int argc, char *const *argv, const char **snap_name_out,
-		       bool * should_setns_out, bool * process_user_fstab, int * uid_out)
+		       bool * should_setns_out, bool * process_user_fstab, unsigned long * uid_out)
 {
 	// Find the name of the called program. If it is ending with ".test" then do nothing.
 	// NOTE: This lets us use cgo/go to write tests without running the bulk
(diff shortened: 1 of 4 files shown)


cmd/snap-update-ns/bootstrap.c
```diff
@@ -327,7 +327,7 @@ int validate_instance_name(const char *instance_name)
 // process_arguments parses given a command line
 // argc and argv are defined as for the main() function
 void process_arguments(int argc, char *const *argv, const char **snap_name_out,
-		       bool * should_setns_out, bool * process_user_fstab)
+		       bool * should_setns_out, bool * process_user_fstab, int * uid_out)
```
bboozzoo: This should probably be `uid_t`

cmd/snap-update-ns/bootstrap.c
```diff
@@ -368,6 +368,40 @@ void process_arguments(int argc, char *const *argv, const char **snap_name_out,
...
+			} else if (!strcmp(arg, "-u")) {
+				const char *uid_text = argv[i + 1];
+				if (i + 1 == argc || uid_text == NULL) {
+					bootstrap_msg = "-u requires an argument";
+					bootstrap_errno = 0;
+					return;
+				}
+				errno = 0;
+				char *uid_text_end = NULL;
+				int parsed_uid = (int)strtol(uid_text, &uid_text_end, 10);
```
bboozzoo: `strtoul`?

Also, I think the string should be pre-validated first. `strto*` allow all kind of weird things, eg. `"    0x123", "0123", "\n123"` are valid strings for `strto*`. In contrast, we should allow only `[0-9]+`.

cmd/snap-update-ns/bootstrap.c
```diff
@@ -368,6 +368,40 @@ void process_arguments(int argc, char *const *argv, const char **snap_name_out,
 				// Processing the user-fstab file implies we're being
 				// called from snap-confine.
 				should_setns = false;
+			} else if (!strcmp(arg, "-u")) {
+				const char *uid_text = argv[i + 1];
```
stolowski: I think wee should check `if i+1==argc` before referencing argv[i+1], not below.

cmd/snap-update-ns/bootstrap.c
```diff
@@ -368,6 +368,40 @@ void process_arguments(int argc, char *const *argv, const char **snap_name_out,
...
+				if (i + 1 == argc || uid_text == NULL) {
```
bboozzoo: I think this should look like this:
```
if (i + 1 == argc || strlen(argv[i+1][0]) == 0) {
	bootstrap_msg = "-u requires an argument";
	bootstrap_errno = 0;
	return;
}
const char* uid_test = argv[i+1];
```

cmd/snap-update-ns/bootstrap.c
```diff
@@ -368,6 +368,40 @@ void process_arguments(int argc, char *const *argv, const char **snap_name_out,
...
+							&& *uid_text_end != '\0')) {
+					bootstrap_msg = "cannot parse user id";
+					bootstrap_errno = errno;
+					return;
+				}
+				if (parsed_uid < 0) {
+					bootstrap_msg = "user id cannot be negative";
+					bootstrap_errno = 0;
+					return;
+				}
```
stolowski: IMHO this parsing code could be extracted into a helper, resulting in more compact body of process_arguments.

cmd/snap-update-ns/bootstrap.c
zyga: The argument vector is always terminated with a NULL pointer so there will always be one extra slot to look at but I agree that for extra sanity it should be done after the check.

cmd/snap-update-ns/bootstrap.go
```diff
@@ -114,19 +114,21 @@ func validateInstanceName(instanceName string) int {
 //
 // This function is here only to make the C.validate_instance_name
 // code testable from go.
-func processArguments(args []string) (snapName string, shouldSetNs bool, processUserFstab bool) {
+func processArguments(args []string) (snapName string, shouldSetNs bool, processUserFstab bool, uid int) {
```
bboozzoo: .. and Golang which does not cover the full range of possible user IDs. I think we should use `uint` regardless of the type returned by functions in `os`.

cmd/snap-update-ns/bootstrap.c
zyga: I will expand the test coverage and see what happens. I think we validate enough because we ensure the whole string is parsed and we provide a base. If there's anything remaining it will indicate failure.

cmd/snap-update-ns/bootstrap.c
```diff
@@ -368,6 +400,17 @@ void process_arguments(int argc, char *const *argv, const char **snap_name_out,
 				// Processing the user-fstab file implies we're being
 				// called from snap-confine.
 				should_setns = false;
+			} else if (!strcmp(arg, "-u")) {
+				if (parse_arg_u(argc, argv, &i, uid_out)) {
```
stolowski: Thank you, that's much easier to read.

cmd/snap-update-ns/bootstrap.c
zyga: Note: I switched to unsigned long. It should be sufficient and explicit (I can format it).

cmd/snap-update-ns/bootstrap.c
zyga: I moved the code to `strtoul` and added way more tests now.

PR #6057, opened 2018-10-29
[RFC] snap-exec: add comment about usage of ReadInfoExceptSize()

mvo5:
In https://github.com/snapcore/snapd/pull/6048 we added
ReadInfoExceptSize(). This PR adds a small comment in the
code about why its used instead of the regular ReadInfo().

Note sure if we *really* need it, just thought it would be nice when
reading the original PR.


chipaca:
We just talked about this with zyga, I think we're going to push a pr that reverts #6048 and instead does  a Lstat

chipaca:
Closing this in favour of #6059.

PR #6059, opened 2018-10-29
snap: use Lstat to determine snap size, remove ReadSnapInfoExceptSize

zyga:
Since we can os.Lstat any symlink safely we can determine that it is
actually a symbolic link and then decide not to follow it or use any
information about it.

This simplifies the API, keeping the fix to LP: #1800004 fixed.
Thanks to Chipaca for the idea.


chipaca:
Credit goes to @kyrofa for mentioning lstat (in the original pr!). Sorry I didn't have the bandwidth to look into this before now.

PR #6060, opened 2018-10-29
overlord/snapshotstate/backend: be more verbose when SNAPPY_TESTING=1

chipaca:
We've seen some very weird errors from tests/main/snapshot*
occasionally pop up in spread, with tar failing with weird
errors. This change makes tar's stderr reach stderr, so we can see the
error itself.

It also makes the reader and writer errors more similar.

### DIFF
--- a/overlord/snapshotstate/backend/reader.go
+++ b/overlord/snapshotstate/backend/reader.go
@@ -309,11 +309,20 @@ func (r *Reader) Restore(ctx context.Context, current snap.Revision, usernames [
 			"--directory", tempdir)
 		cmd.Env = []string{}
 		cmd.Stdin = tr
-		cmd.Stderr = os.Stderr
+		matchCounter := &strutil.MatchCounter{N: 1}
+		cmd.Stderr = matchCounter
 		cmd.Stdout = os.Stderr
+		if isTesting {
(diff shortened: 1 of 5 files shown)


overlord/snapshotstate/backend/reader.go
```diff
@@ -309,11 +309,19 @@ func (r *Reader) Restore(ctx context.Context, current snap.Revision, usernames [
 			"--directory", tempdir)
 		cmd.Env = []string{}
 		cmd.Stdin = tr
-		cmd.Stderr = os.Stderr
-		cmd.Stdout = os.Stderr
+		matchCounter := &strutil.MatchCounter{Regexp: matchAll, N: 1}
```
zyga: Using a match counter that matches everything is a little weird. Perhaps this could use a comment in the code to explain the rationale?

chipaca: with a regexp that matches everything, `strutil.MatchCounter` is a `strutil.LimitedBuffer` with no line length limit, that remembers how many lines it got.

chipaca: I've updated it to make use of the new regexpless match counter

PR #6062, opened 2018-10-29
tests,store,daemon: ensure proxy settings are honored in auth/userinfo too

mvo5:
We got a bugreport (LP: 1791587) that the proxy.http setting is
not honored in snapd on classic systems when no core is installed.

This PR adds a integration test to ensure this works and fixes the
issues around proxy support that prevent this from working (mostly
device and user session auth).


tests/main/proxy-no-core/task.yaml
```diff
@@ -0,0 +1,42 @@
+summary: Ensure that the core.proxy.* settings are honored without core
+
+# only needs a test on classic
+systems: [ubuntu-16.04-64, ubuntu-18.04-64]
+
+restore: |
+    iptables -D OUTPUT -m owner --uid-owner "$(id -u test)" -j ACCEPT -p tcp
```
zyga: Perhaps instead of hand-crafted restore we should use a dump, modify, load approach where here we would just load exactly what was dumped before?

tests/main/proxy-no-core/task.yaml
```diff
@@ -0,0 +1,42 @@
...
+    iptables -D OUTPUT -j REJECT -p tcp --dport https --reject-with tcp-reset
+
+execute: |
+    if ! command -v python3; then
+       echo "SKIP: need python3"
+       exit 0
+    fi
+    #shellcheck source=tests/lib/systemd.sh
+    . "$TESTSLIB"/systemd.sh
+    systemctl stop snapd.service snapd.socket
```
zyga: AFAIK every time I use this order socket is stopped last and system warns me that snapd can be activated by the socket.

tests/main/proxy-no-core/task.yaml
```diff
@@ -0,0 +1,42 @@
...
+    rm -rf /var/lib/snapd/state.json
+    systemctl start snapd.service snapd.socket
+
+    # only allow test user to connect to http/https
+    iptables -I OUTPUT -j REJECT -p tcp --dport http --reject-with tcp-reset
+    iptables -I OUTPUT -j REJECT -p tcp --dport https --reject-with tcp-reset
+    iptables -I OUTPUT -m owner --uid-owner "$(id -u test)" -j ACCEPT -p tcp
+    systemd-run --uid=test --unit tinyproxy -- python3 "$TESTSLIB/tinyproxy/tinyproxy.py"
```
zyga: What ensures that this is stopped?

tests/main/proxy-no-core/task.yaml
```diff
@@ -0,0 +1,42 @@
...
+        echo "without a proxy core install should fail, test broken"
+        exit 1
+    fi
+
+    echo "Setup proxy config"
+    snap set core proxy.https=http://localhost:3128
+
+    echo "Ensure that snap install works even without an installed core to install core"
+    snap install core
+    check_journalctl_log 'CONNECT fastly.cdn.snapcraft.io' -u tinyproxy
```
zyga: This failed in spread tests, it looks like missing helper "import" somewhere.

```
+ check_journalctl_log 'CONNECT fastly.cdn.snapcraft.io' -u tinyproxy
/bin/bash: line 89: check_journalctl_log: command not found
```

tests/main/proxy-no-core/task.yaml
mvo5: Thanks, this was indeed missing, fixed now.

tests/main/proxy-no-core/task.yaml
mvo5: Yeah, missing as well, thank you!

daemon/api.go
```diff
@@ -416,7 +418,9 @@ func loginUser(c *Command, r *http.Request, user *auth.UserState) Response {
 		}, nil)
 	}
 
-	macaroon, discharge, err := store.LoginUser(loginData.Email, loginData.Password, loginData.Otp)
```
mvo5: The alternative to this would be to make "LoginUser()" part of StoreService - which probably makes sense?

daemon/api.go
```diff
@@ -2185,8 +2187,17 @@ var (
...
+func makeHttpClient(st *state.State) *http.Client {
+	proxyConf := proxyconf.New(st)
+	return httputil.NewHTTPClient(&httputil.ClientOptions{
+		Timeout:    10 * time.Second,
+		MayLogBody: true,
+		Proxy:      proxyConf.Conf,
+	})
+}
+
+func getUserDetailsFromStore(client *http.Client, email string) (string, *osutil.AddUserOptions, error) {
```
mvo5: And for this to add "UserInfo()" part of StoreService

daemon/api.go
mvo5: Thanks, I will probably do this as a followup, it is quite a bit of extra churn :/

dshcherb:
@mvo5 

Did some manual verification with this code - seems to work:

https://bugs.launchpad.net/snapd/+bug/1791587/comments/10

PR #6065, opened 2018-10-30
cmd: make coreSupportsReExec faster

chipaca:
coreSupportsReExec is rather critical performance-wise, as it's used
for every invocation of snap, including every binary in /snap/bin.

So I thought I might try to speed it up a bit. This ended up with some
work in `strutil` which got split into #6074, but there was still room
for improvement here. Even with that on master, this change helps:

```
$ benchcmp /tmp/master.txt /tmp/this.txt
benchmark                    old ns/op     new ns/op     delta
BenchmarkCSRE_real-8         18511         4256          -77.01%

benchmark                    old allocs     new allocs     delta
BenchmarkCSRE_real-8         50             12             -76.00%

benchmark                    old bytes     new bytes     delta
BenchmarkCSRE_real-8         44041         1176          -97.33%
```

HTH, HAND.

### DIFF
--- a/cmd/cmd_linux.go
+++ b/cmd/cmd_linux.go
@@ -20,11 +20,11 @@
 package cmd
 
 import (
+	"bytes"
 	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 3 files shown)


zyga:
As a quick comment I would feel more comfortable reviewing this as separate patches.

strutil/version.go
```diff
@@ -80,6 +73,37 @@ func chOrder(ch uint8) int {
...
+		for j := 0; j < 16; j++ {
+			fmt.Printf("%d, ", chOrder(uint8(i*16+j)))
+		}
+		fmt.Println()
+	}
+	fmt.Println("}")
+}
+
+*/
+var chOrder = [...]int{
```
bboozzoo: Played with this a little and turned it into a generator in case it needs tweaking in the future. See https://github.com/snapcore/snapd/compare/master...bboozzoo:bboozzoo/chrorder-generator

zyga: That's super nice, would you mind pushing the generator into this PR?

chipaca:
@mvo5 @bboozzoo 811eda2dc4f91f2806ae5a363dcaac36776c977b would make arch-ish 1-2-3 versions work as well, I can add it to this if you think it's a good idea.
Or I can split this into a 'revamp versionn.go' and the cmd changes, to keep @zyga happy.

cmd/cmd_linux.go
```diff
@@ -79,27 +79,36 @@ func distroSupportsReExec() bool {
...
+			logger.Noticef("cannot open snapd info file %q: %s", fullInfo, err)
+		}
 		return false
 	}
-	ver := regexp.MustCompile("(?m)^VERSION=(.*)$").FindStringSubmatch(string(content))
-	if len(ver) != 2 {
-		logger.Noticef("cannot find snapd version information in %q", content)
-		return false
+
+	if !bytes.HasPrefix(content, []byte("VERSION=")) {
```
mvo5: (nitpick) I think its slightly easier to read if this goes into a helper function (like versionFromInfoContent() or something). The old code was relatively simple (just a regexp match) but the new one requires more state when reading it.

PR #6067, opened 2018-10-30
tests/main/snap-service-after-before-install: verify after/before in snap install

bboozzoo:
Add spread tests for verification of after/before handling during `snap
install`. To be able to observe the whole process in detail, the after/before
snap had to be modified. Switching to `daemon: notify` allows to have a
controlled time gap between the process leaving inactive and entering active
states.

Further, even before the fix was implemented, the processes could have been
started in the right order. This was internally caused how the application map
of the snap.Info was iterated over in a semi-random fashion. For this reason,
the actual test case is repeated 10 times. With just 3 service applications, it
should be possible to rule out the test passing by accident.

### DIFF
--- a/tests/main/snap-service-after-before-install/task.yaml
+++ b/tests/main/snap-service-after-before-install/task.yaml
@@ -1,0 +1,73 @@
+summary: Check that snap install respects after/before ordering of services
+
+debug: |
+    for name in $(snap list |grep '^test-snapd-after-before-service' | awk '{print $1}'); do
+        for service in before-middle middle after-middle; do
+            systemctl status "snap.$name.$service" || true
+        done
+    done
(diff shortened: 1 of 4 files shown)


tests/main/snap-service-after-before-install/task.yaml
```diff
@@ -0,0 +1,73 @@
+summary: Check that snap install respects after/before ordering of services
+
+debug: |
+    for name in $(snap list |grep '^test-snapd-after-before-service' | awk '{print $1}'); do
```
zyga:
```suggestion
    for name in $(snap list | grep '^test-snapd-after-before-service' | awk '{print $1}'); do
```

PR #6068, opened 2018-10-30
ifacestate/helpers: findConnsForHotplugKey helper

stolowski:
Helper function for finding connState entries matching given hotplug key and interface type.

### DIFF
--- a/overlord/ifacestate/helpers.go
+++ b/overlord/ifacestate/helpers.go
@@ -23,6 +23,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/snapcore/snapd/asserts"
@@ -829,5 +830,6 @@ func findConnsForHotplugKey(conns map[string]connState, ifaceName, hotplugKey st
 		}
 		connsForDevice = append(connsForDevice, id)
 	}
+	sort.Strings(connsForDevice)
 	return connsForDevice
 }
--- a/overlord/ifacestate/helpers.go
+++ b/overlord/ifacestate/helpers.go
@@ -820,3 +820,14 @@ func getHotplugSlots(st *state.State) (map[string]*HotplugSlotInfo, error) {
 func setHotplugSlots(st *state.State, slots map[string]*HotplugSlotInfo) {
 	st.Set("hotplug-slots", slots)
 }
+
+func findConnsForHotplugKey(conns map[string]connState, ifaceName, hotplugKey string) []string {
+	var connsForDevice []string
+	for id, connSt := range conns {
+		if connSt.Interface != ifaceName || connSt.HotplugKey != hotplugKey {
+			continue
+		}
+		connsForDevice = append(connsForDevice, id)
+	}
+	return connsForDevice
+}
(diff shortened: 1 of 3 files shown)


overlord/ifacestate/helpers_test.go
```diff
@@ -213,3 +214,41 @@ func (s *helpersSuite) TestHotplugSlotInfo(c *C) {
 	c.Assert(err, IsNil)
 	c.Assert(slots, DeepEquals, defs)
 }
+
+func (s *helpersSuite) TestFindConnsForHotplugKey(c *C) {
+	st := s.st
+	st.Lock()
+	defer st.Unlock()
+
+	st.Set("conns", map[string]interface{}{
```
chipaca: maybe a comment about how you're setting it on state and then getting it just to avoid having to know the internals `connState`?

overlord/ifacestate/helpers.go
```diff
@@ -820,3 +820,14 @@ func getHotplugSlots(st *state.State) (map[string]*HotplugSlotInfo, error) {
 func setHotplugSlots(st *state.State, slots map[string]*HotplugSlotInfo) {
 	st.Set("hotplug-slots", slots)
 }
+
+func findConnsForHotplugKey(conns map[string]connState, ifaceName, hotplugKey string) []string {
```
zyga: Should the signature be `map[string]*connState` to avoid copying connState around?
Should we sort the result or indicate that ordering is not deterministic? - I would go for `sort.Strings` since the result will likely be very short.

overlord/ifacestate/helpers_test.go
```diff
@@ -213,3 +214,43 @@ func (s *helpersSuite) TestHotplugSlotInfo(c *C) {
...
+
+	conns, err := ifacestate.GetConns(st)
+	c.Assert(err, IsNil)
+
+	hotplugConns := ifacestate.FindConnsForHotplugKey(conns, "iface1", "key1")
+	c.Assert(hotplugConns, DeepEquals, []string{"snap1:plug1 core:slot1"})
+
+	hotplugConns = ifacestate.FindConnsForHotplugKey(conns, "iface2", "key2")
+	sort.Strings(hotplugConns)
+	c.Assert(hotplugConns, DeepEquals, []string{"snap1:plug3 core:slot3", "snap2:plug1 core:slot1"})
```
zyga: Won't this sometimes fail because of ordering of map traversal?

overlord/ifacestate/helpers_test.go
```diff
@@ -213,3 +214,43 @@ func (s *helpersSuite) TestHotplugSlotInfo(c *C) {
...
+	})
+
+	conns, err := ifacestate.GetConns(st)
+	c.Assert(err, IsNil)
+
+	hotplugConns := ifacestate.FindConnsForHotplugKey(conns, "iface1", "key1")
+	c.Assert(hotplugConns, DeepEquals, []string{"snap1:plug1 core:slot1"})
+
+	hotplugConns = ifacestate.FindConnsForHotplugKey(conns, "iface2", "key2")
+	sort.Strings(hotplugConns)
```
zyga: Ah, I missed that!

overlord/ifacestate/helpers.go
stolowski: Will address this in a followup due to changes required in unrelated code (getConns, setConns).

PR #6069, opened 2018-10-30
ifacestate/hotplug: removeDevice helper

stolowski:
removeDevice helper creates two tasks that handle hotplug remove event:
- "hotplug-disconnect" task will create actual disconnect task and disconnect hooks (and handle retry if neccessary if there is a conflicting change in progress).
- "hotplug-remove-slot" task will remove the slot from the repository.

The actual handlers of these two tasks will be proposed in separate PRs (but it's included in the big hotplug PR #5962)


overlord/ifacestate/hotplug.go
```diff
@@ -173,3 +174,17 @@ func suggestedSlotName(devinfo *hotplug.HotplugDeviceInfo, fallbackName string)
 	}
 	return shortestName
 }
+
+// create tasks to disconnect slots of given device and remove affected slots.
```
zyga:
```suggestion
// removeDevice creates tasks to disconnect slots of given device and remove affected slots.
```

PR #6072, opened 2018-10-30
ifacestate/udevmonitor: added callback to signal end of enumeration

stolowski:
This PR adds a callback to signal the end of initial enumeration of devices (via udevadm) to interface manager. This is required to know when to compute the "delta" between currently present devices and the past (not in this PR).
Also made the two existing ifacemgr hotplug callbacks private as they don't need to be exported.


overlord/ifacestate/udevmonitor/udevmon.go
```diff
@@ -88,8 +91,15 @@ func (m *Monitor) Connect() error {
 		return fmt.Errorf("cannot start udev monitor: %s", err)
 	}
 
-	// TODO: consider passing a device filter to reduce noise from irrelevant devices.
-	m.monitorStop = m.netlinkConn.Monitor(m.netlinkEvents, m.netlinkErrors, nil)
+	var filter netlink.Matcher
+	// TODO: extend with other criteria based on the hotplug interfaces
```
mvo5: This is slightly unrelated to this PR it seem(?). Its ok but I couldn't help noticing. I guess because this goes directly to netlinkConn.Monitor() we cannot add a test that the subsystem matching does really work(?).

overlord/ifacestate/udevmonitor/udevmon.go
```diff
@@ -38,13 +38,15 @@ type Interface interface {
 
 type DeviceAddedFunc func(device *hotplug.HotplugDeviceInfo)
 type DeviceRemovedFunc func(device *hotplug.HotplugDeviceInfo)
+type EnumerationDoneFunc func()
 
 // Monitor monitors kernel uevents making it possible to find USB devices.
```
bboozzoo: BTW. this should probably say `hotpluggable devices`, there are more interfaces supporting hotplug than just USB.

overlord/ifacestate/udevmonitor/udevmon.go
```diff
@@ -88,8 +91,15 @@ func (m *Monitor) Connect() error {
 		return fmt.Errorf("cannot start udev monitor: %s", err)
 	}
 
-	// TODO: consider passing a device filter to reduce noise from irrelevant devices.
-	m.monitorStop = m.netlinkConn.Monitor(m.netlinkEvents, m.netlinkErrors, nil)
+	var filter netlink.Matcher
+	// TODO: extend with other criteria based on the hotplug interfaces
+	filter = &netlink.RuleDefinitions{
```
bboozzoo: Shouldn't this be a part of a separate PR?

overlord/ifacestate/udevmonitor/udevmon_test.go
```diff
@@ -60,6 +61,15 @@ func (s *udevMonitorSuite) TestDiscovery(c *C) {
 		remInfo = inf
 		callbackChannel <- struct{}{}
 	}
+	done := func() {
```
mvo5: (nitpick) would be nice to call this "enumerateDone" - when I looked at this I though we could change the code here to only look at done but this test also simulates hotplugs after the enumeration was done. So a slightly different name would probably have given me the extra context :)

overlord/ifacestate/udevmonitor/udevmon.go
stolowski: Indeed, this should be part of another PR, removed.

overlord/ifacestate/udevmonitor/udevmon.go
stolowski: Right, removed.

PR #6074, opened 2018-10-30
strutil: make VersionCompare faster

chipaca:
Per @zyga's request, this splits the `strutil` changes from #6065. It then also adds to them.

With these changes,

    benchmark                   old ns/op     new ns/op     delta
    BenchmarkVersionCompare     19479781      560826        -97.12%

    benchmark                   old allocs     new allocs     delta
    BenchmarkVersionCompare     86629          0              -100.00%

    benchmark                   old bytes     new bytes     delta
    BenchmarkVersionCompare     2976276       0             -100.00%

(it also fixes a bug in VersionCompare that we hadn't hit...)

This is what the cpu profile looked like before:

old

and this is what it looks like now:

new


strutil/version.go
```diff
@@ -32,7 +32,7 @@ func max(a, b int) int {
 	return a
 }
 
-//go:generate go run github.com/snapcore/snapd/strutil/chrorder -package=strutil -output=chrorder.go
+//go:generate go run ./chrorder/main.go -package=strutil -output=chrorder.go
```
bboozzoo: Debian build does not set GOPATH?

chipaca: go run does not take a package, it takes a path to a file

strutil/chrorder.go
```diff
@@ -0,0 +1,21 @@
+// auto-generated, DO NOT EDIT!
```
zyga: Should we commit this file to the tree? I was expecting not to.

bboozzoo: I think we should. The upside is that the code, as checked out, builds without any extra operations and any changes to the file are directly visible.

strutil/version.go
```diff
@@ -90,29 +44,70 @@ func cmpString(as, bs string) int {
...
 	}
-	res := cmpString(a, b)
-	//fmt.Println(a, b, res)
-	return res
+	return ""
 }
 
-func getFragments(a string) []string {
-	return findFrags(a, -1)
+func cmpNumeric(a, b string) int {
```
stolowski: Can you please add a comment stating that a and b assume numbers only.

strutil/version.go
```diff
@@ -90,29 +44,70 @@ func cmpString(as, bs string) int {
...
+
+func matchEpoch(a string) bool {
+	if len(a) == 0 {
+		return false
+	}
+	if a[0] < '0' || a[0] > '9' {
+		return false
+	}
+	i := 0
+	for i = 1; i < len(a) && a[i] >= '0' && a[i] <= '9'; i++ {
```
stolowski: Nitpick, skip double initialization of `i`:
```
var i
for i = 1 ; ...
```
or
```
i := 1
for ; i < len...
```

strutil/version.go
```diff
@@ -90,29 +44,70 @@ func cmpString(as, bs string) int {
...
 	}
 	return 0
 }
 
-func cmpFragment(a, b string) int {
-	intA, errA := strconv.Atoi(a)
-	intB, errB := strconv.Atoi(b)
-	if errA == nil && errB == nil {
-		return cmpInt(intA, intB)
+func trimLeadingZeroes(a string) string {
```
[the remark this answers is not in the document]
chipaca: unfortunately TrimLeft is rather bad for this use case,
```
$ benchcmp /tmp/strutil_trim  /tmp/strings_trim 
benchmark                     old ns/op     new ns/op     delta
BenchmarkVersionCompare-8     539117        23324071      +4226.35%

benchmark                     old allocs     new allocs     delta
BenchmarkVersionCompare-8     0              12461          +Inf%

benchmark                     old bytes     new bytes     delta
BenchmarkVersionCompare-8     0             398778        +Inf%
```

PR #6078, opened 2018-10-31
ifacestate/ifacemgr: don't reload hotplug-gone connections on startup

stolowski:
Don't reload connections with hotplug-gone=true on startup. This flag means the device was unplugged, but we shouldn't forget about the connection; it may potentially be restored if the device is plugged back. This may happen at a later moment during udev enumeration on startup, independent of this connection reloading.

### DIFF
--- a/overlord/ifacestate/helpers.go
+++ b/overlord/ifacestate/helpers.go
@@ -277,7 +277,7 @@ func (m *InterfaceManager) reloadConnections(snapName string) ([]string, error) 
 	}
 	affected := make(map[string]bool)
 	for id, conn := range conns {
-		if conn.Undesired {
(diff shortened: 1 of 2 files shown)


zyga:
I approved this but I would suggest to rename the attribute from `HotplugGone` to `device unplugged` or something similar. We use the term hot plug like a keyword everywhere, sometimes where it feels unnatural and more direct language may be used.

stolowski:
> I approved this but I would suggest to rename the attribute from `HotplugGone` to `device unplugged` or something similar. We use the term hot plug like a keyword everywhere, sometimes where it feels unnatural and more direct language may be used.

It was named 'HotplugRemoved' before and renamed to `HotplugGone` per Gustavo's suggestion.

PR #6086, opened 2018-11-01
overlord/snapshotstate/backend: survive missing directories

chipaca:
Without this change a missing "common" directory is enough to abort a
snapshot.

I'm not sure how a "common" directory can exist when the revisioned
directory exists (need to chase zyga for that), but it happens. So
this deals with that, and with the reverse situation (where common
exists but no revisined directory does). It also logs a bit more about
what's going on (in DEBUG), to help if things misbehave in the future.

### DIFF
--- a/overlord/snapshotstate/backend/backend.go
+++ b/overlord/snapshotstate/backend/backend.go
@@ -230,15 +230,16 @@ var isTesting = osutil.GetenvBool("SNAPPY_TESTING")
 
 func addDirToZip(ctx context.Context, snapshot *client.Snapshot, w *zip.Writer, username string, entry, dir string) error {
 	parent, revdir := filepath.Split(dir)
-	if exists, isDir, err := osutil.DirExists(parent); !(err == nil && exists && isDir) {
-		if err != nil {
-			return err
-		}
-		if exists {
-			logger.Noticef("Not saving directories under %q in snapshot #%d of %q as it is not a directory.", parent, snapshot.SetID, snapshot.Snap)
-		} else {
-			logger.Debugf("Not saving directories under %q in snapshot #%d of %q as it is does not exist.", parent, snapshot.SetID, snapshot.Snap)
-		}
+	exists, isDir, err := osutil.DirExists(parent)
+	if err != nil {
+		return err
+	}
+	if exists && !isDir {
+		logger.Noticef("Not saving directories under %q in snapshot #%d of %q as it is not a directory.", parent, snapshot.SetID, snapshot.Snap)
+		return nil
+	}
+	if !exists {
+		logger.Debugf("Not saving directories under %q in snapshot #%d of %q as it is does not exist.", parent, snapshot.SetID, snapshot.Snap)
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


overlord/snapshotstate/backend/backend.go
```diff
@@ -229,27 +229,68 @@ func Save(ctx context.Context, id uint64, si *snap.Info, cfg map[string]interfac
...
+	}
+	tarArgs := []string{
+		"--create",
+		"--sparse", "--gzip",
+		"--directory", parent,
+	}
+
+	noRev, noCommon := true, true
+
+	if exists, isDir, err := osutil.DirExists(dir); err == nil && exists && isDir {
```
stolowski: Similiar to earlier comment, would it make sense to get rid of extra `if` level here and simply check err first?

overlord/snapshotstate/backend/backend.go
```diff
@@ -229,27 +229,68 @@ func Save(ctx context.Context, id uint64, si *snap.Info, cfg map[string]interfac
...
+		}
+		if exists {
 			logger.Noticef("Not saving %q in snapshot #%d of %q as it is not a directory.", dir, snapshot.SetID, snapshot.Snap)
+		} else {
+			logger.Debugf("Not saving %q in snapshot #%d of %q as it is does not exist.", dir, snapshot.SetID, snapshot.Snap)
+		}
+	}
+
+	common := filepath.Join(parent, "common")
+	if exists, isDir, err := osutil.DirExists(common); err == nil && exists && isDir {
```
stolowski: Same.

overlord/snapshotstate/backend/backend.go
```diff
@@ -229,27 +229,68 @@ func Save(ctx context.Context, id uint64, si *snap.Info, cfg map[string]interfac
...
+
+	if exists, isDir, err := osutil.DirExists(dir); err == nil && exists && isDir {
+		tarArgs = append(tarArgs, revdir)
+		noRev = false
+	} else {
+		if err != nil {
+			return err
+		}
+		if exists {
 			logger.Noticef("Not saving %q in snapshot #%d of %q as it is not a directory.", dir, snapshot.SetID, snapshot.Snap)
```
stolowski: Same remark regarding `if exists` check and the message here.

overlord/snapshotstate/backend/backend.go
```diff
@@ -229,27 +229,68 @@ func Save(ctx context.Context, id uint64, si *snap.Info, cfg map[string]interfac
...
+	common := filepath.Join(parent, "common")
+	if exists, isDir, err := osutil.DirExists(common); err == nil && exists && isDir {
+		tarArgs = append(tarArgs, "common")
+		noCommon = false
+	} else {
+		if err != nil {
+			return err
+		}
+		if exists {
+			logger.Noticef("Not saving %q in snapshot #%d of %q as it is not a directory.", common, snapshot.SetID, snapshot.Snap)
```
stolowski: Same as earlier comment.

PR #6095, opened 2018-11-05
packaging/opensuse: stop using golang-packaging

zyga:
This patch switches the openSUSE packaging to stop using go-packaging.
The state of golang in openSUSE is not perfect. The existing helpers are
very cumbersome, verbose and don't provide useful features. In cases
like passing special linker flags the helpers require us to work around
them.

I've been in contact with Go maintainers in openSUSE and they recommend
to just use go directly. The only real value of the packaging helpers as
they exist today is to pass "-buildmode pie". In this patch I did just
that, switching to pure go for building.

The patch refers to "indigo" that I'm also working on, as a better and
less weird packaging helper for distributions. I used some the same
directory structure during the build process that Indigo uses directly
and I plan to switch to Indigo 0.3 or 0.4, as time permits. This will
simply replace "GOPATH=... go build -buildmode pie" with "indigo build"
and remove some of the setup boilerplate.

### DIFF
--- a/packaging/opensuse/snapd.spec
+++ b/packaging/opensuse/snapd.spec
@@ -209,8 +209,8 @@ popd
 GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=pie %{import_path}/cmd/snap
 GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=pie %{import_path}/cmd/snapctl
 GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=pie %{import_path}/cmd/snap-seccomp
-GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=default --ldflags '-extldflags "-static"' %{import_path}/cmd/snap-update-ns
-GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=default --ldflags '-extldflags "-static"' %{import_path}/cmd/snap-exec
+GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=default -ldflags '-extldflags "-static"' %{import_path}/cmd/snap-update-ns
+GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=default -ldflags '-extldflags "-static"' %{import_path}/cmd/snap-exec
 %if 0%{?with_test_keys}
 GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=pie -tags withtestkeys %{import_path}/cmd/snapd
 %else
--- a/packaging/opensuse/snapd.spec
+++ b/packaging/opensuse/snapd.spec
@@ -204,17 +204,18 @@ popd
 # Build golang executables, with the following exceptions everything is built the same way:
 # - snap-exec and snap-update-ns is built statically.
 # - snapd has a variant that uses test keys instead of production keys.
-
-# build snap-exec and snap-update-ns completely static for base snaps
-GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=pie %{import_path}/cmd/snap
-GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=pie %{import_path}/cmd/snapctl
-GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=pie %{import_path}/cmd/snap-seccomp
-GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=default -ldflags '-extldflags "-static"' %{import_path}/cmd/snap-update-ns
-GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=default -ldflags '-extldflags "-static"' %{import_path}/cmd/snap-exec
+#
+# NOTE: indigo_gopath takes priority over GOPATH. This ensures that we
+# build the code that we intended in case GOPATH points to another copy.
+GOPATH=%{indigo_gopath}:$GOPATH go build -buildmode=pie %{import_path}/cmd/snap
+GOPATH=%{indigo_gopath}:$GOPATH go build -buildmode=pie %{import_path}/cmd/snapctl
+GOPATH=%{indigo_gopath}:$GOPATH go build -buildmode=pie %{import_path}/cmd/snap-seccomp
+GOPATH=%{indigo_gopath}:$GOPATH go build -buildmode=default -ldflags '-extldflags "-static"' %{import_path}/cmd/snap-update-ns
+GOPATH=%{indigo_gopath}:$GOPATH go build -buildmode=default -ldflags '-extldflags "-static"' %{import_path}/cmd/snap-exec
 %if 0%{?with_test_keys}
-GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=pie -tags withtestkeys %{import_path}/cmd/snapd
+GOPATH=%{indigo_gopath}:$GOPATH go build -buildmode=pie -tags withtestkeys %{import_path}/cmd/snapd
 %else
-GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=pie %{import_path}/cmd/snapd
+GOPATH=%{indigo_gopath}:$GOPATH go build -buildmode=pie %{import_path}/cmd/snapd
 %endif
 
 # Build C executables
@@ -222,7 +223,7 @@ GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=pie %{import_path}/cmd/snapd
 
 %check
 # Run tests with fixed locale that is expected by unicode/color code.
-LC_ALL=C.UTF-8 GOPATH=$GOPATH:%{indigo_gopath} go test %{import_path}/...
+LC_ALL=C.UTF-8 GOPATH=%{indigo_gopath}:$GOPATH go test %{import_path}/...
 %make_build -C %{indigo_srcdir}/cmd check
 
 %install
--- a/packaging/opensuse/snapd.spec
+++ b/packaging/opensuse/snapd.spec
@@ -17,6 +17,8 @@
 %bcond_with testkeys
 
 # Enable AppArmor on openSUSE Tumbleweed (post 15.0) or higher
+# N.B.: Prior to openSUSE Tumbleweed in May 2018, the AppArmor userspace in SUSE
+# did not support what we needed to be able to turn on basic integration.
 %if 0%{?suse_version} >= 1550
 %bcond_without apparmor
 %else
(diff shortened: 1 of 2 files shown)


packaging/opensuse/snapd.spec
```diff
@@ -143,18 +150,36 @@ updated as soon as new upstream releases are made and is designed to live in
...
+
+# Patch the source in the place it got extracted to.
+pushd %{indigo_srcdir}
+# Add patch0 -p1 ... as appropriate here.
+popd
+
+# Patch snap-seccomp to use dynamic linking. This is not a problem because
+# snap-seccomp only requires static linking when it runs from the core-snap via
+# re-exec.
+sed -e "s/-Bstatic -lseccomp/-Bstatic/g" -i %{indigo_srcdir}/cmd/snap-seccomp/main.go
```
bboozzoo: This is no longer needed. I think we'll need to update the other packages too.

packaging/opensuse/snapd.spec
```diff
@@ -168,83 +193,66 @@ export CFLAGS
...
-%goprep %{import_path}
+# Build golang executables, with the following exceptions everything is built the same way:
+# - snap-exec and snap-update-ns is built statically.
+# - snapd has a variant that uses test keys instead of production keys.
 
+# build snap-exec and snap-update-ns completely static for base snaps
+GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=pie %{import_path}/cmd/snap
+GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=pie %{import_path}/cmd/snapctl
+GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=pie %{import_path}/cmd/snap-seccomp
+GOPATH=$GOPATH:%{indigo_gopath} go build -buildmode=default --ldflags '-extldflags "-static"' %{import_path}/cmd/snap-update-ns
```
bboozzoo: `-ldflags`, single `-`

packaging/opensuse/snapd.spec
```diff
@@ -27,22 +29,35 @@
 %{!?make_build: %global make_build %{__make} %{?_smp_mflags}}
```
bboozzoo: While at it, you can switch this to `%{?!..}`

packaging/opensuse/snapd.spec
```diff
@@ -91,6 +106,9 @@ BuildRequires:  pkg-config
 BuildRequires:  python-docutils
 BuildRequires:  python3-docutils
 BuildRequires:  squashfs
+# Due to: rpm -q --whatprovides /usr/share/pkgconfig/systemd.pc
```
bboozzoo: Shouldn't that be `/usr/lib/pkgconfig..`?

packaging/opensuse/snapd.spec
```diff
@@ -143,18 +156,37 @@ updated as soon as new upstream releases are made and is designed to live in
...
+# means that while go doesn't care at all the current working directory is not
+# the top-level directory of the source tarball which some people may find
+# unusual.
+
+# Create indigo compatible build layout.
+mkdir -p %{indigo_srcdir}
+tar -axf %{_sourcedir}/%{name}_%{version}.vendor.tar.xz --strip-components=1 -C %{indigo_srcdir}
+
+# Patch the source in the place it got extracted to.
+pushd %{indigo_srcdir}
```
bboozzoo: This is fine and in sync with repo packaging https://build.opensuse.org/package/view_file/system:snappy/snapd/snapd.spec?expand=1

packaging/opensuse/snapd.spec
```diff
@@ -143,18 +156,37 @@ updated as soon as new upstream releases are made and is designed to live in
...
+# Set the version that is compiled into the various executables/
+pushd %{indigo_srcdir}
 ./mkversion.sh %{version}-%{release}
+popd
 
-# Generate autotools build system files
-cd cmd && autoreconf -i -f
+# Sanity check, ensure that systemd system generator directory is in agreement between the build system and packaging.
+if [ "$(pkg-config --variable=systemdsystemgeneratordir systemd)" != "%{_systemdgeneratordir}" ]; then
+  echo "pkg-confing and rpm macros disagree about the location of systemd system generator directory"
```
bboozzoo: Heh, does this really happen?

packaging/opensuse/snapd.spec
```diff
@@ -168,83 +200,67 @@ export CFLAGS
 export CXXFLAGS
 export LDFLAGS
 
-# N.B.: Prior to openSUSE Tumbleweed in May 2018, the AppArmor userspace in SUSE
-# did not support what we needed to be able to turn on basic integration.
+# Generate autotools build system files.
+pushd %{indigo_srcdir}/cmd
+autoreconf -i -f
 %configure \
     %{!?with_apparmor:--disable-apparmor} \
```
bboozzoo: Wondering whether it'd be helpful to have an explicit `--enable-apparmor` for TW.

packaging/opensuse/snapd.spec
```diff
@@ -168,83 +200,67 @@ export CFLAGS
...
+# Build golang executables, with the following exceptions everything is built the same way:
+# - snap-exec and snap-update-ns is built statically.
+# - snapd has a variant that uses test keys instead of production keys.
+#
+# NOTE: indigo_gopath takes priority over GOPATH. This ensures that we
+# build the code that we intended in case GOPATH points to another copy.
+GOPATH=%{indigo_gopath}:$GOPATH go build -buildmode=pie %{import_path}/cmd/snap
+GOPATH=%{indigo_gopath}:$GOPATH go build -buildmode=pie %{import_path}/cmd/snapctl
+GOPATH=%{indigo_gopath}:$GOPATH go build -buildmode=pie %{import_path}/cmd/snap-seccomp
+GOPATH=%{indigo_gopath}:$GOPATH go build -buildmode=default -ldflags '-extldflags "-static"' %{import_path}/cmd/snap-update-ns
```
bboozzoo: I think this should include distro `%_ldflags` in `-extldflags` (probably for all the binaries we build). I know Fedora is quite strict about that, not sure about openSUSE though.

packaging/opensuse/snapd.spec
```diff
@@ -168,83 +200,67 @@ export CFLAGS
...
-mv %{buildroot}%{_bindir}/snapd %{buildroot}%{_libexecdir}/snapd/snapd
-mv %{buildroot}%{_bindir}/snap-exec %{buildroot}%{_libexecdir}/snapd/snap-exec
-mv %{buildroot}%{_bindir}/snap-update-ns %{buildroot}%{_libexecdir}/snapd/snap-update-ns
-mv %{buildroot}%{_bindir}/snap-seccomp %{buildroot}%{_libexecdir}/snapd/snap-seccomp
+# Install all systemd and dbus units, and env files.
+%make_install -C %{indigo_srcdir}/data \
+		BINDIR=%{_bindir} \
+		LIBEXECDIR=%{_libexecdir} \
+		SYSTEMDSYSTEMUNITDIR=%{_unitdir} \
+		SNAP_MOUNT_DIR=%{snap_mount_dir}
```
bboozzoo: I think you could also set `SNAPD_ENVIRONMENT_FILE=%{_sysconfdir}/default/snapd`, provided openSUSE uses `/etc/default`, otherwise it ends up using `/etc/environment` which is a bit annoying.

packaging/opensuse/snapd.spec
zyga: That's what is actually on the system and that is what pkg-config picks up. Why did you expect it to be in /usr/lib?

packaging/opensuse/snapd.spec
zyga: Well, that's the reason I spent some much time on this!

Our default in configure.ac has some value that happens to be used if systemd.pc is not present. It is a silent bug that just happens. To ensure that we don't miss this the next time I added this check.

packaging/opensuse/snapd.spec
zyga: Yeah, I can do that.

packaging/opensuse/snapd.spec
zyga: `rpm -E '%{_ldflags}'` doesn't expand here, perhaps it exits just for fedora?

packaging/opensuse/snapd.spec
zyga: +1 but I'd like to make that as a separate improvement. This PR is what shipped last night.

zyga:
NOTE: The tests are failing due to overeager check for usage of $GOPATH. I will fix that and merge this.

packaging/opensuse/snapd.spec
zyga: Yeah, I plan to make those consistent but in a separate cleanup PR.

packaging/opensuse/snapd.spec
bboozzoo: I've always seen ${libdir}/pkgconfig, but I see ${datadir}/pkgconfig used too. On my system:

```
$ ls /usr/lib/pkgconfig | wc -l
1013
$ ls /usr/share/pkgconfig | wc -l
56
```
Another one with almost vanilla Fedora cloud image:
```
$ ls /usr/lib64/pkgconfig/ | wc -l
2
$ ls /usr/share/pkgconfig/ | wc -l
4
```

Looks like it's yet another thing that may vary across distros.

zyga: I can remove the comment but the point is that we needed that .pc file.

packaging/opensuse/snapd.spec
bboozzoo: Fedora has `redhat-rpm-config` which sets that, but openSUSE does not seem to have anything similar. Wonder whether they actually use any flags enforced by distro policy, maybe not.

packaging/opensuse/snapd.spec
bboozzoo: Feels like this should be a bug against systemd packaging or `systemd-rpm-macros`

zyga: The bug is on our side, we should not guesstimate any directories.

packaging/opensuse/snapd.spec
zyga: Actually, I kept this at --disable-apparmor because we may have headers but the system still doesn't "do" apparmor the way we want.

packaging/opensuse/snapd.spec
zyga: They must have something but it applies to C code via a mechanism I'm not familiar with. When doing initial review of the package submission to factory (long time ago) we got flagged on this and fixed it.

PR #6096, opened 2018-11-06
spread.yaml: add more systems to the autopkgtest and qemu backends

mvo5:
The autopkgtests in disco fail right now because there is no
autopkgtest backend with this version defined in spread.yaml.

To fix this this PR adds disco and the versions up to 20.04
for qemu/autopkgtest into spread.yaml.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -175,6 +175,24 @@ backends:
             - ubuntu-18.10-32:
                 username: ubuntu
                 password: ubuntu
+            - ubuntu-19.04-64:
+                username: ubuntu
+                password: ubuntu
+            - ubuntu-19.04-32:
+                username: ubuntu
+                password: ubuntu
+            - ubuntu-19.10-64:
+                username: ubuntu
+                password: ubuntu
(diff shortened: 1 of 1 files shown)


spread.yaml
```diff
@@ -280,6 +298,64 @@ backends:
             - ubuntu-18.10-arm64:
                 username: ubuntu
                 password: ubuntu
+            # Disco
```
bboozzoo: Should probably read: `# Disco Dingo`

mvo5: So far all of the entries only have the codename, so this is probably fine

spread.yaml
```diff
@@ -175,6 +175,24 @@ backends:
...
+            - ubuntu-19.10-64:
+                username: ubuntu
+                password: ubuntu
+            - ubuntu-19.10-32:
+                username: ubuntu
+                password: ubuntu
+            - ubuntu-20.04-64:
+                username: ubuntu
+                password: ubuntu
+            - ubuntu-20.04-32:
```
zyga: LOL :D

PR #6097, opened 2018-11-06
interfaces/tests: MockHotplugSlot test helper

stolowski:
Introduce a simple test helper for creating hotplug slots in tests of concrete interfaces (followup PR with camera interfaces changes will have example of this).

### DIFF
--- a/interfaces/builtin/utils_test.go
+++ b/interfaces/builtin/utils_test.go
@@ -89,3 +89,16 @@ func MockConnectedSlot(c *C, yaml string, si *snap.SideInfo, slotName string) (*
 	}
 	panic(fmt.Sprintf("cannot find slot %q in snap %q", slotName, info.InstanceName()))
 }
+
+func MockHotplugSlot(c *C, yaml string, si *snap.SideInfo, hotplugKey, ifaceName, slotName string, staticAttrs map[string]interface{}) *snap.SlotInfo {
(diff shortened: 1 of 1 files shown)


stolowski:
@mvo See https://github.com/snapcore/snapd/pull/6098/ for example test using this. Thanks!

interfaces/builtin/utils_test.go
```diff
@@ -89,3 +89,16 @@ func MockConnectedSlot(c *C, yaml string, si *snap.SideInfo, slotName string) (*
 	}
 	panic(fmt.Sprintf("cannot find slot %q in snap %q", slotName, info.InstanceName()))
 }
+
+func MockHotplugSlot(c *C, yaml string, si *snap.SideInfo, hotplugKey, ifaceName, slotName string, staticAttrs map[string]interface{}) *snap.SlotInfo {
+	info := snaptest.MockInfo(c, yaml, si)
```
zyga: Could this become:
```
slotInfo := MockSlot(c, yams, si, slotName)
slotInfo.Attrs = staticAttrs
slotInfo.HotplugKey = hotplugKey
return slotInfo
```

stolowski: No, because `MockSlot` requires the slot to be decleared in the yaml, this is exactly what I don't want here.

PR #6100, opened 2018-11-06
overlord/ifacestate: hotplug-remove-slot task handler

stolowski:
Implementation of "hotplug-remove-slot" task handler. This is strictly related to https://github.com/snapcore/snapd/pull/6069 which creates these tasks.

The hotplug-remove-slot is created as part of device removal handling on unplug and it follows hotplug-disconnect task.
This task performs two operations:
- removes given hotplug slot from the interface repository.
- removes the slot from hotplug-slots map in the state as long as there are no connections to this slot, including connections marked with _hotplug-gone=true_. In other words, we only completely forget about a slot if there is no need to restore any connections of its associated device in the future.


overlord/ifacestate/handlers.go
```diff
@@ -1218,6 +1218,58 @@ func (m *InterfaceManager) doGadgetConnect(task *state.Task, _ *tomb.Tomb) error
...
+	defer st.Unlock()
+
+	ifaceName, hotplugKey, err := getHotplugAttrs(task)
+	if err != nil {
+		return fmt.Errorf("internal error: cannot get hotplug task attributes: %s", err)
+	}
+
+	slot, err := m.repo.SlotForHotplugKey(ifaceName, hotplugKey)
+	if err != nil {
+		return fmt.Errorf("cannot determine slots: %s", err)
```
bboozzoo: This should probably read `internal error: cannot ...`

overlord/ifacestate/handlers.go
```diff
@@ -1218,6 +1218,58 @@ func (m *InterfaceManager) doGadgetConnect(task *state.Task, _ *tomb.Tomb) error
...
+	}
+	if slot != nil {
+		if err := m.repo.RemoveSlot(slot.Snap.InstanceName(), slot.Name); err != nil {
+			return fmt.Errorf("cannot remove slot %s of snap %q: %s", slot.Snap.InstanceName(), slot.Name, err)
+		}
+	}
+
+	stateSlots, err := getHotplugSlots(st)
+	if err != nil {
+		return fmt.Errorf("internal error obtaining hotplug slots: %v", err.Error())
```
bboozzoo:
```
return fmt.Errorf("internal error: cannot obtain hotplug slots: %v", err)
```

overlord/ifacestate/handlers.go
```diff
@@ -1218,6 +1218,58 @@ func (m *InterfaceManager) doGadgetConnect(task *state.Task, _ *tomb.Tomb) error
...
+		return fmt.Errorf("internal error: cannot get hotplug task attributes: %s", err)
+	}
+
+	slot, err := m.repo.SlotForHotplugKey(ifaceName, hotplugKey)
+	if err != nil {
+		return fmt.Errorf("cannot determine slots: %s", err)
+	}
+	if slot != nil {
+		if err := m.repo.RemoveSlot(slot.Snap.InstanceName(), slot.Name); err != nil {
+			return fmt.Errorf("cannot remove slot %s of snap %q: %s", slot.Snap.InstanceName(), slot.Name, err)
```
bboozzoo: `RemoveSlot` already builds a meaningful the error message, I think you can just `return err` or `return fmt.Error("internal error: cannot remove hotplug slot: %v", err)` here.

PR #6101, opened 2018-11-06
switch travis unit tests to xenial

chipaca:


.travis.yml
```diff
@@ -4,7 +4,8 @@ git:
 jobs:
   include:
     - stage: quick
-      name: go 1.6/trusty static and unit test suites
+      name: go 1.6/xenial static and unit test suites
```
bboozzoo: BTW. woulld it make sense if we had a unit tests job with go tip too?

chipaca:
This was done because xenial travis images are going live tomorrow, and we were asked to see if they worked for us (they did).
I also thought xenial might be that little bit faster, but that didn't turn out to be true.

PR #6102, opened 2018-11-06
overlord/snapshots: survive an unknown user

chipaca:
Without this change, a system with a directory /home/foo/snap with a
uid not in /etc/passwd will fail to snapshot, with an error such as

```
- Save data of snap "some-snap" in snapshot set #1 (user: unknown userid 1234)
```

This change instead ignores that data.

### DIFF
--- a/overlord/snapshotstate/backend/helpers.go
+++ b/overlord/snapshotstate/backend/helpers.go
@@ -94,6 +94,15 @@ var (
 	userLookupId = user.LookupId
 )
 
+func isUnknownUser(err error) bool {
+	switch err.(type) {
+	case user.UnknownUserError, user.UnknownUserIdError:
+		return true
+	default:
+		return false
+	}
+}
+
 func usersForUsernames(usernames []string) ([]*user.User, error) {
 	if len(usernames) == 0 {
 		return allUsers()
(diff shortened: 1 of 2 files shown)


tests/regression/lp-1801955/task.yaml
```diff
@@ -0,0 +1,13 @@
+summary: Check that snapshot works with unknown users in /home/*/snap
+
+prepare: |
+    snap install test-snapd-tools
+    ! grep :9999: /etc/passwd
+    mkdir -pv /home/potato/snap/
+    chown -vR 9999:9999 /home/potato
```
stolowski: Nice test.

overlord/snapshotstate/backend/helpers.go
```diff
@@ -102,7 +111,7 @@ func usersForUsernames(usernames []string) ([]*user.User, error) {
 	for _, username := range usernames {
 		usr, err := userLookup(username)
 		if err != nil {
-			if _, ok := err.(*user.UnknownUserError); !ok {
+			if !isUnknownUser(err) {
 				return nil, err
 			}
 			u, e := userLookupId(username)
```
stolowski: Actually... isn't this lookup going to fail as well?

chipaca: if the user said `snap save --users=1000`, then no :-)

stolowski: ah, all clear now! thanks!

PR #6103, opened 2018-11-06
tests: split nested vm suite on core and classic and new snapshots test

sergiocazzolato:
This is split because in the future the setups will be different in case
the nested vm is core or classic.

Also it is included a new test for snapshots which tests in case the
core is refreshed and reverted. Until the snapshots feature is in stable
we need to run from beta to edge.

To run it do: SPREAD_CORE_CHANNEL=beta SPREAD_CORE_REFRESH_CHANNEL=edge
spread
google-nested:ubuntu-16.04-64:tests/nested/classic/snapshots-with-core-refresh-revert

Those tests will be executed on each new core on edge soon as part of a
spread cron job.


spread.yaml
```diff
@@ -705,7 +705,29 @@ suites:
...
+            . "$TESTSLIB"/pkgdb.sh
+            distro_install_package snapd qemu genisoimage sshpass qemu-kvm cloud-image-utils ubuntu-image
+        prepare-each: |
+            "$TESTSLIB"/prepare-restore.sh --prepare-suite-each
+        restore: |
+            "$TESTSLIB"/prepare-restore.sh --restore-suite
+
+    tests/nested/core/:
+        summary: Tests for nested images
+        # Test cases are not yet ported to Fedora/openSUSE that is why
```
mvo5: I'm not sure I understand this comment, the suite is called `nested/core` why would we port this to Fedora/openSUSE? Copy/paste error perhaps?

spread.yaml
```diff
@@ -696,7 +696,7 @@ suites:
         restore: |
             "$TESTSLIB"/prepare-restore.sh --restore-suite
 
-    tests/nested/:
+    tests/nested/classic/:
         summary: Tests for nested images
         # Test cases are not yet ported to Fedora/openSUSE that is why
         # we keep them disabled. A later PR will enable most tests and
```
mvo5: I'm not sure I get the comment fully - it says "A lter PR will enable most tests and drop this blacklist"  - AFAICT there is no blacklist just a different backend. Is the comment maybe outdated?

tests/nested/classic/snapshots-with-core-refresh-revert/task.yaml
```diff
@@ -0,0 +1,62 @@
+summary: test snapshots work when core snap is refreshed and reverted
```
mvo5: Silly question (pardon my ignorance) - whats the reason behind testing this in the nested VM? Could we do this also in the standard tests? It seems we install the snapd deb so couldn't we do the same on a regular spread run?

sergiocazzolato: Thanks for reviewing this, in fact for this test it is not really needed to build the snapd.deb, I just did it because in a next PR I am building snapd during the suite praparation and it will be copied to the nested vm by default. The idea of this test is to validate we can refresh/revert core from edge to stable and snapshots are working. This test is gonna be executed when a new core in edge is released, and not for every change. 
As we refresh/revert the core snap it does not make sense to run it as part of the main suite where we test the snapd.deb  

Any suggestion to change the way I am trying to test that is wellcome

sergiocazzolato:
closing thie Pr due to it has been merged with the main one dor nested tests execution

PR #6110, opened 2018-11-08
spread.yaml: add openSUSE Tumbleweed

zyga:
openSUSE Tumblweed is like Arch, a rolling snapshot release
distribution. Unlike the current Leap release it supports enough
apparmor that we enable it there in a non-devmode manner. For this
reason we should run tests against it as an early warning system of
upcoming changes to Leap itself.


mvo5:
One test failure: `    - google:opensuse-tumbleweed-64:tests/main/create-key` looks real

zyga:
@mvo5 yes, the test failure is real. Tumbleweed has gpg version `2.2.10` so perhaps something new on the horizon waiting to break.

zyga:
Shall we disable this test and file a high severity bug about possible GPG issues to investigate, keep this PR open until we can fix it or close it for now and schedule something for later @pedronis

bboozzoo:
Interesting, I see that GPG 2.2.10 is also in Arch, wonder why the tests don't fail there.

zyga:
This is oddly failing, consistently, on one test:
```
+ echo 'Modify state to make it look like the last refresh just happened'
Modify state to make it look like the last refresh just happened
++ date +%Y-%m-%dT%H:%M:%S%:z
+ jq '.data["last-refresh"] = "2018-11-16T11:54:56+00:00"' /var/lib/snapd/state.json
jq: execute.c:1071: jq_compile_args: Assertion `jv_get_kind(args) == JV_KIND_ARRAY' failed.
/home/gopath/src/github.com/snapcore/snapd/tests/lib/prepare.sh: line 43:  7716 Aborted                 (core dumped) jq ".data[\"last-refresh\"] = \"$(date +%Y-%m-%dT%H:%M:%S%:z)\"" /var/lib/snapd/state.json > /var/lib/snapd/state.json.new
```

bboozzoo:
`jq` is still segfaulting. Any chance it's a bug in `jq` and the package got updated already but the image has not?

zyga:
When testing locally this was also failing on snap-confine permissions for mapping runtime libraries. I think the test may show it is genuinely broken. Now that snap-confine fix is merged it _might_ pass, let's see.

zyga:
Let's return to this later.

PR #4880, opened 2018-03-20
[RFC] cmd, data: plain make 

bboozzoo:
Followup on #4571. Experimenting with replacing autotools cruft with plain makefiles.

In order to not require passing all flags in each make invocation, there is a `configure` target that writes the configuration into a shared `build-config` file. That file is later included in the remaining makefiles.

The new thing, autogen.sh && make takes ~2.5s on my build machine. With `-j17`, it's ~400ms.

The autotools variant of autogen.sh&& make is ~12s, `-j17` yields ~6s.

PR #6111, opened 2018-11-08
packaging/opensuse: move most logic to snapd.mk

zyga:
Packaging in spec files, debian rules, arch and all kinds of other
formats is nice when you can essentially push the bulk of the work to
"make install" or similar. In our project we don't yet have that luxury
because we sport two separate makefile systems (one autotools, one
handmade, using different libexecdir in each time for extra
eyebrow-raising) some go executables that are built with non-default
options, a small set of files that are copied manually around and two
major build modes (core vs classic-except-ubuntu) that differ in the set
of files we need to ship.

To start shaving they Yak I propose that we move most of the packaging
muscle in openSUSE's snapd.spec to snapd.mk.

Where the spec file is ugly is around conditional handling. I was unable
to find a way to have RPM handle bot the %if expr conditional as well as
the %{?foo} and %{!?foo} conditionals sanely. Fortunately the makefile
world is much saner and this is where most of the logic now lives.

The spec file builds correctly on openSUSE Leap 42.3, 15.0 and
Tumbleweed. Spread testing against Leap 42.3 suggests that I didn't
break anything either.

The spec file is much more precise now, claiming ownership of many more
files and directories belonging to snapd that are created dynamically.

I tried to document the makefile and have belts-and-suspenders approach
so that nothing (or as little as I could) would fail silently. I haven't
yet tried using it on Debian. I would attempt to use it for Arch next,
where it should be easiest, and then on Fedora and Amazon Linux that
share packaging.


packaging/snapd.mk
```diff
@@ -0,0 +1,190 @@
...
+prefix:=$(shell rpm -E %{_prefix})
+bindir:=$(shell rpm -E %{_bindir})
+sbindir:=$(shell rpm -E %{_sbindir})
+libexecdir:=$(shell rpm -E %{_libexecdir})
+mandir:=$(shell rpm -E %{_mandir})
+datadir:=$(shell rpm -E %{_datadir})
+localstatedir:=$(shell rpm -E %{_localstatedir})
+unitdir=$(shell rpm -E %{_unitdir})
+else
+$(error expected RPM to be used)
```
zyga: Overeager debugging :)

packaging/opensuse/snapd.spec
```diff
@@ -14,17 +14,28 @@
...
+%endif
 
 # Enable AppArmor on openSUSE Tumbleweed (post 15.0) or higher
 # N.B.: Prior to openSUSE Tumbleweed in May 2018, the AppArmor userspace in SUSE
 # did not support what we needed to be able to turn on basic integration.
 %if 0%{?suse_version} >= 1550
-%bcond_without apparmor
-%else
-%bcond_with apparmor
+%global with_apparmor 1
```
Conan-Kudo: You're breaking bcond flags by doing this... Also, bconds already emit 1 or 0.

Cf. https://github.com/rpm-software-management/rpm/blob/rpm-4.14.x/macros.in#L101-L102

packaging/opensuse/snapd.spec
```diff
@@ -35,12 +46,6 @@
 %{?!_systemd_system_env_generator_dir: %global _systemd_system_env_generator_dir %{_prefix}/lib/systemd/system-environment-generators}
 %{?!_systemd_user_env_generator_dir: %global _systemd_user_env_generator_dir %{_prefix}/lib/systemd/user-environment-generators}
 
-# This is fixed in SUSE Linux 15
-# Cf. https://build.opensuse.org/package/rdiff/Base:System/rpm?linkrev=base&rev=396
-%if 0%{?suse_version} < 1500
-%global _sharedstatedir %{_localstatedir}/lib
-%endif
```
Conan-Kudo: Removing this will break things, unless you intend to drop everything older than Leap 15.0.

zyga: It works because I stopped using sharedstatedir. I did this because it was defined differently in RPM (see what the makefile does) and in the spec file (because of the override).

In the end, this doesn't bring us any value, killing it makes the code simpler.

packaging/opensuse/snapd.spec
zyga: Can you elaborate please? How should this look like?

packaging/opensuse/snapd.spec
Conan-Kudo: It _does_ buy me something, the ability to compare the file lists between the Fedora and openSUSE packages.

And it isn't defined differently anymore, since I fixed that in openSUSE, only in older ones.

packaging/snapd.mk
```diff
@@ -0,0 +1,190 @@
...
+# The list of go binaries we are expected to build.
+go_binaries=snap snapctl snap-seccomp snap-update-ns snap-exec snapd
+
+# NOTE: This *depends* on building out of tree. Some of the built binaries
+# conflict with directory names in the tree.
+.PHONY: all
+all: $(go_binaries) 
+
+snap snapctl snap-seccomp:
+	go build -buildmode=pie $(import_path)/cmd/$@
```
Conan-Kudo: No way to set Go build flags? Because this means that snapd debuginfo subpackages can't be generated properly.

packaging/opensuse/snapd.spec
zyga: It is defined differently because if you ask RPM alone it won't now anything that is defined in the spec. I did this for a reason. This allows us to save considerable copy pasted code across packaging.

You can also use localstatedir in Fedora, sharedstatedir doesn't buy us anything. It's not like we either change it or support values other than /var/lib.

packaging/snapd.mk
zyga: I will try to include Indigo in the next iteration. What are the flags you need to pass on Fedora to get the right stars aligned?

I specifically coded this so that all of the flag handling and policy application is done via indigo.

Conan-Kudo: The flags can be seen here: https://src.fedoraproject.org/rpms/go-compilers/blob/master/f/macros.golang-compiler

packaging/opensuse/snapd.spec
```diff
@@ -203,40 +199,41 @@ export LDFLAGS
...
+	GOPATH=%{indigo_gopath}:$GOPATH \
+	all
 
 %check
-# Run tests with fixed locale that is expected by unicode/color code.
-LC_ALL=C.UTF-8 GOPATH=%{indigo_gopath}:$GOPATH go test %{import_path}/...
 %make_build -C %{indigo_srcdir}/cmd check
+# Use the common packaging helper for testing.
+%make_build -f %{indigo_srcdir}/packaging/snapd.mk \
+	with_core_bits=0 \
```
bboozzoo: With #4880 there was no need to repeat this all the time.

packaging/snapd.mk
bboozzoo: CC can be exported but you need to way to set `-ldflags "<go-ldflags> -extldflags '$(LDFLAGS)'"` and avoid shell exploding due to quoting.

packaging/snapd.mk
```diff
@@ -0,0 +1,190 @@
...
+# nearly-arbitrary mount namespace that does not contain anything we can depend
+# on (no standard library, for example).
+snap-update-ns snap-exec:
+	go build -buildmode=default -ldflags '-extldflags "-static"' $(import_path)/cmd/$@
+
+# Snapd can be built with test keys. This is only used by the internal test
+# suite to add test assertions. Do not enable this in distribution packages.
+snapd:
+ifeq ($(with_test_keys),1)
+	go build -buildmode=pie -tags withtestkeys $(import_path)/cmd/$@
```
bboozzoo: I think that `-buildmode=` should be controlled for binaries other than `snap-update-ns` and `snap-exec`. IIRC Yocto built with `-buildmode=shared`.

packaging/snapd.mk
```diff
@@ -0,0 +1,190 @@
...
+all: $(go_binaries) 
+
+snap snapctl snap-seccomp:
+	go build -buildmode=pie $(import_path)/cmd/$@
+
+# Those two need to be built as static binaries. They run on the inside of a
+# nearly-arbitrary mount namespace that does not contain anything we can depend
+# on (no standard library, for example).
+snap-update-ns snap-exec:
+	go build -buildmode=default -ldflags '-extldflags "-static"' $(import_path)/cmd/$@
```
bboozzoo: At some point we'll also need to add `-tags osusergo` to use pure go implementation of `os/user` to avoid some `getpwent()` calls. I also have some changes for `osutil` that do the same trick.

packaging/snapd.mk
```diff
@@ -0,0 +1,190 @@
...
+install:: | $(DESTDIR)$(bindir)
+	ln -s $(libexecdir)/snapd/snapctl $|/snapctl
+
+# Generate and install man page for snap command
+install:: snap | $(DESTDIR)$(mandir)/man8
+	./snap help --man > $|/snap.8
+
+# Install the directory structure in /var/lib/snapd
+install::
+	install -d $(DESTDIR)$(localstatedir)/lib/snapd/apparmor/profiles
```
bboozzoo: Might as well set some default permissions for the directories.

packaging/snapd.mk
```diff
@@ -0,0 +1,190 @@
...
+install::
+	rm -f $(DESTDIR)$(unitdir)/snapd.apparmor.service
+	rm -f $(DESTDIR)$(libexecdir)/snapd/snapd-apparmor
+endif
+
+# Tests use C.UTF-8 because some some code depend on this for fancy Unicode
+# output that unit tests do not mock.
+.PHONY: check
+check:
+	LC_ALL=C.UTF-8 go test $(import_path)/...
```
bboozzoo: No C.UTF-8 on Arch. Maybe it's best not to set anything here.

packaging/opensuse/snapd.spec
zyga: The need to repeat could go away with a preamble.mk being generated by the spec file.

I didn't want the makefile to make any assumptions because a new variable being introduced would silently change the semantics of old packaging. I want this to fail early and loudly.

packaging/snapd.mk
zyga: Build mode will be under control of indigo, this is fine for the moment but will change soon (to another indirection layer where the distribution policy is injected)

packaging/snapd.mk
zyga: Sounds good.

packaging/snapd.mk
zyga: Hmm? Do you mean `-m 755`

packaging/snapd.mk
zyga: We must, otherwise tests are failing. How come arch has no C.UTF_8 locale?!? It's the best thing libc had introduced in years.

packaging/opensuse/snapd.spec
zyga: Note that the scope of this PR is smaller than a main central makefile.

packaging/snapd.mk
bboozzoo: > It's the best thing libc had introduced in years.
That's the catch, it's not glibc but downstream distros that introduced it.

zyga:
Let's return to this later.

packaging/snapd.mk
zyga: Apparently they are the sane ones then :)

packaging/opensuse/snapd.spec
zyga: We can look at undoing this now. I've implemented the idea with `snapd.defines.mk`

packaging/opensuse/snapd.spec
zyga: This is now improved :)

packaging/snapd.mk
zyga: I think this can happen in a follow-up.

packaging/snapd.mk
zyga: I believe we can tweak this as we get to apply this to Arch.

packaging/opensuse/snapd.spec
```diff
@@ -171,6 +167,28 @@ pushd %{indigo_srcdir}
...
+libexecdir=%{_libexecdir}
+mandir=%{_mandir}
+datadir=%{_datadir}
+localstatedir=%{_localstatedir}
+unitdir=%{_unitdir}
+# Build configuration
+with_core_bits=0
+with_alt_snap_mount_dir=%{!?with_alt_snap_mount_dir:0}%{?with_alt_snap_mount_dir:1}
+with_apparmor=%{!?with_apparmor:0}%{?with_apparmor:1}
+with_test_keys=%{!?with_test_keys:0}%{?with_test_keys:1}
```
bboozzoo: I think you can use `with_test_keys=%{with_testkeys}`

packaging/opensuse/snapd.spec
```diff
@@ -203,40 +221,30 @@ export LDFLAGS
 # Generate autotools build system files.
 pushd %{indigo_srcdir}/cmd
 autoreconf -i -f
+
 %configure \
     %{!?with_apparmor:--disable-apparmor} \
+    %{?with_apparmor:--enable-apparmor} \
```
bboozzoo: If you stay with %bcond_with apparmor, I believe this will need to be refactored info `%if.. %endif`

packaging/snapd.mk
bboozzoo: Yes

packaging/snapd.mk
```diff
@@ -0,0 +1,175 @@
...
+# distribution packaging generating snapd.defines.mk get out of sync.
+
+$(foreach var,$(vars),$(if $(value $(var)),,$(error $(var) is empty or unset, check snapd.defines.mk)))
+
+# ------------------------------------------------
+# There are no more control knobs after this point
+# ------------------------------------------------
+
+# Import path of snapd.
+import_path=github.com/snapcore/snapd
```
bboozzoo: I'm actually in favor of `foo = bar` rather than `foo=bar`. Find the former easier to real.

packaging/opensuse/snapd.spec
zyga: I tried to make the various bcond thing work correctly and I didn't manage to. If you have a patch that makes this what you intend I'm happy to merge it.

packaging/opensuse/snapd.spec
zyga: I am happy to undo this with a follow up.

packaging/opensuse/snapd.spec
```diff
@@ -171,6 +167,28 @@ pushd %{indigo_srcdir}
 # Add patch0 -p1 ... as appropriate here.
 popd
 
+# Generate snapd.defines.mk, this file is included by snapd.mk. It contains a
+# number of variable definitions that are set based on their RPM equivalents.
+# Since we can apply any conditional overrides here in the spec file we can
+# maintain one consistent set of variables across the spec and makefile worlds.
+cat >snapd.defines.mk <<__DEFINES__
+# This file is generated by opemSUSE's snapd.spec
```
stolowski: s/opem/open/

zyga:
```suggestion
# This file is generated by openSUSE's snapd.spec
```

packaging/snapd.mk
```diff
@@ -0,0 +1,175 @@
...
+# inconsistency between packaging of snapd across distributions.
+#
+# -----------------------------------------------------
+# Interface between snapd.mk and distribution packaging
+# -----------------------------------------------------
+#
+# Distribution packaging must generate snapd.defines.mk with a set of makefile
+# variable definitions that are discussed below. This allows the packaging
+# world to define directory layout and build configuration in one place and
+# this makefile to simply obey and implement that configuration.
```
stolowski: This is very welcome and nice change :heart:

packaging/opensuse/snapd.spec
```diff
@@ -171,6 +167,28 @@ pushd %{indigo_srcdir}
...
+bindir = %{_bindir}
+sbindir = %{_sbindir}
+libexecdir = %{_libexecdir}
+mandir = %{_mandir}
+datadir = %{_datadir}
+localstatedir = %{_localstatedir}
+unitdir = %{_unitdir}
+# Build configuration
+with_core_bits = 0
+with_alt_snap_mount_dir = %{!?with_alt_snap_mount_dir:0}%{?with_alt_snap_mount_dir:1}
```
stolowski: Would it make sense to simply pass %snap_mount_dir here?

packaging/snapd.mk
```diff
@@ -0,0 +1,175 @@
...
+
+# This is usually set by %make_install. It is defined here to avoid warnings or
+# errors from referencing undefined variables.
+DESTDIR? =
+
+# Decide which of the two snap mount directories to use. This is not
+# referencing localstatedir because the code copes with only those two values
+# explicitly.
+ifeq ($(with_alt_snap_mount_dir),1)
+snap_mount_dir = /var/lib/snapd/snap
```
stolowski: Related to my earlier question - would it make sense to simply use a preset snap_mount_dir value here and get rid of with_alt_snap_mount_dir flag, with /snap as a default value?

packaging/opensuse/snapd.spec
zyga: I decided against that because that gives the feeling of this being any directory we want. That is not the case and we want to really only support one of two locations.

packaging/snapd.mk
zyga: Answered above.

packaging/opensuse/snapd.spec
stolowski: I see, sure. Thanks

packaging/opensuse/snapd.spec
Conan-Kudo: You should just use what it was before...

packaging/opensuse/snapd.spec
Conan-Kudo: Can we please put this back now, since you've got the defines makefile now?

packaging/opensuse/snapd.spec
```diff
@@ -51,29 +56,18 @@
 # Additional entry of $GOPATH during the build process.
 # This is designed to be a sub-directory of {_builddir}/{name}-{version}
 # because that directory is automatically cleaned-up by the build process.
-%global         indigo_gopath   %{_builddir}/%{name}-%{version}/gopath
+%global indigo_gopath   %{_builddir}/%{name}-%{version}/gopath
```
Conan-Kudo: Why are we calling this indigo? This isn't really the indigo thing, is it?

packaging/opensuse/snapd.spec
```diff
@@ -51,29 +56,18 @@
...
-%global         indigo_srcdir   %{indigo_gopath}/src/%{import_path}
-
-%global with_test_keys  0
-
-%if %{with testkeys}
-%global with_test_keys 1
-%else
-%global with_test_keys 0
-%endif
+%global indigo_srcdir   %{indigo_gopath}/src/%{import_path}
```
Conan-Kudo: Same here?

packaging/opensuse/snapd.spec
```diff
@@ -171,6 +167,28 @@ pushd %{indigo_srcdir}
...
+mandir = %{_mandir}
+datadir = %{_datadir}
+localstatedir = %{_localstatedir}
+unitdir = %{_unitdir}
+# Build configuration
+with_core_bits = 0
+with_alt_snap_mount_dir = %{!?with_alt_snap_mount_dir:0}%{?with_alt_snap_mount_dir:1}
+with_apparmor = %{!?with_apparmor:0}%{?with_apparmor:1}
+with_test_keys = %{!?with_test_keys:0}%{?with_test_keys:1}
+__DEFINES__
```
Conan-Kudo: Wouldn't this be easier if we had something like a configure script where we could pass these options in and this file would be generated?

I don't have anything against this structure, other than it would make it hard to do with packaging systems like Debian's.

packaging/opensuse/snapd.spec
Conan-Kudo: This is fine, because of how the `with_<FOO>` macros are defined.

packaging/opensuse/snapd.spec
```diff
@@ -422,11 +396,16 @@ fi
 %{_datadir}/polkit-1/actions/io.snapcraft.snapd.policy
 %{_sysconfdir}/xdg/autostart/snap-userd-autostart.desktop
 %{_libexecdir}/snapd/snapd.run-from-snap
-%if %{with apparmor}
-%{_sysconfdir}/apparmor.d/usr.lib.snapd.snap-confine
-%endif
 %{_environmentdir}/990-snapd.conf
 %{_systemd_system_env_generator_dir}/snapd-env-generator
 
+%if 0%{?with_apparmor:1}
```
Conan-Kudo: Please change this back to `%{with apparmor}` along with changing back to bconds.

packaging/opensuse/snapd.spec
zyga: I don't see how putting this in a separate script helps (it would defeat the purpose of the makefile and would focus us to have hacks for rpm defines). I also don't see any issues with generating this from `debian/rules`. I expect most systems to simply generate a fixed `snapd.defines.mk`.

Conan-Kudo: That's fair. I could equally have this as an extra source file. This is fine as-is.

packaging/opensuse/snapd.spec
zyga: In a way it is not, indigo is after figuring out how to pass compiler flags that are appropriate. In another way it is. The two variables, `indigo_gopath` and `indigo_srcdir` define the GOPATH for the build process and root of the source tree of the current project. Indigo doesn't use them yet but I am thinking of a way to do so.

Conan-Kudo:
@zyga @mvo5 I have it on my TODO to re-review this soonish (hopefully in a few days or so).

zyga:
@Conan-Kudo feel free to suggest changes, either directly or on top of this. I'd like to land it (may not be perfect but it is a huge improvement) and iterate instead of stalling it as we have now.

packaging/opensuse/snapd.spec
zyga: Indigo may provide this value. Let's see how things evolve. I would keep it because it works.

PR #6116, opened 2018-11-08
cmd/libsnap: add simplified feature flag checker

zyga:
Prior approach to add a bridge for knowledge from snapd to via the facts
system was declined (though it may be brought back for another reason to
expose snap installation flags to snap-confine). Still, to move forward
I'd like to propose a more basic system where presence of a file in
/var/lib/snapd/experimental is considered as the feature being enabled.

Feature flags files will be written and removed by snapd in the
configuration handler for core. This will be done in a subsequent patch.

### DIFF
--- a/cmd/libsnap-confine-private/experimental.h
+++ b/cmd/libsnap-confine-private/experimental.h
@@ -1,0 +1,35 @@
+/*
+ * Copyright (C) 2018 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifndef SNAP_CONFINE_EXPERIMENTAL_H
+#define SNAP_CONFINE_EXPERIMENTAL_H
+
+#include <stdbool.h>
+
+typedef enum sc_feature_flag {
+	SC_PER_USER_MOUNT_NAMESPACE,
+} sc_feature_flag;
+
+/**
+ * sc_experimental_flag_active returns true if a given experimental flag
+ * has been activated by the user via "snap set core experimental.xxx=true".
+ * This is determined by testing the presence of a file in /var/lib/snapd/features/
+ * that is named after the flag name.
+**/
+bool sc_experimental_flag_active(sc_feature_flag flag);
+
+#endif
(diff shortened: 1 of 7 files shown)


cmd/libsnap-confine-private/experimental.h
```diff
@@ -0,0 +1,35 @@
...
+	SC_PER_USER_MOUNT_NAMESPACE,
+} sc_feature_flag;
+
+/**
+ * sc_experimental_flag_active returns true if a given experimental flag
+ * has been activated by the user via "snap set core experimental.xxx=true".
+ * This is determined by testing the presence of a file in /var/lib/snapd/features/
+ * that is named after the flag name.
+**/
+bool sc_experimental_flag_active(sc_feature_flag flag);
```
stolowski: I'd rename this to `sc_feature_flag_active`, per my general comment, and the file to `features.h` or some such. Current file name suggests it's experimental snap-confine stuff.

mvo5:
The code looks fine, I agree with @stolowski that it should probably not limited to "experimental" and more general. The dir name could be something like "features". Don't we already have one of these flag files? Or is that the user-mount thing?

zyga:
I don't mind calling this features though I was expecting this to be a short lived codebase that would be removed soon.

cmd/libsnap-confine-private/feature.h
```diff
@@ -0,0 +1,35 @@
...
+	SC_PER_USER_MOUNT_NAMESPACE,
+} sc_feature_flag;
+
+/**
+ * sc_feature_enabled returns true if a given feature flag has been activated
+ * by the user via "snap set core experimental.xxx=true". This is determined by
+ * testing the presence of a file in /var/lib/snapd/features/ that is named
+ * after the flag name.
+**/
+bool sc_feature_enabled(sc_feature_flag flag);
```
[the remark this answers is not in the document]
zyga: I thought about this but I decided to use an enum instead. There are only so many feature flags and it is easier to cause a typo this way (because you can always pass a string manually). This has the advantage that the code knows the set of flags that exist so it cannot be fooled.

zyga:
@mvo this is the first use of the idea. We just discussed this a while so it is in our mental cache.

I renamed it to features, including the directory name.

cmd/libsnap-confine-private/feature.h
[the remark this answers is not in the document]
mvo5: +1 for typos

cmd/libsnap-confine-private/feature.h
```diff
@@ -0,0 +1,35 @@
...
+ *
+ */
+
+#ifndef SNAP_CONFINE_FEATURE_H
+#define SNAP_CONFINE_FEATURE_H
+
+#include <stdbool.h>
+
+typedef enum sc_feature_flag {
+	SC_PER_USER_MOUNT_NAMESPACE,
```
mvo5: If you change the PR maybe SC_FEATURE_PER_USER_MOUNT_NAMESPACE (i.e. prefix with SC_FEATURE_) but not really important.

zyga: I like the SC_FEATURE idea. I'll change this in the next one.

PR #6118, opened 2018-11-08
tests: add core18 only hooks test and fix running core18 only on classic

mvo5:
This PR tests that we can run hooks like configure on systems that
do not have the core snap, just the core18 base. We got a report
that this breaks and this test verifies that it works correctly.


tests/main/core18-configure-hook/task.yaml
```diff
@@ -0,0 +1,31 @@
...
+    #shellcheck source=tests/lib/systemd.sh
+    . "$TESTSLIB/systemd.sh"
+    echo "Ensure all snaps are gone"
+    /usr/lib/snapd/snap-mgmt --purge
+
+    systemd_stop_units snapd.service
+    rm -f /var/lib/snapd/state.json
+    systemctl start snapd
+    snap wait system seed.loaded
+
```
bboozzoo:
```suggestion
```

PR #6119, opened 2018-11-08
sanity: refuse to try to do things on WSL.

chipaca:
Without this change, snapd could be started on WSL (daemons won't run,
but snapd can be started by hand). This'll put an end to that.

Also, snap would try to communicate with snapd on WSL, and error out
with the rather unhelpful "no such file or directory" on the socket.

This change puts WSL on parity with darwin: you'll see the help, and
can pack, download, and look at version; anything else should give you
an error message explaining the problem:

```
Interacting with snapd is not yet supported on Windows Subsystem for Linux.
This command has been left available for documentation purposes only.
```


release/release.go
```diff
@@ -108,17 +110,34 @@ func readOSRelease() OS {
 	return osRelease
 }
 
+var ioutilReadFile = ioutil.ReadFile
+
+func isWSL() bool {
+	version, err := ioutilReadFile("/proc/version")
```
zyga: Is this documented somehow by Microsoft? I also noticed that WSL has a particular uname output that could be used as an alternative detection strategy.

chipaca: https://github.com/Microsoft/WSL/issues/423#issuecomment-221627364

And, AIUI they themselves have code that checks in this way.

bboozzoo: Any idea whether Cygwin exposes /proc/version too? I tracked down this patch http://cygwin.1069669.n5.nabble.com/PATCH-add-information-to-proc-version-td91560.html but I'm not sure if it got applied or not.

PR #6120, opened 2018-11-08
cmd/snap-seccomp: add full complement of ptrace constants

zyga:
The snap-seccomp executable needs to understand each constant
it is given. When encountering unexpected constants it just returns
an error.

Since patch 093b3662047123fe684a0d1dae27775593f1dc1f
"interfaces/browser-support, cmd/snap-seccomp: Allow read-only ptrace,
for the Breakpad crash reporter." snap-seccomp will choke on the
profiles generated for the browser-support interface.

Since patch 353aa703001d63c5ee81af213b6557420b7188b4 "cmd/snap-seccomp:
only look for PTRACE_GETFPX?REGS where available" snap-seccomp contains
additional logic that handles various ptrace constants with
architecture-specific build tags. I'm not sure why this was done because
all of the constants are available in the C header file sys/ptrace.h

Initially I was thinking about making the parser silently ignore certain
PTRACE values but I have since reconsidered to just add the full since
they are just defined and are available at build time. After all, it
doesn't hurt if they don't work on a given architecture (at a kernel
level). We just want to white-list them in case they do work.

Fixes: https://bugs.launchpad.net/snapd/+bug/1802124


zyga:
@jdstrand tests failed here but I fixed that, will push a fix as soon as one more run is finished.

chipaca:
thank you for this.

PR #6121, opened 2018-11-09
tests: new test for snapshots with more than 1 user

sergiocazzolato:


bboozzoo:
Shellcheck doesn't like some things about the test:
```
2018-11-09 01:47:12 Error executing google:ubuntu-18.04-64:tests/unit/spread-shellcheck : 
-----
+ testdir=/home/gopath/src/github.com/snapcore/snapd/tests/unit/spread-shellcheck
+ CAN_FAIL=/home/gopath/src/github.com/snapcore/snapd/tests/unit/spread-shellcheck/can-fail
+ ./spread-shellcheck spread.yaml tests
ERROR:root:shellcheck failed for file tests/main/snapshot-users/task.yaml in sections: execute; error log follows
ERROR:root:tests/main/snapshot-users/task.yaml: section 'execute':

In - line 109:
cat /home/test/snap/test-snapd-tools/current/canary/canary.txt | MATCH "content updated"
    ^-- SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.


In - line 110:
cat /home/test/snap/test-snapd-tools/current/canary.txt | MATCH "content updated"
    ^-- SC2002: Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.

For more information:
  https://www.shellcheck.net/wiki/SC2002 -- Useless cat. Consider 'cmd < file | ....

ERROR:root:validation failed for the following non-whitelisted files:
 - tests/main/snapshot-users/task.yaml
-----
```

tests/main/snapshot-users/task.yaml
```diff
@@ -0,0 +1,122 @@
+summary: Check that the basic snapshots functionality works for different users
+
+prepare: |
+    snap install test-snapd-tools
+
+execute: |
+    # use the snaps, so they create the dirs
+    test-snapd-tools.echo
+    su -p -c "/snap/bin/test-snapd-tools.echo" test
```
bboozzoo: Fedora and Arch do not have `/snap` by default:
```
. "$TESTSLIB"/dirs.sh
su -p -c "$SNAP_MOUNT_DIR/bin/test-snapd-tools.echo" test
```

zyga: Can we just use `snap run test-snapd-tools.echo` instead?

zyga: This fails on:
```
/bin/bash: line 61: SNAP_MOUNT_DIR: unbound variable
```
We probably need to source dirs.sh

tests/main/snapshot-users/task.yaml
```diff
@@ -0,0 +1,125 @@
...
+    snap check-snapshot "$SET_ID_BOTH"
+    snap check-snapshot "$SET_ID_NONE"
+
+    # remove the canaries for both users
+    rm -f /root/snap/test-snapd-tools/{current,common}/canary.txt
+    rm -f /home/test/snap/test-snapd-tools/{current,common}/canary.txt
+
+    # restore the snapshot for the root user and check the files
+    snap restore "$SET_ID_ROOT" test-snapd-tools
+    test -e /root/snap/test-snapd-tools/current/canary.txt
```
mvo5: Should we check the content of the files here (and in the subsequent restores) as well?

tests/main/snapshot-users/task.yaml
```diff
@@ -0,0 +1,125 @@
...
+
+    # restore the snapshot for both user and check the files
+    snap restore "$SET_ID_NONE" --users=test,root
+    test -e /root/snap/test-snapd-tools/current/canary.txt
+    test -e /root/snap/test-snapd-tools/common/canary.txt
+    test -e /home/test/snap/test-snapd-tools/current/canary.txt
+    test -e /home/test/snap/test-snapd-tools/common/canary.txt
+    test -e /home/test/snap/test-snapd-tools/current/canary/canary.txt
+
+    # check files content
```
mvo5: Should we run similar checks for the previous restores too?

tests/main/snapshot-users/task.yaml
```diff
@@ -0,0 +1,125 @@
...
+    MATCH "content updated" < /home/test/snap/test-snapd-tools/current/canary/canary.txt
+    MATCH "content updated" < /home/test/snap/test-snapd-tools/current/canary.txt
+
+    # check removal works
+    snap forget "$SET_ID_NONE"
+    snap saved --id="$SET_ID_NONE" | grep "No snapshots found"
```
mvo5: It looks like we want the same checks for the `snap forget` calls below. Maybe also a `[ $(snap saved | wc -l) = 0 ]` or something at the end?

sergiocazzolato: I implemented in that way to be consisten with the other snapshots tests, I'll see if there another better way to check that.

tests/main/snapshot-users/task.yaml
sergiocazzolato: done

tests/main/snapshot-users/task.yaml
sergiocazzolato: done

sergiocazzolato:
@mvo5 , thanks for reviewing it, could you please take a look again?

tests/main/snapshot-users/task.yaml
sergiocazzolato: done

sergiocazzolato:
@chipaca, I think the main differences are:
. it is checking not just files but subdirectories are correctly snapshoted
. it is checking the restore for different users and both users and validating the content
. it is validating the option --users for the restore and save

The idea of this test is to complement the snapshot-cross-revno one, perhaps we also could just add some parts this one?

PR #6123, opened 2018-11-09
cmd/snap-confine,snap-update-ns: discard quirks

zyga:
Recent discussion about LXD has resurrected the topic of quirks. Quirks
were created in response to a bug where a snap in devmode being unable
to interact with with LXD running as a classic package. At the time it
was a big deal so we implemented the quirk and everything was fine.

Time passed and LXD moved away from storing the socket in that place. It
also moved away from shipping as a classic package as well. Since we
never allowed access to /var/lib/lxd via any interfaces and it is
unlikely anyone still relies on this quirk I think it should be dropped
now.

Removing should speed up execution of all snaps by a small amount, since
we don't need to construct a writable mimic in /var/lib anymore.  We
also shrink snap-confine by removing a good chunk of C code.

https://bugs.launchpad.net/snap-confine/+bug/1613845


zyga:
I think this needs more tweaks. I suspect I removed too many apparmor permissions.

PR #6126, opened 2018-11-09
dirs, wrappers, overlord/snapstate: make completion + bases work

chipaca:
The old completion code assumed that if a snap was installed `<snap
mount dir>/core/current/usr/lib/snapd/complete.sh` was there, and used
that to complete things.

This is not true now that you can have a purely core18-based system,
for example, so completers wouldn't get created.

This fixes that: complete.sh is from core if the snap has no base,
from the snapd snap if it is installed, and otherwise from the host.

This is the first half of the fix for [lp:1802721](https://bugs.launchpad.net/snapd/+bug/1802721).

### DIFF
--- a/dirs/dirs.go
+++ b/dirs/dirs.go
@@ -98,7 +98,6 @@ var (
 
 	CompletionHelper string
 	CompletersDir    string
-	CompleteSh       string
 
 	SystemFontsDir           string
 	SystemLocalFontsDir      string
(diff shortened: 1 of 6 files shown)


dirs/dirs.go
```diff
@@ -293,3 +291,27 @@ func SetRootDir(rootdir string) {
 	ErrtrackerDbDir = filepath.Join(rootdir, snappyDir, "errtracker.db")
 	SysfsDir = filepath.Join(rootdir, "/sys")
 }
+
+// what inside a (non-classic) snap is /usr/lib/snapd, outside can come from different places
+func libExecOutside(base string) string {
```
zyga: The code is ok but the name of the function feels slightly off. The location of libecexdir on the host is a constant. I understand this really means “one of viable libexecdir locations of *snapd*” but the name doesn’t reflect that.

Perhaps snapdCompletionDir or something like that?

chipaca: but we have `CoreLibExecDir = "/usr/lib/snapd"` and `DistroLibExecDir = filepath.Join(rootdir, "/usr/libexec/snapd")`.

Maybe I don't understand your point.

chipaca: I guess the point is that the directory that is CoreLibExecDir inside a snap is not constant, from outside.

PR #6129, opened 2018-11-12
data/completion: pass documented arguments to completion functions

chipaca:
completion functions (and commands, which we treat the same way) are
documented as receiving three arguments. I hadn't noticed, as almost
nobody uses them because most everything you need is in the different
COMP_ variables, but python's argcomplete module uses it.

This fixes that, and adds a spread test just in case.

This is the second half of the fix for [lp:1802721](https://bugs.launchpad.net/snapd/+bug/1802721).

### DIFF
--- a/data/completion/etelpmoc.sh
+++ b/data/completion/etelpmoc.sh
@@ -180,7 +180,7 @@ if [[ "${_comp[*]}" ]]; then
                 _compfunc="$OPTARG"
                 ;;
             W)
-                readarray -t COMPREPLY < <( builtin compgen -W "$OPTARG" -- "${COMP_WORDS[$COMP_CWORD]}" )
+                readarray -t COMPREPLY < <( builtin compgen -W "$OPTARG" -- "${COMP_WORDS[COMP_CWORD]}" )
                 _compfunc=""
                 ;;
             *)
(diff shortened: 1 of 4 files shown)


data/completion/etelpmoc.sh
```diff
@@ -180,7 +180,7 @@ if [[ "${_comp[*]}" ]]; then
                 _compfunc="$OPTARG"
                 ;;
             W)
-                readarray -t COMPREPLY < <( builtin compgen -W "$OPTARG" -- "${COMP_WORDS[$COMP_CWORD]}" )
+                readarray -t COMPREPLY < <( builtin compgen -W "$OPTARG" -- "${COMP_WORDS[COMP_CWORD]}" )
```
zyga: Is this related to the bug in question? It looks like a bug fix but I don't see how it relates to the 3rd argument.

bboozzoo: I believe it's just a run by fix for not needing `$i` for array indices as those are already treated like arithmetic evaluation.

data/completion/etelpmoc.sh
```diff
@@ -202,10 +202,19 @@ esac
...
+
+        # from https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion.html:
+        #   When the function or command is invoked, the first argument ($1) is
+        #   the name of the command whose arguments are being completed, the
+        #   second argument ($2) is the word being completed, and the third
+        #   argument ($3) is the word preceding the word being completed on the
+        #   current command line.
+        # that's "$1" "${COMP_WORDS[COMP_CWORD]}" and "${COMP_WORDS[COMP_CWORD-1]}"
+        # (probably)
+        $_compfunc "$1" "${COMP_WORDS[COMP_CWORD]}" "${COMP_WORDS[COMP_CWORD-1]}"
```
bboozzoo: Do you know whether COMP_CWORD is always > 0 when this is called?

chipaca: yes, because `COMP_WORDS[0]` is the command you're completing (so you can't get here without `COMP_CWORD` being at least 1)

PR #6130, opened 2018-11-12
snap: make Epoch's MarshalJSON not simplify

chipaca:
Before this change, snap.Epoch's marshaller tried to use the
simplified notation when possible, i.e. it'd write `"1*"` instead of
`{"read":[0,1],"write":[1]}`.

While we still want to do this when printing it to the user, the store
has dropped support for the more compact representation (if it ever
had it), which means we'd need a different marshaller there. Rather
than that, this just drops the smarts from the marshaller. String()
still calls simplify() so that path is unchanged (and we explicitly
don't serialise to YAML).

### DIFF
--- a/snap/epoch.go
+++ b/snap/epoch.go
@@ -182,7 +182,12 @@ func (e *Epoch) simplify() interface{} {
 }
 
 func (e *Epoch) MarshalJSON() ([]byte, error) {
-	return json.Marshal(e.simplify())
+	// note the nil case doesn't happen unless called explicitly
+	if e == nil || (e.Read == nil && e.Write == nil) {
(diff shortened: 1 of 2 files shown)


snap/epoch.go
```diff
@@ -182,7 +182,12 @@ func (e *Epoch) simplify() interface{} {
 }
 
 func (e *Epoch) MarshalJSON() ([]byte, error) {
-	return json.Marshal(e.simplify())
+	// note the nil case doesn't happen unless called explicitly
```
mvo5: Silly question but why do we care about this case if doesn't happen?

PR #6134, opened 2018-11-13
tests: skip opensuse from interfaces-openvswitch-support test

sergiocazzolato:
Next update for opensuse is failing when interfaces-openvswitch-support
test is executed. The cause is the same than for arch system, where the
interface is allowing access to /run/uuidd/request and in these systems
the request is done in /run/run/uuidd/request, making fail the snaps
which try to request a random id.

test error:
https://paste.ubuntu.com/p/bv9xZj36XR/

debug info:
https://paste.ubuntu.com/p/nMF4BR8ZF7/

### DIFF
--- a/tests/main/interfaces-openvswitch-support/task.yaml
+++ b/tests/main/interfaces-openvswitch-support/task.yaml
@@ -5,8 +5,8 @@ details: |
 
 # ubuntu-core, ubuntu-14, fedora, amazon are skipped as /run/uuidd/request file does not
 # exist. On those systems different files are being used instead.
-# arch: uses /run/run/uuidd/request, filed a bug report https://bugs.archlinux.org/task/58122
-systems: [-ubuntu-14.04-*,-ubuntu-core-*,-fedora-*, -arch-*, -amazon-*]
+# arch, opensuse: uses /run/run/uuidd/request, filed a bug report https://bugs.archlinux.org/task/58122
+systems: [-ubuntu-14.04-*,-ubuntu-core-*,-fedora-*, -arch-*, -amazon-*, -opensuse-*]
 
 prepare: |
     snap install test-snapd-openvswitch-support


tests/main/interfaces-openvswitch-support/task.yaml
```diff
@@ -5,8 +5,8 @@ details: |
 
 # ubuntu-core, ubuntu-14, fedora, amazon are skipped as /run/uuidd/request file does not
 # exist. On those systems different files are being used instead.
-# arch: uses /run/run/uuidd/request, filed a bug report https://bugs.archlinux.org/task/58122
-systems: [-ubuntu-14.04-*,-ubuntu-core-*,-fedora-*, -arch-*, -amazon-*]
+# arch, opensuse: uses /run/run/uuidd/request, filed a bug report https://bugs.archlinux.org/task/58122
```
mvo5: Is this comment at the right place? I mean, the bug was against the arch bugtracker is marked as fixed.

bboozzoo: @sergiocazzolato can you file a bug at https://bugzilla.opensuse.org/index.cgi ? I'll update the test for Arch in a separate PR.

mvo5: Sorry, I think I was a bit unclear here - does suse has the same bug that they use "/run/run/uuidd" (i.e. the double /run/run) ?

zyga: Yes, suse has the exact same bug. I tested this in tumbleweed.

zyga: FYI https://github.com/snapcore/snapd/pull/6110/files#diff-913d16e6e81868475136850d59dc29bfR9

PR #6135, opened 2018-11-13
packaging/arch: fix bash completions path

bboozzoo:
Bash completion files are under `$(datadir)/bash-completion/completions`

### DIFF
--- a/packaging/arch/PKGBUILD
+++ b/packaging/arch/PKGBUILD
@@ -120,7 +120,7 @@ package() {
 
   # Install bash completion
   install -Dm644 data/completion/snap \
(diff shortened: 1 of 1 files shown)


packaging/arch/PKGBUILD
```diff
@@ -120,7 +120,7 @@ package() {
 
   # Install bash completion
   install -Dm644 data/completion/snap \
-    "$pkgdir/usr/share/bash-completion/completion/snap"
+    "$pkgdir/usr/share/bash-completion/completions/snap"
```
chipaca: whoops

PR #6136, opened 2018-11-13
tests: run interfaces-openvswitch-support on arch again, the arch bug is fixed

mvo5:
Bug https://bugs.archlinux.org/task/58122 is marked fixed so we can enable this test on arch again (presumably).


mvo5:
@zyga What is the bugnumber?

PR #6139, opened 2018-11-13
snap, store, overlord/snapshotstate: drop epoch pointers

chipaca:
I got tired of the `snap.Epoch` vs `*snap.Epoch` mess.

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -1260,7 +1260,6 @@ func (s *Store) WriteCatalogs(ctx context.Context, names io.Writer, adder SnapAd
 type RefreshCandidate struct {
 	SnapID   string
 	Revision snap.Revision
-	Epoch    snap.Epoch
 	Block    []snap.Revision
 
 	// the desired channel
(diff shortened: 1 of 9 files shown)


store/store.go
```diff
@@ -1980,14 +1978,13 @@ func isValidAction(action string) bool {
 }
 
 type snapActionJSON struct {
-	Action           string      `json:"action"`
-	InstanceKey      string      `json:"instance-key"`
-	Name             string      `json:"name,omitempty"`
-	SnapID           string      `json:"snap-id,omitempty"`
-	Channel          string      `json:"channel,omitempty"`
-	Revision         int         `json:"revision,omitempty"`
-	Epoch            *snap.Epoch `json:"epoch,omitempty"`
```
zyga: Removal of Epoch doesn’t fit in the description of the branch that changes pointer to non pointer type. Is this intentional?

chipaca: they were always nil, without the change that follows this one. I didn't want this pr to start sending bogus epochs if it landed without the followup.

chipaca: The followup is https://github.com/snapcore/snapd/commit/cd34903524215c4f96f8841c3c1bf1f8ff944cd4

PR #6145, opened 2018-11-13
cmd/libsnap: add sc_verify_snap_lock

zyga:
This new API allows ensuring that a lock is held. This is needed by
tools that are invoked from snap-confine that would otherwise normally
grab a lock themselves. Instead when they are invoked from snap-confine
they check that a lock is held.

### DIFF
--- a/cmd/libsnap-confine-private/locking.c
+++ b/cmd/libsnap-confine-private/locking.c
@@ -176,14 +176,11 @@ void sc_verify_snap_lock(const char *snap_name)
 		errno = 0;
 		die("unexpectedly managed to acquire exclusive lock over snap %s", snap_name);
 	}
-	if (retval < 0) {
-		if (errno == EWOULDBLOCK) {
-			/* We tried but failed to grab the lock because the file is already
-			 * locked. Good, this is what we expected. */
-			return;
-		}
-		die("cannot acquire exclusive lock over snap %s", snap_name);
+	if (retval < 0 && errno != EWOULDBLOCK) {
+		die("cannot verify exclusive lock over snap %s", snap_name);
 	}
+	/* We tried but failed to grab the lock because the file is already locked.
+	 * Good, this is what we expected. */
 }
 
 int sc_lock_snap_user(const char *snap_name, uid_t uid)
--- a/cmd/libsnap-confine-private/locking.c
+++ b/cmd/libsnap-confine-private/locking.c
@@ -166,7 +166,8 @@ void sc_verify_snap_lock(const char *snap_name)
 	int lock_fd, retval;
 
 	lock_fd = open_lock(snap_name, 0);
-	debug("trying to acquire exclusive lock over snap %s", snap_name);
+	debug("trying to verify whether exclusive lock over snap %s is held",
+	      snap_name);
 	retval = flock(lock_fd, LOCK_EX | LOCK_NB);
 	if (retval == 0) {
 		/* We managed to grab the lock, the lock was not held! */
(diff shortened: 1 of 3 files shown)


zyga:
NOTE: This is an early version of the patch, I will likely split it. I opened this to see test results tomorrow.

cmd/libsnap-confine-private/feature.c
```diff
@@ -42,7 +42,9 @@ bool sc_feature_enabled(sc_feature_flag flag)
 	}
 
 	int dirfd SC_CLEANUP(sc_cleanup_close) = -1;
-	dirfd = open(feature_flag_dir, O_CLOEXEC | O_DIRECTORY | O_NOFOLLOW | O_PATH);
+	dirfd =
```
zyga: The more I `make fmt` the less I like `indent` as a formatting tool.

cmd/libsnap-confine-private/locking.c
```diff
@@ -143,6 +161,30 @@ int sc_lock_snap(const char *snap_name)
...
+	}
+	if (retval < 0) {
+		if (errno == EWOULDBLOCK) {
+			/* We tried but failed to grab the lock because the file is already
+			 * locked. Good, this is what we expected. */
+			return;
+		}
+		die("cannot acquire exclusive lock over snap %s", snap_name);
+	}
+}
```
stolowski: I've just verified flock manual, >0 is not returned, so this is fine. Nonetheless, should we die here just in case?

zyga: Flock is documented to return zero -1 and set errno. I don't think we need to handle >0, we certainly don't do that in other system calls.

cmd/libsnap-confine-private/locking.c
```diff
@@ -143,6 +161,30 @@ int sc_lock_snap(const char *snap_name)
...
+	debug("trying to acquire exclusive lock over snap %s", snap_name);
+	retval = flock(lock_fd, LOCK_EX | LOCK_NB);
+	if (retval == 0) {
+		/* We managed to grab the lock, the lock was not held! */
+		flock(lock_fd, LOCK_UN);
+		close(lock_fd);
+		errno = 0;
+		die("unexpectedly managed to acquire exclusive lock over snap %s", snap_name);
+	}
+	if (retval < 0) {
```
bboozzoo: I'd wrap this into:
```
void sc_verify_snap_lock(const char *snap_name) {
    ...
    if (retval < 0 && errno != EWOULDBLOCK) {
       die("failed to verify exclusive lock over snap %s", snap_name);
    }

    /* We tried but failed to grab the lock because the file is already
     * locked. Good, this is what we expected. */
}
```

cmd/libsnap-confine-private/locking.c
```diff
@@ -143,6 +161,30 @@ int sc_lock_snap(const char *snap_name)
...
+		if (errno == EWOULDBLOCK) {
+			/* We tried but failed to grab the lock because the file is already
+			 * locked. Good, this is what we expected. */
+			return;
+		}
+		die("cannot acquire exclusive lock over snap %s", snap_name);
```
bboozzoo: Maybe `cannot verify whether exclusive lock over snap %s is held`?

cmd/libsnap-confine-private/locking.c
```diff
@@ -143,6 +161,30 @@ int sc_lock_snap(const char *snap_name)
 	return sc_lock_generic(snap_name, 0);
 }
 
+void sc_verify_snap_lock(const char *snap_name)
+{
+	int lock_fd, retval;
+
+	lock_fd = open_lock(snap_name, 0);
+	debug("trying to acquire exclusive lock over snap %s", snap_name);
```
bboozzoo: maybe `trying to verify whether exclusive lock over snap %s is held`

PR #6146, opened 2018-11-14
ifacestate/helpers: added SystemSnapName mapper helper method

stolowski:
Added `SystemSnapName` helper method to mapper, followup to #6138.

### DIFF
--- a/overlord/ifacestate/helpers.go
+++ b/overlord/ifacestate/helpers.go
@@ -679,11 +679,6 @@ func (m *IdentityMapper) RemapSnapFromRequest(snapName string) string {
 	return snapName
 }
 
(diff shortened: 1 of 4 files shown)


overlord/ifacestate/helpers.go
```diff
@@ -677,6 +679,11 @@ func (m *IdentityMapper) RemapSnapFromRequest(snapName string) string {
 	return snapName
 }
 
+// Returns actual name of the system snap - needs to be reimplemented by concrete mapper.
+func (m *IdentityMapper) SystemSnapName() string {
```
bboozzoo: You don't need to provide this. If it's just for the tests, you can put it in `export_test.go`.

stolowski: Good idea, thanks! Done.

PR #6149, opened 2018-11-14
cmd/snap-confine: capture initialized per-user mount ns

zyga:
When the experimental feature with persistent per-user mount namespaces
is enabled then a new mount namespace is created in /run/snapd/ns,
alongside $SNAP_NAME.mnt we now get $SNAP_NAME.$UID.mnt.

Note that per-user mount namespaces are not supported for the root user.
As a design decision root user is synonymous with the per-snap mount
namespace.

Preserved per-user mount namespaces are not inspected to detect
staleness. Instead the per-snap mount namespace, if stale, is discarded
alongside the mount namespaces of all the users of that snap on the
system, by calling snap-discard-ns.

### DIFF
--- a/cmd/snap-confine/ns-support.c
+++ b/cmd/snap-confine/ns-support.c
@@ -566,6 +566,7 @@ int sc_join_preserved_per_user_ns(struct sc_mount_ns *group,
 		die("cannot inspect preserved mount namespace file");
 	}
 #ifndef NSFS_MAGIC
+	/* Define NSFS_MAGIC for Ubuntu 14.04 and other older systems. */
 #define NSFS_MAGIC 0x6e736673
 #endif
 	if (ns_statfs_buf.f_type == NSFS_MAGIC
--- a/cmd/snap-confine/ns-support.c
+++ b/cmd/snap-confine/ns-support.c
@@ -728,10 +728,9 @@ static void helper_capture_per_user_ns(struct sc_mount_ns *group, pid_t parent)
 	sc_must_snprintf(src, sizeof src, "/proc/%d/ns/mnt", (int)parent);
 	sc_must_snprintf(dst, sizeof dst, "%s.%d.mnt", group->name, (int)uid);
 	if (mount(src, dst, NULL, MS_BIND, NULL) < 0) {
-		die("cannot preserve mount namespace of process %d as %s",
-		    (int)parent, dst);
+		die("cannot preserve per-user mount namespace of process %d as %s", (int)parent, dst);
 	}
-	debug("mount namespace of process %d preserved as %s",
+	debug("per-user mount namespace of process %d preserved as %s",
 	      (int)parent, dst);
 }
 
--- a/cmd/snap-confine/ns-support.c
+++ b/cmd/snap-confine/ns-support.c
@@ -75,7 +75,8 @@ static const char *sc_ns_dir = SC_NS_DIR;
 
 enum {
 	HELPER_CMD_EXIT,
-	HELPER_CMD_CAPTURE_NS,
+	HELPER_CMD_CAPTURE_MOUNT_NS,
+	HELPER_CMD_CAPTURE_PER_USER_MOUNT_NS,
 };
 
 /**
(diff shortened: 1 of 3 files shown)


zyga:
This is failing because it depends on https://github.com/snapcore/snapd/pull/6147

cmd/snap-confine/ns-support.c
```diff
@@ -541,6 +543,56 @@ int sc_join_preserved_ns(struct sc_mount_ns *group, struct sc_apparmor
...
+	}
+	struct statfs ns_statfs_buf;
+	if (fstatfs(mnt_fd, &ns_statfs_buf) < 0) {
+		die("cannot inspect filesystem of preserved mount namespace file");
+	}
+	struct stat ns_stat_buf;
+	if (fstat(mnt_fd, &ns_stat_buf) < 0) {
+		die("cannot inspect preserved mount namespace file");
+	}
+#ifndef NSFS_MAGIC
```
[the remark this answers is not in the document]
zyga: This is not defined on 14.04 last time I looked.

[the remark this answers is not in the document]
stolowski: OK, can you add a one-line comment about 14.04 there?

cmd/snap-confine/ns-support.c
```diff
@@ -541,6 +543,57 @@ int sc_join_preserved_ns(struct sc_mount_ns *group, struct sc_apparmor
...
+		if (vanilla_cwd == NULL) {
+			die("cannot get the current working directory");
+		}
+		if (setns(mnt_fd, CLONE_NEWNS) < 0) {
+			die("cannot join preserved per-user mount namespace %s",
+			    group->name);
+		}
+		debug("joined preserved mount namespace %s", group->name);
+		if (chdir(vanilla_cwd) != 0) {
+			debug("cannot enter %s, moving to void", vanilla_cwd);
```
stolowski: Should we make this more visible than debug?

zyga: No, this is by design.

Users cannot do anything about this, the extra output would be visible (unexpectedly) from other programs and might break their behaviour.

cmd/snap-confine/ns-support.c
```diff
@@ -662,6 +718,23 @@ static void helper_capture_ns(struct sc_mount_ns *group, pid_t parent)
...
 
+static void helper_capture_per_user_ns(struct sc_mount_ns *group, pid_t parent)
+{
+	char src[PATH_MAX] = { 0 };
+	char dst[PATH_MAX] = { 0 };
+	uid_t uid = getuid();
+
+	debug("capturing per-snap, per-user mount namespace");
+	sc_must_snprintf(src, sizeof src, "/proc/%d/ns/mnt", (int)parent);
+	sc_must_snprintf(dst, sizeof dst, "%s.%d.mnt", group->name, (int)uid);
```
bboozzoo: To keep with convention, that should probably be:
```
sc_must_snprintf(dst, sizeof dst, "%s.%d%s", group->name, (int)uid, SC_NS_MNT_FILE);
```

cmd/snap-confine/ns-support.c
```diff
@@ -662,6 +718,23 @@ static void helper_capture_ns(struct sc_mount_ns *group, pid_t parent)
...
+	if (mount(src, dst, NULL, MS_BIND, NULL) < 0) {
+		die("cannot preserve mount namespace of process %d as %s",
```
bboozzoo: `s/mount/per-user mount/`

cmd/snap-confine/ns-support.c
```diff
@@ -662,6 +718,23 @@ static void helper_capture_ns(struct sc_mount_ns *group, pid_t parent)
...
+		    (int)parent, dst);
+	}
+	debug("mount namespace of process %d preserved as %s",
```
bboozzoo: same here

cmd/snap-confine/ns-support.c
zyga: I'd like to drop that SC_NS_MNT_FILE define. It is not universally used and it clouds the file name more than it helps.

bboozzoo: Sounds good to me. With the way it's currently defined the printf format string looks a bit awkward.

zyga: I'll nuke that define.

PR #6151, opened 2018-11-14
snapd: allow snap-update-ns to read /proc/version

zyga:
When doing some casual exploratory testing on my branches I noticed an
unrelated apparmor denial.

    type=AVC msg=audit(1542197569.186:1140): apparmor="DENIED"
    operation="open" profile="snap-update-ns.snapd-hacker-toolbelt"
    name="/proc/version" pid=37237 comm="3" requested_mask="r"
    denied_mask="r" fsuid=0 ouid=0

It seems that release/release.go's WSL check now affects snap-update-ns.
While we could do lazy init the denial should go a way as well.

### DIFF
--- a/interfaces/apparmor/template.go
+++ b/interfaces/apparmor/template.go
@@ -625,6 +625,8 @@ profile snap-update-ns.###SNAP_INSTANCE_NAME### (attach_disconnected) {
 
   # Allow reading file descriptor paths
   @{PROC}/@{pid}/fd/* r,
+  # Allow reading /proc/version. For release.go WSL detection.
+  @{PROC}/version r,
 
   # Allow reading the os-release file (possibly a symlink to /usr/lib).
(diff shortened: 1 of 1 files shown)


interfaces/apparmor/template.go
```diff
@@ -699,6 +699,9 @@ profile snap-update-ns.###SNAP_INSTANCE_NAME### (attach_disconnected) {
   /tmp/ r,
   /tmp/.snap/{,**} rw,
 
+  # Allow reading /proc/version. For release.go WSL detection.
+  /proc/version r,
```
jdstrand: Please use `@{PROC}/version`, btw otherwise LGTM. Would be nice if this was moved up near the other @{PROC} accesses.

zyga: Will do, thank you for the quick review!

PR #6153, opened 2018-11-14
tests: use mock-gpio.py in enable-disable-units-gpio test

mvo5:
This is build on top of https://github.com/snapcore/snapd/pull/6128

Ths PR makes the enable-disable-units-gpio test use the fake-gpio.py service which simplifies the test.

### DIFF
--- a/tests/lib/fakegpio/fake-gpio.py
+++ b/tests/lib/fakegpio/fake-gpio.py
@@ -6,6 +6,7 @@ import logging
 import os
 import selectors
 import shutil
+import socket
 import subprocess
 import sys
 import tempfile
@@ -59,6 +60,15 @@ def dispatch(sel):
     return True
 
 
+def maybe_sd_notify(s: str) -> None:
+    addr = os.getenv('NOTIFY_SOCKET')
+    if not addr:
+        return
+    soc = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
+    soc.connect(addr)
+    soc.sendall(s.encode())
+
+
 def main():
     """the main method"""
     if os.getuid() != 0:
@@ -83,6 +93,8 @@ def main():
     ufd = os.open("unexport", os.O_RDWR | os.O_NONBLOCK)
     sel.register(efd, selectors.EVENT_READ, export_ready)
     sel.register(ufd, selectors.EVENT_READ, unexport_ready)
+    # notify
+    maybe_sd_notify("READY=1")
     while True:
         if not dispatch(sel):
             break
--- a/tests/lib/systemd.sh
+++ b/tests/lib/systemd.sh
@@ -2,12 +2,36 @@
 
 # Use like systemd_create_and_start_unit(fakestore, "$(which fakestore) -start -dir $top_dir -addr localhost:11028 $@")
 systemd_create_and_start_unit() {
-    printf '[Unit]\nDescription=For testing purposes\n[Service]\nType=simple\nExecStart=%s\n' "$2" > "/run/systemd/system/$1.service"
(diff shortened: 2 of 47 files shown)


zyga:
This failed with:
```

In tests/lib/systemd.sh line 22:
        printf "$3" >> "/etc/systemd/system/$1.service.d/override.conf"
               ^-- SC2059: Don't use variables in the printf format string. Use printf "..%s.." "$foo".
```

tests/lib/systemd.sh
```diff
@@ -2,14 +2,38 @@
...
+# Create and start a persistent systemd unit that survives reboots. Use as:
+#   systemd_create_and_start_persistent_unit "name" "my-service --args"
+# The third arg supports "overrides" which allow to customize the service
+# as needed, e.g.:
+#   systemd_create_and_start_persistent_unit "name" "start" "[Unit]\nAfter=foo"
+systemd_create_and_start_persistent_unit() {
+    printf '[Unit]\nDescription=For testing purposes\n[Service]\nType=simple\nExecStart=%s\n[Install]\nWantedBy=multi-user.target\n' "$2" > "/etc/systemd/system/$1.service"
+    if [ -n "${3:-}" ]; then
+        mkdir -p "/etc/systemd/system/$1.service.d"
+        printf "$3" >> "/etc/systemd/system/$1.service.d/override.conf"
```
zyga:
```suggestion
        printf "%s" "$3" >> "/etc/systemd/system/$1.service.d/override.conf"
```

tests/lib/fakegpio/fake-gpio.py
```diff
@@ -59,6 +60,15 @@ def dispatch(sel):
...
 
+def maybe_sd_notify(s: str) -> None:
+    addr = os.getenv('NOTIFY_SOCKET')
+    if not addr:
+        return
+    soc = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
+    soc.connect(addr)
+    soc.sendall(s.encode())
+
+
```
stolowski: Nitpick, double newline.

tests/lib/systemd.sh
```diff
@@ -10,6 +10,31 @@ systemd_create_and_start_unit() {
     systemctl start "$1"
 }
 
+# Create and start a persistent systemd unit that survives reboots. Use as:
+#   systemd_create_and_start_persistent_unit "name" "my-service --args"
+# The third arg supports "overrides" which allow to customize the service
+# as needed, e.g.:
+#   systemd_create_and_start_persistent_unit "name" "start" "[Unit]\nAfter=foo"
+systemd_create_and_start_persistent_unit() {
```
stolowski: Nice!

tests/lib/systemd.sh
```diff
@@ -10,6 +10,31 @@ systemd_create_and_start_unit() {
...
+    printf '[Unit]\nDescription=For testing purposes\n[Service]\nType=simple\nExecStart=%s\n[Install]\nWantedBy=multi-user.target\n' "$2" > "/etc/systemd/system/$1.service"
```
stolowski: If we could automatically insert the name of current test into the description, that would be a nice touch, perahps spread exposes it somehow?

tests/lib/fakegpio/fake-gpio.py
mvo5: I added those because pep8 wants double newlines between functions that are not part of a class.

tests/lib/systemd.sh
mvo5: I like this a lot, will do a followup with this idea.

PR #6154, opened 2018-11-14
snap: enforce minimal snap name len of 2

mvo5:
Trivial drive-by after looking at LP: #1763048


snap/validate.go
```diff
@@ -93,7 +93,7 @@ func ValidateInstanceName(instanceName string) error {
 func ValidateName(name string) error {
 	// NOTE: This function should be synchronized with the two other
 	// implementations: sc_snap_name_validate and validate_snap_name .
```
stolowski: Per this comment, it seems we need similiar change in sc_snap_name_validate?

cmd/libsnap-confine-private/snap-test.c
```diff
@@ -201,6 +201,8 @@ static void test_sc_snap_or_instance_name_validate(gconstpointer data)
 	const char *invalid_names[] = {
 		// name cannot be empty
 		"",
+		// too short
+		"a",
```
zyga: Nice touch :)

cmd/libsnap-confine-private/snap-test.c
```diff
@@ -317,8 +319,8 @@ static void test_sc_instance_name_validate(void)
 	sc_error_free(err);
 
 	const char *valid_names[] = {
-		"a", "aa", "aaa", "aaaa",
-		"a_a", "aa_1", "a_123", "a_0123456789",
```
zyga: Interesting that short snap name with long instance key is still too short. Useful test.

cmd/snap-update-ns/bootstrap.c
```diff
@@ -239,6 +239,10 @@ int validate_snap_name(const char *snap_name)
 		bootstrap_msg = "snap name must contain at least one letter";
 		return -1;
 	}
+        if (n < 2) {
```
zyga: Formatting seems off, I'll correct that.

snap/validate.go
```diff
@@ -93,7 +93,7 @@ func ValidateInstanceName(instanceName string) error {
...
-	if len(name) > 40 || !isValidName(name) {
+	if len(name) < 2 || len(name) > 40 || !isValidName(name) {
 		return fmt.Errorf("invalid snap name: %q", name)
```
zyga: I find it funny that obscure C code has better error messages than the main go code. Not something for this PR to fix though.

PR #6159, opened 2018-11-15
cmd/snap-confine: handle mounted shared /run/snapd/ns

zyga:
The code that handles special sharing mode of /run/snapd/ns did cope
with /run/snapd/ns being a private mount point and not being a mount
point at all. When it was a mount point but one with shared mount event
propagation all kinds of hell would break loose. This patch fixes this.

The existing unit tests were nixed and replaced with a much more
readable and comprehensive spread test.

Fixes: https://bugs.launchpad.net/snapd/+bug/1803542

### DIFF
--- a/tests/regression/lp-1803542/task.yaml
+++ b/tests/regression/lp-1803542/task.yaml
@@ -1,0 +1,43 @@
+summary: Regression test for https://bugs.launchpad.net/snapd/+bug/1803542
+environment:
+    VARIANT/gone: gone
+    VARIANT/present: present 
+    VARIANT/shared: shared
+    VARIANT/private: private 
+prepare: |
+    #shellcheck source=tests/lib/snaps.sh
+    . "$TESTSLIB/snaps.sh"
+    install_local test-snapd-sh
+    #shellcheck source=tests/lib/dirs.sh
+    . "$TESTSLIB/dirs.sh"
+    # Ensure that every snap namespace is discarded.
+    for name in $(snap list | awk '!/Name/ {print $1}'); do
+        "$LIBEXECDIR"/snapd/snap-discard-ns "$name"
+    done
+    if findmnt --noheadings /run/snapd/ns >/dev/null; then
+        umount /run/snapd/ns
+    fi
+    ! findmnt --noheadings /run/snapd/ns
+execute: |
+    case "$VARIANT" in
+       absent)
+            rmdir /run/snapd/ns
+            ;;
+       present)
+            mkdir -p /run/snapd/ns
+            ;;
+       shared)
+            mkdir -p /run/snapd/ns
+            mount --bind /run/snapd/ns /run/snapd/ns
+            mount --make-shared /run/snapd/ns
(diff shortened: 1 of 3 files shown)


tests/regression/lp-1803542/task.yaml
```diff
@@ -0,0 +1,45 @@
...
+    for name in $(snap list | awk '!/Name/ {print $1}'); do
+        "$LIBEXECDIR"/snapd/snap-discard-ns "$name"
+    done
+    if findmnt --noheadings /run/snapd/ns >/dev/null; then
+        umount /run/snapd/ns
+    fi
+    ! findmnt --noheadings /run/snapd/ns
+execute: |
+    case "$VARIANT" in
+       absent)
```
zyga: Oh, this should say "gone"

bboozzoo:
```
+ /usr/lib/snapd/snap-discard-ns test-snapd-sh
+ findmnt --noheadings /run/snapd/ns
+ umount /run/snapd/ns
umount: /run/snapd/ns: target is busy
        (In some cases useful info about processes that
         use the device is found by lsof(8) or fuser(1).)
```

zyga:
I saw the failure last night, I need to look at that.

zyga:
I reproduced the bug, looks like something didn't clean up after itself in a prior test.

zyga:
This is blocked by a bug fixed by https://github.com/snapcore/snapd/pull/6203

tests/regression/lp-1803542/task.yaml
```diff
@@ -0,0 +1,53 @@
...
+            mkdir -p /run/snapd/ns
+            mount --bind /run/snapd/ns /run/snapd/ns
+            mount --make-shared /run/snapd/ns
+            # verify mount tags to ensure the mount point has shared propagation (format shared:<nnnn>)
+            test "$(grep '/snapd/ns /run/snapd/ns' < /proc/self/mountinfo | awk '{print $7}' | cut -d : -f 1)" = shared
+            ;;
+        private)
+            mkdir -p /run/snapd/ns
+            mount --bind /run/snapd/ns /run/snapd/ns
+            mount --make-private /run/snapd/ns
```
stolowski: Should we do anything about these mount points in `restore:`?

tests/regression/lp-1803542/task.yaml
```diff
@@ -0,0 +1,53 @@
+summary: Regression test for https://bugs.launchpad.net/snapd/+bug/1803542
+environment:
+    VARIANT/absent: absent
+    VARIANT/present: present
+    VARIANT/shared: shared
+    VARIANT/private: private
```
stolowski: Nice

tests/regression/lp-1803542/task.yaml
zyga: Not really, snap-confine restores the correct setup on each execution. We don't have to unmount them. The code here is what snap-confine does internally.

tests/regression/lp-1803542/task.yaml
zyga: Yeah, worked better than the unit test :)

PR #6160, opened 2018-11-15
nvidia, interfaces/builtin: OpenCL fixes

bboozzoo:
Two patches related to OpenCL and nvidia. The first one allows picking up `libnvidia-opencl.so` from the host. The second one opens `/etc/OpenCL/vendor` for reading in the `opengl` interface. This is needed when one wants to reasonably implement OpenCL support with nvidia.


jdstrand:
This was merged before I had a chance to review, but LGTM.

PR #6161, opened 2018-11-15
tests: new test suite to run snapd tests on a google remote instance

sergiocazzolato:
It is needed to run with -debug because so far there is not sa.json
available to fully automate the process. So far, it is needed to
manually login to gce by running the next command:
> gcloud auth application-default login

Example to run a test:
SPREAD_RUN_PARAMS="-repeat 50 -debug"
SPREAD_RUN_TESTS="google:ubuntu-16.04-64:tests/main/find-private" spread
-debug google:fedora-28-64:tests/remote/run-task


sergiocazzolato:
It is gonna be closed until we have a sa.json to be used

PR #6164, opened 2018-11-16
wrappers: rename the desktop file to their apps

mvo5:
Followup for https://github.com/snapcore/snapd/pull/6156

This renames the desktop files instead of blindly trusting the name that comes from the snap. 

Only 5d46f8c08c077a3bdfdf4243af50603aa4ca28d3 is new in this PR the other bits are from the previous PR.

### DIFF
--- a/wrappers/desktop.go
+++ b/wrappers/desktop.go
@@ -190,6 +190,21 @@ func updateDesktopDatabase(desktopFiles []string) error {
 	return nil
 }
 
+// desktopPrefix returns the prefix string for the desktop files that
+// belongs to the given snapInstance. We need to do something custom
+// here because a) we need to be compatible with the world before we had
+// parallel installs b) we can't just use the usual "_" parallel installs
+// separator because that is already used as the separator between snap
+// and desktop filename.
+func desktopPrefix(s *snap.Info) string {
+	if s.SnapName() == s.InstanceName() {
+		return s.SnapName()
+	}
+	// we cannot use the usual "_" separator because that is also used
+	// to separate "$snap_$desktopfile"
+	return fmt.Sprintf("%s+%s", s.SnapName(), s.InstanceKey)
+}
+
 // AddSnapDesktopFiles puts in place the desktop files for the applications from the snap.
 func AddSnapDesktopFiles(s *snap.Info) (err error) {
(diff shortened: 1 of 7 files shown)


wrappers/desktop.go
```diff
@@ -190,6 +190,47 @@ func updateDesktopDatabase(desktopFiles []string) error {
...
+		if bytes.HasPrefix(bline, []byte("Exec=")) {
+			snapApp, _, err = rewriteExecLine(s, df, string(bline))
+			if err != nil {
+				return "", err
+			}
+		}
+	}
+
+	for i := 0; i < 1000; i++ {
+		dfBaseName := fmt.Sprintf("%s_%s_%v.desktop", desktopPrefix(s), snapApp, i)
```
bboozzoo: I think this may end up a little confusing in the end. For instance when manipulating mime handlers via `xdg-mime` you need to use desktop file names. If you have 2 desktop files, say `foo.desktop` and `foo-some-handler.desktop`, both for app `foo` they would end up being named `snap_foo_0.desktop` and `snap_foo_1.desktop`. IIRC `filepath.Glob` (as used by `AddSnapDesktopFiles`) result order is not deterministic, so there's a chance you'd end up with different names for each install.

zyga:
It seems that at least some tests need adjusting:
```
+ snap install --dangerous ./basic-desktop_1.0_all.snap
basic-desktop 1.0 installed
+ diff -u /dev/fd/63 -
++ head -n4 /var/lib/snapd/desktop/applications/basic-desktop_echo.desktop
head: cannot open '/var/lib/snapd/desktop/applications/basic-desktop_echo.desktop' for reading: No such file or directory
--- /dev/fd/63	2018-11-16 11:00:22.329974251 +0000
+++ -	2018-11-16 11:00:22.338171176 +0000
@@ -0,0 +1,4 @@
+[Desktop Entry]
+Name=Echo
+Comment=It echos stuff
+Exec=env BAMF_DESKTOP_FILE_HINT=/var/lib/snapd/desktop/applications/basic-desktop_echo.desktop /snap/bin/basic-desktop.echo
-----
```

mvo5:
Closing until there is clear concensus

PR #6167, opened 2018-11-18
cmd/snap-confine: allow mapping more libc shards

zyga:
Libc seems to split into ever-growing list of shared libraries,
unfortunately we need to know about all of them. It was reported that
the latest update of openSUSE tumbleweed snap-confine fails to start
because it cannot map libmount due to self-imposed apparmor confinement.

    type=AVC msg=audit(1542580432.818:874): apparmor="DENIED"
    operation="file_mmap" profile="/usr/lib/snapd/snap-confine"
    name="/usr/lib64/libmount.so.1.1.0" pid=63751 comm="snap-confine"
    requested_mask="m" denied_mask="m" fsuid=0 ouid=0

While investigating this I noticed we also link to (but do not allow
mapping libblkid and libuuid. This patch addresses all three missing
shared libraries.

Fixes: https://bugs.launchpad.net/snapd/+bug/1803903


jdstrand:
The patch is fine as far as apparmor rules go. One of the reasons we are specific with the libraries though is because we are trying to limit the attack surface available to this setuid application. I suggest only adding libblkid and libuuid after they are used. +1 for libmount, -.1 to libblkid and libuuid.

zyga:
Thanks, I'll do that @jdstrand

PR #6168, opened 2018-11-19
spread, tests, snap-mgmt: add CentOS support

bboozzoo:
Add CentOS 7.5 to the spread suite.

Apply tweaks to `snap-mgmt` to discard the mount ns before we remote the mounts, but after stopping snap services. This was a partial workaround for https://access.redhat.com/articles/3128691

Add `data/sysctl/99-snap.conf` which enables lazy mount detach. 

@Conan-Kudo can you please take a look at the packaging bits? Could we apply that to the EPEL package?


Conan-Kudo:
@bboozzoo We might want to give that sysctl file a name that indicates it's rhel7 specific, so people don't think it should be generally installed at a glance.

Conan-Kudo:
@bboozzoo Is the workaround necessary if the sysctl is applied already? I would think not?

bboozzoo:
> @bboozzoo Is the workaround necessary if the sysctl is applied already? I would think not?

Let me drop the `snap-mgmt` patch locally and run the whole test suite again.

Conan-Kudo:
@bboozzoo Also, if we enable the "CR" repository, we'll have the EL7.6 content available in our CI. So, we don't need to disable SELinux for CentOS.

bboozzoo:
@Conan-Kudo I tried that earlier and got this:
```
Error: Package: snapd-selinux-2.36-2.el7.noarch (epel-testing)
           Requires: selinux-policy-base >= 3.13.1-229.el7_6.5
           Installed: selinux-policy-targeted-3.13.1-229.el7.noarch (@cr)
               selinux-policy-base = 3.13.1-229.el7
```
Is the package in RHEL proper built with different `%dist` ?

tests/lib/reset.sh
```diff
@@ -87,9 +88,11 @@ reset_classic() {
 
         # wait for snapd listening
         EXTRA_NC_ARGS="-q 1"
-        if [[ "$SPREAD_SYSTEM" = fedora-* || "$SPREAD_SYSTEM" = amazon-* ]]; then
-            EXTRA_NC_ARGS=""
-        fi
+        case "$SPREAD_SYSTEM" in
```
zyga: +1

tests/main/classic-ubuntu-core-transition/task.yaml
```diff
@@ -1,10 +1,10 @@
...
 
-# we never test on core because the transition can only happen on "classic"
-# we disable on ppc64el because the downloads are very slow there
-# Fedora, openSUSE and Arch are disabled at the moment as there is something
-# fishy going on and the snapd service gets terminated during the process.
-systems: [-ubuntu-core-*, -ubuntu-*-ppc64el, -fedora-*, -opensuse-*, -ubuntu-*-i386, -arch-*, -amazon-*]
+# we never test on core because the transition can only happen on "classic" we
+# disable on ppc64el because the downloads are very slow there Fedora, openSUSE,
+# Arch, CentOS are disabled at the moment as there is something fishy going on
+# and the snapd service gets terminated during the process.
```
zyga: Curious to know what happened here, no need to block on this though.

tests/main/confinement-classic/task.yaml
```diff
@@ -11,17 +11,22 @@ details: |
 prepare: |
     #shellcheck source=tests/lib/dirs.sh
     . "$TESTSLIB"/dirs.sh
-    if [[ "$SPREAD_SYSTEM" == fedora-* || "$SPREAD_SYSTEM" == arch-* ]]; then
-        # although classic snaps do not work out of the box on fedora,
-        # we still want to verify if the basics do work if the user
-        # symlinks /snap to $SNAP_MOUNT_DIR themselves
-        ln -sf $SNAP_MOUNT_DIR /snap
-    fi
+    case "$SPREAD_SYSTEM" in
```
zyga: This really feels like a helper is needed. Feel free to do in a follow up if you want.

```
. "$TESTSLIB/classic.sh"
enable_classic_support
```

tests/main/confinement-classic/task.yaml
```diff
@@ -11,17 +11,22 @@ details: |
...
+            ln -sf $SNAP_MOUNT_DIR /snap
+            ;;
+    esac
+
 
 restore: |
-    if [[ "$SPREAD_SYSTEM" == fedora-* || "$SPREAD_SYSTEM" == arch-* ]]; then
-        rm -f /snap
-    fi
+    case "$SPREAD_SYSTEM" in
```
zyga: Another thing for the helper to handle.

Conan-Kudo:
@bboozzoo I think that's actually a bug in CentOS, because custom DistTag changes in the release are supposed to be mimicked by CentOS as well.

bboozzoo:
@Conan-Kudo let's take it to the forum https://forum.snapcraft.io/t/snapd-updates-in-fedora-epel-for-enterprise-linux/8310/7

PR #6173, opened 2018-11-19
sanity: extend the kernel version check to cover CentOS/RHEL kernels

bboozzoo:
Extend the check to cover kernel versions in CentOS/RHEL 7.x. Probe for a known
sysctl that needs to be enabled.

### DIFF
--- a/sanity/version.go
+++ b/sanity/version.go
@@ -20,9 +20,13 @@
 package sanity
 
 import (
+	"bytes"
 	"fmt"
+	"io/ioutil"
+	"path/filepath"
 	"strings"
 
+	"github.com/snapcore/snapd/dirs"
 	"github.com/snapcore/snapd/logger"
 	"github.com/snapcore/snapd/osutil"
 	"github.com/snapcore/snapd/release"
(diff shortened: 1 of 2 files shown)


sanity/version.go
```diff
@@ -33,21 +37,65 @@ func init() {
 	checks = append(checks, checkKernelVersion)
 }
 
+// supportsMayDetachMounts checks whether a RHEL 7.4+ specific kernel knob is present
+// and set to proper value
+func supportsMayDetachMounts(kver string) error {
+	p := filepath.Join(dirs.GlobalRootDir, "/proc/sys/fs/may_detach_mounts")
+	value, err := ioutil.ReadFile(p)
+	if err != nil {
+		return fmt.Errorf("cannot read fs.may_detach_mounts state: %v", err)
```
zyga: Nitpick: I would use the real path we tried to read here.

sanity/version.go
```diff
@@ -33,21 +37,65 @@ func init() {
...
-		kver := osutil.KernelVersion()
-		// a kernel version looks like this: "4.4.0-112-generic" and
-		// we are only interested in the bits before the "-"
-		kver = strings.SplitN(kver, "-", 2)[0]
-		cmp, err := strutil.VersionCompare(kver, "3.13.0")
+	if !release.OnClassic {
+		return nil
+	}
+
+	switch release.ReleaseInfo.ID {
```
zyga: This was bound to happen :)

+1

sanity/version.go
```diff
@@ -33,21 +37,65 @@ func init() {
...
+				return nil
+			}
+			if cmp <= 0 {
+				return fmt.Errorf("you need to reboot into a 4.4 kernel to start using snapd")
+			}
+		}
+	case "rhel", "centos":
+		// check for kernel tweaks on RHEL/CentOS 7.5+
+		// CentoS 7.5 has VERSION_ID="7", RHEL 7.6 has VERSION_ID="7.6"
+		if release.ReleaseInfo.VersionID == "" || release.ReleaseInfo.VersionID[0] != '7' {
```
zyga: We will have a RHEL 70 problem here. Could we please use strings.HasPrefix(release.ReleaseInfo.VersionID, "7.")
```suggestion
		if release.ReleaseInfo.VersionID == "" || strings.HasPrefix(release.ReleaseInfo.VersionID, "7.") {
```

zyga: As a separate question, what is the point of the empty version ID test?

bboozzoo: Maybe I should update the comment under the `if` block. CentOS uses `VERSION_ID="7"`since 7.0 until now. It's enough to know that the first rune is `'7'`, since 6 lacks systemd, and RHEL 8 beta seems fine.

sanity/version.go
bboozzoo: How about I rephrase that to `cannot read the value of fs.may_detach_mounts kernel parameter: %v`?

PR #6174, opened 2018-11-19
many: add snap debug refresh-catalogs

zyga:
This command can be used to force the refresh of the "catalogs", namely
some of the store side meta-data collection that is cached locally for
performance.

The command is used to improve one of the existing tests that fails on
timeout if the store responds with a 403, like it does now.

### DIFF
--- a/daemon/api.go
+++ b/daemon/api.go
@@ -923,6 +923,7 @@ var (
 	snapstateInstall           = snapstate.Install
 	snapstateInstallPath       = snapstate.InstallPath
 	snapstateRefreshCandidates = snapstate.RefreshCandidates
+	snapstateRefreshCatalogs   = snapstate.RefreshCatalogs
 	snapstateTryPath           = snapstate.TryPath
 	snapstateUpdate            = snapstate.Update
 	snapstateUpdateMany        = snapstate.UpdateMany
@@ -2549,8 +2550,12 @@ func postDebug(c *Command, r *http.Request, user *auth.UserState) Response {
 			}
 		}
 		sort.Strings(status.Unreachable)
-
 		return SyncResponse(status, nil)
+	case "refresh-catalogs":
+		if err := snapstateRefreshCatalogs(st); err != nil {
+			return InternalError("cannot refresh catalogs: %s", err)
+		}
+		return SyncResponse(true, nil)
 	default:
 		return BadRequest("unknown debug action: %v", a.Action)
 	}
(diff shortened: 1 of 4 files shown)


zyga:
I think this works as intended:
```
2018-11-19 19:38:14 Error executing google:ubuntu-18.10-64:tests/main/apt-hooks : 
-----
+ echo 'Ensure we have a snap catalog in our cache'
Ensure we have a snap catalog in our cache
+ snap debug refresh-catalogs
error: cannot refresh catalogs: cannot decode new commands catalog: got
       unexpected HTTP status code 403 via GET to
       "https://api.snapcraft.io/api/v1/snaps/names?confinement=strict%2Cclassic"
```

This is much better than long chain of sleep calls.

tests/main/apt-hooks/task.yaml
```diff
@@ -13,9 +13,8 @@ debug: |
     
 execute: |
     echo "Ensure we have a snap catalog in our cache"
-    while ! test -s /var/cache/snapd/commands.db; do
```
mvo5: I think we need to keep a test that checks if we do the catalog update on startup.

overlord/snapstate/catalogrefresh.go
```diff
@@ -79,6 +79,10 @@ func (r *catalogRefresh) Ensure() error {
 	return err
 }
 
+func RefreshCatalogs(st *state.State) error {
```
chipaca: eh, you might as well just expose `refreshCatalogs`, and do `snapstateRefreshCatalogs(st, snapstate.Store(st))` in `daemon`

tests/main/apt-hooks/task.yaml
chipaca: that could be done as a unit test (i think we don't have it right now)

daemon/api.go
```diff
@@ -2549,8 +2550,12 @@ func postDebug(c *Command, r *http.Request, user *auth.UserState) Response {
 			}
 		}
 		sort.Strings(status.Unreachable)
-
 		return SyncResponse(status, nil)
+	case "refresh-catalogs":
+		if err := snapstateRefreshCatalogs(st); err != nil {
+			return InternalError("cannot refresh catalogs: %s", err)
```
chipaca: you could actually return a much richer error (there are a bunch of `switch err {` in `api.go` for inspiration). It might not be worth it for a debug command though...

zyga:
Let's return to this later.

mvo5:
I close this one for now,, please reopen once Johns comments and the conflicts are addressed.

PR #6175, opened 2018-11-20
tests: make tinyproxy support systemd notify

mvo5:
The tinyproxy has a bit of a startup delay (python init and all that).
This means that sometimes in tests the proxy is not fully ready when
the test tries to talk to it. This PR makes the proxy a "notify"
service that only notifies after the proxy is ready.

### DIFF
--- a/tests/lib/tinyproxy/tinyproxy.py
+++ b/tests/lib/tinyproxy/tinyproxy.py
@@ -5,6 +5,7 @@
 # Ported to py3 and modified to remove the bits we don't need
 # and modernized.
 
+import os
 import http.server
 import select
 import socket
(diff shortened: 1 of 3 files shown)


tests/lib/tinyproxy/tinyproxy.py
```diff
@@ -111,9 +112,22 @@ def _read_write(self, soc, max_idling=20):
     do_PUT  = do_GET
     do_DELETE=do_GET
 
+
+def maybe_sd_notify(s: str) -> None:
```
zyga: Do you think we should start having snapd.tests package where we can stash things like that?

PR #6178, opened 2018-11-21
snap,snap-exec: add SNAP_DEBUG_START_TIME environment

mvo5:
This helps analyzing the startup overhead that snap-{run,confine,exec}
impose on top of the actual application. Its important to measure
this to know what part of the startup is the snapcraft wrappers and
what part is our own namespace setup.

### DIFF
--- a/cmd/snap-exec/main.go
+++ b/cmd/snap-exec/main.go
@@ -23,8 +23,10 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/jessevdk/go-flags"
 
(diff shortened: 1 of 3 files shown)


cmd/snap-exec/main.go
```diff
@@ -214,6 +216,13 @@ func execApp(snapApp, revision, command string, args []string) error {
 
 	fullCmd = append(absoluteCommandChain(app.Snap, app.CommandChain), fullCmd...)
 
+	if osutil.GetenvBool("SNAP_DEBUG_START_TIME") {
+		if runTimeStart, err := strconv.ParseInt(os.Getenv("SNAP_RUN_TIME_START"), 10, 64); err == nil {
+			d := time.Now().Sub(time.Unix(0, runTimeStart))
+			fmt.Fprintf(os.Stderr, "Seconds until exec %q: %v\n", fullCmd, d.Seconds())
```
bboozzoo: With command chains `fullCmd` can get quite long, pushing duration far to the right. How about this format:
```
fmt.Fprintf("command: %q\ntime until exec: %v\n", fullCmd, d)
```
BTW `time.Duration.String()` produces nice, human readable output, but I'm fine if you prefer using `d.Seconds()`.

zyga:
This is failing in tests:
```
+ cat stderr
/usr/sbin/strace: Can't stat 'invalid': No such file or directory
error: exit status 1
```

stolowski:
The MATCH expression needs a little tweak (remove the colon):
```
+ MATCH 'Seconds until exec: '
Seconds until exec ["/snap/test-snapd-tools/x1/bin/echo" "foo"]: 0.010493841
```

mvo5:
WE can probbly kill this in favour of https://github.com/snapcore/snapd/pull/6185

PR #6179, opened 2018-11-21
snap: epoch lists must contain no duplicate entries

chipaca:
Without this, `{[0,0]}` would be considered a valid epoch. This puts
an end to that.

### DIFF
--- a/snap/epoch.go
+++ b/snap/epoch.go
@@ -22,7 +22,6 @@ package snap
 import (
 	"encoding/json"
 	"fmt"
-	"sort"
 	"strconv"
 
 	"github.com/snapcore/snapd/logger"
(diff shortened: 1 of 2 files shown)


snap/epoch.go
```diff
@@ -342,6 +337,18 @@ func (z *uint32slice) UnmarshalJSON(bs []byte) error {
 	return nil
 }
 
+func isIncreasing(z []uint32) bool {
+	if len(z) == 0 {
```
bboozzoo: Since you are already doing the check:
```
if len(z) < 2 { return true }
```

chipaca: heh, good point. I'll tweak it in this PR if there are other related changes, otherwise I'll slide it in to the checker pr (upcoming).

PR #6182, opened 2018-11-21
Revert "cmd/snap-confine: don't allow mapping lib{uuid,blkid}"

zyga:
It turns out that without ability to map the two libraries the
permissions snap-confine no longer starts on current snapshots of
openSUSE Tumbleweed.

Fixes: https://bugs.launchpad.net/snapd/+bug/1803903

This reverts commit 38df5f9d2ac93a85258491e31346f1c8e0dd3684.


jdstrand:
This was merged before I could take a look at it, but LGTM.

PR #6183, opened 2018-11-21
sanity, spread, tests: add CentOS (2.36)

bboozzoo:
Backports of #6168 and #6173 to 2.36

(cc @Conan-Kudo )


sanity/version.go
```diff
@@ -33,21 +37,65 @@ func init() {
...
+	case "rhel", "centos":
+		// check for kernel tweaks on RHEL/CentOS 7.5+
+		// CentoS 7.5 has VERSION_ID="7", RHEL 7.6 has VERSION_ID="7.6"
+		if release.ReleaseInfo.VersionID == "" || release.ReleaseInfo.VersionID[0] != '7' {
+			return nil
+		}
+		fullKver := osutil.KernelVersion()
+		// kernel version looks like this: "3.10.0-957.el7.x86_64"
+		kver := strings.SplitN(fullKver, "-", 2)[0]
+		cmp, err := strutil.VersionCompare(kver, "3.18.0")
```
Conan-Kudo: This logic looks like we'd never consider a CentOS kernel to be valid?

bboozzoo: We know that the behavior was changed in 3.18. All earlier versions, including stock CentOS kernels (all on 3.10x), will be verified. If they ever update to something more recent (unlikely), or the host is running a newer custom build (why not?), or it's one of the alternative architectures (RHEL on ppc64el?), the check will not be applied.

Conan-Kudo: @bboozzoo So this won't block CentOS 7 systems running the stock 3.10.x kernel?

bboozzoo: It will. That's why we're shipping the sysctl conf file as part of packaging. In case someone removes it, or builds a half-decent package out of the source tree, they will get a nice error message in the logs. The sanity check just verifies that the knob has been set to the proper value.

Conan-Kudo: Ah, so it does if the knob isn't set?

bboozzoo: Yes.

packaging/fedora/snapd.spec
```diff
@@ -587,12 +592,18 @@ pushd ./data
               SYSTEMDSYSTEMUNITDIR="%{_unitdir}" \
               SNAP_MOUNT_DIR="%{_sharedstatedir}/snapd/snap" \
               SNAPD_ENVIRONMENT_FILE="%{_sysconfdir}/sysconfig/snapd"
+popd
+
+%if 0%{?rhel} == 7
+# Install kernel tweaks
+# See: https://access.redhat.com/articles/3128691
+install -m 644 -D data/sysctl/rhel7-snap.conf %{buildroot}%{_sysconfdir}/sysctl.d/99-snap.conf
```
Conan-Kudo: We should install to `%_sysctldir` instead of `/etc/sysctl.d`.

Conan-Kudo:
@bboozzoo I'm getting this failure when backporting this patch on top of 2.36.1:

```
/usr/lib/golang/pkg/tool/linux_amd64/compile -o $WORK/b062/_pkg_.a -trimpath $WORK/b062 -shared -p github.com/snapcore/snapd/daemon -complete -installsuffix shared -buildid XZaYVkh8Fr0dz0KVuXyX/XZaYVkh8Fr0dz0KVuXyX -goversion go1.11.2 -D "" -importcfg $WORK/b062/importcfg -pack ./api.go ./api_json.go ./api_snapshots.go ./daemon.go ./response.go ./snap.go ./ucrednet.go
# github.com/snapcore/snapd/daemon
src/github.com/snapcore/snapd/daemon/daemon.go:341:40: too many arguments in call to activation.Listeners
        have (bool)
        want ()
```

Conan-Kudo:
@bboozzoo It did, to version 18: https://koji.fedoraproject.org/koji/buildinfo?buildID=1164158

Conan-Kudo:
This also happened in F29, too...

bboozzoo:
> This also happened in F29, too...
Thought this doesn't happen in released versions, unless there's a really good reason to do so, but the changelog is rather brief.

sanity/version.go
```diff
@@ -33,21 +37,65 @@ func init() {
...
 		if err != nil {
 			logger.Noticef("cannot check kernel: %v", err)
 			return nil
 		}
-		if cmp <= 0 {
-			return fmt.Errorf("you need to reboot into a 4.4 kernel to start using snapd")
+		if cmp < 0 {
+			// pre 3.18 kernels here
+			if idx := strings.Index(fullKver, ".el7."); idx == -1 {
```
mvo5: You could have used `strings.Contains()` here, no? But its fine :)

PR #6191, opened 2018-11-22
cmd/snap: fix missing newline in "snap keys" error message

stolowski:
While testing if a fix for an old bug for snap keys landed, I noticed the error message is missing a newline.

### DIFF
--- a/cmd/snap/cmd_keys.go
+++ b/cmd/snap/cmd_keys.go
@@ -66,7 +66,7 @@ func outputJSON(keys []Key) error {
 
 func outputText(keys []Key) error {
 	if len(keys) == 0 {
-		fmt.Fprintf(Stdout, "No keys registered, see `snapcraft create-key`")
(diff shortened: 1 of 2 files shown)


cmd/snap/cmd_keys.go
```diff
@@ -66,7 +66,7 @@ func outputJSON(keys []Key) error {
 
 func outputText(keys []Key) error {
 	if len(keys) == 0 {
-		fmt.Fprintf(Stdout, "No keys registered, see `snapcraft create-key`")
+		fmt.Fprintf(Stdout, "No keys registered, see `snapcraft create-key`\n")
```
chipaca: hm. I think in most other places this sort of message goes to stderr. Could you change it to `fmt.Fprintln(Stderr, ...)` instead?

zyga:
+1, I also found this in my tumbleweed branch but didn't propose separately. Thank you :)

PR #6193, opened 2018-11-22
spread: drop Fedora 27, add Fedora 29

bboozzoo:
Add Fedora 29, while at it drop Fedora 27.


zyga:
I think this is blocked on the repackaging design bug in our test suite. We cannot land this without addressing that.

zyga:
The remark is no longer a blocker because F29 is manual and thus won't break anything. We still need to change the way we rebuild and repackage snapd snap for tests as it is fundamentally broken.

PR #6196, opened 2018-11-22
many: validate title

chipaca:
Three commits for your consideration:
1. `strutil.ElliptRight` (need I say more?)
2. ellipt titles coming from the store (in json), both for v1 and v2 paths, to 40 codepoints.
3. reject as invalid snaps that have titles >40 codepoints. This affects `snap pack` as well as install etc (but the review tools rejected titles altogether until really recently so that's ok).

[relevant forum topic](https://forum.snapcraft.io/t/title-length-in-snapcraft-yaml-snap-yaml/8625?u=chipaca).

### DIFF
--- a/snap/validate_test.go
+++ b/snap/validate_test.go
@@ -1439,6 +1439,17 @@ func (s *validateSuite) TestValidateDescription(c *C) {
 	c.Check(ValidateDescription(strings.Repeat("x", 4000)), IsNil)
 }
 
+func (s *validateSuite) TestValidateTitle(c *C) {
+	for _, s := range []string{
+		"xx", // boringest ASCII
+		"🐧🐧", // len("🐧🐧") == 8
+		"á", // á (combining)
+	} {
+		c.Check(ValidateTitle(strings.Repeat(s, 21)), ErrorMatches, `title can have up to 40 codepoints, got 42`)
+		c.Check(ValidateTitle(strings.Repeat(s, 20)), IsNil)
+	}
+}
+
 func (s *validateSuite) TestValidatePlugSlotName(c *C) {
 	validNames := []string{
 		"a", "aa", "aaa", "aaaa",
--- a/store/details.go
+++ b/store/details.go
@@ -22,6 +22,7 @@ package store
 import (
 	"github.com/snapcore/snapd/jsonutil/safejson"
 	"github.com/snapcore/snapd/snap"
+	"github.com/snapcore/snapd/strutil"
 )
 
 // snapDetails encapsulates the data sent to us from the store as JSON.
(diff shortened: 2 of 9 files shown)


strutil/strutil.go
```diff
@@ -182,3 +183,18 @@ func CommaSeparatedList(str string) []string {
 	}
 	return filtered
 }
+
+// ElliptRight returns a string that is at most n runes long,
+// replacing the last rune with an ellipsis if necessary. If N is less
+// than 1 it's treated as a 1.
+func ElliptRight(str string, n int) string {
+	if n < 1 {
```
sergiusens: maybe if n > 1 default to the last rune in that string? And return the original size

strutil/strutil.go
```diff
@@ -182,3 +183,18 @@ func CommaSeparatedList(str string) []string {
 	}
 	return filtered
 }
+
+// ElliptRight returns a string that is at most n runes long,
+// replacing the last rune with an ellipsis if necessary. If N is less
```
sergiusens: uppercase N, should be n

store/details.go
```diff
@@ -75,7 +76,10 @@ func infoFromRemote(d *snapDetails) *snap.Info {
 	info.RealName = d.Name
 	info.SnapID = d.SnapID
 	info.Revision = snap.R(d.Revision)
-	info.EditedTitle = d.Title.Clean()
+
+	// https://forum.snapcraft.io/t/title-length-in-snapcraft-yaml-snap-yaml/8625/10
+	info.EditedTitle = strutil.ElliptRight(d.Title.Clean(), 40)
```
zyga: Sanity check: stratal.ElliptRight uses runes, not bytes, right?

strutil/strutil.go
```diff
@@ -182,3 +183,18 @@ func CommaSeparatedList(str string) []string {
...
+func ElliptRight(str string, n int) string {
+	if n < 1 {
+		n = 1
+	}
+	if utf8.RuneCountInString(str) <= n {
+		return str
+	}
+
+	// this is expensive; look into a cheaper way maybe sometime
+	return string([]rune(str)[:n-1]) + "…"
```
zyga: Ah, yes it does :) +1

zyga: Will we have terminal issues with the fancy unicode version of "..."?

snap/validate_test.go
```diff
@@ -1439,6 +1439,17 @@ func (s *validateSuite) TestValidateDescription(c *C) {
 	c.Check(ValidateDescription(strings.Repeat("x", 4000)), IsNil)
 }
 
+func (s *validateSuite) TestValidateTitle(c *C) {
+	for _, s := range []string{
+		"xx", // boringest ASCII
+		"🐧🐧", // len("🐧🐧") == 8
+		"á", // á (combining)
```
mvo5: This one counts as two codepoints?

zyga: Because it is combining. It would be one code point after normalisation.

strutil/strutil.go
chipaca: it's a'ight. On the default linux console (on Ubuntu) that one works. If it's got a different font loaded maybe not :man_shrugging:

PR #6201, opened 2018-11-23
tests: remove test-snapd-with-configure on restore

zyga:
The tests/main/ubuntu-core-gadget-config-defaults test, when executed,
leaves the mount namespace of the test-snapd-with-configure snap. For
some tests that is not a problem so the mistake was left unnoticed for a
long time. For a new test I was writing it does matter though hence the
chase for leftovers in the test suite.


zyga:
Eh, the snap is not removable. Let's see what we can do instead.

PR #6202, opened 2018-11-23
tests: restore in restore, not in prepare

zyga:
The test suite is not doing any restore logic in per-test-case restore
section. Instead that is done in the front of the per-test-case prepare
section. The reason for that is unclear but it makes no apparent sense
on a quick inspection.

Putting actual restore in the restore section will open the path to
adding safeguards that system state before prepare is measurably the
same as after restore, that is, that test doesn't affect the system.


tests/lib/prepare-restore.sh
```diff
@@ -488,6 +487,9 @@ restore_project_each() {
         dmesg
         exit 1
     fi
+
+    # Domething is hosing the filesystem so look for signs of that
```
chipaca: are you dure domething id?

PR #6205, opened 2018-11-23
many: staticcheck fixes

chipaca:
This is the result of running

    staticcheck -tests=false $(go list ./... | grep -v '/vendor/')

and then fixing each issue.

One commit per file/issue (with one exception where the same issue was present twice in the same file).

`-tests=false` was necessary because we have way more dubious stuff going on in the tests... :-)

### DIFF
--- a/cmd/snap/cmd_help.go
+++ b/cmd/snap/cmd_help.go
@@ -99,7 +99,11 @@ func (w *manfixer) Write(buf []byte) (int, error) {
 	if !w.done {
 		w.done = true
 		if bytes.HasPrefix(buf, []byte(".TH snap 1 ")) {
-			buf[9] = '8'
+			// io.Writer.Write must not modify the buffer, even temporarily
+			n, _ := Stdout.Write(buf[:9])
+			Stdout.Write([]byte{'8'})
+			m, err := Stdout.Write(buf[10:])
+			return n + m + 1, err
 		}
 	}
 	return Stdout.Write(buf)
--- a/cmd/snapd/main.go
+++ b/cmd/snapd/main.go
@@ -91,7 +91,7 @@ func runWatchdog(d *daemon.Daemon) (*time.Ticker, error) {
 				//       replies with valid data
 				systemd.SdNotify("WATCHDOG=1")
 			case <-d.Dying():
-				break
+				return
 			}
 		}
 	}()
--- a/mkversion.sh
+++ b/mkversion.sh
@@ -33,7 +33,7 @@ if [ "$1" = "--output-only" ]; then
 fi
 
 # If the version is passed in as an argument to mkversion.sh, let's use that.
-if [ ! -z "$1" ]; then
+if [ -n "$1" ]; then
     v="$1"
     o=shell
 fi
--- a/osutil/udev/netlink/conn.go
+++ b/osutil/udev/netlink/conn.go
@@ -112,12 +112,10 @@ func (c *UEventConn) Monitor(queue chan UEvent, errors chan error, matcher Match
 	}
 
 	go func() {
-		loop := true
-		for loop {
+		for {
 			select {
 			case <-quit:
-				loop = false
-				break
+				return
 			default:
 				uevent, err := c.ReadUEvent()
 				if err != nil {
--- a/run-checks
+++ b/run-checks
@@ -218,7 +218,7 @@ if [ "$STATIC" = 1 ]; then
     echo Checking all interfaces have minimal spread test
     missing_interface_spread_test
 
-    if command -v staticcheck >/dev/null || [[ "$(go version)" =~ go1\.(9|1[0-9])\.[0-9] ]]; then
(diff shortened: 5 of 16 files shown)


cmd/snap/cmd_help.go
```diff
@@ -99,7 +99,11 @@ func (w *manfixer) Write(buf []byte) (int, error) {
 	if !w.done {
 		w.done = true
 		if bytes.HasPrefix(buf, []byte(".TH snap 1 ")) {
-			buf[9] = '8'
+			// io.Writer.Write must not modify the buffer, even temporarily
+			n, _ := Stdout.Write(buf[:9])
+			Stdout.Write([]byte{'8'})
+			m, err := Stdout.Write(buf[10:])
```
zyga: Nice! Did the helper find this error?

cmd/snapd/main.go
```diff
@@ -91,7 +91,7 @@ func runWatchdog(d *daemon.Daemon) (*time.Ticker, error) {
 				//       replies with valid data
 				systemd.SdNotify("WATCHDOG=1")
 			case <-d.Dying():
-				break
+				return
```
zyga: Is that something we should back port to stable? CC @mvo5

spdx/parser.go
```diff
@@ -25,13 +25,11 @@ import (
 	"strings"
 )
 
-type operator string
```
zyga: Not sure I understand why this is incorrect.

osutil/udev/netlink/conn.go
```diff
@@ -112,12 +112,10 @@ func (c *UEventConn) Monitor(queue chan UEvent, errors chan error, matcher Match
 	}
 
 	go func() {
-		loop := true
```
zyga: This also looks like a stable back port candidate.

cmd/snapd/main.go
mvo5: I think cherry-picking this makes sense

mkversion.sh
```diff
@@ -33,7 +33,7 @@ if [ "$1" = "--output-only" ]; then
 fi
 
 # If the version is passed in as an argument to mkversion.sh, let's use that.
-if [ ! -z "$1" ]; then
+if [ -n "$1" ]; then
```
mvo5: Haha, I think this one was found by zyga as well.

run-checks
```diff
@@ -217,6 +217,15 @@ if [ "$STATIC" = 1 ]; then
 
     echo Checking all interfaces have minimal spread test
     missing_interface_spread_test
+
+    if command -v staticcheck >/dev/null || ( go version | grep -E 'go1\.(9|1[0-9])\.[0-9]' ); then
```
mvo5: :+1:

zyga:
```
asserts/header_checks.go:199:35: regular expression does not contain any meta characters (SA6004)
cmd/snap/cmd_run.go:519:31: os.SEEK_SET is deprecated: Use io.SeekStart, io.SeekCurrent, and io.SeekEnd.  (SA1019)
cmd/snap/cmd_run.go:532:29: os.SEEK_SET is deprecated: Use io.SeekStart, io.SeekCurrent, and io.SeekEnd.  (SA1019)
interfaces/apparmor/backend.go:411:38: regular expression does not contain any meta characters (SA6004)
osutil/buildid.go:67:14: os.SEEK_SET is deprecated: Use io.SeekStart, io.SeekCurrent, and io.SeekEnd.  (SA1019)
overlord/snapshotstate/snapshotmgr.go:139:36: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (SA1012)
overlord/snapshotstate/snapshotmgr.go:182:59: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (SA1012)
overlord/snapshotstate/snapshotmgr.go:281:43: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (SA1012)
overlord/snapstate/handlers.go:460:40: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (SA1012)
overlord/snapstate/handlers.go:463:40: do not pass a nil Context, even if a function permits it; pass context.TODO if you are unsure about which Context to use (SA1012)
store/store.go:1365:26: os.SEEK_SET is deprecated: Use io.SeekStart, io.SeekCurrent, and io.SeekEnd.  (SA1019)
store/store.go:1384:22: os.SEEK_SET is deprecated: Use io.SeekStart, io.SeekCurrent, and io.SeekEnd.  (SA1019)
store/store.go:1455:27: os.SEEK_SET is deprecated: Use io.SeekStart, io.SeekCurrent, and io.SeekEnd.  (SA1019)

Crushing failure and despair.
```

chipaca:
@zyga stop. Using. 1.11. We can't support both 1.6 and 1.11. Just stop.

chipaca:
I will fix the ones that I can fix (io.Seek* isn't a thing in 1.6), some of those are very reasonable.

chipaca:
The continually failing spread tests make me think there's something deeper going on. Closing for now, will re-propose in smaller batches later.

PR #6208, opened 2018-11-23
run-checks: assorted fixes

zyga:
run-checks: remove useless DEPRECATED check
run-checks: expand newlines in error message
run-checks: use conditional expansion for TRAVIS_BUILD_NUMBER
run-checks: run the whole script with set -eu
run-checks: add newlines between functions


run-checks
```diff
@@ -1,14 +1,13 @@
-#!/bin/sh
+#!/bin/sh -eu
```
mvo5: Thanks for this!

PR #6210, opened 2018-11-23
many: run go fmt ./...

zyga:
Apparently go 1.11.1 changed format style because I get formatting
errors when running ./run-checks --static


chipaca:
I think we need to not run go fmt if go is > 1.10.
You can write that like
```sh
[[ "$(go version)" =~ go1\.(1[1-9])\.[0-9] ]]
```
fwiw

zyga:
Perhaps we should drop this checker then. It prevents a team of collaborators using both go 1.6 and 1.11 from being "clean" for local testing.

mvo5:
Maybe - there will be flip-flop changes in the files that are touched by go1.6 and go1.11. I.e. my editor automatically runs go fmt as pre-save so if two people work on the same file with two version things will always be ugly. I think we need to discuss what to do. Ideally go1.11 would have a compat option.

zyga:
I'll close this for now. I didn't expect to merge it straight away, just wanted to raise awareness.

PR #6212, opened 2018-11-26
tests: measure kernel properties across tests

zyga:
This patch enables measurements of kernel resources across tests:
 - kernel version
 - loaded kernel modules
   - except for changes in binfmt_misc
 - mount table
   - except for mounted binfmt_misc in /proc/sys/fs/binfmt_misc
 - started processes
 - sysctl table
   - except random and counter-like data

Use: https://github.com/zyga/spread/tree/measure-each to run this


zyga:
Let's return to this later.

PR #6213, opened 2018-11-26
interfaces: let NM access ifindex/ifupdown files

alfonsosanchezbeato:
Add some new permissions needed by NetworkManager 1.10: access to device
interface index and to ifupdown state.


alfonsosanchezbeato:
The spread failures seem unrelated to this MP, could somebody please re-trigger CI for this MP?

PR #6214, opened 2018-11-26
spread.yaml: disable _/funcarg test for now

mvo5:
This is a band-aid solution to unblock tests until this is properly fixed by knowledgable people (like @chipaca)


mvo5:
Lets wait until we reverted the reset.sh change from last week.

mvo5:
This may not be needed with https://github.com/snapcore/snapd/pull/6216#pullrequestreview-178325586

PR #6219, opened 2018-11-27
overlord/tests: fix panic in managers test

stolowski:
Mock snap-confine profile of mocked core snaps used in the tests to avoid errors in setup-profiles tasks.

### DIFF
--- a/overlord/managers_test.go
+++ b/overlord/managers_test.go
@@ -25,6 +25,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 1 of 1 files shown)


zyga:
I just started a run of this on my machine. Let's see if I can hit the panic.

overlord/managers_test.go
```diff
@@ -113,6 +115,16 @@ func (ms *mgrsSuite) SetUpTest(c *C) {
 	err := os.MkdirAll(filepath.Dir(dirs.SnapStateFile), 0755)
 	c.Assert(err, IsNil)
 
+	// mock snap-confine profile inside core snap rev 1 and 30, required by setup-profiles tasks.
```
mvo5: Why 1-30 ? Do we really use so many mocked snap-confines?

stolowski: No, we use only 2  revisions: "1" and "30". It's not a 1..30 range :)

mvo5: Uhhh - silly me - sorry for the noise.

mvo5:
This is superseded by https://github.com/snapcore/snapd/pull/6227 now - yes?

zyga:
I think we should close this and instead merge https://github.com/snapcore/snapd/pull/6226 with the rationale given there in the commit message.

PR #6221, opened 2018-11-27
interfaces: return security setup errors (2.36)

zyga:
Snapd used to ignore errors from backend.Setup calls
and within apparmor backend it also used to ignore errors
from a special code path generating snap-confine profile for when
snap-confine is used from {core,snapd} snaps.

We are observing a peculiar error, where snapd is trying to compile and
load the profile for snap-confine, an operation involving executing
apparmor_parser process, but that operation fails because apparmor
parser is killed with SIGTERM. The origin of the signal is unknown
but being vocal about the problem occurring might help us understand the
problem better.

The problem is only visible in log files as a following message:

    Nov 27 10:58:09 nov271039-251208 snapd[24167]: backend.go:312:
    cannot create host snap-confine apparmor configuration: cannot
    reload snap-confine apparmor profile: cannot load apparmor profiles:
    signal: terminated

There is no output from apparmor_parser.

Note: this issue is very likely not new. The existing test setup would
mask it. It is only visible now because snapd in 2.36 release branch
requires more permissions to execute snap-confine than it does with
edge. In the past if this problem had occurred the test machine would
carry on running with the apparmor profile from the edge channel.

### DIFF
--- a/interfaces/apparmor/backend.go
+++ b/interfaces/apparmor/backend.go
@@ -305,11 +305,9 @@ func (b *Backend) Setup(snapInfo *snap.Info, opts interfaces.ConfinementOptions,
 	spec.(*Specification).AddLayout(snapInfo)
 
 	// core on classic is special
-	//
-	// TODO: we need to deal with the "snapd" snap here soon
 	if snapName == "core" && release.OnClassic && release.AppArmorLevel() != release.NoAppArmor {
 		if err := setupSnapConfineReexec(snapInfo); err != nil {
-			logger.Noticef("cannot create host snap-confine apparmor configuration: %s", err)
+			return fmt.Errorf("cannot create host snap-confine apparmor configuration: %s", err)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


zyga:
This has uncovered real errors:
```
- Setup snap "core" (6016) security profiles (cannot setup apparmor for snap "core": cannot create host snap-confine apparmor configuration: cannot reload snap-confine apparmor profile: cannot load apparmor profiles: exit status 1
apparmor_parser output:
AppArmor parser error for /var/lib/snapd/apparmor/profiles/snap-confine.core.6016 in /var/lib/snapd/apparmor/profiles/snap-confine.core.6016 at line 103: Exec condition must begin with '/'.
)
- Setup snap "core" (6016) security profiles (cannot create host snap-confine apparmor configuration: cannot reload snap-confine apparmor profile: cannot load apparmor profiles: exit status 1
apparmor_parser output:
AppArmor parser error for /var/lib/snapd/apparmor/profiles/snap-confine.core.6016 in /var/lib/snapd/apparmor/profiles/snap-confine.core.6016 at line 103: Exec condition must begin with '/'.
```

zyga:
When considering this, please ensure that whatever other errors we may have, the critical property is that snapd should be able to refresh itself. As such breaking in overlord/ifacestate/InterfaceManager.iniitalise is probably not desirable. Perhaps a warning should be added instead.

mvo5:
It looks like this is now superseded by https://github.com/snapcore/snapd/pull/6235 ? Is that correct? If not, please reopen.

PR #4435, opened 2018-01-03
snap: do not leak internal network errors to the user

mvo5:
We currently present a rather ugly error message to the user when the snap store is not available. We will need to do something similar for "cmd_snap_op.go" (I will do a followup).

This addresses LP: 1688720.

PR #6222, opened 2018-11-27
cmd/snap: handle DNS error gracefully

stolowski:
Followup to https://github.com/snapcore/snapd/pull/4435 (and bug https://bugs.launchpad.net/snapd/+bug/1688720).

This change fixes one more case where network issue can cause an ugly error.


cmd/snap/cmd_find.go
```diff
@@ -234,7 +234,7 @@ func (x *cmdFind) Execute(args []string) error {
 	}
 
 	snaps, resInfo, err := x.client.Find(opts)
-	if e, ok := err.(*client.Error); ok && e.Kind == client.ErrorKindNetworkTimeout {
+	if e, ok := err.(*client.Error); ok && (e.Kind == client.ErrorKindNetworkTimeout || e.Kind == client.ErrorKindDNSFailure) {
```
zyga: Looking at this, should we retry on DNS errors everywhere where we retry on timeouts?

daemon/api.go
```diff
@@ -687,6 +688,17 @@ func searchStore(c *Command, r *http.Request, user *auth.UserState) Response {
 	case store.ErrUnauthenticated, store.ErrInvalidCredentials:
 		return Unauthorized(err.Error())
 	default:
+		if e, ok := err.(*url.Error); ok {
```
zyga: Looking at this code makes me want to use https://github.com/pkg/errors - CC @mvo5 and @bboozzoo

tests/main/snap-network-errors/task.yaml
```diff
@@ -0,0 +1,32 @@
...
+# autopkgtest run only a subset of tests that deals with the integration
+# with the distro
+backends: [-autopkgtest]
+
+# no iptables on core18 yet
+systems: [-ubuntu-core-18-*]
+
+restore: |
+    echo "Remove iptables rules"
+    iptables -D OUTPUT -p udp --dport 53 -j REJECT --reject-with icmp-port-unreachable || true
```
zyga: I would feel safer if the restore logic loaded a dump of iptables that was made in prepare.

daemon/api.go
```diff
@@ -687,6 +688,17 @@ func searchStore(c *Command, r *http.Request, user *auth.UserState) Response {
...
+			if neterr, ok := e.Err.(*net.OpError); ok {
```
mvo5: Do we need the same if we want to handle `snap install` gracefully?

zyga:
This is failing with:
```
2018-11-27 16:07:42 Error executing google:ubuntu-16.04-32:tests/main/snap-network-errors : 
-----
+ echo 'Disabling DNS queries'
Disabling DNS queries
+ iptables -I OUTPUT -p udp --dport 53 -j REJECT --reject-with icmp-port-unreachable
+ iptables -I OUTPUT -p tcp --dport 53 -j REJECT --reject-with tcp-reset
+ systemctl start snapd.socket snapd.service
++ snap find test
++ true
+ OUT='error: cannot get nonce from store: Post
       https://api.snapcraft.io/api/v1/snaps/auth/nonces: dial tcp: lookup
       api.snapcraft.io on 169.254.169.254:53: write udp
       10.240.0.37:33842->169.254.169.254:53: write: operation not permitted'
+ iptables -D OUTPUT -p udp --dport 53 -j REJECT --reject-with icmp-port-unreachable
+ iptables -D OUTPUT -p tcp --dport 53 -j REJECT --reject-with tcp-reset
+ MATCH 'error: unable to contact snap store'
+ echo 'error: cannot get nonce from store: Post
       https://api.snapcraft.io/api/v1/snaps/auth/nonces: dial tcp: lookup
       api.snapcraft.io on 169.254.169.254:53: write udp
       10.240.0.37:33842->169.254.169.254:53: write: operation not permitted'
error: cannot get nonce from store: Post
       https://api.snapcraft.io/api/v1/snaps/auth/nonces: dial tcp: lookup
       api.snapcraft.io on 169.254.169.254:53: write udp
       10.240.0.37:33842->169.254.169.254:53: write: operation not permitted
-----
```

stolowski:
The spread test error indicates that network errors may manifest themselves in some other ways, so I need to reconsider that test. Marking blocked for now.

tests/main/snap-network-errors/task.yaml
bboozzoo: Myabe something like:
```
prepare: |
    iptables-save > rules

restore: |
    iptables-restore < rules
```

tests/main/snap-network-errors/task.yaml
```diff
@@ -0,0 +1,32 @@
...
+debug: |
+    echo "iptables rules"
+    iptables -L -n -v
+
+execute: |
+    systemctl stop snapd.{socket,service}
+
+    echo "Disabling DNS queries"
+    iptables -I OUTPUT -p udp --dport 53 -j REJECT --reject-with icmp-port-unreachable
+    iptables -I OUTPUT -p tcp --dport 53 -j REJECT --reject-with tcp-reset
```
bboozzoo: You shouldn't need to do anything with TCP.

tests/main/snap-network-errors/task.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+debug: |
+    echo "iptables rules"
+    iptables -L -n -v
+
+execute: |
+    systemctl stop snapd.{socket,service}
+
+    echo "Disabling DNS queries"
+    iptables -I OUTPUT -p udp --dport 53 -j REJECT --reject-with icmp-port-unreachable
```
bboozzoo: My guess is `-j REJECT` is what's causing `operation not permitted`, have you tried `-j DROP`?

stolowski: Yeah, I tried, that's not it.

daemon/api.go
stolowski: No, afacit from quickly looking, install is different because on the api side we only call `snapstate.Install` and get a taskset (change), and it's a `download` task that will fail. In general we have many different error paths in the code and they all need to be investigated and addressed case-by-case if we want this to be robust everywhere :(

cmd/snap/cmd_find.go
stolowski: Good point, probably, to be investigated - retry logic in httputil would be the candidate to look at, and it's used by `store.Find`. I think it's a material for possible followup though.

tests/main/snap-network-errors/task.yaml
stolowski: Allright, after some more testing it seems a different kind of failure could happen do to forced device auth refresh (I saw that consistently when running on 14.04), so I added `snap refresh` to avoid that.

robert-ancell:
Supported in snapd-glib 1.46 https://github.com/snapcore/snapd-glib/commit/11fb22be1926ec6c9960ea0aadd66cbe36224f12

PR #6224, opened 2018-11-27
interfaces: return security setup errors

zyga:
Snapd used to ignore errors from backend.Setup calls
and within apparmor backend it also used to ignore errors
from a special code path generating snap-confine profile for when
snap-confine is used from {core,snapd} snaps.

We are observing a peculiar error, where snapd is trying to compile and
load the profile for snap-confine, an operation involving executing
apparmor_parser process, but that operation fails because apparmor
parser is killed with SIGTERM. The origin of the signal is unknown
but being vocal about the problem occurring might help us understand the
problem better.

The problem is only visible in log files as a following message:

    Nov 27 10:58:09 nov271039-251208 snapd[24167]: backend.go:312:
    cannot create host snap-confine apparmor configuration: cannot
    reload snap-confine apparmor profile: cannot load apparmor profiles:
    signal: terminated

There is no output from apparmor_parser.

Note: this issue is very likely not new. The existing test setup would
mask it. It is only visible now because snapd in 2.36 release branch
requires more permissions to execute snap-confine than it does with
edge. In the past if this problem had occurred the test machine would
carry on running with the apparmor profile from the edge channel.


zyga:
This is the master version of https://github.com/snapcore/snapd/pull/6221

PR #6225, opened 2018-11-27
tests: fix for failover test on how logs are checked

sergiocazzolato:
Otherwise logs used are since last reboot.

error:
https://api.travis-ci.org/v3/job/460209244/log.txt

### DIFF
--- a/tests/main/failover/task.yaml
+++ b/tests/main/failover/task.yaml
@@ -68,6 +68,9 @@ execute: |
     . "$TESTSLIB"/names.sh
     #shellcheck source=tests/lib/boot.sh
     . "$TESTSLIB"/boot.sh
+    #shellcheck source=tests/lib/journalctl.sh
+    . "$TESTSLIB"/journalctl.sh
+
     if [ "$TARGET_SNAP" = kernel ]; then
         TARGET_SNAP_NAME=$kernel_name
     else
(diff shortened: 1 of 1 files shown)


tests/main/failover/task.yaml
```diff
@@ -87,10 +90,17 @@ execute: |
...
+        if check_journalctl_log "Waiting for system reboot"; then
+            echo "Already waiting for system reboot, exiting..."
+            exit 1
+        fi
+
         # install new target snap
         snap install --dangerous failing.snap
 
-        while ! journalctl -b -u snapd|grep -q "Waiting for system reboot" ; do sleep 1 ; done
+        while ! check_journalctl_log "Waiting for system reboot"; do
```
bboozzoo: Would be a bit easier to understand with the comment:
```
# use journalctl wrapper to grep only the logs collected while the test is running
```

tests/main/failover/task.yaml
```diff
@@ -87,10 +90,17 @@ execute: |
...
 
+        if check_journalctl_log "Waiting for system reboot"; then
+            echo "Already waiting for system reboot, exiting..."
+            exit 1
+        fi
+
         # install new target snap
         snap install --dangerous failing.snap
 
-        while ! journalctl -b -u snapd|grep -q "Waiting for system reboot" ; do sleep 1 ; done
```
bboozzoo: Wonder why `journalctl -b` didn't work as expected. In theory those are the logs from current boot only.

mvo5:
Nice, this eliminates a class of spurious errors we saw, right?

sergiocazzolato:
@bboozzoo, thanks for the review!! could you please take a look again?

PR #6226, opened 2018-11-28
overlord,daemon: mock security backends for testing (2.36)

zyga:
Testing the overlord involves initializing the overlord, which in turn
does the same to each of the managers. The interface manager is
particularly active during the initialization phase. It will initialize
all the security backends, some which deeply interrogate the system .
It will compute and compare the system key. Lastly it will regenerate
security profiles for all the snaps if said profile is mismatching. When
setting up core the apparmor security backend performs special handling
for the snap-confine program running from core or from snapd snaps.

All of that interacts with the system. While we could, with enough
effort, mock it away at a very fine grained level we could just replace
all the real security backends with a test backend. This is easy and has
no consequences because we were not measuring anything about the
interactions of the security backend anyway.

This patch does just that. In addition, now-useless mocking of various
system commands has been removed.

### DIFF
--- a/daemon/daemon_test.go
+++ b/daemon/daemon_test.go
@@ -43,6 +43,7 @@ import (
 	"github.com/snapcore/snapd/dirs"
 	"github.com/snapcore/snapd/logger"
 	"github.com/snapcore/snapd/overlord/auth"
+	"github.com/snapcore/snapd/overlord/ifacestate"
 	"github.com/snapcore/snapd/overlord/snapstate"
 	"github.com/snapcore/snapd/overlord/standby"
 	"github.com/snapcore/snapd/overlord/state"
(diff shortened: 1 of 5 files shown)


zyga:
The patch is now handling all instances of the overlord being constructed. Proposed for master separately.

zyga:
This failed with:
```
+ snap install --dangerous /home/gopath/src/github.com/snapcore/snapd/tests/lib/snaps/test-snapd-xdg-autostart/test-snapd-xdg-autostart_1.0_all.snap
test-snapd-xdg-autostart 1.0 installed
+ snap run test-snapd-xdg-autostart.foo
cannot create temporary directory for /var/lib/snapd mount point: Permission denied
```

PR #6232, opened 2018-11-28
overlord/snapstate: support for pre-remove hook

stolowski:
Support for `pre-remove` hook, executed before stop-snap-services (contrary to `remove` hook which runs later). This addresses https://bugs.launchpad.net/snapd/+bug/1777121

Note, the case where pre-remove hook is not run at all is covered by existing test case (that checks no hooks were run on remove with 2 revisions), so not visible in the diff.


stolowski:
Marking blocked as the bug report and use case need further consideration.

stolowski:
Closing for now, will re-open if we have clarity and decide to revive it.

PR #6235, opened 2018-11-28
overlord,apparmor: new syskey behaviour + non-ignored snap-confine profile errors (2.36)

zyga:
This branch combines https://github.com/snapcore/snapd/pull/6234 with another
patch that makes snap-confine profile failures visible as errors from backend.Setup
and yet another patch that temporarily disables openSUSE Leap 42.3.

This branch cannot land yet, please see the commit messages for detailed description.
It can land once the issue affecting openSUSE Leap 42.3 is fixed.


interfaces/apparmor/backend.go
```diff
@@ -309,7 +309,7 @@ func (b *Backend) Setup(snapInfo *snap.Info, opts interfaces.ConfinementOptions,
 	// TODO: we need to deal with the "snapd" snap here soon
 	if snapName == "core" && release.OnClassic && release.AppArmorLevel() != release.NoAppArmor {
 		if err := setupSnapConfineReexec(snapInfo); err != nil {
-			logger.Noticef("cannot create host snap-confine apparmor configuration: %s", err)
+			return fmt.Errorf("cannot create host snap-confine apparmor configuration: %s", err)
```
mvo5: Just to be on the safe side - I think we should add a spread test that does something like:
```
systemctl stop snapd.socket snapd
mount -o bind /bin/false /sbin/apparmor_parser
systemctl start snapd snapd.socket
journalctl -u snapd | MATCH "cannot regenerate apparmor profile"
snap list | MATCH core
```
to ensure failures in the apparmor (re)generation do not prevent snapd from starting.

One aspect of this change makes me slightly nervous - it may prevent core refreshes. This can be a good thing, i.e. if we would write something that is unusable but we need to be mindful about it. Its one of the few thngs that can prevent a core refresh.

overlord/ifacestate/helpers.go
```diff
@@ -163,6 +165,9 @@ func (m *InterfaceManager) regenerateAllSecurityProfiles() error {
 		}
 	}
 
+	writeSystemKey := true
+	os.Remove(dirs.SnapSystemKeyFile)
```
mvo5: This should not be needed - we only end up in this code path if `profilesNeedRegeneration()` detected a systemKey mismatch. So the change to only write the systemKey if the profile was successful should be sufficient (this is the only place in the code that writes the system key). Or am I missing something?

zyga: Did you read the commit message of the patch adding this?

interfaces/apparmor/backend.go
zyga: +1 but I will do this in https://github.com/snapcore/snapd/pull/6233 where this feature is "landing", then I will back port here to ensure it all works in the 2.36 world.

overlord/ifacestate/helpers.go
mvo5: Thanks, seems worthwhile to add a comment about this into the code though (copy/paste of the relevant bits from the commit message should be fine).

zyga:
This is no longer blocked, it contains all the patches need to unbreak 2.36

interfaces/apparmor/backend.go
stolowski: 6233 doesn't have this change, or was it landed separately and to visible there?

zyga: I can back port all the tests as well. Let me do that now.

zyga: I've back ported all the things from https://github.com/snapcore/snapd/pull/6233

PR #6236, opened 2018-11-29
Staticcheck fixes

chipaca:
Trying this again (maybe the continual test failures weren't this one's fault :-) ).

### DIFF
--- a/asserts/header_checks.go
+++ b/asserts/header_checks.go
@@ -196,6 +196,10 @@ func checkDigest(headers map[string]interface{}, name string, h crypto.Hash) ([]
 	return b, nil
 }
 
+// checkStringListInMap returns the `name` entry in the `m` map as a (possibly nil) `[]string`
+// if `m` has an entry for `name` and it isn't a `[]string`, an error is returned
+// if pattern is not nil, all the strings must match that pattern, otherwise an error is returned
+// `what` is a descriptor, used for error messages
 func checkStringListInMap(m map[string]interface{}, name, what string, pattern *regexp.Regexp) ([]string, error) {
 	value, ok := m[name]
 	if !ok {
--- a/asserts/header_checks.go
+++ b/asserts/header_checks.go
@@ -196,8 +196,6 @@ func checkDigest(headers map[string]interface{}, name string, h crypto.Hash) ([]
 	return b, nil
 }
 
-var anyString = regexp.MustCompile("")
-
 func checkStringListInMap(m map[string]interface{}, name, what string, pattern *regexp.Regexp) ([]string, error) {
 	value, ok := m[name]
 	if !ok {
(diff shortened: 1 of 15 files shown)


osutil/udev/netlink/conn.go
```diff
@@ -112,12 +112,10 @@ func (c *UEventConn) Monitor(queue chan UEvent, errors chan error, matcher Match
...
 
 	go func() {
-		loop := true
-		for loop {
+		for {
 			select {
 			case <-quit:
-				loop = false
-				break
+				return
```
bboozzoo: Nice!

run-checks
```diff
@@ -218,7 +218,7 @@ if [ "$STATIC" = 1 ]; then
     echo Checking all interfaces have minimal spread test
     missing_interface_spread_test
 
-    if command -v staticcheck >/dev/null || [[ "$(go version)" =~ go1\.(9|1[0-9])\.[0-9] ]]; then
+    if command -v staticcheck >/dev/null || ( go version | grep -E 'go1\.(9|1[0-9])\.[0-9]' ); then
```
bboozzoo: You probably meant `grep -E 'go1\.(9|1[0-9])(\.[0-9])?'` since Go 1.10 reported `go version go1.10 linux..`

bboozzoo: Looks good.

cmd/snapd/main.go
```diff
@@ -91,7 +91,7 @@ func runWatchdog(d *daemon.Daemon) (*time.Ticker, error) {
 				//       replies with valid data
 				systemd.SdNotify("WATCHDOG=1")
 			case <-d.Dying():
-				break
+				return
```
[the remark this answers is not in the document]
chipaca: should be, yes. Wondering if systemd considering multiple restarts as failures is because of this, but it shouldn't be.

asserts/header_checks.go
```diff
@@ -196,8 +196,6 @@ func checkDigest(headers map[string]interface{}, name string, h crypto.Hash) ([]
 	return b, nil
 }
 
-var anyString = regexp.MustCompile("")
-
 func checkStringListInMap(m map[string]interface{}, name, what string, pattern *regexp.Regexp) ([]string, error) {
```
mvo5: Maybe a small comment here (and/or below) that `nil` matches any string?

interfaces/apparmor/backend.go
```diff
@@ -408,10 +408,12 @@ func (b *Backend) Remove(snapName string) error {
 
 var (
 	templatePattern = regexp.MustCompile("(###[A-Z_]+###)")
-	attachPattern   = regexp.MustCompile(`\(attach_disconnected,mediate_deleted\)`)
```
mvo5: Haha - nice find!

asserts/header_checks.go
mvo5: Not a blocker though - only add it if you modify this branch anyway.

chipaca:
```suggestion
// checkStringListInMap returns the `name` entry in the `m` map as a (possibly nil) `[]string`
// if `m` has an entry for `name` and it isn't a `[]string`, an error is returned
// if pattern is not nil, all the strings must match that pattern, otherwise an error is returned
// `what` is a descriptor, used for error messages
func checkStringListInMap(m map[string]interface{}, name, what string, pattern *regexp.Regexp) ([]string, error) {
```

cmd/snapd/main.go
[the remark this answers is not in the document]
chipaca: confirmed it isn't :-)

PR #6237, opened 2018-11-29
client, store: don't use store from client (use client from store)

chipaca:
This moves the definition of BuyOptions and BuyResult from store to
client, and inverts the dependency.

The dependency graph goes from

dependency graph before the refactor

to

dependency graph after the refactor

(everything there is probably OK for "client" 🙌

### DIFF
--- a/overlord/snapstate/backend.go
+++ b/overlord/snapstate/backend.go
@@ -25,6 +25,7 @@ import (
 	"golang.org/x/net/context"
 
 	"github.com/snapcore/snapd/asserts"
+	"github.com/snapcore/snapd/client"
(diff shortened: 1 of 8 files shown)


overlord/snapstate/backend.go
```diff
@@ -25,6 +25,7 @@ import (
 	"golang.org/x/net/context"
 
 	"github.com/snapcore/snapd/asserts"
+	"github.com/snapcore/snapd/client"
```
mvo5: This feels slightly strange - makes me wonder if something like `store/buy` would be better but then it would be a tiny tiny package so probably not quite right?

PR #6238, opened 2018-11-29
[WIP] many: add minimal SELinux support, refactor the policy

bboozzoo:
A week long crash course in SELinux gave birth to this larg(-ish) branch that introduces a minimal SELinux support and does a major refactoring to SELinux targeted policy for snapd.

At this point I am able to start/stop snapd cleanly and install/remove snaps, both simple tools and test-snapd-service, cleanly. All of the denials that were logged to audit are gone, meaning stuff was mostly accounted for, fixed, or the policy got extended where needed.

Major changes:
- policy introduces a separate domain for snapd and each type of helper (mount, confine, cli)
- each domain has a separate policy and allows only specific transitions
- there is a separate file type for data in mounted snap images (`snappy_snap_t`, for the lack of a better name)
- snaps are applied a SELinux context on mount, though I am not sure yet, how we are going to upgrade people to the new mount units
- `snap-confine` learned some SELinux tricks, relabeling of `/run/snapd` which caused many issues, and process transitions on exec
- `snap` does relabeling of `~/snap` in case there is data there, but otherwise the policy contains the necessary file transition
- a minimal detection of SELinux was added to snapd `release` package
- sanity checks apply proper SELinux context option when mounting the canary image
- `snapd` also applies a proper SELinux context for generated mount units

Some notes about transitions and starting snaps. When a regular user starts a snap, they are typically in `unconfined_u:unconfined_r:unconfined_t` context. This proceeds without any special transitions or limitations, since the user is already unconfined, and there is no targeted policy to close it down. The snaps that are started by systemd end up transitioning to `system_u:system_r:unconfined_service_t` domain, where they should not raise any more denials. While doing so, the transitions are:
```
init_t -> snappy_cli_t (snap run ..) -> snappy_confine_t -> ..
   ... -> snappy_mount_t (snap-update-ns) and back -> 
   ... -> unconfined_service_t
```
Domains permissions are limited to just the things each tool does. With the split in place, it's way easier to track which piece is doing what.

I will be splitting this up into smaller PRs, but some  patches depend on one another. Specifically, pathches introducing the use of SELinux context labels need the policy to exist before.

On top of this, I used Fedora 29 as dev environment. When transitioning to CentOS 7.6, the policy should be up to date, but it may end up needing some tweaks.

There is a spread test for checking if audit logs are clean while doing basic snap operations coming up.

I'm by no means expert in SELinux, so feedback on the policy and libselinux integration is welcome.

(cc @Conan-Kudo)


data/selinux/snappy.fc
```diff
@@ -18,30 +18,38 @@
 
 
 HOME_DIR/snap(/.*)?			gen_context(system_u:object_r:snappy_home_t,s0)
+/root/snap(/.*)?			gen_context(system_u:object_r:snappy_home_t,s0)
```
Conan-Kudo: Why is this necessary? For the `root` user, `HOME_DIR` should equal `/root`...

data/selinux/snappy.fc
```diff
@@ -18,30 +18,38 @@
...
+/usr/lib/snapd/snap-confine		--	gen_context(system_u:object_r:snappy_confine_exec_t,s0)
+/usr/lib/snapd/snap-update-ns		--	gen_context(system_u:object_r:snappy_mount_exec_t,s0)
+/usr/lib/snapd/snap-discard-ns		--	gen_context(system_u:object_r:snappy_mount_exec_t,s0)
 /usr/lib/snapd/.*  		--	gen_context(system_u:object_r:snappy_exec_t,s0)
 /etc/default/snapd		-- 	gen_context(system_u:object_r:snappy_config_t,s0)
 /lib/systemd/system/snapd.* 	--	gen_context(system_u:object_r:snappy_unit_file_t,s0)
 ')
 
-/var/run/snapd(/.*)?	        --	gen_context(system_u:object_r:snappy_var_run_t,s0)
+/var/run/snapd(/.*)?	        gen_context(system_u:object_r:snappy_var_run_t,s0)
```
Conan-Kudo: Why'd we drop the `--` here? We want to match on everything in here now?

cmd/Makefile.am
```diff
@@ -271,6 +271,18 @@ snap_confine_snap_confine_extra_libs += $(APPARMOR_LIBS)
 endif  # STATIC_LIBAPPARMOR
 endif  # APPARMOR
 
+if SELINUX
+snap_confine_snap_confine_SOURCES += \
+	snap-confine/selinux-support.c \
+	snap-confine/selinux-support.h
+snap_confine_snap_confine_CFLAGS += $(SELINUX_CFLAGS)
+if STATIC_LIBSELINUX
+snap_confine_snap_confine_STATIC += $(shell $(PKG_CONFIG) --static --libs libselinux)
```
zyga: Is linking to selinux statically a safe thing to do?

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,94 @@
...
+#include <selinux/restorecon.h>
+#include <selinux/selinux.h>
+#include <selinux/context.h>
+
+#include "../libsnap-confine-private/utils.h"
+#include "../libsnap-confine-private/string-utils.h"
+
+int sc_selinux_relabel_run_dir(void)
+{
+	if (is_selinux_enabled() < 1) {
```
zyga: https://linux.die.net/man/3/is_selinux_enabled says that `< 0` is an error. Should we die on that case?

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,94 @@
...
+		return 0;
+	}
+
+	int ret = selinux_restorecon("/run/snapd",
+				     SELINUX_RESTORECON_RECURSE |
+				     SELINUX_RESTORECON_IGNORE_MOUNTS |
+				     SELINUX_RESTORECON_XDEV);
+	if (ret == -1) {
```
zyga: Nitpick, most of snap-confine tries to be consistent in using `< 0` for such check, would you mind changing that across the patch please?

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,94 @@
...
+ **/
+int sc_selinux_set_snap_execcon(void)
+{
+	/* die("foo"); */
+	if (is_selinux_enabled() < 1) {
+		debug("selinux not enabled");
+		return 0;
+	}
+
+	char *ctx_str = NULL;
```
zyga: Can you please add a selinux specific cleanup handler that calls `freecon`

This is based on https://linux.die.net/man/3/getcon`

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,94 @@
...
+	if (getcon(&ctx_str) == -1) {
+		die("failed to obtain current process context");
+	}
+	debug("current context: %s", ctx_str);
+
+	context_t ctx = context_new(ctx_str);
```
zyga: Could you please initialise this to NULL, add a selinux specific cleanup handler that calls `context_free`. This is based on http://man7.org/linux/man-pages/man3/context_new.3.html

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,94 @@
...
+	if (ctx == NULL) {
+		die("failed to create context from context string %s", ctx_str);
+	}
+
+	if (sc_streq(context_type_get(ctx), "snappy_confine_t")) {
```
zyga: Nitpick: I would call this `snapd_t` or `snap_confine_t` but definitely not `snappy_confine_t`

Conan-Kudo: The context label matches the convention for the snappy module that governs snapd, snap-confine, and snaps. So, I think that's perfectly fine.

zyga: I just ... `snappy_confine_t` is hideous :)

Conan-Kudo: Well, we could do `snappy_snap_confine_t` if you want...

data/selinux/snappy.fc
wrabcak: @Conan-Kudo, 

This will match also all objects inside the directory /var/run/snapd.  This is good fix.

Conan-Kudo:
@bboozzoo This commit history is really confusing. Could you rebase instead, perhaps?

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,77 @@
...
+#include "../libsnap-confine-private/utils.h"
+#include "../libsnap-confine-private/string-utils.h"
+
+/**
+ * Set security context for the snap
+ *
+ **/
+int sc_selinux_set_snap_execcon(void)
+{
+	/* die("foo"); */
```
stolowski: This comment can go away

bboozzoo:
@Conan-Kudo you'll probably want to take a look at https://github.com/snapcore/snapd/pull/6238/commits/bc95ea5d408b738541bbd387d9e72d2cc48d8358 I've had some trouble with properly handling the downgrade path. I'm afraid we wont' be able to do the transition cleanly without asking the users to reboot.

Conan-Kudo:
@bboozzoo The changes look okay, but there's conflicts with current master that need to be resolved.

Conan-Kudo:
@bboozzoo This needs rebasing again...

bboozzoo:
The branch served it's purpose. The final bit up for a review in #6711.

PR #6242, opened 2018-11-29
overlord/snapstate: use file timestamp to initialize timer

chipaca:
Without this change, the catalog refresh wass done every time snapd
started. Usually not too bad, but in tests or dev it could get
onerous.

This change makes it so that the time of the last catalog refresh is
initialized from the timestamp of the sections file, which is a
sensible lower bound for the last time the catalog refresh was done
(unless the system clock is broken -- against which we protect by
ignoring times that are in the future).

### DIFF
--- a/overlord/snapstate/catalogrefresh.go
+++ b/overlord/snapstate/catalogrefresh.go
@@ -56,8 +56,15 @@ func (r *catalogRefresh) Ensure() error {
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


bboozzoo:
`tests/main/catalog-update` apparently needs to be updated.

mvo5:
Very nice, thank you! I'm sure @sergiocazzolato will also like it as it will speed up tests.

PR #6244, opened 2018-11-29
release: detect too old apparmor_parser

zyga:
This patch re-factors apparmor probing logic so that we can detect older
apparmor parser (with fewer features than required to run snapd in any
way). The API has been streamlined, tests simplified.

The main change is that probing of kernel and parser features happens in
tandem. If some kernel features are present but required parser features
are absent then "level" is set to unusable. This is a new level that
that sits between none and partial.

Most importantly this is the level that is not causing the apparmor
backend to be added to all security backends.

In practical terms this means we run without apparmor on openSUSE Leap
42.3 which in turn fixes a bug that prevented snapd from working
correctly with stricter error reporting during startup phase when system
key is considered.

Extra care has been taken to preserve the property that comparing the system
key does not necessarily need to run apparmor_parser, so that fast path of
"snap run <snap>" remains efficient.

A simple spread test ensures that openSUSE Leap 42.3 is not using
apparmor anymore.

Fixes: https://bugs.launchpad.net/snapd/+bug/1805485

### DIFF
--- a/interfaces/apparmor/backend.go
+++ b/interfaces/apparmor/backend.go
@@ -522,7 +522,8 @@ func addContent(securityTag string, snapInfo *snap.Info, opts interfaces.Confine
 		case "###PROFILEATTACH###":
 			return fmt.Sprintf("profile \"%s\"", securityTag)
 		case "###CHANGEPROFILE_RULE###":
-			for _, f := range parserFeatures() {
+			features, _ := parserFeatures()
+			for _, f := range features {
 				if f == "unsafe" {
 					return fmt.Sprintf("change_profile unsafe /**,")
 				}
@@ -587,7 +588,7 @@ func (b *Backend) SandboxFeatures() []string {
 	}
 
 	features := kernelFeatures()
-	pFeatures := parserFeatures()
+	pFeatures, _ := parserFeatures()
 	tags := make([]string, 0, len(features)+len(pFeatures))
 	for _, feature := range features {
 		// Prepend "kernel:" to apparmor kernel features to namespace them and
--- a/interfaces/apparmor/backend.go
+++ b/interfaces/apparmor/backend.go
@@ -60,7 +60,7 @@ var (
 	procSelfExe           = "/proc/self/exe"
 	isHomeUsingNFS        = osutil.IsHomeUsingNFS
 	isRootWritableOverlay = osutil.IsRootWritableOverlay
-	kernelFeatures        = release.AppArmorFeatures
+	kernelFeatures        = release.AppArmorKernelFeatures
 	parserFeatures        = release.AppArmorParserFeatures
 )
 
@@ -587,10 +587,10 @@ func (b *Backend) SandboxFeatures() []string {
 		return nil
 	}
 
-	features := kernelFeatures()
+	kFeatures, _ := kernelFeatures()
 	pFeatures, _ := parserFeatures()
-	tags := make([]string, 0, len(features)+len(pFeatures))
-	for _, feature := range features {
+	tags := make([]string, 0, len(kFeatures)+len(pFeatures))
+	for _, feature := range kFeatures {
 		// Prepend "kernel:" to apparmor kernel features to namespace them and
 		// allow us to introduce our own tags later.
 		tags = append(tags, "kernel:"+feature)
--- a/release/apparmor.go
+++ b/release/apparmor.go
@@ -123,12 +123,6 @@ func AppArmorParserFeatures() ([]string, error) {
 	return appArmorParserFeatures, appArmorParserError
 }
 
-// AppArmorFeatures is a deprecated version for AppArmorKernelFeatures.
-func AppArmorFeatures() []string {
-	features, _ := AppArmorKernelFeatures()
-	return features
-}
-
 // AppArmorParserMtime returns the mtime of the parser, else 0.
 func AppArmorParserMtime() int64 {
 	var mtime int64
--- a/release/apparmor.go
+++ b/release/apparmor.go
@@ -197,7 +197,7 @@ var (
 	// requiredAppArmorKernelFeatures denotes the features that must be present in the kernel.
 	// Absence of any of those features results in the effective level be at most UnusableAppArmor.
 	requiredAppArmorKernelFeatures = []string{
-		// TODO: what is the minimal set?
+		// For now, require at least file and simply prefer the rest.
 		"file",
 	}
 	// preferredAppArmorKernelFeatures denotes the features that should be present in the kernel.
--- a/release/apparmor.go
+++ b/release/apparmor.go
@@ -28,6 +28,8 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+
+	"github.com/snapcore/snapd/strutil"
 )
 
 // ApparmorLevelType encodes the kind of support for apparmor
@@ -39,6 +41,8 @@ const (
 	UnknownAppArmor AppArmorLevelType = iota
 	// NoAppArmor indicates that apparmor is not enabled.
 	NoAppArmor
+	// UnusableAppArmor indicates that apparmor is enabled but cannot be used.
+	UnusableAppArmor
 	// PartialAppArmor indicates that apparmor is enabled but some
 	// features are missing.
 	PartialAppArmor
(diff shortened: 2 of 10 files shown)


release/apparmor.go
```diff
@@ -39,54 +41,141 @@ const (
...
+// cached but is is somewhat costly to perform.
+func AppArmorParserFeatures() []string {
+	if appArmorParserFeatures == nil {
+		appArmorParserFeatures = probeAppArmorParserFeatures()
+	}
+	return appArmorParserFeatures
+}
+
+// AppArmorFeatures is a deprecated name for AppArmorKernelFeatures.
+func AppArmorFeatures() []string {
```
bboozzoo: If it's deprecated, do we still need to have it around?

tests/regression/lp-1805485/task.yaml
```diff
@@ -0,0 +1,10 @@
+summary: apparmor backend is not used on openSUSE Leap 42.3
+details: |
+    AppArmor on openSUSE Leap 43.2 is too old to parse the snap-confine
+    apparmor profile.  Because of "relaxed" error handling snapd was enabling
+    apparmor backend even though it resulted in constant problems with setting
+    up security for the core snap. With this bug fixed the old version of
+    apparmor_parser makes snapd disable the relevant security backend. 
+systems: [opensuse-42.3-64]
```
bboozzoo: Shouldn't the test check the supported systems have not regressed after the fix was introduced?
```
systems: [ubuntu-*, opensuse-*, arch-linux-*]
execute:
    if [[ "$SPREAD_SYSTEM" == opensuse-42.3* ]]; then
        # openSUSE 43.2 has AppArmor parser in version too old to parse snap-confine profile
        snap debug sandbox-features | MATCH -v "^apparmor:"
    else
        snap debug sandbox-features | MATCH "^apparmor:"
    fi

File: release/apparmor.go (hunk 1 above)
Comment (zyga) [reply release/apparmor.go]: The parts that remove it are in another branch. I wanted to limit this for just the release package.

File: tests/regression/lp-1805485/task.yaml (hunk 2 above)
Comment (zyga) [reply tests/regression/lp-1805485/task.yaml]: We have plenty of places that do that but I can add a safety check yeah,

File: release/apparmor.go [hunk 3]
```diff
@@ -39,54 +41,139 @@ const (
...
+	case NoAppArmor:
+		return "none"
+	case UnusableAppArmor:
+		return "unusable"
+	case PartialAppArmor:
+		return "partial"
+	case FullAppArmor:
+		return "full"
+	}
+	return "???"
```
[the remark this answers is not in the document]
zyga: We could just panic.

release/apparmor.go
```diff
@@ -39,54 +41,139 @@ const (
...
 
-// AppArmorSummary describes how well apparmor is supported on the
-// current kernel.
+// AppArmorSummary describes how well apparmor is supported on the current
+// kernel. The computation is costly to perform. The result is cached
+// internally.
 func AppArmorSummary() string {
 	if appArmorLevel == UnknownAppArmor {
-		appArmorLevel, appArmorSummary = probeAppArmor()
+		assessAppArmor()
```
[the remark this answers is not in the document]
zyga: Those were always determined together. There should be no special order defined if we can get both at the same time.

release/apparmor.go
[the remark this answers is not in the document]
zyga: I didn't like the panic, this is now returning a proper name for unknown values.

release/apparmor.go
```diff
@@ -97,117 +184,100 @@ var (
...
-	return stat.IsDir()
-}
 
-func probeAppArmor() (AppArmorLevelType, string) {
-	if !isDirectory(appArmorFeaturesSysPath) {
-		return NoAppArmor, "apparmor not enabled"
+	parserFeatures := AppArmorParserFeatures()
+	var missingParserFeatures []string
+	for _, feature := range requiredAppArmorParserFeatures {
+		if !strutil.SortedListContains(parserFeatures, feature) {
```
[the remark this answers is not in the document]
jdstrand: > IME we should return UnusableAppArmor if `len(parserFeatures) == 0` since AppArmorParserFeatures() no longer returns an error but instead an empty list when the parser is not found. As it stands, this PR is operationally correct but there is an 'almost bug' here in that if we ever decide to remove 'unsafe' as a required feature then there is no way to discern between a missing parser and no required features.

Thanks for all the updates! This is the only remaining item for my +1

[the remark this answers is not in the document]
zyga: +1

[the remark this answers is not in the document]
jdstrand: What is the idea? How involved is it?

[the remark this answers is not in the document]
zyga: To have two sets: features that presence of yields full level and features that absence of yields unsupported level. It's not too involved I suspect.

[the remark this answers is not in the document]
jdstrand: Well, the current code has a concept of missing parser error vs empty features so removing that to only add it back in a separate PR seems unnecessary (especially since it should be very complicated). Am I missing something?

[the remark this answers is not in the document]
zyga: I believe this is ready now.

release/apparmor.go
```diff
@@ -184,10 +184,25 @@ func MockAppArmorFeatures(kernelFeatures, parserFeatures []string) (restore func
...
 var (
+	// requiredAppArmorParserFeatures denotes the features that must be present in the parser.
+	// Absence of any of those features results in the effective level be at most UnusableAppArmor.
 	requiredAppArmorParserFeatures = []string{
 		"unsafe",
 	}
+	// preferredAppArmorParserFeatures denotes the features that should be present in the parser.
+	// Absence of any of those features results in the effective level be at most PartialAppArmor.
+	preferredAppArmorParserFeatures = []string{
+		"unsafe",
```
jdstrand: It doesn't make sense (to me anyway) to have something in preferred also in required. Shouldn't this just be empty?

release/apparmor.go
```diff
@@ -184,10 +184,25 @@ func MockAppArmorFeatures(kernelFeatures, parserFeatures []string) (restore func
...
+	}
+	// requiredAppArmorKernelFeatures denotes the features that must be present in the kernel.
+	// Absence of any of those features results in the effective level be at most UnusableAppArmor.
 	requiredAppArmorKernelFeatures = []string{
+		// TODO: what is the minimal set?
+		"file",
```
jdstrand: I think you can remove the TODO and say: `// For now, require at least file and simply prefer the rest`

release/apparmor.go
```diff
@@ -97,117 +213,130 @@ var (
...
+	return features
 }
 
-var (
-	requestedParserFeatures = []apparmorParserFeature{
-		{"unsafe", "change_profile unsafe /**,"},
+func probeAppArmorParserFeatures() []string {
+	parser := findAppArmorParser()
+	if parser == "" {
+		return []string{}
```
jdstrand: This is the thing that still bothers me. We are returning an empty list to denote that either we can't find a parser but also a parser that was found but doesn't support any special features. With the first, that is always at best UnusableAppArmor while the second could be other things. As it stands, your code elsewhere has "unsafe" as required so this always ends up as UnsableAppArmor, but I'm uncomfortable with this at an API level.

zyga: I can refactor this to have return an explicit error if you'd prefer that.

release/apparmor.go
zyga: From one point of view you could say that what we require is a subset of what we prefer and not a distinct set. From another you can say that this can be empty since we already require that anyway. I don't think either is more correct than the other. In the implementation I opted into the first of those interpretations.

release/apparmor.go
zyga: And to be clear, this can be an internal error that is never surfaced (`errNotFound` or something like that).

jdstrand: IME, it can return an error, like it did before, but the caller shouldn't carry that error forward. The caller should make that UnusableAppArmor. With a change along these lines, I'll +1 the PR.

zyga: Thanks, doing that now.

zyga: Done

release/apparmor.go
```diff
@@ -292,43 +315,43 @@ func assessAppArmor() {
...
+	return features, nil
 }
 
-func probeAppArmorParserFeatures() []string {
-	parser := findAppArmorParser()
-	if parser == "" {
-		return []string{}
+func probeAppArmorParserFeatures() ([]string, error) {
+	parser, err := findAppArmorParser()
+	if os.IsNotExist(err) {
```
jdstrand: If findAppArmorParser() is updated to return something other than IsNotExist, then we won't return an error. Why not just check if there is an error?

interfaces/apparmor/backend.go
```diff
@@ -587,7 +588,7 @@ func (b *Backend) SandboxFeatures() []string {
 	}
 
 	features := kernelFeatures()
```
jdstrand: Why isn't this `features, _ := kernelFeatures()`?

zyga: Because for the sake of landing a smaller patch, originally, there's a compatibility function that ignores the error. It was present in the first iteration of this patch and I kept it. I have also a small set of patches on top of the initial one that discard that function.

zyga: I removed the old function now.

release/apparmor.go
zyga: Fixed, thank you for being thorough :-)

PR #6246, opened 2018-11-30
spread: show AVC audits when debugging, start auditd on Fedora

bboozzoo:
Systems that have SELinux enabled typically use audit to log AV denials. Dump the AVC log when debugging.

For some reason, auditd is not started in our Fedora GCP images. It is started in Fedora Cloud images pulled from https://alt.fedoraproject.org/cloud/ though. Make sure that auditd is running when spread takes over the instance.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -414,8 +414,15 @@ debug-each: |
 
         echo '# journal messages for snapd'
         get_journalctl_log -u snapd
-        echo '# apparmor denials '
-        dmesg --ctime | grep DENIED || true
+        case "$SPREAD_SYSTEM" in
+            fedora-*|centos-*|amazon-*)
+                ausearch -m AVC
+                ;;
+            *)
+               echo '# apparmor denials '
+               dmesg --ctime | grep DENIED || true
+               ;;
+        esac
         echo '# seccomp denials (kills) '
         dmesg --ctime | grep type=1326 || true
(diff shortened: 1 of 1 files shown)


spread.yaml
```diff
@@ -414,8 +414,15 @@ debug-each: |
...
-        dmesg --ctime | grep DENIED || true
+        case "$SPREAD_SYSTEM" in
+            fedora-*|centos-*|amazon-*)
+                ausearch -m AVC
+                ;;
+            *)
+               echo '# apparmor denials '
+               dmesg --ctime | grep DENIED || true
+               ;;
+        esac
```
zyga: +1

zyga: Do you think we are a ready for a spread test that only passes if one can snap install hello on fedora with 0 warnings?

bboozzoo: Yes, that's part of my branch.

spread.yaml
```diff
@@ -414,8 +414,15 @@ debug-each: |
 
         echo '# journal messages for snapd'
         get_journalctl_log -u snapd
-        echo '# apparmor denials '
-        dmesg --ctime | grep DENIED || true
+        case "$SPREAD_SYSTEM" in
+            fedora-*|centos-*|amazon-*)
+                ausearch -m AVC
```
stolowski: :heart:

PR #6247, opened 2018-11-30
tests/lib: do not pick up xatts when repacking core snap

bboozzoo:
Avoid picking up xattrs when repacking core snap. On systems supporting SELinux
this may accidentally pick up the file context from where the core snap was
initially unpack. In case of tests, the core snap is unpacked to test home
directory and the repacked snap files have `user_home_t` context.

### DIFF
--- a/tests/lib/prepare.sh
+++ b/tests/lib/prepare.sh
@@ -141,7 +141,8 @@ update_core_snap_for_classic_reexec() {
 
     # repack, cheating to speed things up (4sec vs 1.5min)
     mv "$snap" "${snap}.orig"
-    mksnap_fast "squashfs-root" "$snap"
+    # make sure that SELinux context is not picked up by accident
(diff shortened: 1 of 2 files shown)


tests/lib/prepare.sh
```diff
@@ -141,7 +141,8 @@ update_core_snap_for_classic_reexec() {
 
     # repack, cheating to speed things up (4sec vs 1.5min)
     mv "$snap" "${snap}.orig"
-    mksnap_fast "squashfs-root" "$snap"
+    # make sure that SELinux context is not picked up by accident
+    mksnap_fast "squashfs-root" "$snap" -no-xattrs
```
zyga: Should -no-xattrs be inside mksnap_fast?

bboozzoo:
Will replace that with a smarter version.

PR #6248, opened 2018-11-30
tests/lib/reset: restore context of removed snapd directories

bboozzoo:
SELinux context of filesystem objects is the same as their parent. Once the top
level snapd directories are purged and remade, they will end up with a context
that stems off respective parent directories, which can end up being `var_t`,
`var_lib_t` (or even `root_t` if `/snap` was used). Make sure to restore the context
of these locations according to our targeted policy.

### DIFF
--- a/tests/lib/reset.sh
+++ b/tests/lib/reset.sh
@@ -61,6 +61,12 @@ reset_classic() {
         exit 1
     fi
 
+    case "$SPREAD_SYSTEM" in
+        fedora-*|centos-*)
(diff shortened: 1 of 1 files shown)


tests/lib/reset.sh
```diff
@@ -61,6 +61,12 @@ reset_classic() {
         exit 1
     fi
 
+    case "$SPREAD_SYSTEM" in
```
sergiocazzolato: could you please add a short comment to explain that line?

bboozzoo: Added now.

PR #6249, opened 2018-11-30
cmd/libsnap: introduce and use sc_strdup

zyga:
We have a small but common pattern where we need to duplicate a string,
check for memory allocation errors and carry on. This patch introduces
sc_strdup that does that internally.

All of the instances of strdup and error checking in the code were
updated to use it.

### DIFF
--- a/cmd/libsnap-confine-private/string-utils.c
+++ b/cmd/libsnap-confine-private/string-utils.c
@@ -56,6 +56,22 @@ bool sc_endswith(const char *str, const char *suffix)
 	return strncmp(str - xlen + slen, suffix, xlen) == 0;
 }
 
+char *sc_strdup(const char *str)
+{
+	size_t len;
+	char *copy;
+	if (str == NULL) {
+		die("cannot duplicate NULL string");
+	}
+	len = strlen(str);
+	copy = malloc(len + 1);
+	if (copy == NULL) {
+		die("cannot allocate string copy (len: %zd)", len);
+	}
+	memcpy(copy, str, len + 1);
+	return copy;
+}
+
 int sc_must_snprintf(char *str, size_t size, const char *format, ...)
 {
 	int n;
--- a/cmd/snap-confine/mount-support-nvidia.c
+++ b/cmd/snap-confine/mount-support-nvidia.c
@@ -166,12 +166,9 @@ static void sc_populate_libgl_with_hostfs_symlinks(const char *libgl_dir,
 		char prefix_dir[512] = { 0 };
 		const char *pathname = glob_res.gl_pathv[i];
 		char *pathname_copy1
-		    SC_CLEANUP(sc_cleanup_string) = strdup(pathname);
+		    SC_CLEANUP(sc_cleanup_string) = sc_strdup(pathname);
 		char *pathname_copy2
-		    SC_CLEANUP(sc_cleanup_string) = strdup(pathname);
-		if (pathname_copy1 == NULL || pathname_copy2 == NULL) {
-			die("failed to copy pathname");
-		}
+		    SC_CLEANUP(sc_cleanup_string) = sc_strdup(pathname);
 		// POSIX dirname() and basename() may modify their input arguments
 		char *filename = basename(pathname_copy1);
 		char *directory_name = dirname(pathname_copy2);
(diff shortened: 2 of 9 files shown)


cmd/libsnap-confine-private/string-utils.c
```diff
@@ -56,6 +56,22 @@ bool sc_endswith(const char *str, const char *suffix)
...
 }
 
+char *sc_strdup(const char *str)
+{
+	size_t len;
+	char *copy;
+	if (str == NULL) {
+		die("cannot duplicate NULL string");
+	}
+	len = strlen(str);
```
bboozzoo: `stdup()` does exactly that: strlen(str), malloc(len+1), memcpy(), why not use it and do:
```
char * sc_strdup(const char *str)
{
    char *copy = stdup(str);
    if (copy == NULL) {
        die("cannot allocate string copy (len: %zd)", len);
    }
    return copy
}
```

cmd/libsnap-confine-private/string-utils.c
```diff
@@ -56,6 +56,22 @@ bool sc_endswith(const char *str, const char *suffix)
 	return strncmp(str - xlen + slen, suffix, xlen) == 0;
 }
 
+char *sc_strdup(const char *str)
```
bboozzoo: Considering that it calls to `die()` on failure, I'd call it `sc_must_strdup()`, similarly to `sc_must_snprintf()`.

cmd/snap-confine/mount-support-nvidia.c
```diff
@@ -166,12 +166,9 @@ static void sc_populate_libgl_with_hostfs_symlinks(const char *libgl_dir,
 		char prefix_dir[512] = { 0 };
 		const char *pathname = glob_res.gl_pathv[i];
 		char *pathname_copy1
-		    SC_CLEANUP(sc_cleanup_string) = strdup(pathname);
+		    SC_CLEANUP(sc_cleanup_string) = sc_strdup(pathname);
```
bboozzoo: Wondering if maybe `strndup()` wrapper would be useful too. The upper bound of size is usually known, and the implementations I've seen all used `strnlen()` so as not to allocate the max size unnecessarily.

cmd/libsnap-confine-private/string-utils.c
zyga: Because it doesn't tell use the length :D

cmd/libsnap-confine-private/string-utils.c
zyga: Virtually all functions die, I think the must thing is a useless add-on that doesn't reflect the permadeath mode of the code.

cmd/snap-confine/mount-support-nvidia.c
zyga: I'm happy to add one once we use strndup more than once.

cmd/libsnap-confine-private/string-utils.c
stolowski: I'm slightly agreeing with @bboozzoo here: similiar to `sc_must_sprintf`, this new function belongs to primitives that you normally don't expect to abort, so emphasizing it makes sense (also for consistency with sc_must_snprintf).

cmd/libsnap-confine-private/string-utils.c
stolowski: Is `len` useful though? Just wondering if it helps in any way to show it in the error message.

cmd/snap-confine/mount-support.c
```diff
@@ -75,10 +75,7 @@ static void setup_private_mount(const char *snap_name)
 	}
 	// now we create a 1777 /tmp inside our private dir
 	mode_t old_mask = umask(0);
-	char *d = strdup(tmpdir);
```
stolowski: Nice reduction of boilerplate down there!

cmd/libsnap-confine-private/string-utils.c
zyga: There are many dozen functions using the sc_ prefix that all die an any error. The only one of those that uses the word _must_ is `sc_must_snprintf`. I'd really rather invent an opposite word if we want to have this sort of consistency.

cmd/libsnap-confine-private/string-utils.c
zyga: In case we ever want to malloc -1 bytes, I suspect it would be useful.

bboozzoo: In that case it makes sense indeed.

PR #6250, opened 2018-11-30
data: set KillMode=process for snapd

mvo5:
When snapd is restarted via systemd there may be processes like
apparmor_parser or snap-seccomp running that (re)generate profiles. The
current behavior of systemd is right now that on restart all processes
in the cgroup are killed which means that the security profile generation
may be incomplete. To avoid this as a first step this commit sets the
KillMode=process to ensure that only snapd itself is restarted and
the seccomp/apparmor helpers can continue doing their job.

This needs some extra work to write a /etc/systemd/system/snapd.service.d/kill-mode.conf override for the re-exec case.


mvo5:
Thanks @zyga

zyga:
This is green which is remarkable given how the tests are looking today. I'm very happy to merge this and iterate.

PR #6251, opened 2018-11-30
cmd/snap-confine: refactor calling snapd tools into helper module

zyga:
The module allows opening tool by name, opening the two tools we
are known to open (snap-{update,discard}-ns) and calling a function
by open descriptor, with apparmor confinement if desired.

There's a small skew that tool.c includes snap-confine/apparmor-support.h
but that's a low price to pay for this cleanup.

### DIFF
--- a/cmd/libsnap-confine-private/tool.c
+++ b/cmd/libsnap-confine-private/tool.c
@@ -50,7 +50,7 @@ static int sc_open_snapd_tool(const char *tool_name);
  * anymore. The caller establishes an open file descriptor in one namespace and later on
  * performs the call in another mount namespace.
  *
- * The environment vector has special support for expanding he string "SNAPD_DEBUG=x".
+ * The environment vector has special support for expanding the string "SNAPD_DEBUG=x".
  * If such string is present, the "x" is replaced with either "0" or "1" depending on
  * the result of is_sc_debug_enabled().
  **/
--- a/cmd/libsnap-confine-private/tool.c
+++ b/cmd/libsnap-confine-private/tool.c
@@ -1,0 +1,193 @@
+/*
+ * Copyright (C) 2018 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "tool.h"
+
+#include <fcntl.h>
+#include <libgen.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+#include "../libsnap-confine-private/apparmor-support.h"
+#include "../libsnap-confine-private/cleanup-funcs.h"
+#include "../libsnap-confine-private/string-utils.h"
+#include "../libsnap-confine-private/utils.h"
+
+/**
+ * sc_open_snapd_tool returns a file descriptor of the given internal executable.
+ *
+ * The executable is located based on the location of the currently executing process.
+ * The returning file descriptor can be used with fexecve function, like in sc_call_snapd_tool.
+**/
+static int sc_open_snapd_tool(const char *tool_name);
+
+/**
+ * sc_call_snapd_tool calls a snapd tool by file descriptor.
+ *
+ * The idea with calling with an open file descriptor is to allow calling executables
+ * across mount namespaces, where the executable may not be visible in the new filesystem
+ * anymore. The caller establishes an open file descriptor in one namespace and later on
+ * performs the call in another mount namespace.
+ *
+ * The environment vector has special support for expanding he string "SNAPD_DEBUG=x".
+ * If such string is present, the "x" is replaced with either "0" or "1" depending on
+ * the result of is_sc_debug_enabled().
+ **/
+static void sc_call_snapd_tool(int tool_fd, const char *tool_name, char **argv,
+			       char **envp);
+
+/**
+ * sc_call_snapd_tool_with_apparmor calls a snapd tool by file descriptor,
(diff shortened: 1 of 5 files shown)


cmd/libsnap-confine-private/tool.c
```diff
@@ -0,0 +1,193 @@
...
+		     env != NULL && *env != NULL && **env != '\0'; env++) {
+			if (sc_streq(*env, "SNAPD_DEBUG=x")) {
+				/* NOTE: this is not released, on purpose. */
+				char *entry = sc_strdup(*env);
+				entry[strlen("SNAPD_DEBUG=x") - 1] =
+				    sc_is_debug_enabled()? '1' : '0';
+				*env = entry;
+			}
+		}
+		/* As the child ensure that the file descriptor is remains open across exec. */
```
bboozzoo: s/is remains/remains/

cmd/libsnap-confine-private/tool.c
```diff
@@ -0,0 +1,193 @@
...
+		if (apparmor != NULL && aa_profile != NULL) {
+			sc_maybe_aa_change_onexec(apparmor, aa_profile);
+		}
+		fexecve(tool_fd, argv, envp);
+		die("cannot execute snapd tool %s", tool_name);
+	} else {
+		int status = 0;
+		debug("waiting for snapd tool %s to terminate", tool_name);
+		if (waitpid(child, &status, 0) < 0) {
+			die("cannot wait for snapd tool %s to terminate",
```
bboozzoo: Maybe `failed to wait for snapd ..`. It's probably subjective, but because of the _cannot wait_ having more than one meaning I found this message a bit unclear.

cmd/libsnap-confine-private/tool.c
```diff
@@ -0,0 +1,193 @@
...
+
+/**
+ * sc_call_snapd_tool calls a snapd tool by file descriptor.
+ *
+ * The idea with calling with an open file descriptor is to allow calling executables
+ * across mount namespaces, where the executable may not be visible in the new filesystem
+ * anymore. The caller establishes an open file descriptor in one namespace and later on
+ * performs the call in another mount namespace.
+ *
+ * The environment vector has special support for expanding he string "SNAPD_DEBUG=x".
```
bboozzoo: s/he/the/

cmd/snap-confine/ns-support.c
```diff
@@ -304,44 +305,6 @@ enum sc_discard_vote {
 	SC_DISCARD_YES = 2,
 };
 
-static void call_snap_discard_ns(int snap_discard_ns_fd, const char *snap_name)
```
bboozzoo: Nice to see the duplication go away.

cmd/libsnap-confine-private/tool.c
zyga: Yeah, I agree. Maybe, `cannot get child termination status via waitpid`?

bboozzoo: Yup, sounds good

cmd/libsnap-confine-private/tool.c
```diff
@@ -0,0 +1,192 @@
...
+				    sc_is_debug_enabled()? '1' : '0';
+				*env = entry;
+			}
+		}
+		/* As the child ensure that the file descriptor remains open across exec. */
+		int flags = fcntl(tool_fd, F_GETFD);
+		if (flags < 0) {
+			die("cannot get file descriptor flags");
+		}
+		flags &= ~FD_CLOEXEC;
```
bboozzoo: Wondering about this. We don't seem to have had this pattern before. I suppose the actual fd to the executable if off-limits for clearing on exec anyway, so why would we want to leave it open?

zyga: This is for extra safety. Notice that the descriptor obtained by opening the tools has the CLOEXEC flag set. In case you just open it and forget about it. This function knows this and ensures that the flag is cleared in the forked child, just before we exec.

If you remove this code funny error show up because the executed child has no descriptor to use.

bboozzoo: What I mean is that before we also called `open(.., .. | O_CLOEXEC)` but the flag was not cleared before `fexecve()` in the child. Looking at the manpage they actually mention close-on-exec as a thing to do (unless one is execing a script).

zyga: Hmmmm, maybe something else was at fault. I will check this.

zyga: Removed now, nice catch! Thank you.

cmd/libsnap-confine-private/tool.c
```diff
@@ -0,0 +1,183 @@
...
+}
+
+void sc_call_snap_update_ns(int snap_update_ns_fd, const char *snap_name,
+			    struct sc_apparmor *apparmor)
+{
+	char *snap_name_copy SC_CLEANUP(sc_cleanup_string) = NULL;
+	snap_name_copy = sc_strdup(snap_name);
+	char *argv[] =
+	    { "snap-update-ns", "--from-snap-confine", snap_name_copy, NULL };
+	char *envp[] = { "SNAPD_DEBUG=x", NULL };
```
stolowski: Nitpick: I find `SNAPD_DEBUG=1` more widespread.

cmd/libsnap-confine-private/tool.c
```diff
@@ -0,0 +1,183 @@
...
+	return sc_open_snapd_tool("snap-discard-ns");
+}
+
+void sc_call_snap_discard_ns(int snap_discard_ns_fd, const char *snap_name)
+{
+	char *snap_name_copy SC_CLEANUP(sc_cleanup_string) = NULL;
+	snap_name_copy = sc_strdup(snap_name);
+	char *argv[] =
+	    { "snap-discard-ns", "--from-snap-confine", snap_name_copy, NULL };
+	char *envp[] = { "SNAPD_DEBUG=x", NULL };
```
stolowski: Same.

cmd/libsnap-confine-private/tool.c
```diff
@@ -0,0 +1,183 @@
...
+	if (child < 0) {
+		die("cannot fork to run snapd tool %s", tool_name);
+	}
+	if (child == 0) {
+		/* If the caller provided template environment entry for SNAPD_DEBUG
+		 * then expand it to the actual value. */
+		for (char **env = envp;
+		     /* Mama mia, that's a spicy meatball. */
+		     env != NULL && *env != NULL && **env != '\0'; env++) {
+			if (sc_streq(*env, "SNAPD_DEBUG=x")) {
```
stolowski: Ah, I think I get it now, mind adding an extra comment above whe `SNAPD_DEBUG=x` is set?

zyga: Not at all, I'll do that straight away :)

PR #6253, opened 2018-11-30
Members of canonical LP group should pass CLA check

kenvandine:
cla_check should pass for members of canonical on LP


kenvandine:
Yeah, that is a direct copy from the check that was already there.

PR #6254, opened 2018-11-30
tests: improve how the log is checked to see if the system is waiting for a reboot

sergiocazzolato:
The previous function retries several times checking in the logs, but
this method retrieve the logs and we match, which is much better.


sergiocazzolato:
@mvo5, it is ok to change the names but I don't get the idea from the example, could you please explain a bit more that?

mvo5:
@sergiocazzolato Happy to do this as a followup, this can be merged. I was mostly thinking that the names  check_journalctl_log and get_journalctl_log are very similar and when reading just the function names its not clear why we have two and what the difference is. So if one tries several times, maybe we name it something like "get_journal_log_retry_until_ready" or something (assuming this is what it does, I have not looked at the details of the functions yet.

PR #6257, opened 2018-12-03
testutils: split checkers, tweak tests

zyga:
The checkers were all in one big module and that was unwieldy for upstreaming
or for adding more. To avoid having the of those huge files that only gets appended
to I decided to split things up.

While I was at it, I made two small changes:
 - dropped ppc exception, since ppc is not a problem anymore
 - moved all the tests into a `testutil_test` package


testutil/containschecker.go
```diff
@@ -0,0 +1,152 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2015 Canonical Ltd
```
stolowski: Let's update it since you're on it.

zyga: Shall I bump all of those to 2018?

testutil/filepresencechecker.go
```diff
@@ -0,0 +1,59 @@
...
+package testutil
+
+import (
+	"fmt"
+	"os"
+
+	"gopkg.in/check.v1"
+)
+
+type filePresenceChecker struct {
```
mvo5: I wonder if this (and the IntLessThan, IntEqual etc) checks should simply become something like:
```
c.Check(osutil.FileExists(foo), IsTrue)
c.Check(osutil.FileExists(bar), IsFalse)
c.Check(i < 7, Istrue)
``` 
It means less code and also I personally find this easier to read.

zyga: While unrelated to the branch, the reason for using those is that there's a useful failure message, otherwise we could do all checks like this:

```
c.Check(whatever boolean expression, Equals, true)
```

mvo5: Aha, this is done so that people don't have to do: `c.Check(bool-expr, Equals, true, Commentf("reasons!"))`

PR #6259, opened 2018-12-04
tests: reduce verbosity around package installation

zyga:
Package installation is particularly verbose, not because of actual
invocation of the system package manager, but because there's a pack of
indirection layers to simplify cross distribution portability.

Recently we ran into an issue where a Travis job is killed after
exceeding log size. As the package trace logs are rarely a problem I've
made as small patch that silences shell trace output around that logic.

To maintain minimal trace we now echo the set of packages to install
along with the name of the function that was invoked.

Thanks to Maciej Borzecki for the trick with "set -o"

### DIFF
--- a/tests/lib/pkgdb.sh
+++ b/tests/lib/pkgdb.sh
@@ -191,7 +191,7 @@ distro_install_local_package() {
 }
 
 distro_install_package() {
-    orig_xtrace=$(set -o | grep xtrace | cut -f 2)
+    orig_xtrace=$(set -o | awk '/xtrace / { print $2 }')
     set +x
     echo "distro_install_package $*"
     # Parse additional arguments; once we find the first unknown
--- a/tests/lib/pkgdb.sh
+++ b/tests/lib/pkgdb.sh
@@ -191,6 +191,9 @@ distro_install_local_package() {
 }
 
 distro_install_package() {
+    orig_xtrace=$(set -o | grep xtrace | cut -f 2)
+    set +x
+    echo "distro_install_package $*"
     # Parse additional arguments; once we find the first unknown
     # part we break argument parsing and process all further
     # arguments as package names.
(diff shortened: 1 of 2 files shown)


tests/lib/pkgdb.sh
```diff
@@ -191,6 +191,9 @@ distro_install_local_package() {
 }
 
 distro_install_package() {
+    orig_xtrace=$(set -o | grep xtrace | cut -f 2)
```
bboozzoo: Further reading `bash` manpage, you could do do `[[ -o xtrace ]] && orig_xtrace=on`

zyga: Yeah but this works in shell and bash, safer to copy-paste around :)

bboozzoo: Ah, then `set -o | awk '/xtrace / { print $2 }'` is one fork less.

zyga:
It looks like we've uncovered some real issues that were, somehow, previously hidden. It looks like before importing pkgdb would nuke "set -e" state.

```
2018-12-04 11:41:49 Error executing google:ubuntu-18.10-64:tests/upgrade/snapd-xdg-open : 
-----
+ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/pkgdb.sh
++ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/quiet.sh
+ ver=0.0.0~16.04
+ distro_install_package snapd-xdg-open=0.0.0~16.04
++ awk '/xtrace / { print $2 }'
++ set -o
+ orig_xtrace=on
+ set +x
distro_install_package snapd-xdg-open=0.0.0~16.04
quiet: apt-get install -y snapd-xdg-open=0.0.0~16.04
quiet: exit status 100. Output follows:
Reading package lists...
Building dependency tree...
Reading state information...
E: Version '0.0.0~16.04' for 'snapd-xdg-open' was not found
quiet: end of output.
++ dpkg-query '--showformat=${Version}' --show snapd-xdg-open
dpkg-query: no packages found matching snapd-xdg-open
+ prevsnapdxdgver=
-----
```

mvo5:
This is very interessting - does `-x` also implicitly set `-e` ?

PR #6261, opened 2018-12-04
tests/lib/prepare: make sure that SELinux context of repacked core snap is controlled

bboozzoo:
When the core snap in unpacked to $HOME, user_home_t label will be automatically
applied on all its content. Since we cannot drop xattrs when calling mksquashfs,
as that could introduce breakage in the future, make sure that we relabel the
contents in way that a squashfs image without any labels would look like:
`system_u:object_r:unlabeled_t`

### DIFF
--- a/tests/lib/prepare.sh
+++ b/tests/lib/prepare.sh
@@ -141,7 +141,7 @@ update_core_snap_for_classic_reexec() {
 
     case "$SPREAD_SYSTEM" in
         fedora-*|centos-*|amazon-*)
-            # on these systems just unpacking core snap to $HOME will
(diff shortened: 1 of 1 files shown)


tests/lib/prepare.sh
```diff
@@ -139,6 +139,17 @@ update_core_snap_for_classic_reexec() {
             ;;
     esac
 
+    case "$SPREAD_SYSTEM" in
+        fedora-*|centos-*|amazon-*)
+            # on these systems just unpacking core snap to $HOME will
```
zyga:
```suggestion
            # On these systems just unpacking core snap to $HOME will
```

PR #6263, opened 2018-12-04
Retrieving all pages for 'find' command, not only one

WiseRaccoon:
Snapd API call GET /v2/find returns not all necessary snap packages. For example if I request '/v2/find?name=*', I must receive a full list of packages. But response is not full.
The problem exists, because snapd sends a request to store (https://api.snapcraft.io/api/v1/snaps/search) and returns a list of packages from the response. But response from the store contains not only a list of packages, but also a number of more pages! This data ignored by snapd.
Now, if response from the store contains links to other pages, snapd requests other lists.

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -303,6 +303,17 @@ type searchResults struct {
 	Payload struct {
 		Packages []*snapDetails `json:"clickindex:package"`
 	} `json:"_embedded"`
+	Links struct {
+		First searchResultsLink `json:"first"`
+		Last  searchResultsLink `json:"last"`
+		Next  searchResultsLink `json:"next"`
+		Prev  searchResultsLink `json:"prev"`
+		Self  searchResultsLink `json:"self"`
+	} `json:"_links"`
+}
+
+type searchResultsLink struct {
+	HRef string `json:"href"`
 }
 
 type sectionResults struct {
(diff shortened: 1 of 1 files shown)


WiseRaccoon:
> It will make snap find worse, by making searches slower, showing less relevant results, and needlessly bogging down the backend.

> 
I agree, as I tested, snapd more slowly, than it shows only first page. But I disagree with 'showing less relevant results'. For example, I remember, that snap package name begins with 'a' and searches 'a*', but the result may not be relevant, because my package (for example) can exist on page 2 or 3.

> I think we should block empty name searches entirely to disabuse people of the notion that they can see a list of all snaps.

> 

Now 'snap find' with empty name don't starting and show an error, so this problem will not exist. But If user specially calls a search for a big amount of packages, may be we need to allow it for him?

chipaca:
FWIW snapd will sometime soon grow support for pagination, not by returning the whole set but by exposing the same kind of controls that the store exposes to snapd.

This should not impact `snap` though.

WiseRaccoon:
> FWIW snapd will sometime soon grow support for pagination, not by returning the whole set but by exposing the same kind of controls that the store exposes to snapd.

This should not impact snap though.

> 
If so, you can delete this PR. Also a question. Does 'pagination' support for snapd was scheduled for someone? Or I can try to develop it?

zyga:
Closing as requested.

PR #6264, opened 2018-12-05
spread, tests: add Fedora 29

bboozzoo:
Add Fedora 29 to spread suite and try to workaround different quirks identified in the process.

New issues:
- default umask seems to be `0077` now, causing some issues in snap try tests
- `xdg-desktop-portal` seems to be installed by default
- strace-static (stable is 4.20) does not work with Fedora's kernel (4.19.5 atm), tried the edge version (4.23) but still no luck
- RPM seems to rewrite the shebang line to `#!/usr/bin/sh`, added a workaround for the case when we repack the core snap

The PR includes bits from #6260 

@sergiocazzolato would it be possible to update the image and remove `xdg-desktop-portal` ?


tests/main/degraded/task.yaml
```diff
@@ -13,7 +13,7 @@ debug: |
         units="$(systemctl --failed --type=$unit_type --no-pager | grep -o -E ".*.$unit_type" | tr '●' ' ' )"
         for unit in $units; do
             echo " -- systemctl status $unit --"
-            systemctl status "$unit"
+            systemctl status "$unit" || true
```
zyga: What is the motivation behind this change? When would status fail?

tests/main/document-portal-activation/task.yaml
```diff
@@ -27,6 +27,11 @@ prepare: |
     . "$TESTSLIB"/pkgdb.sh
     #shellcheck source=tests/lib/snaps.sh
     . "$TESTSLIB"/snaps.sh
+    if [[ "$SPREAD_SYSTEM" == fedora-29 ]]; then
```
zyga: It is likely that fedora-30 will have the same property. Perhaps we could have a negative test instead.

tests/main/try-with-hooks/task.yaml
```diff
@@ -1,7 +1,7 @@
 summary: Reproduce a known issue of snap try failure with hooks
 
 prepare: |
-  cp -R "$TESTSLIB"/snaps/basic ./
+  cp -a "$TESTSLIB"/snaps/basic ./
```
zyga: +1 on the change but curious on the motivation that lead to it

tests/main/document-portal-activation/task.yaml
bboozzoo: I'd prefer to have the test image updated. Will work with @sergiocazzolato to do that and then I'll drop the `if..` entirely

tests/main/try-with-hooks/task.yaml
bboozzoo: Default umask got changed to `0077`. The permissions are already set properly in our tree, but without `-a` they'd get reset to defaults. Then snap verification would bail because the directories are not world readable and so on.

tests/main/degraded/task.yaml
bboozzoo: It's in the commit message. To recap, `systemctl status <svc>` exits with non-0 status when the service has failed.

tests/main/document-portal-activation/task.yaml
zyga: That's unrealistic with comparison to what the regular system contains. I'm mildly opposed to crafting images that pass tests rather than have the test declare what is assumed. I will not block on this but I want to make my position clear.

tests/main/degraded/task.yaml
zyga: Interesting, thanks.

tests/main/document-portal-activation/task.yaml
bboozzoo: I'm looking into this more. A default cloud image does not contain `xdg-desktop-portal` package. This must be pulled in by some of the packages we install in `pkg_dependencies()`.

bboozzoo: An interesting find, apparently when we `dnf install` the built RPM, `dnf` pulls in `flatpak` and `xdg-desktop-portal`. I have no idea how this dependency came to be. Once installed I can actually remove `flatpak` and other packages and `snapd` is not removed. Playing with `dnf deplist` I did not found why exactly this happens. Maybe @Conan-Kudo  can suggest something.

Conan-Kudo: It's most likely a rich weak dependency being triggered. Do we run with weak deps enabled? If so, you could try adding `--setopt=install_weak_deps=False` to not make it happen.

bboozzoo: Thanks, that does the trick

tests/lib/pkgdb.sh
```diff
@@ -172,7 +172,7 @@ distro_install_local_package() {
             apt install $flags "$@"
             ;;
         fedora-*)
-            quiet dnf -y install "$@"
+            quiet dnf -y install --setopt=install_weak_deps=False "$@"
```
zyga: I wonder if this will impact anything that only worked because of a weak dependency that was installed before. Let's see.

tests/main/auth-errors/task.yaml
```diff
@@ -5,21 +5,22 @@ systems: [-ubuntu-core-*]
...
     mkdir -p /home/test/.snap
     echo -n "{\"macaroon\":\"yummy\",\"discharges\":[ \"some \"]}" > /home/test/.snap/auth.json
+    chown -R test:test /home/test/.snap
 
 restore: |
     rm -rf /home/test/.snap install.output connect.output
 
 execute: |
     echo "An unauthenticated user cannot install snaps"
-    if su - -c "snap install test-snapd-tools 2>${PWD}/install.output" test; then
```
zyga: Ouch, how did this ever work before?

tests/main/auth-errors/task.yaml
```diff
@@ -5,21 +5,22 @@ systems: [-ubuntu-core-*]
...
+    if su - -c "snap install test-snapd-tools" test 2>${PWD}/install.output ; then
         echo "Expected error installing snap from unauthenticated account"
         exit 1
     fi
     expected="error: access denied (try with sudo)"
     [ "$(cat install.output)" = "$expected" ]
 
     echo "An unauthenticated user cannot connect plugs to slots"
-    if su - -c "snap connect foo:bar baz:fromp 2>${PWD}/connect.output" test; then
+    if su - -c "snap connect foo:bar baz:fromp" test 2>${PWD}/connect.output ; then
```
zyga: And this?

tests/main/auth-errors/task.yaml
bboozzoo: Yeah, I was wondering that too.

zyga: Can you try to understand that? I think there's something fishy in our test suite still.

tests/main/document-portal-activation/task.yaml
Conan-Kudo: This makes me wonder if we're not disabling weak deps on all targets for CI...

tests/main/auth-errors/task.yaml
bboozzoo: My bad, I ran su - -c "snap install test-snapd-tools 2>${PWD}/install.output" in spread shell, which ended up creating the file owned by root. Another observation is that although the tests are run by root, /home/gopath is owned by test:test so it all works out in the test user in the end, unless we create files in the test itself and are not careful about changing the permissions. There are more tests that do the su -c ".. ${PWD}>somepath" test pattern, so I'll drop that hunk.

bboozzoo: Pushed a commit to fix it in other places too. Conveniently, those tests did not execute on Fedora, either by explicit `systems` entry or requiring strict confinement.

tests/main/econnreset/task.yaml
```diff
@@ -38,7 +38,7 @@ execute: |
     iptables -I INPUT -p tcp --match hashlimit --hashlimit-mode srcip,dstip,srcport,dstport --hashlimit-above 512kb/s \
              --hashlimit-name 'econnreset' -j DROP
 
-    su -c "/usr/bin/env SNAPD_DEBUG=1 SNAPD_DEBUG_HTTP=3 snap download --edge test-snapd-huge 2>snap-download.log" test &
```
stolowski: Nothing wrong with this change afacit, but I wonder what was the motivation for it?

stolowski: Ah, ignore me, I see similiar question from zyga above.

PR #6266, opened 2018-12-05
tests: make security-device-cgroups-{devmode,jailmode} work on arm devices

sergiocazzolato:
Currently we use mem device on the tests but it is not allowed to access
on arm device such as rpi3 and dragonboard. The idea is to use random
device instead.

example for rpi3:

test@localhost:~$ dd if=/dev/mem of=/dev/null count=10
dd: failed to open '/dev/mem': Permission denied

test@localhost:~$ dd if=/dev/random of=/dev/null count=10
10+0 records in
10+0 records out
5120 bytes (5.1 kB, 5.0 KiB) copied, 0.00265199 s, 1.9 MB/s

### DIFF
--- a/tests/main/security-device-cgroups-devmode/task.yaml
+++ b/tests/main/security-device-cgroups-devmode/task.yaml
@@ -31,7 +31,7 @@ execute: |
     test-devmode-cgroup.read-dev fb0 2>&1 | MATCH -v '(Permission denied|Operation not permitted)'
 
     echo "the devmode snap can access other devices"
-    test-devmode-cgroup.read-dev mem
+    test-devmode-cgroup.read-dev random
 
     echo "And the framebuffer plug is disconnected"
     snap disconnect test-devmode-cgroup:framebuffer
@@ -40,4 +40,4 @@ execute: |
     test-devmode-cgroup.read-dev fb0 2>&1 | MATCH -v '(Permission denied|Operation not permitted)'
 
     echo "the devmode snap can access other devices"
-    test-devmode-cgroup.read-dev mem
+    test-devmode-cgroup.read-dev random
--- a/tests/main/security-device-cgroups-devmode/task.yaml
+++ b/tests/main/security-device-cgroups-devmode/task.yaml
@@ -15,7 +15,7 @@ prepare: |
 
     # Create a test char device so we can verify the test snap cannot read.
     if [ ! -e /dev/test1 ]; then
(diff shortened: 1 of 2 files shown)


tests/main/security-device-cgroups-devmode/task.yaml
```diff
@@ -13,6 +13,12 @@ prepare: |
         touch /dev/fb0.spread
     fi
 
+    # Create a test char device so we can verify the test snap cannot read.
+    if [ ! -e /dev/test1 ]; then
+        mknod /dev/test1 c 29 0
```
bboozzoo: Just FYI, 29:0 char is framebuffer, the same as /dev/fb0

tests/main/security-device-cgroups-devmode/task.yaml
```diff
@@ -31,7 +40,7 @@ execute: |
     test-devmode-cgroup.read-dev fb0 2>&1 | MATCH -v '(Permission denied|Operation not permitted)'
 
     echo "the devmode snap can access other devices"
-    test-devmode-cgroup.read-dev mem
+    test-devmode-cgroup.read-dev test1 | MATCH -v '(Permission denied|Operation not permitted)'
```
bboozzoo: Can you check if those kernels have CONFIG_STRICT_DEVMEM? eg `zgrep /proc/config.gz`

tests/main/security-device-cgroups-devmode/task.yaml
sergiocazzolato: yes, I just wanted to make sure the char device is supported by the kernel, I'll try with a different one.

tests/main/security-device-cgroups-devmode/task.yaml
sergiocazzolato: Yes, but for this test I think best solution is to use a fake char davice which we know in all the images/boards, if I make that check I'll end up skipping the validation on some boards. Does it make sense?

bboozzoo: Sorry for adding confusion. I was just curious whether `CONFIG_STRICT_DEVMEM` is enabled in those kernels and I suspected it could be causing the issues with `/dev/mem`.

sergiocazzolato: If I remember correctly CONFIG_STRICT_DEVMEM is enable on rpi kernel. I'll manually check it today.

PR #6270, opened 2018-12-06
data/selinux, tests: refactor SELinux policy, add minimal tests

bboozzoo:
Followup on #6238. The PR contains a refactored SELinux policy for snapd. 

A week long crash course in SELinux gave birth to this larg(-ish) branch that introduces a minimal SELinux support and does a major refactoring to SELinux targeted policy for snapd.

Major changes:
- policy introduces a separate domain for snapd and each type of helper (mount, confine, cli)
- each domain has a separate policy and allows only specific transitions
- there is a separate file type for data in mounted snap images (`snappy_snap_t`, for the lack of a better name)

Some notes about transitions and starting snaps. When a regular user starts a snap, they are typically in `unconfined_u:unconfined_r:unconfined_t` context. This proceeds without any special transitions or limitations, since the user is already unconfined, and there is no targeted policy to close it down. The snaps that are started by systemd end up transitioning to `system_u:system_r:unconfined_service_t` domain, where they should not raise any more denials. While doing so, the transitions are:
```
init_t -> snappy_cli_t (snap run ..) -> snappy_confine_t -> ..
   ... -> snappy_mount_t (snap-update-ns) and back -> 
   ... -> unconfined_service_t
```
Domains permissions are limited to just the things each tool does. With the split in place, it's way easier to track which piece is doing what.

(cc @Conan-Kudo)


data/selinux/snappy.fc
```diff
@@ -18,30 +18,38 @@
 
 
 HOME_DIR/snap(/.*)?			gen_context(system_u:object_r:snappy_home_t,s0)
+/root/snap(/.*)?			gen_context(system_u:object_r:snappy_home_t,s0)
```
Conan-Kudo: This should not be necessary, since `HOME_DIR` applies to `/root` too.

tests/main/selinux-data-context/task.yaml
```diff
@@ -0,0 +1,36 @@
+summary: Check that SELinux file context transitions work
+
+description: |
+    Verify that automatic file context transitions work and cover typical use cases
+
+# TODO: enable centos once we are able to build the policy
+systems: [fedora-*, centos-*]
```
Conan-Kudo: We should be able to build the policy for CentOS now.

tests/upgrade/selinux-relabel/task.yaml
```diff
@@ -0,0 +1,36 @@
+summary: Check that correct SELinux file contexts were assigned on upgrade
+
+# TODO: enable centos once we build the policy there too
+systems: [fedora-*]
```
Conan-Kudo: We should be able to build the policy for CentOS now.

data/selinux/snappy.fc
```diff
@@ -18,30 +18,38 @@
...
-/usr/bin/snapctl		--	gen_context(system_u:object_r:snappy_exec_t,s0)
+/usr/bin/snap			--	gen_context(system_u:object_r:snappy_cli_exec_t,s0)
+/usr/bin/snapctl		--	gen_context(system_u:object_r:snappy_cli_exec_t,s0)
 
 ifdef(`distro_redhat',`
+/usr/libexec/snapd/snapctl		--	gen_context(system_u:object_r:snappy_cli_exec_t,s0)
+/usr/libexec/snapd/snap-confine		--	gen_context(system_u:object_r:snappy_confine_exec_t,s0)
+/usr/libexec/snapd/snap-update-ns		--	gen_context(system_u:object_r:snappy_mount_exec_t,s0)
+/usr/libexec/snapd/snap-discard-ns		--	gen_context(system_u:object_r:snappy_mount_exec_t,s0)
 /usr/libexec/snapd/.*		--	gen_context(system_u:object_r:snappy_exec_t,s0)
```
Conan-Kudo: Would the ordering matter here? Should the generic entry be first, and then the specific entries after?

data/selinux/snappy.fc
```diff
@@ -18,30 +18,38 @@
...
 /etc/sysconfig/snapd		-- 	gen_context(system_u:object_r:snappy_config_t,s0)
 /usr/lib/systemd/system/snapd.* --	gen_context(system_u:object_r:snappy_unit_file_t,s0)
 ')
 
 ifdef(`distro_debian',`
+/usr/lib/snapd/snapctl		--	gen_context(system_u:object_r:snappy_cli_exec_t,s0)
+/usr/lib/snapd/snap-confine		--	gen_context(system_u:object_r:snappy_confine_exec_t,s0)
+/usr/lib/snapd/snap-update-ns		--	gen_context(system_u:object_r:snappy_mount_exec_t,s0)
+/usr/lib/snapd/snap-discard-ns		--	gen_context(system_u:object_r:snappy_mount_exec_t,s0)
 /usr/lib/snapd/.*  		--	gen_context(system_u:object_r:snappy_exec_t,s0)
```
Conan-Kudo: Would the ordering matter here? Should the generic entry be first, and then the specific entries after?

data/selinux/snappy.fc
bboozzoo: I wasn't sure about that. I recall seeing separate rules for both `HOME_DIR` and `/root` in the core policy.

tests/main/selinux-data-context/task.yaml
bboozzoo: Yup, waiting for @sergiocazzolato to greenlight CentOS 7.6. There was some hiccup with `golang` packaging that made building `snapd` fail, but `golang-1.11.2-3.el7` is in EPEL now, so we should be fine.

data/selinux/snappy.fc
bboozzoo: Right here: https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/services/ssh.fc

Conan-Kudo: Ah, okay.

data/selinux/snappy.fc
bboozzoo: No, the order doesn't matter AFAIK. I'd leave the `/usr/libexec/snapd/.*` last to not confuse people.

tests/upgrade/selinux-relabel/task.yaml
Conan-Kudo: This looks like the only thing that's left that needs to be fixed here...

Conan-Kudo: Though since 2.36.3 is out, I should go ahead and build that and update everybody...

bboozzoo: Even better. Ping me when you upload a new update. I'll play with it in a VM and will post karma.

Conan-Kudo: snapd 2.36.3 is in the EPEL7 repo now.

tests/upgrade/selinux-relabel/task.yaml
```diff
@@ -0,0 +1,33 @@
+summary: Check that correct SELinux file contexts were assigned on upgrade
```
stolowski: ❤️

tests/main/selinux-data-context/task.yaml
```diff
@@ -0,0 +1,78 @@
+summary: Check that SELinux file context transitions work
```
stolowski: Great test!

bboozzoo:
@Conan-Kudo I hope to be able to start landing this stuff now. Can you take a look again at this PR and #6238 ?

Conan-Kudo:
🎆

bboozzoo:
Yeah, the first bit is out. Now I need to push the rest.

PR #6271, opened 2018-12-07
tests/lib/pkgdb: disable weak deps on Fedora

bboozzoo:
Fedora images we use come with a number of preinstalled package, among
them gtk3. Those packages are needed to run the tests. The
xdg-desktop-portal-gtk package uses this in the spec:
```
  Supplements:    (gtk3 and (flatpak or snapd))
```
As a result, when snapd is installed, we will unintentionally pull in
xdg-desktop-portal-gtk and its dependencies breaking tests. For this
reason, disable weak deps altogether.

(cc @Conan-Kudo )

### DIFF
--- a/tests/lib/pkgdb.sh
+++ b/tests/lib/pkgdb.sh
@@ -199,6 +199,18 @@ distro_install_package() {
     # arguments as package names.
     APT_FLAGS=
     DNF_FLAGS=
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
The alternative, of course, is to uninstall `gtk3` if we don't need it.

PR #6269, opened 2018-12-06
cmd, dirs, interfaces/apparmor: update distro identification to support ID="archlinux"

bboozzoo:
In filesystem-2018.12-1 Arch is switching its /etc/os-release to:
```
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=archlinux
ANSI_COLOR="0;36"
HOME_URL="https://www.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
```

Before we had:
```
ID=arch
ID_LIKE=archlinux
...
```

Update relevant locations to support the new IDs.

Relevant change in Arch packaging: https://git.archlinux.org/svntogit/packages.git/commit/trunk?h=packages/filesystem&id=6fa932aa685da87e78e0f845bd890645607d302e

PR #6274, opened 2018-12-07
cmd, dirs, interfaces/apparmor: update distro identification to support ID="archlinux" (2.36)

bboozzoo:
Cherry pick of #6269 to 2.36

In filesystem-2018.12-1 Arch is switching its /etc/os-release to:

  NAME="Arch Linux"
  PRETTY_NAME="Arch Linux"
  ID=archlinux
  ANSI_COLOR="0;36"
  HOME_URL="https://www.archlinux.org/"
  SUPPORT_URL="https://bbs.archlinux.org/"
  BUG_REPORT_URL="https://bugs.archlinux.org/"

Before we had:
  ID=arch
  ID_LIKE=archlinux

Update relevant locations to support the new IDs.


zyga:
@Conan-Kudo right? :) At least I'm happy that now Arch follows the os-release spec closer than before so the direction of the change is good.

PR #6281, opened 2018-12-10
interfaces: add multipass-support interface

gerboland:
Multipass is a tool allowing easy creation and management of VMs. This PR is to add the necessary interfaces to allow Multipass to run fully confined.

Multipass has a client/server design - the server (multipassd) will use this `multipass-support` interface. The client can use existing interfaces.

Multipass uses KVM & Qemu to run VMs, qemu-img for image conversion and dnsmasq for IP allocation. Qemu needs careful confinement, but adding those rules to the `multipass-support` policy would make it very broad and thus not very secure. 

Our solution is to emulate the LXD confinement design: we have multipassd generate and load the apparmor profile for each Qemu VM instance, so it is as strictly confined as possible (likewise for dnsmasq and qemu-img). This does mean the `multipass-support` interface grants multipassd the ability to load apparmor profiles and execute child processes under those new profiles.

The qemu profile looks like this:
https://github.com/CanonicalLtd/multipass/pull/498/files#diff-d417c995bccbbc968f5786343d364d0eR81
but we shall request a security team member to review it (and other all profiles) before deploying it as a fully confined multipass snap.


interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,95 @@
...
+change_profile unsafe /var/snap/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}/usr/bin/* -> multipass.*,
+signal (send) peer=multipass.*,
+ptrace (read, trace) peer=multipass.*,
+`
+
+const multipassSupportConnectedPlugSecComp = `
+# Description: allow operating as the Multipass daemon. The Multipass daemon
+# by design requires extensive access to the system and cannot be effectively
+# confined against malicious activity.
+@unrestricted
```
zyga: I'm sceptical that multipass daemon cannot have a meaningful seccomp profile.

gerboland: Ack, I'll give it a look so. Since lxd got away with it, I thought i might too :)

zyga: LXD is different that it runs end user payloads directly on the kernel. Multipass runs a virtual machine that should have a more or less fixed set of system calls, even if they are broad, regardless of the payload inside.

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,108 @@
...
+capability kill, # multipass needs to be able to kill child processes
+capability sys_module, # checking for kvm will involve loading module
+
+# Profiles Multipass generates have a naming scheme. Need unsafe to prevent env scrubbing
+change_profile unsafe /var/snap/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}/usr/bin/* -> multipass.*,
+signal (send) peer=multipass.*,
+ptrace (read, trace) peer=multipass.*,
+`
+
+const multipassSupportConnectedPlugSecComp = `
```
zyga: Not so many system calls after all. Thank you for doing those.

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,108 @@
...
+/etc/apparmor/parser.conf r,
+/etc/apparmor/subdomain.conf r,
+/sys/kernel/security/apparmor/.replace rw,
+/sys/kernel/security/apparmor/.remove w,
+/sys/kernel/security/apparmor/{,**} r,
+
+capability net_admin,
+capability mac_admin,
+capability kill, # multipass needs to be able to kill child processes
+capability sys_module, # checking for kvm will involve loading module
```
zyga: Hmm, I'm mildly sceptical about this. Is there any other way?

interfaces/builtin/multipass.go
```diff
@@ -0,0 +1,51 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package builtin
+
+const multipassSummary = `allows access to the Multipass socket`
+
+const multipassBaseDeclarationSlots = `
+  multipass:
```
zyga: Can you please comment on the design of the policy? Why cannot the multipass interface be installed or connected?

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,108 @@
...
+
+const multipassSupportSummary = `allows operating as the Multipass service`
+
+const multipassSupportBaseDeclarationPlugs = `
+  multipass-support:
+    allow-installation: false
+    deny-auto-connection: true
+`
+
+const multipassSupportBaseDeclarationSlots = `
```
zyga: This is fine

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,108 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package builtin
+
+const multipassSupportSummary = `allows operating as the Multipass service`
+
+const multipassSupportBaseDeclarationPlugs = `
```
zyga: This is fine

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,108 @@
...
+/usr/sbin/aa-exec ixr,
+/sbin/apparmor_parser ixr,
+/etc/apparmor.d/cache/ r,
+/etc/apparmor.d/cache/.features r,
+/etc/apparmor.d/{,cache/}multipass* rw,
+/etc/apparmor.d/tunables/{,**} r,
+/etc/apparmor.d/abstractions/{,**} r,
+/etc/apparmor/parser.conf r,
+/etc/apparmor/subdomain.conf r,
+/sys/kernel/security/apparmor/.replace rw,
```
zyga: Do you really need to `r` here?

interfaces/builtin/multipass_support.go
gerboland: No prob, wasn't as bad as I was expecting

interfaces/builtin/multipass_support.go
gerboland: Nope, you're right, it's a write only interface. I copied it from docker.

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,108 @@
...
+/sys/kernel/security/apparmor/.remove w,
+/sys/kernel/security/apparmor/{,**} r,
+
+capability net_admin,
+capability mac_admin,
+capability kill, # multipass needs to be able to kill child processes
+capability sys_module, # checking for kvm will involve loading module
+
+# Profiles Multipass generates have a naming scheme. Need unsafe to prevent env scrubbing
+change_profile unsafe /var/snap/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}/usr/bin/* -> multipass.*,
```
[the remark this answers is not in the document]
gerboland: I think I'll need to add "{f,}init_module" to the seccomp rules then too...

interfaces/builtin/multipass.go
gerboland: The intent is for Multipass to only provide a stable command line interface. 

We don't want to expose the Multipass server socket to other snaps that might try ship their own multipass client binary, as the binary protocol can and will change.

I believe we can achieve this by preventing the multipass interface from being installed or connected.

gerboland:
I removed private email from launchpad, will pass CLA test on next pass

interfaces/builtin/multipass_support.go
gerboland: Thanks for letting me know about the kmod interface in snapd, it should do the job. I'll update this when ready

zyga: We discussed this briefly and I recommended to use the `kmod` backend where snapd does the module load for the snap.

interfaces/builtin/multipass.go
jdstrand: In addition to the reasons @gerboland gave, we want to mimic the lxd and docker interfaces since access to this socket gives quite privileged access so IMO the base declaration is correctly setup here.

interfaces/builtin/multipass.go
```diff
@@ -0,0 +1,50 @@
...
+    allow-installation: false
+    deny-connection: true
+    deny-auto-connection: true
+`
+
+const multipassConnectedPlugAppArmor = `
+# Description: allow access to the Multipass daemon socket.
```
jdstrand: Can you adjust to include: `This gives privileged access to the system via the mulipass socket API`?

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,105 @@
...
+ */
+
+package builtin
+
+const multipassSupportSummary = `allows operating as the Multipass service`
+
+const multipassSupportBaseDeclarationPlugs = `
+  multipass-support:
+    allow-installation: false
+    deny-auto-connection: true
```
jdstrand: This looks good.

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,105 @@
...
+`
+
+const multipassSupportBaseDeclarationSlots = `
+  multipass-support:
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
```
jdstrand: This looks good.

interfaces/builtin/multipass.go
```diff
@@ -0,0 +1,50 @@
...
+
+package builtin
+
+const multipassSummary = `allows access to the Multipass socket`
+
+const multipassBaseDeclarationSlots = `
+  multipass:
+    allow-installation: false
+    deny-connection: true
+    deny-auto-connection: true
```
jdstrand: While this is mimicking the lxd and docker interfaces, it occurred to me that, iirc, multipass isn't installed outside of the snap, is it? Ie, we could limit this to only being exposed if the slot side provides it.

interfaces/builtin/multipass.go
```diff
@@ -0,0 +1,50 @@
...
+`
+
+const multipassConnectedPlugAppArmor = `
+# Description: allow access to the Multipass daemon socket.
+
+/run/multipass_socket rw,
```
jdstrand: Related to the above question on the base declaration and the providing slot-- is there any reason this is here rather than SNAP_COMMON, a content slot  or similar? You mentioned that the socket protocol is not stable, maybe we don't need to expose the socket in a place for others to use?

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,105 @@
...
+`
+
+const multipassSupportConnectedPlugAppArmor = `
+# Description: allow operating as the Multipass daemon. This policy intentionally
+# allows the Multipass daemon to configure AppArmor, and we expect Multipass to
+# create and apply correct policies for the child processes it spawns.
+
+# Allow socket
+/run/multipass_socket  rw,
```
jdstrand: There is an extra space here.

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,105 @@
...
+capability chown,
+capability fsetid,
```
jdstrand: Is there a reason the daemon is written in a way to require CAP_FSETID? (see man capabilities)

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,105 @@
...
+# Allow it to add these profiles and run the child under that profile.
+/usr/sbin/aa-exec ixr,
+/sbin/apparmor_parser ixr,
+/etc/apparmor.d/cache/ r,
+/etc/apparmor.d/cache/.features r,
+/etc/apparmor.d/{,cache/}multipass* rw,
+/etc/apparmor.d/tunables/{,**} r,
+/etc/apparmor.d/abstractions/{,**} r,
+/etc/apparmor/parser.conf r,
+/etc/apparmor/subdomain.conf r,
```
jdstrand: There are some changes coming in apparmor 2.13 and apparmor 3 that will change some of this. For future compatibility, perhaps collapse this down to:
```
/etc/appamor{,.d}/{,**} r,
```

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,105 @@
...
+/sys/kernel/security/apparmor/.replace w,
+/sys/kernel/security/apparmor/.remove w,
+/sys/kernel/security/apparmor/{,**} r,
```
jdstrand: For style reasons, can you move this real up above the '.replace' rule?

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,105 @@
...
+
+capability net_admin,
```
jdstrand: Why is net_admin needed?

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,105 @@
...
+capability mac_admin,
+capability kill, # multipass needs to be able to kill child processes
```
jdstrand: This capability is for bypassing DAC checks when sending signals to other processes. When is multipass sending signals to other UIDs?

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,105 @@
...
+/etc/apparmor.d/tunables/{,**} r,
+/etc/apparmor.d/abstractions/{,**} r,
+/etc/apparmor/parser.conf r,
+/etc/apparmor/subdomain.conf r,
+/sys/kernel/security/apparmor/.replace w,
+/sys/kernel/security/apparmor/.remove w,
+/sys/kernel/security/apparmor/{,**} r,
+
+capability net_admin,
+capability mac_admin,
```
jdstrand: This is obviously needed since you are loading your own policy. That said, it allows you to modify your own policy. I sort of want to recommend using a child profile, but that doesn't help a lot since you will need to call apparmor_parser to load it and it of course needs this.

We spoke about privilege separation in the multipassd code. Are you doing any of that with regard to loading policy? If so, it would be nice to add a brief comment on that.

interfaces/builtin/multipass_support.go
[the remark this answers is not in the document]
jdstrand: Regarding the apparmor policy allowing module loading, yes, multipassd could write out policy and then aa-exec into it, but as @gerboland said, multipassd would also need to be allowed some syscalls.

In the end, multipassd is going to have to be trusted which is why the base declaration is the way it is. That said, the multipass team is looking to harden the design and implementation in various ways to that the internal interfaces of multipass are separated and less subject to abuse (see my last question regarding privilege separation).

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,105 @@
...
+signal (send) peer=multipass.*,
+ptrace (read, trace) peer=multipass.*,
+`
+
+const multipassSupportConnectedPlugSecComp = `
+# Description: allow operating as the Multipass daemon, and enable its various child
+# processes (qemu, qemu-img, dnsmasq)
+
+# Multipassd needs these. Qemu also could use them, but multipass will execute Qemu
+# with a custom profile that will lock down its filesystem access.
```
jdstrand: This comment seems out of place with the 4 subsequent socket rules. Is it meant to be here? If so, perhaps add a newline and then a new comment about the socket rules?

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,105 @@
...
+accept
+accept4
+bind
+listen
+
+# dnsmasq fails unless it can drop supplementary groups
+setgroups 0 -
+setgroups32 0 -
```
jdstrand: FYI, I don't know if you are patching dnsmasq for this, but calling setgroups() with zero groups (ie, `setgroups(0, NULL)`, which is what you want for the snap) is only guaranteed to work on Linux, so if you patched and that is meant to be portable, then you'll want to check if Linux and call in the above manner, otherwise when not Linux, call with size as '1' and give the egid as the first of the list of 1 supplementary groups.

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,105 @@
...
+
+# multipassd needs to chown its socket to a non-root user
+chown
+chown32
```
jdstrand: Curious for something I'll be working on soon, if we allowed chowning to the `daemon` uid/gid, would that be helpful?

jdstrand:
FYI, I scanned the qemu profile and didn't see anything shocking. Please ping me when your ready to have it reviewed.

interfaces/builtin/multipass_support.go
gerboland: Replaced with kmod support in kvm

interfaces/builtin/multipass.go
gerboland: After another pass, this client interface was completely removed, as the socket is now kept in $SNAP_COMMON

interfaces/builtin/multipass_support.go
gerboland: Good to know. We're just using dnsmasq from the repo in the snap. And we intend only to distribute multipass with the snap. On mac/windows we don't use dnsmasq at all, and don't plan unix support at this time.

interfaces/builtin/multipass_support.go
gerboland: I don't think so, as the multipass client and server will usually be run as different users, and so the socket needs re-owning.

interfaces/builtin/multipass_support.go
gerboland: Yes that's the ultimate goal, privilege separation along with apparmor confinement. I've updated the intro comment to suggest this (is "all possible security meachanisms" too vague?)

I've not got the privilege separation stuff working yet though ([1]). I hope that doesn't delay this review.

[1] https://github.com/CanonicalLtd/multipass/compare/strict-snap-part1?expand=1#diff-87e6e776daf42b26accf0a1d40020cbfR74

interfaces/builtin/multipass_support.go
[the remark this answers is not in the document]
gerboland: Correct, I've gone into the overall architecture in the comment, and it'll use privilege sep along with apparmor to be secure.

gerboland:
Hi folks, I've iterated this a few times and have converged on this. I'd appreciate another pass.

interfaces/builtin/multipass.go
gerboland: This is gone...

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,156 @@
...
+# to confine them all in a single profile would result in an extremely broad AppArmor
+# profile.
+
+# Instead we defer to Multipass the responsibility of generating custom AppArmor
+# profiles for each of these utilities, and trust it launches each utility with
+# all possible security mechanisms enabled. The Multipass daemon itself will run
+# under this restricted policy.
+
+# Therefore this policy intentionally allows the Multipass daemon to configure
+# AppArmor.
```
jdstrand: This comment is great but a little much for a policy comment. I suggest moving it (everything from under Description to here) to a code comment and then expand on the inline comment slightly.

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,156 @@
...
+
+
+# Multipass has a server/client design, using a socket for IPC. The daemon runs
+# as root, but makes the socket accessible to anyone in the sudo group.
+# Need to permit chown and chgrp of the socket.
+capability chown,
+
+# Multipass uses gRPC for IPC, which calls shutdown on the socket
+unix ( shutdown) addr="@snap.@{SNAP_INSTANCE_NAME}.**",
```
jdstrand: There is an unneeded space before `shutdown`. Also, you technically only need one `*` assuming you don't start adding '/' into your abstract socket path. Furthermore, the default template includes `unix (getattr, getopt, setopt, shutdown),` so why is this needed at all?

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,156 @@
...
+
+# Multipass generates AppArmor profiles for the utility processes it spawns.
+/sbin/apparmor_parser ixr,
+/etc/apparmor{,.d}/{,**} r,
+/etc/apparmor.d/{,cache/}multipass* rw,
```
jdstrand: You may not want to store these here in the system file location and instead store them in SNAP_DATA or SNAP_COMMON. Who is multipass managing its cache files? Do you expect VMs/etc service to be launched separately from multipassd?

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,156 @@
...
+change_profile unsafe /var/snap/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}/usr/bin/* -> multipass.*,
+signal (send) peer=multipass.*,
+ptrace (read, trace) peer=multipass.*,
+`
+
+const multipassSupportConnectedPlugSecComp = `
+# Description: allow operating as the Multipass daemon, and enable its various utilites
+# (qemu, qemu-img, dnsmasq, sshfs_server)
+
+# Note that this profile contains the union of all the necessary syscalls each of the
```
jdstrand: s/contains/necessarily contains/

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,156 @@
...
+# for each child process, to further restrict their abilities.
+
+# Multipass has server/client design, using a socket for IPC.
+# Note: Qemu has several features that also open sockets, but the AppArmor profile
+# Multipass will execute Qemu under locks down its filesystem access dramatically
+# and prevents it creating these sockets.
+accept
+accept4
+bind
+listen
```
jdstrand: So, can you not use network-bind for this? (Did we discuss this before?)

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,156 @@
...
+
+# dnsmasq fails unless it can drop supplementary groups
+setgroups 0 -
+setgroups32 0 -
+
+# Multipassd needs to chown its socket to a non-root user. sshfs_server also
+# will need these (see later).
+chown
+chown32
```
jdstrand: Which non-root user? I suggest `daemon` since we are going to start supporting that soonish.

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,156 @@
...
+# syscalls with the knowledge that Multipass will generate a specific AppArmor
+# profile for sshfs_server, restricting any filesystem access to just the
+# user-specified directory.
+
+# The default seccomp profile already includes these syscalls with the note:
+# "until we have syscall arg filtering (LP: #1446748)"
+# Am specifying them here as sshfs_server needs these syscalls unfiltered.
+chmod
+fchmod
+fchmodat
```
jdstrand: These are already allowed in the default template.

interfaces/builtin/multipass_support.go
gerboland: You're correct, it isn't needed, will remove

interfaces/builtin/multipass_support.go
gerboland: You're correct, there's no need to have cache global. Will remove.

interfaces/builtin/multipass_support.go
gerboland: I can, this is my error, will be removed.

interfaces/builtin/multipass_support.go
gerboland: Right. I was paranoid as the default template has these listed with a precursor that reads like "they might be removed in future". I'll remove these and trust they won't :)

interfaces/builtin/multipass_support.go
gerboland: We chgrp the socket to the sudo group (or adm on Arch/Centos), so that a user in that group can run the multipass client without being prompted for a password.

The daemon itself for now will have to retain root privileges as it configures ip and iptables on cleanup, and launching qemu needs root as qemu needs kvm access. Its other utility processes should be possible to run as a less privileged user, `daemon` will be prime candidate.

gerboland:
All review comments addressed, this is ready for another pass please

interfaces/builtin/multipass.go
zyga: Gone or done?

interfaces/builtin/multipass.go
zyga: What is the response to Jamie's question?

interfaces/builtin/multipass_support.go
zyga: Please answer this question.

interfaces/builtin/multipass.go
gerboland: I've relocated the socket to $SNAP_COMMON, so this is unnecessary. That meant a client-specific interface became unecessary, so was removed

interfaces/builtin/multipass_support.go
zyga: Please answer this question as well.

interfaces/builtin/multipass_support.go
gerboland: done

interfaces/builtin/multipass_support.go
gerboland: removed

interfaces/builtin/multipass_support.go
gerboland: Comments added as requested

interfaces/builtin/multipass_support.go
gerboland: removed

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package builtin
+
+/*
+ * Multipass is a tool to create and manage Virtual Machines and their images.
```
zyga: This is a very nice description. Thank you for this.

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,129 @@
...
+ * Each VM runs as a separate "qemu" process (on Linux). VM images are automatically
+ * downloaded, but need conversion using "qemu-img". Networking is provided by
+ * configuring a TUN/TAP network on the host, with DHCP provided by a shared
+ * "dnsmasq" process. File sharing between the VM and the host is provided by a
+ * "sshfs_server" utility. All these utilities are shipped in the snap.
+ *
+ * Each of these utilities have a very different different purpose, and an attempt
```
zyga:
```suggestion
 * Each of these utilities have a different purpose, and an attempt
```

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,129 @@
...
+ * to confine them all in a single profile would result in an extremely broad AppArmor
+ * profile.
+ *
+ * Instead we defer to Multipass the responsibility of generating custom AppArmor
+ * profiles for each of these utilities, and trust it launches each utility with
+ * all possible security mechanisms enabled. The Multipass daemon itself will run
+ * under this restricted policy.
```
zyga: As a meta-comment. I would love to explore a framework where snapd would be aware of this and could be in the loop. Specifically where snapd would receive the profile from multipass, associate it with a security tag (also sent by multipass) and then generate the profiles. I don't mean to block this by any chance but if we grow more systems like that (perhaps LXD already qualifies) I think it could be beneficial to the overall integrity of the sandbox mechanism.

interfaces/builtin/multipass_support.go
zyga: Yes, this looks good.

interfaces/builtin/multipass_support.go
zyga: Agreed.

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,129 @@
...
+# Multipass has a server/client design, using a socket for IPC. The daemon runs
+# as root, but makes the socket accessible to anyone in the sudo group.
+capability chown,
+
+# Multipass will also use privilege separation when running utility processes
+capability setuid,
+capability setgid,
+
+# Some utility process (e.g. dnsmasq) will drop root privileges after startup and
+# change to another user. Multipass needs ability to stop them.
```
zyga: Is _multipass_ here the multipass daemon? Would it still run as root and therefore have the ability to kill them? Can you please clarify this.

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,129 @@
...
+capability kill,
+
+# Profiles Multipass generates have a naming scheme, restrict any profile changes to
+# those matching that scheme. Need unsafe to prevent env scrubbing.
+change_profile unsafe /var/snap/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}/usr/bin/* -> multipass.*,
+signal (send) peer=multipass.*,
```
zyga: Is one-way signal sufficient?

interfaces/builtin/multipass_support.go
```diff
@@ -0,0 +1,129 @@
...
+ptrace (read, trace) peer=multipass.*,
+`
+
+const multipassSupportConnectedPlugSecComp = `
+# Description: allow operating as the Multipass daemon, and enable its various utilites
+# (qemu, qemu-img, dnsmasq, sshfs_server)
+
+# Note that this profile necessarily contains the union of all the syscalls each of the
+# utilities requires. We rely on Multipass to generate specific AppArmor profiles
+# for each child process, to further restrict their abilities.
```
zyga: I'm happy to see that multipass can run under seccomp confinement.

interfaces/builtin/multipass_support_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+var _ = Suite(&MultipassSupportInterfaceSuite{
+	iface: builtin.MustInterface("multipass-support"),
+})
+
+func (s *MultipassSupportInterfaceSuite) SetUpTest(c *C) {
+	s.slotInfo = &snap.SlotInfo{
+		Snap: &snap.Info{
+			SuggestedName: "core",
+			Type:          snap.TypeOS},
```
zyga: Nitpick: unusual formatting, perhaps this could fit on one line?

interfaces/builtin/multipass_support_test.go
```diff
@@ -0,0 +1,110 @@
...
+	c.Assert(err, IsNil)
+	c.Assert(seccompSpec.Snippets(), HasLen, 1)
+}
+
+func (s *MultipassSupportInterfaceSuite) TestConnectedPlugSnippet(c *C) {
+	apparmorSpec := &apparmor.Specification{}
+	err := apparmorSpec.AddConnectedPlug(s.iface, s.plug, s.slot)
+	c.Assert(err, IsNil)
+	c.Assert(apparmorSpec.SecurityTags(), DeepEquals, []string{"snap.multipass.app"})
+	c.Assert(apparmorSpec.SnippetForTag("snap.multipass.app"), testutil.Contains, "/sbin/apparmor_parser ixr,\n")
```
zyga: Thank you for checking for those.

interfaces/builtin/multipass_support.go
gerboland: Yes agreed, I would like to see something like this.

interfaces/builtin/multipass_support.go
gerboland: Me too :)

interfaces/builtin/multipass_support_test.go
gerboland: np

interfaces/builtin/multipass_support_test.go
gerboland: Fixed. I see now gofmt doesn't add/remove lines to improve formatting (clang-format for C++ does)

interfaces/builtin/multipass_support.go
gerboland: Yep, I meant the multipass daemon (comment updated to clarify). The daemon runs continually as root as it needs to:
- launch qemu as root (unfortunate, but I don't see alternative until snapd can ensure a kvm user exists which permits access to /dev/kvm)
- change networking interfaces and firewall rules at any time (and clean them up on shutdown)

interfaces/builtin/multipass_support.go
gerboland: It is, I see no reason for the daemon to want to receive a signal. Any of the child process that the multipass daemon spawns under its own apparmor rules have a corresponding "receive", e.g.:
https://github.com/CanonicalLtd/multipass/blob/strict-snap-part1/src/platform/backends/qemu/qemu_process_spec.cpp#L108
(note that branch hasn't landed, and the apparmor profiles multipass will be generating will be reviewed before landing)

interfaces/builtin/multipass_support.go
gerboland: ta

interfaces/builtin/multipass_support.go
gerboland: Maybe it's too strong coupling, but multipass could work with something like this: the snapd multipass interface contains AA & seccomp profile templates, which have placeholders for strings/file paths. snapd would be able to check those strings/file paths for safety. 

Multipass could then ask snapd to run a process it specifies, under those AA/seccomp profiles with placeholders it requests.

interfaces/builtin/multipass.go
gerboland: Gone. This was a multipass client interface, which became unnecessary after refactoring.

gerboland:
> Thank you for the extensive work that went into this interface.
> 
> LGTM, I left some comments (there's a suggestion that fixes one repeated word inline) and some questions but I think this is good as is. Please feel free to merge when you feel like it.

Thanks for the review pass, I hope I've answered or addressed your concerns. 
Note I don't have ability to merge, when you are happy I'd appreciate you merging it.

interfaces/builtin/multipass_support.go
jdstrand: Whenever you are at the point of thinking through this, please include me. I think I'd rather see snapd providing the template for the VMs and then multipassd asking snapd to load one with a given profile name. Then multipassd doesn't need mac_admin at all and it isn't supplying any policy itself.

PR #6282, opened 2018-12-10
tests: force test-snapd-daemon-notify exit 0 when the interface is not connected

sergiocazzolato:
This snap is exiting 1 when the interface is not connected, so systemd
retries trying to start the service making reach the start-limit-hit.

This test is mostly failing on boards where the delay between the snap
installation and the interface is connected is much higher than in cloud
images.


chipaca:
@bboozzoo it's a race between systemd giving up on the service, and the test script connecting the interface

PR #6284, opened 2018-12-11
spread: make opensuse-42.3 manual

bboozzoo:
openSUSE 42.3 does not build again. Needs investigation.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -83,6 +83,7 @@ backends:
                 workers: 4
             - opensuse-42.3-64:
                 workers: 4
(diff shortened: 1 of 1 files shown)


spread.yaml
```diff
@@ -83,6 +83,7 @@ backends:
                 workers: 4
             - opensuse-42.3-64:
                 workers: 4
+                manual: true
```
zyga:
```suggestion
                manual: true // golang stack cannot compile anything, needs investigation
```

PR #6286, opened 2018-12-11
release: support probing SELinux state

bboozzoo:
The PR adds support for probing SELinux status.

Contains #6283 and #6285. Please review those first. The changes relevant to the PR are in release package only.


release/selinux.go
```diff
@@ -0,0 +1,79 @@
...
+var (
+	selinuxLevel   SELinuxLevelType
+	selinuxSummary string
+
+	selinuxIsEnabled   = selinux.IsEnabled
+	selinuxIsEnforcing = selinux.IsEnforcing
+)
+
+func init() {
+	selinuxLevel, selinuxSummary = probeSELinux()
```
zyga: Can this be done lazily please

Conan-Kudo: Why?

zyga: Because importing release doesn't mean you need to know the status of SELinux or Apparmor or Seccomp or anything else. There are functions for that.

bboozzoo: What makes me think all the slow, feature testing pieces, should probably be in a separate package.

zyga: I tried that with apparmor making it a separate package but we ended up putting it here in release.go.

release/selinux.go
```diff
@@ -0,0 +1,90 @@
...
+	if !enabled {
+		return NoSELinux, ""
+	}
+
+	enforcing, err := selinuxIsEnforcing()
+	if err != nil {
+		return NoSELinux, fmt.Sprintf("SELinux is enabled, but status cannot be determined: %v", err)
+	}
+	if !enforcing {
+		return SELinuxPermissive, "SELinux is enabled and in permissive mode"
```
zyga: Just a small nitpick about wording in this case:

```suggestion
		return SELinuxPermissive, "SELinux is enabled but in permissive mode"
```

release/selinux_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+func (s *selinuxSuite) TestProbePermissive(c *C) {
+	restore := release.MockSELinuxIsEnabled(func() (bool, error) { return true, nil })
+	defer restore()
+	restore = release.MockSELinuxIsEnforcing(func() (bool, error) { return false, nil })
+	defer restore()
+
+	level, status := release.ProbeSELinux()
+	c.Assert(level, Equals, release.SELinuxPermissive)
+	c.Assert(status, Equals, "SELinux is enabled and in permissive mode")
```
zyga: To match the suggestion above:

```suggestion
	c.Assert(status, Equals, "SELinux is enabled but in permissive mode")
```

PR #6288, opened 2018-12-11
cmd/snap-confine: refactor call to snap-update-ns --user-mounts

zyga:
Having introduced tool helper module we have cut a significant amount of
boiler-plate code related to calling snap-update-ns and snap-discard-ns.
Somehow I totally neglected the call to snap-update-ns --user-mounts,
so here it is.

### DIFF
--- a/cmd/libsnap-confine-private/tool.c
+++ b/cmd/libsnap-confine-private/tool.c
@@ -105,7 +105,7 @@ void sc_call_snap_update_ns_as_user(int snap_update_ns_fd,
 			 snap_name);
 
 	const char *xdg_runtime_dir = getenv("XDG_RUNTIME_DIR");
-	char xdg_runtime_dir_env[PATH_MAX];
+	char xdg_runtime_dir_env[PATH_MAX+strlen("XDG_RUNTIME_DIR=")];
 	if (xdg_runtime_dir != NULL) {
 		sc_must_snprintf(xdg_runtime_dir_env,
 				 sizeof(xdg_runtime_dir_env),
--- a/cmd/libsnap-confine-private/tool.c
+++ b/cmd/libsnap-confine-private/tool.c
@@ -72,7 +72,7 @@ int sc_open_snap_update_ns(void)
 }
 
 void sc_call_snap_update_ns(int snap_update_ns_fd, const char *snap_name,
-			    struct sc_apparmor *apparmor, bool is_user_profile)
+			    struct sc_apparmor *apparmor)
 {
 	char *snap_name_copy SC_CLEANUP(sc_cleanup_string) = NULL;
 	snap_name_copy = sc_strdup(snap_name);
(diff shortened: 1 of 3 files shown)


cmd/libsnap-confine-private/tool.c
```diff
@@ -72,20 +72,52 @@ int sc_open_snap_update_ns(void)
...
-	char profile[PATH_MAX] = { 0 };
-	sc_must_snprintf(profile, sizeof profile, "snap-update-ns.%s",
+
+	char aa_profile[PATH_MAX] = { 0 };
+	sc_must_snprintf(aa_profile, sizeof aa_profile, "snap-update-ns.%s",
 			 snap_name);
-	sc_call_snapd_tool_with_apparmor(snap_update_ns_fd, "snap-update-ns",
-					 apparmor, profile, argv, envp);
+
+	if (is_user_profile) {
```
bboozzoo: You could refactor this to:
```
char *argv[] = {
	"snap-update-ns",
	/* This tells snap-update-ns we are calling from snap-confine and locking is in place */
	"--from-snap-confine",
	snap_name_copy, NULL
};
char *envp[] =  { "SNAPD_DEBUG=x", NULL };
if (is_user_profile) {
    argv = ...;
    envp = ....;
}
sc_call_snapd_tool_with_apparmor(snap_update_ns_fd,
				 "snap-update-ns", apparmor,
				 aa_profile, argv, envp);
```

cmd/snap-confine/mount-support.c
```diff
@@ -699,55 +699,5 @@ void sc_setup_user_mounts(struct sc_apparmor *apparmor, int snap_update_ns_fd,
...
-	if (waitpid(child, &status, 0) < 0) {
-		die("waitpid() failed for snap-update-ns process");
-	}
-	if (WIFEXITED(status) && WEXITSTATUS(status) != 0) {
-		die("snap-update-ns failed with code %i", WEXITSTATUS(status));
-	} else if (WIFSIGNALED(status)) {
-		die("snap-update-ns killed by signal %i", WTERMSIG(status));
-	}
-	debug("snap-update-ns finished successfully");
+	sc_call_snap_update_ns(snap_update_ns_fd, snap_name, apparmor, true);
```
bboozzoo: There is always some confusion with bare true/false values. Wonder if we could use a wrapper with more explicit name, or maybe replace the bool with a more descriptive enum.

cmd/libsnap-confine-private/tool.c
zyga: Tried that initially and it turned out to be worse because of how scoping and memory management works.

cmd/snap-confine/mount-support.c
zyga: I can easily fork this to two functions. No strong opinions about it.

cmd/libsnap-confine-private/tool.c
```diff
@@ -76,16 +76,59 @@ void sc_call_snap_update_ns(int snap_update_ns_fd, const char *snap_name,
...
+	snap_name_copy = sc_strdup(snap_name);
+
+	char aa_profile[PATH_MAX] = { 0 };
+	sc_must_snprintf(aa_profile, sizeof aa_profile, "snap-update-ns.%s",
 			 snap_name);
-	sc_call_snapd_tool_with_apparmor(snap_update_ns_fd, "snap-update-ns",
-					 apparmor, profile, argv, envp);
+
+	const char *xdg_runtime_dir = getenv("XDG_RUNTIME_DIR");
+	char xdg_runtime_dir_env[PATH_MAX];
```
stolowski: Since we're putting `XDG_RUNTIME_DIR=%s` in here, I think it should be PATH_MAX + len(XDG_RUNTIME_DIR) to be on the safe side (I know `sc_must_snprintf` will die and not overflow, nonethelss I think the intention should be made clear when declaring size of the buffer).

zyga:
```suggestion
	char xdg_runtime_dir_env[PATH_MAX+strlen("XDG_RUNTIME_DIR=")];
```

PR #6292, opened 2018-12-12
spread: record each tests/upgrade job

bboozzoo:
Spread jobs under tests/upgrade/ suite were not being recorded in the 'runs'
file.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -636,6 +636,12 @@ suites:
         # autopkgtest runs against localhost which causes problems with
         # this test prepare
         backends: [-autopkgtest]
+        prepare-each: |
+            # shellcheck source=tests/lib/state.sh
+            "$TESTSLIB"/state.sh
+            mkdir -p "$RUNTIME_STATE_PATH"
+            # save the job which is going to be executed in the system
+            echo -n "$SPREAD_JOB " >> "$RUNTIME_STATE_PATH/runs"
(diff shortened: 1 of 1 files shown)


spread.yaml
```diff
@@ -636,6 +636,12 @@ suites:
         # autopkgtest runs against localhost which causes problems with
         # this test prepare
         backends: [-autopkgtest]
+        prepare-each: |
+            # shellcheck source=tests/lib/state.sh
+            "$TESTSLIB"/state.sh
+            mkdir -p "$RUNTIME_STATE_PATH"
+            # save the job which is going to be executed in the system
+            echo -n "$SPREAD_JOB " >> "$RUNTIME_STATE_PATH/runs"
```
stolowski: This is done in `prepare-restore.sh` in `prepare_suite_each`, isn't it effective for these tests?

bboozzoo: We are not running `"$TESTSLIB"/prepare-restore.sh --prepare-suite-each` for `tests/upgrade/` suite for some undocumented reason. I briefly looked inside `prepare_suite_each` and it probably needs some updates before it can be used here. With how brittle the test setup is, I didn't want to add these changes in the PR.

bboozzoo:
Ive merged this branch into #6270. The actual verification should confirm my suspicions why that other PR fails.

PR #6294, opened 2018-12-12
packaging/ubuntu: build with golang 1.10

zyga:
This patch changes build dependency for snapd to pull in at least go
1.10. This way we can get off the 1.6 bandwagon (which has stopped and
is not going anywhere) and get the benefit of new and exciting features
of go 1.10.

The real hard code compatibility is go 1.9, up from 1.6 but not fully up
to 1.10 because some distributions are still used with only 1.9
available.

This branch _may_ help in chasing a core18 bug related to socket and
seeding. Fingers crossed.


zyga:
This passed, surprisingly so.

packaging/ubuntu-16.04/control
```diff
@@ -17,7 +17,7 @@ Build-Depends: autoconf,
                gettext,
                grub-common,
                gnupg2,
-               golang-any (>=2:1.6) | golang-1.6,
+               golang-any (>=2:1.10) | golang-1.10,
```
bboozzoo: Yay!

mvo5:
Very cool! We can merge this only after 2.37 is released (to make sure 2.37 does not get delayed because of this) but its great that this looks so straightforward.

PR #6298, opened 2018-12-13
packaging/{fedora,opensuse}: own /var/lib/snapd/cookie

bboozzoo:
Otherwise the directory won't be removed with the package.

cc @Conan-Kudo you may want to pick this up for the spec too.


Conan-Kudo:
@zyga That wouldn't fix making sure the directories were owned properly in the package. Empty directories would just get filtered out if they're not listed in `%files`.

PR #6299, opened 2018-12-13
travis: short circuit failures in static and unit tests travis job

bboozzoo:
It's super annoying to have the Travis job execute all scripts, while the first
one has failed. This makes it hard to find out the failure that occurred in
--static checks as it doesn't really stand out and it followed by a couple of
pages of unit tests log.

Alternative to #6209


bboozzoo:
@chipaca pushed a change you suggested

PR #6300, opened 2018-12-13
travis: use more recent version of gofmt

bboozzoo:
Use a recent gofmt to workaround people working with the code using recent versions of Go that may do the formatting differently.


bboozzoo:
The rules of formatting seem to handled elsewhere.

PR #6318, opened 2018-12-18
release-tools: display self-help

zyga:
Doing the release once in blue moon, one would appreciate the release
tools hand-holding one through the process.

### DIFF
--- a/release-tools/repack-debian-tarball.sh
+++ b/release-tools/repack-debian-tarball.sh
@@ -23,6 +23,8 @@ set -ue
 debian_tarball="${1:-}"
 if [ "$debian_tarball" = "" ]; then
 	echo "Usage: repack-debian-tarball.sh <snapd-debian-tarball>"
+	echo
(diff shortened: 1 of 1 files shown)


release-tools/repack-debian-tarball.sh
```diff
@@ -23,6 +23,8 @@ set -ue
 debian_tarball="${1:-}"
 if [ "$debian_tarball" = "" ]; then
 	echo "Usage: repack-debian-tarball.sh <snapd-debian-tarball>"
+	echo
+	head -n 19 "$0" | tail -n 18 | sed -e 's/#[ ]*//g'
```
mvo5: Maybe a small variation that avoids the 19,18 contants? Just print everything until the first non `#`. Something like:
```
sed -n '2,/^[^#]/p' $0 | sed -e 's/#[ ]*//g' | head -n-1
```
(the tail at the end is a bit ugly but my sed foo is not that strong to figure out how to skip it).

zyga: I made something even simpler

PR #6325, opened 2019-01-04
Allowing avahi-observer/control slots from app snap also on classic

kubiko:
On classis assumption is avahi is installed as deb and provided by default
however this if often not case for server images

From avahi snap metrics, this assumption turned up to be flawed.
Since avahi socked is now moved to same place as with deb, avahi snap does not have sufficient permissions to run in classic systems

Signed-off-by: Ondrej Kubik <ondrej.kubik@canonical.com>

### DIFF
--- a/interfaces/builtin/avahi_control.go
+++ b/interfaces/builtin/avahi_control.go
@@ -25,7 +25,6 @@ import (
 	"github.com/snapcore/snapd/interfaces"
 	"github.com/snapcore/snapd/interfaces/apparmor"
 	"github.com/snapcore/snapd/interfaces/dbus"
-	"github.com/snapcore/snapd/release"
 	"github.com/snapcore/snapd/snap"
 )
 
@@ -119,9 +118,11 @@ func (iface *avahiControlInterface) StaticInfo() interfaces.StaticInfo {
 func (iface *avahiControlInterface) AppArmorConnectedPlug(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
 	old := "###SLOT_SECURITY_TAGS###"
 	var new string
-	if release.OnClassic {
-		// If we're running on classic Avahi will be part
-		// of the OS snap and will run unconfined.
+	// If we're running on classic, Avahi may be installed either as a snap of
+	// as part of the OS. If it is part of the OS, it will not have a security
+	// label like it would when installed as a snap.
+	if implicitSystemConnectedSlot(slot) {
+		// avahi from the OS is typically unconfined but known to sometimes be confined
 		new = "unconfined"
 	} else {
 		new = slotAppLabelExpr(slot)
@@ -135,7 +136,9 @@ func (iface *avahiControlInterface) AppArmorConnectedPlug(spec *apparmor.Specifi
 }
 
 func (iface *avahiControlInterface) AppArmorPermanentSlot(spec *apparmor.Specification, slot *snap.SlotInfo) error {
-	if !release.OnClassic {
+	// Only apply slot snippet when running as application snap
+	// on classic, slot side can be system or application
+	if !implicitSystemPermanentSlot(slot) {
 		// NOTE: this is using avahi-observe permanent slot as it contains
 		// base declarations for running as the avahi service.
 		spec.AddSnippet(avahiObservePermanentSlotAppArmor)
@@ -144,7 +147,9 @@ func (iface *avahiControlInterface) AppArmorPermanentSlot(spec *apparmor.Specifi
 }
 
 func (iface *avahiControlInterface) AppArmorConnectedSlot(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
-	if !release.OnClassic {
+	// Only apply slot snippet when running as application snap
+	// on classic, slot side can be system or application
+	if !implicitSystemConnectedSlot(slot) {
 		old := "###PLUG_SECURITY_TAGS###"
 		new := plugAppLabelExpr(plug)
 		// avahi-control implies avahi-observe, so add snippets for both here
@@ -157,7 +162,9 @@ func (iface *avahiControlInterface) AppArmorConnectedSlot(spec *apparmor.Specifi
 }
 
 func (iface *avahiControlInterface) DBusPermanentSlot(spec *dbus.Specification, slot *snap.SlotInfo) error {
-	if !release.OnClassic {
+	// Only apply slot snippet when running as application snap
+	// on classic, slot side can be system or application
+	if !implicitSystemPermanentSlot(slot) {
 		// NOTE: this is using avahi-observe permanent slot as it contains
 		// base declarations for running as the avahi service.
 		spec.AddSnippet(avahiObservePermanentSlotDBus)
--- a/interfaces/builtin/avahi_control.go
+++ b/interfaces/builtin/avahi_control.go
@@ -123,7 +123,7 @@ func (iface *avahiControlInterface) AppArmorConnectedPlug(spec *apparmor.Specifi
 	// label like it would when installed as a snap.
 	if implicitSystemConnectedSlot(slot) {
 		// avahi from the OS is typically unconfined but known to sometimes be confined
-		new = "unconfined"
+		new = "\"{unconfined,/usr/sbin/avahi-daemon}\""
 	} else {
 		new = slotAppLabelExpr(slot)
 	}
--- a/interfaces/builtin/avahi_observe.go
+++ b/interfaces/builtin/avahi_observe.go
@@ -434,7 +434,7 @@ func (iface *avahiObserveInterface) AppArmorConnectedPlug(spec *apparmor.Specifi
 	// label like it would when installed as a snap.
 	if implicitSystemConnectedSlot(slot) {
 		// avahi from the OS is typically unconfined but known to sometimes be confined
-		new = "unconfined"
+		new = "\"{unconfined,/usr/sbin/avahi-daemon}\""
 	} else {
 		new = slotAppLabelExpr(slot)
 	}
--- a/interfaces/builtin/avahi_observe.go
+++ b/interfaces/builtin/avahi_observe.go
@@ -25,7 +25,6 @@ import (
 	"github.com/snapcore/snapd/interfaces"
 	"github.com/snapcore/snapd/interfaces/apparmor"
 	"github.com/snapcore/snapd/interfaces/dbus"
-	"github.com/snapcore/snapd/release"
 	"github.com/snapcore/snapd/snap"
 )
 
@@ -430,9 +429,11 @@ func (iface *avahiObserveInterface) StaticInfo() interfaces.StaticInfo {
 func (iface *avahiObserveInterface) AppArmorConnectedPlug(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
 	old := "###SLOT_SECURITY_TAGS###"
 	var new string
-	if release.OnClassic {
-		// If we're running on classic Avahi will be part
-		// of the OS snap and will run unconfined.
+	// If we're running on classic, Avahi may be installed either as a snap of
+	// as part of the OS. If it is part of the OS, it will not have a security
+	// label like it would when installed as a snap.
+	if implicitSystemConnectedSlot(slot) {
+		// avahi from the OS is typically unconfined but known to sometimes be confined
 		new = "unconfined"
 	} else {
 		new = slotAppLabelExpr(slot)
@@ -443,14 +444,18 @@ func (iface *avahiObserveInterface) AppArmorConnectedPlug(spec *apparmor.Specifi
 }
 
 func (iface *avahiObserveInterface) AppArmorPermanentSlot(spec *apparmor.Specification, slot *snap.SlotInfo) error {
-	if !release.OnClassic {
+	// Only apply slot snippet when running as application snap
+	// on classic, slot side can be system or application
+	if !implicitSystemPermanentSlot(slot) {
 		spec.AddSnippet(avahiObservePermanentSlotAppArmor)
 	}
 	return nil
 }
 
 func (iface *avahiObserveInterface) AppArmorConnectedSlot(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
-	if !release.OnClassic {
+	// Only apply slot snippet when running as application snap
+	// on classic, slot side can be system or application
+	if !implicitSystemConnectedSlot(slot) {
 		old := "###PLUG_SECURITY_TAGS###"
 		new := plugAppLabelExpr(plug)
 		snippet := strings.Replace(avahiObserveConnectedSlotAppArmor, old, new, -1)
@@ -460,7 +465,9 @@ func (iface *avahiObserveInterface) AppArmorConnectedSlot(spec *apparmor.Specifi
 }
 
 func (iface *avahiObserveInterface) DBusPermanentSlot(spec *dbus.Specification, slot *snap.SlotInfo) error {
-	if !release.OnClassic {
+	// Only apply slot snippet when running as application snap
+	// on classic, slot side can be system or application
+	if !implicitSystemPermanentSlot(slot) {
 		spec.AddSnippet(avahiObservePermanentSlotDBus)
 	}
 	return nil
--- a/interfaces/builtin/utils.go
+++ b/interfaces/builtin/utils.go
@@ -25,6 +25,7 @@ import (
 	"sort"
 
 	"github.com/snapcore/snapd/interfaces"
+	"github.com/snapcore/snapd/release"
 	"github.com/snapcore/snapd/snap"
 )
 
@@ -95,4 +96,27 @@ func sanitizeSlotReservedForOSOrApp(iface interfaces.Interface, slot *snap.SlotI
 		return fmt.Errorf("%s slots are reserved for the core and app snaps", iface.Name())
 	}
 	return nil
+}
+
+// determine if permanent slot side is provided by the system
+// on classic system some implicit slots can be provided by system or by
+// application snap e.g. avahi (it can be installed as deb or snap)
+// - slot owned by the system (core,snapd snap)  usually requires no action
+// - slot owned by application snap typically requires rules update
+func implicitSystemPermanentSlot(slot *snap.SlotInfo) bool {
+	if release.OnClassic &&
+		(slot.Snap.Type == snap.TypeOS || slot.Snap.Type == snap.TypeSnapd) {
+		return true
+	}
+	return false
+}
+
+// determine if connected slot side is provided by the system
+// as for isPermanentSlotSystemSlot() slot can be owned by app or system
+func implicitSystemConnectedSlot(slot *interfaces.ConnectedSlot) bool {
+	if release.OnClassic &&
+		(slot.Snap().Type == snap.TypeOS || slot.Snap().Type == snap.TypeSnapd) {
+		return true
+	}
+	return false
 }
--- a/interfaces/builtin/utils_test.go
+++ b/interfaces/builtin/utils_test.go
@@ -32,19 +32,25 @@ import (
 )
 
 type utilsSuite struct {
-	iface      interfaces.Interface
-	slotOS     *snap.SlotInfo
-	slotApp    *snap.SlotInfo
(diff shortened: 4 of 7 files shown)


kubiko:
For reference here is forum post about fail
https://forum.snapcraft.io/t/avahi-daemon-wont-start/8990/2
For reference, from snap metrics I can see over 90% of installs being on classic system (which is now broken since we moved socket to more predictable place)

interfaces/builtin/avahi_control.go
```diff
@@ -120,9 +120,12 @@ func (iface *avahiControlInterface) AppArmorConnectedPlug(spec *apparmor.Specifi
 	old := "###SLOT_SECURITY_TAGS###"
 	var new string
 	if release.OnClassic {
-		// If we're running on classic Avahi will be part
+		// If we're running on classic Avahi can be installed as snap or be part
 		// of the OS snap and will run unconfined.
-		new = "unconfined"
+		new = slotAppLabelExpr(slot)
+		if new == "\"snap.core.*\"" {
```
jdstrand: This doesn't look like it would work for core16/core18-based systems...

interfaces/builtin/avahi_control.go
```diff
@@ -120,9 +120,12 @@ func (iface *avahiControlInterface) AppArmorConnectedPlug(spec *apparmor.Specifi
...
+			new = "unconfined"
+		}
 	} else {
 		new = slotAppLabelExpr(slot)
 	}
```
jdstrand: This if/else could be reorganized to be a little more clean I think with:
```
new = slotAppLabelExpr(slot)
if release.OnClassic && new == "\"snap.core.*\"" { # fix for core16/core18
	new = "unconfined"
}
```

interfaces/builtin/avahi_control.go
```diff
@@ -135,33 +138,27 @@ func (iface *avahiControlInterface) AppArmorConnectedPlug(spec *apparmor.Specifi
...
 
 func (iface *avahiControlInterface) AppArmorPermanentSlot(spec *apparmor.Specification, slot *snap.SlotInfo) error {
-	if !release.OnClassic {
-		// NOTE: this is using avahi-observe permanent slot as it contains
-		// base declarations for running as the avahi service.
-		spec.AddSnippet(avahiObservePermanentSlotAppArmor)
-	}
+	// NOTE: this is using avahi-observe permanent slot as it contains
+	// base declarations for running as the avahi service.
+	spec.AddSnippet(avahiObservePermanentSlotAppArmor)
```
jdstrand: The PermanentSlot policy should not be applied when the slotting snap is core, core18, etc.

interfaces/builtin/avahi_control.go
```diff
@@ -135,33 +138,27 @@ func (iface *avahiControlInterface) AppArmorConnectedPlug(spec *apparmor.Specifi
...
-		snippet = strings.Replace(avahiControlConnectedSlotAppArmor, old, new, -1)
-		spec.AddSnippet(snippet)
-	}
+	old := "###PLUG_SECURITY_TAGS###"
+	new := plugAppLabelExpr(plug)
+	// avahi-control implies avahi-observe, so add snippets for both here
+	snippet := strings.Replace(avahiObserveConnectedSlotAppArmor, old, new, -1)
+	spec.AddSnippet(snippet)
+	snippet = strings.Replace(avahiControlConnectedSlotAppArmor, old, new, -1)
+	spec.AddSnippet(snippet)
```
jdstrand: Likewise, the ConnectedSlot policy should not be applied when the slotting snap is core, core18, etc.

interfaces/builtin/avahi_control.go
```diff
@@ -135,33 +138,27 @@ func (iface *avahiControlInterface) AppArmorConnectedPlug(spec *apparmor.Specifi
...
 
 func (iface *avahiControlInterface) DBusPermanentSlot(spec *dbus.Specification, slot *snap.SlotInfo) error {
-	if !release.OnClassic {
-		// NOTE: this is using avahi-observe permanent slot as it contains
-		// base declarations for running as the avahi service.
-		spec.AddSnippet(avahiObservePermanentSlotDBus)
-	}
+	// NOTE: this is using avahi-observe permanent slot as it contains
+	// base declarations for running as the avahi service.
+	spec.AddSnippet(avahiObservePermanentSlotDBus)
```
jdstrand: Likewise, the PermanentSlot policy for dbus should not be applied when the slotting snap is core, core18, etc. The deb/rpm will already have dbus policy on the system and we shouldn't try to override that.

interfaces/builtin/avahi_observe.go
```diff
@@ -431,9 +431,12 @@ func (iface *avahiObserveInterface) AppArmorConnectedPlug(spec *apparmor.Specifi
...
 	var new string
 	if release.OnClassic {
-		// If we're running on classic Avahi will be part
+		// If we're running on classic Avahi can be installed as snap or be part
 		// of the OS snap and will run unconfined.
-		new = "unconfined"
+		new = slotAppLabelExpr(slot)
+		if new == "\"snap.core.*\"" {
+			new = "unconfined"
+		}
```
jdstrand: See comments for avahi_control.go.

interfaces/builtin/avahi_observe.go
```diff
@@ -443,26 +446,20 @@ func (iface *avahiObserveInterface) AppArmorConnectedPlug(spec *apparmor.Specifi
 }
 
 func (iface *avahiObserveInterface) AppArmorPermanentSlot(spec *apparmor.Specification, slot *snap.SlotInfo) error {
-	if !release.OnClassic {
-		spec.AddSnippet(avahiObservePermanentSlotAppArmor)
-	}
+	spec.AddSnippet(avahiObservePermanentSlotAppArmor)
```
jdstrand: See comments for avahi_control.go.

interfaces/builtin/avahi_observe.go
```diff
@@ -443,26 +446,20 @@ func (iface *avahiObserveInterface) AppArmorConnectedPlug(spec *apparmor.Specifi
...
-	if !release.OnClassic {
-		old := "###PLUG_SECURITY_TAGS###"
-		new := plugAppLabelExpr(plug)
-		snippet := strings.Replace(avahiObserveConnectedSlotAppArmor, old, new, -1)
-		spec.AddSnippet(snippet)
-	}
+	old := "###PLUG_SECURITY_TAGS###"
+	new := plugAppLabelExpr(plug)
+	snippet := strings.Replace(avahiObserveConnectedSlotAppArmor, old, new, -1)
+	spec.AddSnippet(snippet)
```
jdstrand: See comments for avahi_control.go.

interfaces/builtin/avahi_observe.go
```diff
@@ -443,26 +446,20 @@ func (iface *avahiObserveInterface) AppArmorConnectedPlug(spec *apparmor.Specifi
...
 	return nil
 }
 
 func (iface *avahiObserveInterface) DBusPermanentSlot(spec *dbus.Specification, slot *snap.SlotInfo) error {
-	if !release.OnClassic {
-		spec.AddSnippet(avahiObservePermanentSlotDBus)
-	}
+	spec.AddSnippet(avahiObservePermanentSlotDBus)
```
jdstrand: See comments for avahi_control.go.

om26er:
@kubiko Do you think you can revert the avahi snap to create the sock in a "non-standard" directory for now ? The snap is currently broken on classic, which as a result breaks my project that is supposed to be discoverable on the local network (both UbuntuCore and Classic)

kubiko:
@jdstrand thanks for great comments and chat about right way to handle this. I realised previous implementation wasn't really approaching it right way, it went from one extreme to another.
I added now helper functions to determine this specific case when slot is implicit on classic, but could be also provided by app snap. Just as you suggested.
I also moved from using snap names and rule to determine slot provider and instead used snap type, this seems more reliable approach going forward

kubiko:
@om26er let's give it second try before reverting

kubiko:
@stolowski can you please check chase I did in `interfaces/builtin/utils_test.go`
I updated slotSnapd type from `snap.TypeApp` -> `snap.TypeApp`
```- slotSnapd:  &snap.SlotInfo{Snap: &snap.Info{Type: snap.TypeApp, SuggestedName: "snapd"}},
+ slotSnapd:    &snap.SlotInfo{Snap: &snap.Info{Type: snap.TypeSnapd, SuggestedName: "snapd"}},

File: interfaces/builtin/utils_test.go [hunk 10]
```diff
@@ -32,19 +32,25 @@ import (
...
 var _ = Suite(&utilsSuite{
-	iface:      &ifacetest.TestInterface{InterfaceName: "iface"},
-	slotOS:     &snap.SlotInfo{Snap: &snap.Info{Type: snap.TypeOS}},
-	slotApp:    &snap.SlotInfo{Snap: &snap.Info{Type: snap.TypeApp}},
-	slotSnapd:  &snap.SlotInfo{Snap: &snap.Info{Type: snap.TypeApp, SuggestedName: "snapd"}},
-	slotGadget: &snap.SlotInfo{Snap: &snap.Info{Type: snap.TypeGadget}},
+	iface:        &ifacetest.TestInterface{InterfaceName: "iface"},
+	slotOS:       &snap.SlotInfo{Snap: &snap.Info{Type: snap.TypeOS}},
+	slotApp:      &snap.SlotInfo{Snap: &snap.Info{Type: snap.TypeApp}},
+	slotSnapd:    &snap.SlotInfo{Snap: &snap.Info{Type: snap.TypeSnapd, SuggestedName: "snapd"}},
```
stolowski: This change is fine although although technically it makes no difference for sanitizeSlotReservedForOS() as at the moment it is based on snap name check. But it's the right thing to do. Thanks.

interfaces/builtin/avahi_control.go
```diff
@@ -119,9 +118,10 @@ func (iface *avahiControlInterface) StaticInfo() interfaces.StaticInfo {
 func (iface *avahiControlInterface) AppArmorConnectedPlug(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
 	old := "###SLOT_SECURITY_TAGS###"
 	var new string
-	if release.OnClassic {
-		// If we're running on classic Avahi will be part
-		// of the OS snap and will run unconfined.
+	// If we're running on classic Avahi can be installed as snap or be part
+	// of the OS and it will run unconfined.
```
mvo5: Does this comment need to be expanded? Something like:
```
	// If we're running on classic Avahi can be installed as snap or be part
	// of the OS and it will run unconfined *or* it can be installed as a snap
        // with the normal confined snap label
```

interfaces/builtin/avahi_control.go
```diff
@@ -119,9 +118,10 @@ func (iface *avahiControlInterface) StaticInfo() interfaces.StaticInfo {
...
+	if isConnectedSlotSystemSlot(slot) {
```
mvo5: (nitpick) I don't really like the name - at the same time I'm not great with names. Maybe some suggestions:
- implicitSystemSlot(slot)
- ... ?

interfaces/builtin/avahi_control.go
```diff
@@ -135,7 +135,9 @@ func (iface *avahiControlInterface) AppArmorConnectedPlug(spec *apparmor.Specifi
 }
 
 func (iface *avahiControlInterface) AppArmorPermanentSlot(spec *apparmor.Specification, slot *snap.SlotInfo) error {
-	if !release.OnClassic {
+	// only apply slot snipped when running as application snap
+	// on classic, slot side can be system or application
+	if !isSlotSystemSlot(slot) {
```
mvo5: Same as above, maybe:
- implicitSystemSlot(slot) 
- ?
(also not great but…)

interfaces/builtin/utils.go
```diff
@@ -96,3 +97,26 @@ func sanitizeSlotReservedForOSOrApp(iface interfaces.Interface, slot *snap.SlotI
...
 	return nil
 }
+
+// determine if permanent slot side is provided by the system
+// on classic system some implicit slots can be provided by system or by
+// application snap e.g. avahi (it can be installed as deb or snap)
+// - slot owned by the system (core,snapd snap)  usually requires no action
+// - slot owned by application snap typically requires rules update
+func isSlotSystemSlot(slot *snap.SlotInfo) bool {
+	if release.OnClassic &&
```
mvo5: Do we actually need this? The function today is used in a way that we can't have "avahi" on !classic, right?

interfaces/builtin/utils.go
```diff
@@ -96,3 +97,26 @@ func sanitizeSlotReservedForOSOrApp(iface interfaces.Interface, slot *snap.SlotI
...
+		(slot.Snap.Type == snap.TypeOS || slot.Snap.Type == snap.TypeSnapd) {
```
mvo5: This is good but probably needs to include a check for the snap name "snapd" - we don't use the snapd type yet (mostly because in the past the store was not ready and now I think it is but we dropped the ball).

kubiko: @mvo5 instead of adding name check to many different places, wouldn't it be cleaner to simply choose correct enum when populating snap info? So we have name check only in one place?

interfaces/builtin/utils.go
kubiko: we need it to determine how is avahi installed.
Avahi can be on classic installed as deb or as snap. And clients should be able to connect to either one.
Trick is when installing avahi on classic as snap, we need to give it right permissions, but without giving those permissions to core snap in case avahi is installed as deb....
So this function should determine if slot in question is system slot ( avahi installed as deb) or not (avahi installed as snap)
We do not have this case yet, but I can image blues, network manager being another good examples....

interfaces/builtin/avahi_control.go
kubiko: updated

interfaces/builtin/avahi_control.go
kubiko: yeah same here I don't know what's best name here I changed it already good 5 times :)
- implicitSystemSlot
- connectedImplicitSystemSlot ?

interfaces/builtin/avahi_control.go
```diff
@@ -119,9 +118,11 @@ func (iface *avahiControlInterface) StaticInfo() interfaces.StaticInfo {
...
 	var new string
-	if release.OnClassic {
-		// If we're running on classic Avahi will be part
-		// of the OS snap and will run unconfined.
+	// If we're running on classic Avahi can be installed as snap or be part
+	// of the OS and it will run unconfined *or* it can be installed as a snap
+	// with the normal confined snap label
+	if isConnectedSlotSystemSlot(slot) {
+		// avahi is part of the OS and runs unconfined.
 		new = "unconfined"
```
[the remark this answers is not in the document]
kubiko: I notice those apparmor profiles in avahi deb package, so let's add it here
code updated

interfaces/builtin/avahi_control.go
```diff
@@ -119,9 +118,11 @@ func (iface *avahiControlInterface) StaticInfo() interfaces.StaticInfo {
 func (iface *avahiControlInterface) AppArmorConnectedPlug(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
 	old := "###SLOT_SECURITY_TAGS###"
 	var new string
-	if release.OnClassic {
-		// If we're running on classic Avahi will be part
-		// of the OS snap and will run unconfined.
+	// If we're running on classic Avahi can be installed as snap or be part
+	// of the OS and it will run unconfined *or* it can be installed as a snap
+	// with the normal confined snap label
```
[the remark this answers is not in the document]
kubiko: Updated with proposed comment

interfaces/builtin/avahi_control.go
kubiko: Used proposal from Jamie, thank you :)

interfaces/builtin/avahi_control.go
kubiko: I like those names, updated based on proposal from Jamie, thank you @jdstrand

interfaces/builtin/avahi_control.go
```diff
@@ -135,7 +136,9 @@ func (iface *avahiControlInterface) AppArmorConnectedPlug(spec *apparmor.Specifi
 }
 
 func (iface *avahiControlInterface) AppArmorPermanentSlot(spec *apparmor.Specification, slot *snap.SlotInfo) error {
-	if !release.OnClassic {
+	// only apply slot snipped when running as application snap
```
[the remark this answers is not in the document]
kubiko: ups, good find!
Fixed

interfaces/builtin/avahi_observe.go
```diff
@@ -430,9 +429,10 @@ func (iface *avahiObserveInterface) StaticInfo() interfaces.StaticInfo {
...
 	old := "###SLOT_SECURITY_TAGS###"
 	var new string
-	if release.OnClassic {
-		// If we're running on classic Avahi will be part
-		// of the OS snap and will run unconfined.
+	// If we're running on classic Avahi can be installed as snap or be part
+	// of the OS and it will run unconfined.
+	if isConnectedSlotSystemSlot(slot) {
+		// avahi is part of the OS and runs unconfined.
 		new = "unconfined"
```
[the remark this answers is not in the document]
kubiko: updated here as well

interfaces/builtin/utils.go
mvo5: Yeah, what I mean I that I wonder if we need the explicit check for `release.OnClassic` here? Given the current uses/use-cases for this its fine mostly wondering if at some point down the line we write code like "implicitSystemPermanentSlot(kernelModuleCOntrolSlot)" and wonder why we get different results on core vs classic. Maybe @pedronis has an opinion?

mvo5:
The test failures look real:
```
AppArmor parser error for /var/lib/snapd/apparmor/profiles/snap.test-snapd-policy-app-consumer.avahi-control in /var/lib/snapd/apparmor/profiles/snap.test-snapd-policy-app-consumer.avahi-control at line 492: syntax error, unexpected TOK_CONDID, expecting TOK_EQUALS or TOK_IN
```

interfaces/builtin/avahi_observe.go
```diff
@@ -430,10 +429,12 @@ func (iface *avahiObserveInterface) StaticInfo() interfaces.StaticInfo {
...
-	if release.OnClassic {
-		// If we're running on classic Avahi will be part
-		// of the OS snap and will run unconfined.
-		new = "unconfined"
+	// If we're running on classic, Avahi may be installed either as a snap of
+	// as part of the OS. If it is part of the OS, it will not have a security
+	// label like it would when installed as a snap.
+	if implicitSystemConnectedSlot(slot) {
+		// avahi from the OS is typically unconfined but known to sometimes be confined
+		new = "{unconfined,/usr/sbin/avahi-daemon}"
```
jdstrand: Please change this to:
```
		new = "\"{unconfined,/usr/sbin/avahi-daemon}\""
```

interfaces/builtin/avahi_control.go
```diff
@@ -119,10 +118,12 @@ func (iface *avahiControlInterface) StaticInfo() interfaces.StaticInfo {
...
-	if release.OnClassic {
-		// If we're running on classic Avahi will be part
-		// of the OS snap and will run unconfined.
-		new = "unconfined"
+	// If we're running on classic, Avahi may be installed either as a snap of
+	// as part of the OS. If it is part of the OS, it will not have a security
+	// label like it would when installed as a snap.
+	if implicitSystemConnectedSlot(slot) {
+		// avahi from the OS is typically unconfined but known to sometimes be confined
+		new = "{unconfined,/usr/sbin/avahi-daemon}"
```
jdstrand: Please change this to:
```
		new = "\"{unconfined,/usr/sbin/avahi-daemon}\""
```

kubiko: Thank you @jdstrand 
This is now changed and tested. Now it works when I use avahi-client and connect its plug to implicit slot

interfaces/builtin/avahi_observe.go
kubiko: Thank you @jdstrand 
This is now changed and tested. Now it works when I use avahi-client and connect its plug to implicit slot

jdstrand:
Assuming manual testing shows this to still work, LGTM.

kubiko:
> Assuming manual testing shows this to still work, LGTM.

@jdstrand yes, I have tested it also manually on classic and core system. Running as expected

PR #6329, opened 2019-01-07
cmd/snap-confine, packaging: support SELinux

bboozzoo:
The PR adds support for SELinux in snap-confine. Since we do not support SELinux security backend fully yet, transition the service to unconfined_service_t. Note, that the transition only happens if snap-confine is executed under `snappy_confine_t` context. For now, this can only happen if the snap application is a service started by systemd. 

The PR depends on bits in #6270 that allow `snappy_confine_t` -> `unconfined_service_t transition`.


cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,77 @@
...
+		die("failed to obtain current process context");
+	}
+	debug("current context: %s", ctx_str);
+
+	context_t ctx = context_new(ctx_str);
+	if (ctx == NULL) {
+		die("failed to create context from context string %s", ctx_str);
+	}
+
+	if (sc_streq(context_type_get(ctx), "snappy_confine_t")) {
```
zyga: I really strongly dislike this name. Can we do something to mend the situation?

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,77 @@
...
+#include "../libsnap-confine-private/utils.h"
+#include "../libsnap-confine-private/string-utils.h"
+
+/**
+ * Set security context for the snap
+ *
+ **/
+int sc_selinux_set_snap_execcon(void)
+{
+	/* die("foo"); */
```
zyga: Leftover from debugging?

bboozzoo: Haha, right

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,77 @@
...
+	if (is_selinux_enabled() < 1) {
+		debug("selinux not enabled");
+		return 0;
+	}
+
+	char *ctx_str = NULL;
+	if (getcon(&ctx_str) == -1) {
+		die("failed to obtain current process context");
```
zyga:
```suggestion
		die("cannot obtain SELinux process context");
```

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,77 @@
...
+	}
+	debug("current context: %s", ctx_str);
+
+	context_t ctx = context_new(ctx_str);
+	if (ctx == NULL) {
+		die("failed to create context from context string %s", ctx_str);
```
zyga:
```suggestion
		die("cannot create SELinux context from context string %s", ctx_str);
```

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,77 @@
...
+		 * to snappy_confine_t domain, at this point we are right before
+		 * executing snap-exec. However we do not have a full SELinux support
+		 * for services running in snaps, only the snapd bits and helpers are
+		 * covered by the policy.
+		 *
+		 * At this point transition to the unconfined_service_t domain (allowed
+		 * by snap_confine_t policy) upon the next exec() call.
+		 */
+		if (context_type_set(ctx, "unconfined_service_t") != 0) {
+			die("failed to update context %s type to unconfined_service_t",
```
zyga:
```suggestion
			die("cannot update SELinux context %s type to unconfined_service_t",
```

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,77 @@
...
+			    ctx_str);
+		}
+
+		char *new_ctx_str = context_str(ctx);
+		if (new_ctx_str == NULL) {
+			die("failed to obtain string of an updated context");
```
zyga:
```suggestion
			die("cannot obtain updated SELinux context string");
```

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,77 @@
...
+		}
+		if (setexeccon(new_ctx_str) == -1) {
+			die("failed to set exec context to %s", new_ctx_str);
```
zyga:
```suggestion
			die("cannot set SELinux exec context to %s", new_ctx_str);
```

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,77 @@
...
+		/* We are running under a targeted policy which ended up transitioning
+		 * to snappy_confine_t domain, at this point we are right before
+		 * executing snap-exec. However we do not have a full SELinux support
+		 * for services running in snaps, only the snapd bits and helpers are
+		 * covered by the policy.
+		 *
+		 * At this point transition to the unconfined_service_t domain (allowed
+		 * by snap_confine_t policy) upon the next exec() call.
+		 */
+		if (context_type_set(ctx, "unconfined_service_t") != 0) {
```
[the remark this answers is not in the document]
bboozzoo: Tha's why I introduced `snappy_unconfined_snap_t` in the policy. The idea was to have a starting point for implementing this. However, at this point it's somewhat unusable. AFAIU fixing issues requires altering the policy of other components, such as `init_t`, that would require changes in the reference policy.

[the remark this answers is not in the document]
Conan-Kudo: 👍

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,76 @@
...
+		}
+
+		char *new_ctx_str = context_str(ctx);
+		if (new_ctx_str == NULL) {
+			die("cannot obtain updated SELinux context string");
+		}
+		if (setexeccon(new_ctx_str) == -1) {
+			die("cannot set SELinux exec context to %s", new_ctx_str);
+		}
+		debug("SELinux context after next exec: %s", new_ctx_str);
```
stolowski: Missing `freecon(new_ctx_str)`?

cmd/snap-confine/selinux-support.h
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+#ifndef SNAP_CONFINE_SELINUX_SUPPORT_H
+#define SNAP_CONFINE_SELINUX_SUPPORT_H
+
+/**
+ * Set security context for the snap
```
stolowski: Could you please extend this comment and say e.g. "Set security context for the snap to support transition from snappy_confine_t to unconfined_service_t"?

Conan-Kudo: I'm pretty sure the goal is to eventually make this do more than that in the future...

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,76 @@
...
+			    ctx_str);
+		}
+
+		char *new_ctx_str = context_str(ctx);
+		if (new_ctx_str == NULL) {
+			die("cannot obtain updated SELinux context string");
+		}
+		if (setexeccon(new_ctx_str) == -1) {
+			die("cannot set SELinux exec context to %s", new_ctx_str);
+		}
```
jdstrand: I'm unfamiliar with the selinux API and it would probably be wise to have someone more knowledgeable take a look. That said, setexeccon takes a security_context_t whereas context_str return a const char *, so there seems to be a type mismatch. If setexeccon can simply take a string, why create context_str at all and just simply pass is 'unconfined_service_t'?

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,76 @@
...
+		die("cannot obtain current SELinux process context");
+	}
+	debug("current SELinux process context: %s", ctx_str);
+
+	context_t ctx = context_new(ctx_str);
+	if (ctx == NULL) {
+		die("cannot create SELinux context from context string %s", ctx_str);
+	}
+
+	if (sc_streq(context_type_get(ctx), "snappy_confine_t")) {
```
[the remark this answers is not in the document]
bboozzoo: Looking at libselinux source code, I see that `context_type_get()` returns a pointer to a char buffer allocated by `context_new()` when a context string is parsed. While doing so, the seem to null terminate each component https://github.com/SELinuxProject/selinux/blob/master/libselinux/src/context.c#L75 Type is mandatory, so a context without type will not parse correctly.

cmd/snap-confine/selinux-support.c
bboozzoo: That must depend on the library version you are looking at. The from libselinux-2.8.6 has `int setexeccon(char *context);`. The source code also has this note, which suggest the type applies to older releases of libselinux:
```
/* No longer used; here for compatibility with legacy callers. */
typedef char *security_context_t;
```

The reason the code goes through the trouble of using `context_new()` & friends is that we need to create the whole context string `user:role:type` or `user:role:type:sens1:cata-sens2:catb` and just switch the type to one we want. The full form can than be passed to `setexeccon()`.

jdstrand: Ok, thank you for the explanation.

cmd/snap-confine/selinux-support.c
bboozzoo: AFAICT no free call is needed. The storage of the context string is managed internally by `context_*` calls. It is freed automatically with `context_free(ctx)` just below.

bboozzoo: I'll leave a comment just in case.

stolowski: Thanks, I double-checked with man pages, looks good.

cmd/Makefile.am
```diff
@@ -273,6 +273,18 @@ snap_confine_snap_confine_extra_libs += $(APPARMOR_LIBS)
 endif  # STATIC_LIBAPPARMOR
 endif  # APPARMOR
 
+if SELINUX
+snap_confine_snap_confine_SOURCES += \
+	snap-confine/selinux-support.c \
```
zyga: Could you ensure that selinux-support is a part of the new-style formatting (clang-format). Just look at the fmt target to see how this happens.

cmd/autogen.sh
```diff
@@ -39,7 +39,7 @@ case "$ID" in
 		fi
 		;;
 	fedora|centos|rhel)
-		extra_opts="--libexecdir=/usr/libexec/snapd --with-snap-mount-dir=/var/lib/snapd/snap --enable-merged-usr --disable-apparmor"
+		extra_opts="--libexecdir=/usr/libexec/snapd --with-snap-mount-dir=/var/lib/snapd/snap --enable-merged-usr --disable-apparmor --enable-selinux"
```
zyga: Thank you for this.

cmd/configure.ac
```diff
@@ -77,6 +77,17 @@ AC_ARG_ENABLE([seccomp],
     esac], [enable_seccomp=yes])
 AM_CONDITIONAL([SECCOMP], [test "x$enable_seccomp" = "xyes"])
 
+# Allow to build with SELinux support by calling:
```
zyga: Reexec idea: I'd like to depend on selinux and just optionally enable it (or not) somehow. This doesn't have to happen now but it would would strongly simplify the set of possible builds of snap-confine that we must consider. Ideally it would be just (exactly) one so that we can really support re-exec everywhere.

cmd/configure.ac
```diff
@@ -209,6 +226,15 @@ AC_ARG_ENABLE([static-libseccomp],
     esac], [enable_static_libseccomp=no])
 AM_CONDITIONAL([STATIC_LIBSECCOMP], [test "x$enable_static_libseccomp" = "xyes"])
 
+AC_ARG_ENABLE([static-libselinux],
```
zyga: Can you add a comment that explains when we might want to statically link libselinux.

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,77 @@
...
+ *
+ **/
+int sc_selinux_set_snap_execcon(void)
+{
+	if (is_selinux_enabled() < 1) {
+		debug("SELinux not enabled");
+		return 0;
+	}
+
+	char *ctx_str = NULL;
```
zyga: Could we please have a cleanup function for this. It's not an invasive change but would simplify the set of logic one needs to track in review.

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,77 @@
...
+		 * At this point transition to the unconfined_service_t domain (allowed
+		 * by snap_confine_t policy) upon the next exec() call.
+		 */
+		if (context_type_set(ctx, "unconfined_service_t") != 0) {
+			die("cannot update SELinux context %s type to unconfined_service_t",
+			    ctx_str);
+		}
+
+		/* freed by context_free(ctx) */
+		char *new_ctx_str = context_str(ctx);
```
zyga: Same request here (cleanup). God why did selinux people not prefix their functions with `selinux_`.

Conan-Kudo: @zyga It was 2000, and the ABI hasn't changed since... 🤷‍♂️

cmd/configure.ac
bboozzoo:
```
    if (is_selinux_enabled() < 1) {
        debug("SELinux not enabled");
        return 0;
    }
```
This should do the trick. AA has similar functionality, so in theory we should be able to reexec on either AA or SELinux enabled systems (only if it wasn't for the mess caused by paths).

cmd/configure.ac
bboozzoo: Reexec only. We should probably have a similar comment for libapparmor and libseccomp bits.

cmd/snap-confine/selinux-support.c
bboozzoo: Added

cmd/configure.ac
zyga: That was not my point: can we just. not have the "--enable-selinux" flag at all and _always_ link to it?

cmd/configure.ac
zyga: Is it safe to statically link to libselinux?

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,88 @@
...
+#include "config.h"
+
+#include <selinux/context.h>
+#include <selinux/selinux.h>
+
+#include "../libsnap-confine-private/cleanup-funcs.h"
+#include "../libsnap-confine-private/string-utils.h"
+#include "../libsnap-confine-private/utils.h"
+
+static void sc_freecon(char **ctx) {
```
zyga: Thank you for those!

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,88 @@
...
+static void sc_context_free(context_t *ctx) {
+    if (ctx != NULL && *ctx != NULL) {
+        context_free(*ctx);
+        *ctx = NULL;
+    }
+}
+
+/**
+ * Set security context for the snap
+ *
```
zyga: Stray newline?

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,88 @@
...
+        die("cannot obtain current SELinux process context");
+    }
+    debug("current SELinux process context: %s", ctx_str);
+
+    context_t ctx SC_CLEANUP(sc_context_free) = context_new(ctx_str);
+    if (ctx == NULL) {
+        die("cannot create SELinux context from context string %s", ctx_str);
+    }
+
+    if (sc_streq(context_type_get(ctx), "snappy_confine_t")) {
```
zyga: `context_type_get` can return NULL on failure. Could you guard against that please?

cmd/snap-confine/selinux-support.h
```diff
@@ -0,0 +1,26 @@
...
+ *
```
zyga: Stray newline.

cmd/snap-confine/selinux-support.c
bboozzoo: This was discussed here: https://github.com/snapcore/snapd/pull/6329#discussion_r257646913 I can add another check though

cmd/configure.ac
bboozzoo: There are distros out there that don't support SELinux but have snapd available (Arch for instance). So this would have to degrade gracefully. Then we wouldn't know when you actually wanted SELinux enabled but it wasn't because libselinux is missing.

I think we could revisit this once we actually have an idea of making more universal use of SELinux support.

cmd/configure.ac
bboozzoo: I can't really address this without more hands-on experience. There is a static library in Fedora repositories. Looking at the source code, the library looks more like a wrapper around the kernel interfaces in /sys/fs/selinux. I think we'll only know what it lands and someone starts making use of it.

cmd/snap-confine/selinux-support.c
zyga: While I don't disagree it behaving this way now it may change later without us realising this. It's better to be safe than sorry.

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,94 @@
...
+static void sc_context_free(context_t *ctx) {
+    if (ctx != NULL && *ctx != NULL) {
+        context_free(*ctx);
+        *ctx = NULL;
+    }
+}
+
+/**
+ * Set security context for the snap
+ **/
```
pedronis: have a second of  line mentioning  unconfined_service_t  here as well for now

cmd/snap-confine/selinux-support.h
pedronis: Here that is fine I think.

cmd/snap-confine/snap-confine.c
```diff
@@ -243,6 +246,9 @@ int main(int argc, char **argv)
 		 * profile as well. */
 		sc_apply_global_seccomp_profile();
 	}
+#ifdef HAVE_SELINUX
+	sc_selinux_set_snap_execcon();
+#endif
```
pedronis: can we have  a note about that this is done both for confined and classic snaps, in any case the context of the snap != context of snap-confine

bboozzoo:
@jdstrand would you mind taking another look?

bboozzoo:
@jdstrand could you take one last look at this one?

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,96 @@
...
+        if (context_type_set(ctx, "unconfined_service_t") != 0) {
+            die("cannot update SELinux context %s type to unconfined_service_t", ctx_str);
+        }
+
+        /* freed by context_free(ctx) */
+        char *new_ctx_str = context_str(ctx);
+        if (new_ctx_str == NULL) {
+            die("cannot obtain updated SELinux context string");
+        }
+        if (setexeccon(new_ctx_str) == -1) {
```
jdstrand: Same nitpick as before: we typically use `< 0` in other parts of the codebase for things like this. Not a blocker.

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /* freed by context_free(ctx) */
+    const char *ctx_type = context_type_get(ctx);
+
+    if (ctx_type == NULL) {
+        die("cannot obtain type from SELinux context string %s", ctx_str);
+    }
+
+    if (sc_streq(ctx_type, "snappy_confine_t")) {
```
jdstrand: Where is this policy defined?

cmd/snap-confine/selinux-support.c
```diff
@@ -0,0 +1,96 @@
...
+        /* We are running under a targeted policy which ended up transitioning
+         * to snappy_confine_t domain, at this point we are right before
+         * executing snap-exec. However we do not have a full SELinux support
+         * for services running in snaps, only the snapd bits and helpers are
+         * covered by the policy.
+         *
+         * At this point transition to the unconfined_service_t domain (allowed
+         * by snap_confine_t policy) upon the next exec() call.
+         */
+        if (context_type_set(ctx, "unconfined_service_t") != 0) {
```
jdstrand: Is this guaranteed to be everywhere where `is_selinux_enabled()` is `1`? Isn't this specific to the reference policy?

cmd/snap-confine/selinux-support.c
Conan-Kudo: This is in the SELinux policy module in `data/selinux` in the tree.

cmd/snap-confine/selinux-support.c
Conan-Kudo: Technically, yes, but I don't know of a policy not derived from refpolicy or otherwise that doesn't have `unconfined_service_t` in place?

jdstrand: If that's so, then no objection. *Perhaps* that should be a code comment, but not a blocker.

jdstrand:
Based on the answers to the questions I have, marking as Approved, but please consider my nitpicks.

cmd/snap-confine/snap-confine.c
bboozzoo: Moved that around to be closed to sc_maybe_aa_change_onexec(). I see that the AppArmor support code is in libsnap-confine-private (?). I'd be happy to propose a follow-up PR moving selinux code around to that location.

PR #6335, opened 2019-01-08
tests: fix "No space left on device" issue on amazon-linux

sergiocazzolato:
The "|| true" is hiding a problem when we umount /tmp which is causing the error "No space left on device" when /tmp is used by another test. 

This is an example about how to reproduce it (I commented the umount time): https://paste.ubuntu.com/p/zsrqFDgD4V

This is an actual error reproduced on a travis execution:
https://api.travis-ci.org/v3/job/476382472/log.txt

The error can be reproduced by running the following tests (regression/lp-1800004, install-store-laaaarge and main/xdg-open) and the error is the following: https://travis-ci.org/snapcore/snapd/jobs/477104762#L1064

The fix is retrying the umount until the device is not busy anymore.

### DIFF
--- a/tests/main/install-store-laaaarge/task.yaml
+++ b/tests/main/install-store-laaaarge/task.yaml
@@ -1,5 +1,7 @@
 summary: snap install a large snap from the store (bigger than tmpfs)
 
+priority: 100
+
 prepare: |
     #shellcheck source=tests/lib/systemd.sh
     . "$TESTSLIB/systemd.sh"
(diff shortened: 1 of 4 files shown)


tests/main/install-store-laaaarge/task.yaml
```diff
@@ -13,7 +13,17 @@ restore: |
     . "$TESTSLIB/systemd.sh"
 
     systemd_stop_units snapd.service snapd.socket
-    umount /tmp || true
```
mvo5: Could we do a `umount --lazy /tmp` here? This might be sufficient to fix the issue.

bboozzoo: BTW. I recall a note that `--lazy` does not work on 14.04 and you need to use `-l` instead.

sergiocazzolato: Sure, I'll make the change and test it.

mvo5:
Thank you @sergiocazzolato

PR #6342, opened 2019-01-10
use major() and minor() in <sys/sysmacros.h> instead of MAJOR and MINOR…

timchen119:
… macro in <linux/kdev_t.h>

While I'm creating a snap messing with /dev/tpmrm0 which major device id is 253 and minor id is 65536, snap-confine keep giving wrong major and minor id, that makes my snap always had wrong device id in /sys/fs/cgroup/devices/snap*/devices.list at runtime and thus un-accessible from the strict confined snap. 

The debug message when setting SNAP_CONFINE_DEBUG=1 is something like this:
```
DEBUG: run_snappy_app_dev_add: /sys/devices/platform/MSFT0101:00/tpm/tpm0 snap_tpm2-simulator_clear
DEBUG: running snap-device-helper add snap_tpm2-simulator_clear /sys/devices/platform/MSFT0101:00/tpm/tpm0 10:224
DEBUG: run_snappy_app_dev_add: /sys/devices/platform/MSFT0101:00/tpmrm/tpmrm0 snap_tpm2-simulator_clear
DEBUG: running snap-device-helper add snap_tpm2-simulator_clear /sys/devices/platform/MSFT0101:00/tpmrm/tpmrm0 1048829:0
```

After reading a bit <bits/sysmacros.h> it looks to me we should use the major() and minor() macros in <sys/sysmacros.h> instead of using <linux/kdev_t.h> directly.

Using the macros in <sys/sysmacros.h> solve my original snap udev cgroup issue.

### DIFF
--- a/cmd/snap-confine/udev-support.c
+++ b/cmd/snap-confine/udev-support.c
@@ -19,7 +19,7 @@
 #include <ctype.h>
 #include <errno.h>
 #include <limits.h>
-#include <linux/kdev_t.h>
+#include <sys/sysmacros.h>
 #include <sched.h>
 #include <string.h>
 #include <sys/stat.h>
@@ -108,9 +108,7 @@ void run_snappy_app_dev_add(struct snappy_udev *udev_s, const char *path)
 	dev_t devnum = udev_device_get_devnum(d);
 	udev_device_unref(d);
 
-	unsigned major = MAJOR(devnum);
-	unsigned minor = MINOR(devnum);
-	_run_snappy_app_dev_add_majmin(udev_s, path, major, minor);
+	_run_snappy_app_dev_add_majmin(udev_s, path, major(devnum), minor(devnum));
 }
 
 /*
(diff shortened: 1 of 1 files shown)


mvo5:
Thanks for this PR and thanks for finding this. I can reproduce this with this trivial reproducer:
```
cat $0
cat > devt.c <<'EOF'
#include<linux/kdev_t.h>
#include<sys/sysmacros.h>
#include<sys/stat.h>
#include<stdio.h>

int main() {
	dev_t tpm0rm0 = makedev(253, 65536);
        dev_t MK = MKDEV((dev_t)253,(dev_t)65536);
	printf("makedev %li MKDEV() %li \n", tpm0rm0, MK);
	printf("MAJOR(): %li MINOR(): %li\n", MAJOR(tpm0rm0), MINOR(tpm0rm0));
	printf("major(): %i minor(): %i\n", major(tpm0rm0), minor(tpm0rm0));

}
EOF
gcc -o devt devt.c
./devt
makedev 268500224 MKDEV() 130304 
MAJOR(): 1048829 MINOR(): 0
major(): 253 minor(): 65536
```
based on that :+1:

zyga:
I second mvo's comment. I grepped for the offending macros in the tree and found some more. I pushed a fix to this branch. LGTM when green.

mvo5:
Thanks @zyga !

PR #6347, opened 2019-01-10
many: allow snap-update-ns to write user mount profile

zyga:
When snap-confine constructs a per-user mount namespace it delegates the
task to snap-update-ns invoked with a special option. The code in
snap-update-ns switches to the appropriate user ID while retaining some
capabilities required for mounting.

Before snap-update-ns was never saving the per-user mount profile so it
didn't need anything. Now it may start saving it (when the appropriate
feature is enabled) but it runs under a non-root UID. To allow the save
to really happen allow the file path pattern via apparmor template of
snap-update-ns and retain more capabilities to bypass filesystem checks.

### DIFF
--- a/cmd/snap-update-ns/bootstrap.c
+++ b/cmd/snap-update-ns/bootstrap.c
@@ -96,8 +96,16 @@ static int switch_to_privileged_user()
 	    { _LINUX_CAPABILITY_VERSION_3, 0 };
 	struct __user_cap_data_struct data[2] = { {0} };
 
-	data[0].effective = (CAP_TO_MASK(CAP_SYS_ADMIN) |
-			     CAP_TO_MASK(CAP_SETUID) | CAP_TO_MASK(CAP_SETGID));
+	data[0].effective = (
+		/* To perform mount operations. */
+		CAP_TO_MASK(CAP_SYS_ADMIN) |
+		/* To change user and group IDs. */
+		CAP_TO_MASK(CAP_SETUID) |
+		CAP_TO_MASK(CAP_SETGID) |
+		/* To ignore filesystem permissions. */
+		CAP_TO_MASK(CAP_DAC_OVERRIDE) |
+		CAP_TO_MASK(CAP_DAC_READ_SEARCH)
+	);
 	data[0].permitted = data[0].effective;
 	data[0].inheritable = 0;
 	data[1].effective = 0;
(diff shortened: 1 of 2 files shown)


zyga:
We need to write to the freezer cgroup and we need to write to `/run/snapd/ns`. I don't know if we can reliably change users after the bootstrap C code has ran since Go clones new threads for virtually all system calls. We can pin threads but I don't know if that's enough.

zyga:
@jdstrand let's discuss this next week at the sprint.

zyga:
Gentle ping @jdstrand

cmd/snap-update-ns/bootstrap.c
```diff
@@ -136,7 +144,13 @@ static int switch_to_privileged_user()
 	}
 	// After changing uid, our effective capabilities were dropped.
 	// Reacquire CAP_SYS_ADMIN, and discard CAP_SETUID/CAP_SETGID.
-	data[0].effective = CAP_TO_MASK(CAP_SYS_ADMIN);
+	data[0].effective = (
+		/* To perform mount operations. */
+		CAP_TO_MASK(CAP_SYS_ADMIN) |
+		/* To ignore filesystem permissions. */
+		CAP_TO_MASK(CAP_DAC_OVERRIDE) |
+		CAP_TO_MASK(CAP_DAC_READ_SEARCH)
```
jdstrand: I really dislike keeping this for the duration of snap-update-ns because it severely weakens the benefit of dropping privileges. Between SYS_ADMIN and these, snap-update-ns can do so much... (though on systems with apparmor, snap-update-ns is quite constrained wrt writes by a snap-specific apparmor profile, but on systems without apparmor we don't).

Needing DAC_OVERRIDE and DAC_READ_SEARCH typically indicates a design flaw with the privilege dropping and should be fixable without requiring keeping the capability. Is it not possible to open the fd prior to dropping privileges here in the bootstrap code and then passing it to snap-update-ns? I guess part of the problem with that is that then the bootstrap code might have to parse things.... What if snap-update-ns instead works on a temporary file and then after it returns the calling process puts it into place?

Considering the snap-update-ns apparmor profile, I'm grudgingly willing to accept this as is, but not without understanding why the couple of ideas I had (and any others you had) for avoiding it are worse than keeping the caps. We may want to include those as a code comment, but let's see what happens with the discussion.

Btw, if you have DAC_OVERRIDE, you probably shouldn't need DAC_READ_SEARCH; did you verify you need both?

zyga:
I will return to this branch, along with it's peer siblings as soon as the refactoring branch lands. Due to the inter-dependencies I cannot easily test the runtime behaviour without having more of the parts in master.

stolowski:
Is this ready for re-reviews?

zyga:

PR #6351, opened 2019-01-10
tests: review/fix the autopkgtest failures in disco

mvo5:
This fixes the issues found by the latest "disco" upload:
http://people.canonical.com/~ubuntu-archive/proposed-migration/disco/update_excuses.html#snapd

The amd64 testrun runs into a timeout, so there maybe more that we need to disable in ADT to stay below this limit.

### DIFF
--- a/tests/lib/snaps/test-snapd-service-watchdog/bin/direct
+++ b/tests/lib/snaps/test-snapd-service-watchdog/bin/direct
@@ -41,7 +41,7 @@ def main(opts):
     else:
         sleep_time = watchdog_timeout / 2
 
-    logging.info('watchdog notification every %us', sleep_time)
+    logging.info('watchdog notification every %s', sleep_time)
 
     with closing(socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)) as sock:
         sock.connect(notify_addr)
--- a/tests/lib/snaps/test-snapd-service-watchdog/bin/direct
+++ b/tests/lib/snaps/test-snapd-service-watchdog/bin/direct
@@ -41,7 +41,7 @@ def main(opts):
     else:
         sleep_time = watchdog_timeout / 2
 
-    logging.info('watchdog notification every %s', sleep_time)
+    logging.info('watchdog notification every %ss', sleep_time)
 
     with closing(socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)) as sock:
         sock.connect(notify_addr)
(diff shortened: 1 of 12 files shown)


tests/lib/snaps/test-snapd-service-watchdog/bin/direct
```diff
@@ -41,7 +41,7 @@ def main(opts):
     else:
         sleep_time = watchdog_timeout / 2
 
-    logging.info('watchdog notification every %us', sleep_time)
+    logging.info('watchdog notification every %s', sleep_time)
```
chipaca: why this one? It changes it from logging e.g. `... every 30s` to `every 30`

tests/main/network-retry/task.yaml
```diff
@@ -20,6 +20,12 @@ restore: |
     fi
 
 execute: |
+    if [ -n "$http_proxy" ] || [ -n "$https_proxy" ]; then
```
chipaca: don't you also need to check the uppercase ones?

tests/lib/snaps/test-snapd-service-watchdog/bin/direct
mvo5: Indeed, silly me, this should be: `%ss` now - the issue is really that `%u` gives us `0` for a `0.5` float value.

chipaca: if you know it's going to be less than 10000 you could use `%.4g`

tests/main/network-retry/task.yaml
```diff
@@ -20,6 +20,12 @@ restore: |
     fi
 
 execute: |
+    if [ -n "${http_proxy:-}" ] || [ -n "${https_proxy:-}" ] || [ -n "${HTTPS_PROXY:-}" ] || [ -n "${HTTPS_PROXY:-}" ]; then
```
bboozzoo: Breaking that line like in ` tests/main/proxy/task.yaml` would be nice, but probably not worth another spread run.

tests/main/install-store/task.yaml
```diff
@@ -3,6 +3,9 @@ summary: Checks for special cases of snap install from the store
 # run on ubuntu-14,16,18,20+ but not on ubuntu-core, fedora etc
 systems: [ubuntu-1*, ubuntu-2*, ubuntu-3*]
 
+# takes >2min to run in total
+backends: [-autopkgtest]
```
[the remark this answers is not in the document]
mvo5: Ok, I re-enabled this one and disabled two other ones.

PR #6358, opened 2019-01-11
cmd/snap-update-ns: close internal lock on error

zyga:
This patch fixes an error where a lock would internally leak until the
snap-update-ns process quits in the case where we would either:

 - fail to freeze the freezer cgroup
 - fail to verify a lock was already locked

The actual error is harmless since if either of those fail we just quit
anyway. The tests for this are already covered in the bigger refactor
branch where the locking code is broken out to a dedicated method for
easier testing.

### DIFF
--- a/cmd/snap-update-ns/system.go
+++ b/cmd/snap-update-ns/system.go
@@ -47,6 +47,7 @@ func applySystemFstab(instanceName string, fromSnapConfine bool) error {
 		// namespace is locked. This is used by snap-confine to use
 		// snap-update-ns to apply mount profiles.
 		if err := lock.TryLock(); err != osutil.ErrAlreadyLocked {
+			lock.Close()
 			return fmt.Errorf("mount namespace of snap %q is not locked but --from-snap-confine was used", instanceName)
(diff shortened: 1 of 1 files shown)


cmd/snap-update-ns/system.go
```diff
@@ -47,6 +47,7 @@ func applySystemFstab(instanceName string, fromSnapConfine bool) error {
 		// namespace is locked. This is used by snap-confine to use
 		// snap-update-ns to apply mount profiles.
 		if err := lock.TryLock(); err != osutil.ErrAlreadyLocked {
+			lock.Close()
```
bboozzoo: Isn't the defer above handling both places already?

zyga: Indeed! This only became an issue after I switched the logic to return a restore-like function. In this code, before the refactoring, locking is done as a part of larger body of work so there is indeed a defer there.

zyga: Since the refactored version already has this thing fixed. I can close this.

PR #6359, opened 2019-01-11
tests: fix listing tests to match "snap list --unicode=never"

sergiocazzolato:
Currently the test is failing when we run "snap list --unicode=never"
with external backend trying to match using single quotes:
> MATCH '^core .* [0-9]{2}-[0-9.]+(~[a-z0-9]+)?(\+git[0-9]+\.[0-9a-f]+)?
+[0-9]+ +(edge|beta|candidate|stable) +canonical\\* +core.*$'

### DIFF
--- a/tests/main/listing/task.yaml
+++ b/tests/main/listing/task.yaml
@@ -26,20 +26,20 @@ execute: |
     #shellcheck disable=SC2166
     if [ "$SPREAD_BACKEND" = "linode" -o "$SPREAD_BACKEND" = "google" -o "$SPREAD_BACKEND" == "qemu" ] && [ "$SPREAD_SYSTEM" = "ubuntu-core-16-64" ]; then
         echo "With customized images the core snap is sideloaded"
-        expected="^core .* [0-9]{2}-[0-9.]+(~[a-z0-9]+)?(\+git[0-9]+\.[0-9a-f]+)? +x[0-9]+ +- +- +core.*$"
+        expected='^core .* [0-9]{2}-[0-9.]+(~[a-z0-9]+)?(\+git[0-9]+\.[0-9a-f]+)? +x[0-9]+ +- +- +core.*$'
     elif [ "$SRU_VALIDATION" = "1" ]; then
         echo "When sru validation is done the core snap is installed from the store"
-        expected="^core .* [0-9]{2}-[0-9.]+(~[a-z0-9]+)?(\+[0-9]+\.[0-9a-f]+)? +[0-9]+ +stable +canonical\\* +core.*$"
+        expected='^core .* [0-9]{2}-[0-9.]+(~[a-z0-9]+)?(\+[0-9]+\.[0-9a-f]+)? +[0-9]+ +stable +canonical\* +core.*$'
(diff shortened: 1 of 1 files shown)


sergiocazzolato:
@zyga, 

Before we were matching this:
core              16-2.37~pre1   6229  beta      canonical✓  core

Now we need to match this:
core               16-2.37~pre1   6229  beta      canonical*  core

zyga:
@sergiocazzolato I was asking about how the arguments to MATCH are expanded with single and double quotes. I believe everything should work fine with an appropriate single-quoted argument.

Can you run something like:
```
echo "foo..."
echo 'foo...'
```
...and paste the results here?

sergiocazzolato:
@zyga, ah, ok, this is https://paste.ubuntu.com/p/T9t5WYwzvQ/ all the info 

Thanks for reviewing this

tests/main/listing/task.yaml
```diff
@@ -29,9 +29,9 @@ execute: |
         expected='^core .* [0-9]{2}-[0-9.]+(~[a-z0-9]+)?(\+git[0-9]+\.[0-9a-f]+)? +x[0-9]+ +- +- +core.*$'
     elif [ "$SRU_VALIDATION" = "1" ]; then
         echo "When sru validation is done the core snap is installed from the store"
-        expected='^core .* [0-9]{2}-[0-9.]+(~[a-z0-9]+)?(\+[0-9]+\.[0-9a-f]+)? +[0-9]+ +stable +canonical\\* +core.*$'
+        expected='^core .* [0-9]{2}-[0-9.]+(~[a-z0-9]+)?(\+[0-9]+\.[0-9a-f]+)? +[0-9]+ +stable +canonical\* +core.*$'
```
bboozzoo: The regex looks ok now. The old one would match something like `canonical\\\\`, but we actually wanted to match `canonical*`.

tests/main/listing/task.yaml
```diff
@@ -29,9 +29,9 @@ execute: |
...
     elif [ "$SPREAD_BACKEND" = "external" ] || [ "$SPREAD_BACKEND" = "autopkgtest" ]; then
-        expected='^core .* [0-9]{2}-[0-9.]+(~[a-z0-9]+)?(\+git[0-9]+\.[0-9a-f]+)? +[0-9]+ +(edge|beta|candidate|stable) +canonical\\* +core.*$'
+        expected='^core .* [0-9]{2}-[0-9.]+(~[a-z0-9]+)?(\+git[0-9]+\.[0-9a-f]+)? +[0-9]+ +(edge|beta|candidate|stable) +canonical\* +core.*$'
     else
         expected="^core .* [0-9]{2}-[0-9.]+(~[a-z0-9]+)?(\\+git[0-9]+\\.[0-9a-f]+)? +[0-9]+ +$CORE_CHANNEL +canonical\\* +core.*$"
```
mvo5: Won't we need a a similar change here? I mean, not strictly needed because the " vs ' means this is interpreted differently by the shell so we get the right outcome in the test. But it seems strange, so either we change all of the related regexps to use ' or " IMO. WDYT?

bboozzoo: `$CORE_CHANNEL` needs to be expanded here.

PR #6361, opened 2019-01-11
kvm: load required kernel modules if necessary

gerboland:
Have the KVM interface load the suitable KVM kernel modules.

Supports checking the required CPU capabilities for Intel and AMD CPUs and loading suitable modules if supported, falls back to generic KVM module otherwise.

Ref: https://www.linux-kvm.org/page/Processor_support

### DIFF
--- a/interfaces/builtin/kvm.go
+++ b/interfaces/builtin/kvm.go
@@ -81,7 +81,7 @@ func (iface *kvmInterface) KModConnectedPlug(spec *kmod.Specification, plug *int
 	m := "kvm"
 	cpu_flags, err := getCpuFlags()
 	if err != nil {
-		logger.Debugf("kvm: fetching cpu info failed:", err)
+		logger.Debugf("kvm: fetching cpu info failed: %v", err)
 	}
 
 	if strutil.ListContains(cpu_flags, "vmx") {
--- a/interfaces/builtin/kvm.go
+++ b/interfaces/builtin/kvm.go
@@ -55,6 +55,7 @@ type kvmInterface struct {
 }
 
 var procCpuinfo = "/proc/cpuinfo"
+var flagsMatcher = regexp.MustCompile(`(?m)^flags\s+:\s+(.*)$`).FindSubmatch
 
 func getCpuFlags() (flags []string, err error) {
 	buf, err := ioutil.ReadFile(procCpuinfo)
@@ -64,8 +65,7 @@ func getCpuFlags() (flags []string, err error) {
 	}
 
 	// want to capture the text after 'flags:' entry
-	pattern := regexp.MustCompile(`(?m)^flags\s+:\s+(.*)$`)
-	match := pattern.FindSubmatch(buf)
+	match := flagsMatcher(buf)
 	if len(match) == 0 {
 		return nil, fmt.Errorf("%v does not contain a 'flags:' entry", procCpuinfo)
 	}
--- a/interfaces/builtin/kvm.go
+++ b/interfaces/builtin/kvm.go
@@ -1,7 +1,7 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 
 /*
- * Copyright (C) 2017 Canonical Ltd
+ * Copyright (C) 2017-2019 Canonical Ltd
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 3 as
@@ -18,6 +18,16 @@
  */
 
 package builtin
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"strings"
+
+	"github.com/snapcore/snapd/interfaces"
+	"github.com/snapcore/snapd/interfaces/kmod"
+)
 
 const kvmSummary = `allows access to the kvm device`
 
@@ -38,8 +48,60 @@ const kvmConnectedPlugAppArmor = `
 
 var kvmConnectedPlugUDev = []string{`KERNEL=="kvm"`}
 
+type kvmInterface struct {
+	commonInterface
+}
+
+var procCpuinfo = "/proc/cpuinfo"
+
+func getCpuFlags() (flags string, err error) {
+	buf, err := ioutil.ReadFile(procCpuinfo)
+	if err != nil {
+		// if we can't read cpuinfo, we want to know _why_
+		return "", fmt.Errorf("error: %v", err)
+	}
+
+	pattern := []byte("\nflags\t\t:")
+	idx := bytes.LastIndex(buf, pattern)
+	if idx >= 0 {
+		offset := idx + len(pattern)
+		endidx := bytes.Index(buf[offset:], []byte("\n"))
+		return string(buf[offset : endidx+offset]), nil
+	}
+
+	// if not found (which will happen on non-x86 architectures, which is ok
+	// because they'd typically not have the same info over and over again),
+	// return whole buffer; otherwise, return from just after the \n
+	return string(buf[idx+1:]), nil
+}
+
+func (iface *kvmInterface) KModConnectedPlug(spec *kmod.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	// Check CPU capabilities to load suitable module
+	// TODO: this only considering x86 CPUs, but some ARM, PPC and S390 CPUs also support KVM
+	m := "kvm"
+	cpu_flags, err := getCpuFlags()
+	if err != nil {
+		fmt.Println("kvm: fetching cpu info failed:", err)
+	}
+
+	if strings.Contains(cpu_flags, "vmx") {
+		m = "kvm_intel"
+	} else if strings.Contains(cpu_flags, "svm") {
+		m = "kvm_amd"
+	} else {
+		// CPU appears not to support KVM extensions, fall back to bare kvm module as it appears
+		// sufficient for some architectures
+		fmt.Println("kvm: failed to detect CPU specific KVM support, will attempt to modprobe generic KVM support")
+	}
+
+	if err := spec.AddModule(m); err != nil {
+		return nil
+	}
+	return nil
+}
+
 func init() {
-	registerIface(&commonInterface{
+	registerIface(&kvmInterface{commonInterface{
 		name:                  "kvm",
 		summary:               kvmSummary,
 		implicitOnCore:        true,
@@ -48,5 +110,5 @@ func init() {
 		connectedPlugAppArmor: kvmConnectedPlugAppArmor,
 		connectedPlugUDev:     kvmConnectedPlugUDev,
 		reservedForOS:         true,
-	})
+	}})
 }
--- a/interfaces/builtin/kvm.go
+++ b/interfaces/builtin/kvm.go
@@ -20,13 +20,15 @@
 package builtin
 
 import (
-	"bytes"
 	"fmt"
 	"io/ioutil"
+	"regexp"
 	"strings"
 
 	"github.com/snapcore/snapd/interfaces"
 	"github.com/snapcore/snapd/interfaces/kmod"
+	"github.com/snapcore/snapd/logger"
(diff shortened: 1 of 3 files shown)


gerboland:
Spread failed for some other reason:

```
Cloning into '/home/gopath/.cache/govendor/golang.org/x/crypto'...
fatal: unable to access 'https://go.googlesource.com/crypto/': Failed to connect to go.googlesource.com port 443: Connection timed out
Error: Remotes failed for:
	Failed for "golang.org/x/crypto/cast5" (failed to clone repo): exit status 128
```

interfaces/builtin/kvm.go
```diff
@@ -38,8 +48,60 @@ const kvmConnectedPlugAppArmor = `
...
+	if err != nil {
+		// if we can't read cpuinfo, we want to know _why_
+		return "", fmt.Errorf("error: %v", err)
+	}
+
+	pattern := []byte("\nflags\t\t:")
+	idx := bytes.LastIndex(buf, pattern)
+	if idx >= 0 {
+		offset := idx + len(pattern)
+		endidx := bytes.Index(buf[offset:], []byte("\n"))
```
zyga: This can return `-1`, it's not super likely but we should check for it.

interfaces/builtin/kvm.go
```diff
@@ -38,8 +48,60 @@ const kvmConnectedPlugAppArmor = `
...
+		return string(buf[offset : endidx+offset]), nil
+	}
+
+	// if not found (which will happen on non-x86 architectures, which is ok
+	// because they'd typically not have the same info over and over again),
+	// return whole buffer; otherwise, return from just after the \n
+	return string(buf[idx+1:]), nil
```
zyga: I don't think this is sensible behaviour. Just return an empty string in this case.

interfaces/builtin/kvm.go
```diff
@@ -38,8 +48,60 @@ const kvmConnectedPlugAppArmor = `
...
+}
+
+func (iface *kvmInterface) KModConnectedPlug(spec *kmod.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	// Check CPU capabilities to load suitable module
+	// TODO: this only considering x86 CPUs, but some ARM, PPC and S390 CPUs also support KVM
```
zyga: Please at the very least support x86_64, i386 and amd64. I don't know of any easily available ARMv7 boards that have kvm (none of the reference platforms do) so I think it is fine to skip that.

interfaces/builtin/kvm.go
```diff
@@ -38,8 +48,60 @@ const kvmConnectedPlugAppArmor = `
...
+	m := "kvm"
+	cpu_flags, err := getCpuFlags()
+	if err != nil {
+		fmt.Println("kvm: fetching cpu info failed:", err)
+	}
+
+	if strings.Contains(cpu_flags, "vmx") {
```
zyga: I would feel better if cpu_flags was `[]string` or if this was entirely encapsulated in a helper function.

interfaces/builtin/kvm.go
```diff
@@ -38,8 +48,60 @@ const kvmConnectedPlugAppArmor = `
...
+		m = "kvm_intel"
+	} else if strings.Contains(cpu_flags, "svm") {
+		m = "kvm_amd"
+	} else {
+		// CPU appears not to support KVM extensions, fall back to bare kvm module as it appears
+		// sufficient for some architectures
+		fmt.Println("kvm: failed to detect CPU specific KVM support, will attempt to modprobe generic KVM support")
```
zyga: Please use a logger, don't print this way.

interfaces/builtin/kvm_test.go
```diff
@@ -57,8 +66,23 @@ slots:
...
 
 func (s *kvmInterfaceSuite) SetUpTest(c *C) {
+	s.BaseTest.SetUpTest(c)
+
 	s.plug, s.plugInfo = MockConnectedPlug(c, kvmConsumerYaml, nil, "kvm")
 	s.slot, s.slotInfo = MockConnectedSlot(c, kvmCoreYaml, nil, "kvm")
+
+	// Need to Mock output of /proc/cpuinfo
+	s.tmpdir = c.MkDir()
+	dirs.SetRootDir(s.tmpdir)
```
zyga: Please restore original root directory in `TearDownTest()`

interfaces/builtin/kvm.go
```diff
@@ -38,8 +48,60 @@ const kvmConnectedPlugAppArmor = `
...
+var procCpuinfo = "/proc/cpuinfo"
+
+func getCpuFlags() (flags string, err error) {
+	buf, err := ioutil.ReadFile(procCpuinfo)
+	if err != nil {
+		// if we can't read cpuinfo, we want to know _why_
+		return "", fmt.Errorf("error: %v", err)
+	}
+
+	pattern := []byte("\nflags\t\t:")
```
mvo5: I'm slightly worried about `\t\t`, i.e. this pattern is pretty static, if /proc/flags ever gets something like `cache_alignment_size_offset_table_address : 0xdeadbeef` this may break. As much as I dislike regexp I wonder if they will be more appropriate here so that we can simply ignore the whitespace?

interfaces/builtin/kvm.go
```diff
@@ -38,8 +48,60 @@ const kvmConnectedPlugAppArmor = `
...
+	commonInterface
+}
+
+var procCpuinfo = "/proc/cpuinfo"
+
+func getCpuFlags() (flags string, err error) {
+	buf, err := ioutil.ReadFile(procCpuinfo)
+	if err != nil {
+		// if we can't read cpuinfo, we want to know _why_
+		return "", fmt.Errorf("error: %v", err)
```
mvo5: Please provide a bit more context to the user here, something like: `cannot read cpu info: %v"` for the error string.

interfaces/builtin/kvm.go
gerboland: Ack. I was duplicating procCpuinfoMinimal() in the errtracker package, but I prefer your thinking

interfaces/builtin/kvm.go
```diff
@@ -38,8 +48,60 @@ const kvmConnectedPlugAppArmor = `
 
 var kvmConnectedPlugUDev = []string{`KERNEL=="kvm"`}
 
+type kvmInterface struct {
+	commonInterface
+}
+
+var procCpuinfo = "/proc/cpuinfo"
+
+func getCpuFlags() (flags string, err error) {
```
mvo5: How about returning a list of strings here with the flags? Something like: `func getCpuFlags() (flags []string, err error` ? We have strutil.ListContains() (and the same for sorted lists) as a helper so that you can easily test if a string is in the list. Or even a map[string]bool  (or interface{}) ? Of course with the list approach once the flags are identified an additional `flags = strings.Fields(flagsString)` will be required but I think thats not too bad.

interfaces/builtin/kvm.go
gerboland: I need to double-check i386 (think it uses same cpu flag and kernel module), but x86_64 and amd64 are covered.

interfaces/builtin/kvm.go
mvo5: Yeah, or if we go with the []string just `nil` for the empty list.

interfaces/builtin/kvm.go
```diff
@@ -38,8 +48,60 @@ const kvmConnectedPlugAppArmor = `
...
+	return string(buf[idx+1:]), nil
+}
+
+func (iface *kvmInterface) KModConnectedPlug(spec *kmod.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	// Check CPU capabilities to load suitable module
+	// TODO: this only considering x86 CPUs, but some ARM, PPC and S390 CPUs also support KVM
+	m := "kvm"
+	cpu_flags, err := getCpuFlags()
+	if err != nil {
+		fmt.Println("kvm: fetching cpu info failed:", err)
```
mvo5: If we want to print anything (and not error here) we should either push this via the logger or generate a snapd warning.

interfaces/builtin/kvm.go
gerboland: I share your discomfort, but I was also avoiding regex :) I'll see what I can do

interfaces/builtin/kvm.go
gerboland: Ack. /me confesses he didn't notice the logger implementation.

interfaces/builtin/kvm.go
gerboland: Replaced with regexp, and function returns []string, this is gone

interfaces/builtin/kvm.go
gerboland: vmx is available on some i386. Happy to skip the others. Comment updated to be clearer

interfaces/builtin/kvm.go
gerboland: Done, Please see if you agree with what logger priority I use.

interfaces/builtin/kvm_test.go
gerboland: Done

interfaces/builtin/kvm.go
gerboland: I'm using regexp, is cleaner and not in a hot code path.

interfaces/builtin/kvm.go
gerboland: Done

interfaces/builtin/kvm.go
gerboland: Done, but please check if you agree with the logger method I use

interfaces/builtin/kvm.go
```diff
@@ -38,8 +50,58 @@ const kvmConnectedPlugAppArmor = `
...
+
+func getCpuFlags() (flags []string, err error) {
+	buf, err := ioutil.ReadFile(procCpuinfo)
+	if err != nil {
+		// if we can't read cpuinfo, we want to know _why_
+		return nil, fmt.Errorf("unable to read %v: %v", procCpuinfo, err)
+	}
+
+	// want to capture the text after 'flags:' entry
+	pattern := regexp.MustCompile(`(?m)^flags\s+:\s+(.*)$`)
```
bboozzoo: This should probably be package level `var` so that we don't compile the regex on each call.

gerboland: Done

interfaces/builtin/kvm.go
```diff
@@ -38,8 +50,58 @@ const kvmConnectedPlugAppArmor = `
...
+	return cpu_flags, nil
+}
+
+func (iface *kvmInterface) KModConnectedPlug(spec *kmod.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	// Check CPU capabilities to load suitable module
+	// NOTE: this only considers i386, x86_64 and amd64 CPUs, but some ARM, PPC and S390 CPUs also support KVM
+	m := "kvm"
+	cpu_flags, err := getCpuFlags()
+	if err != nil {
+		logger.Debugf("kvm: fetching cpu info failed:", err)
```
bboozzoo:
```suggestion
		logger.Debugf("kvm: fetching cpu info failed: %v", err)
```

interfaces/builtin/kvm_test.go
```diff
@@ -57,8 +66,25 @@ slots:
...
+	s.AddCleanup(func() { dirs.SetRootDir("/") })
+
+	mockCpuinfo := filepath.Join(s.tmpdir, "cpuinfo")
+	c.Assert(ioutil.WriteFile(mockCpuinfo, []byte(`
```
bboozzoo: I think the PR could use another unit test where `cpuinfo` is empty or doesn't exist.

gerboland: Done

mvo5:
Thanks for your work on this!

PR #6363, opened 2019-01-11
cmd/snap-update-ns: save errno from strtoul

zyga:
This is not required now but it helps in someone adds printf-style
debugging messages that may clobber errno.

### DIFF
--- a/cmd/snap-update-ns/bootstrap.c
+++ b/cmd/snap-update-ns/bootstrap.c
@@ -340,6 +340,7 @@ static int parse_arg_u(int argc, char * const *argv, int *optind, unsigned long 
 	errno = 0;
 	char *uid_text_end = NULL;
 	unsigned long parsed_uid = strtoul(uid_text, &uid_text_end, 10);
+	int saved_errno = errno;
 	if (
 			/* Reject overflow in parsed representation */
 			(parsed_uid == ULONG_MAX && errno != 0)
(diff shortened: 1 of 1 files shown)


cmd/snap-update-ns/bootstrap.c
```diff
@@ -351,7 +352,7 @@ static int parse_arg_u(int argc, char * const *argv, int *optind, unsigned long
 			|| (*uid_text != '\0' && uid_text_end != NULL
 				&& *uid_text_end != '\0')) {
 		bootstrap_msg = "cannot parse user id";
-		bootstrap_errno = errno;
```
mvo5: Sorry, I'm not sure I understand this - AFAICT nothing in the `if( (parsed_uid ...)` changes errno, so why do we need to safe it?

zyga: We don't need to save it yet, as I said above this is in case someone forgets about this fact and adds any code in between, like I did while debugging isspace.

mvo5: You mean if someone adds debug code into the `if()` ? Sorry for being so slow today, it seems the save/restore is so close together that I don't quite see a use case. Maybe an example would help me :) ?

zyga: Yes, exactly for that kind of case. It's not mandatory, we can drop this patch if you feel like it.

PR #6364, opened 2019-01-11
cmd/snap-update-ns: let the go parser know we are parsing -u

zyga:
Ironically the go side doesn't strictly care about this because the C
bootstrap code will set up everything so that existing calls to
os.Getuid are sufficient but we should let the parser know so that it
doesn't complain.


zyga:
@mvo, the test is indeed coming but that is further down the pipe. As for the -u parsing question, there is already a PR that uses it: https://github.com/snapcore/snapd/pull/6367

PR #6365, opened 2019-01-11
cmd/snap-update-ns: manually implement isspace

zyga:
I wasn't able to find any other cause but using the isspace macro,
outside of tests (which pass just fine), causes ./snap-update-ns -u 1000
to crash. The debugger points to issspace(*uid_text) being the culprit
despite it not being a NULL pointer. It is correctly pointing to the
string "1000" when the crash happens. I was unable to solve it with a
static function that calls isspace internally (which is probably a
macro). As such, implement isspace ourselves to avoid the crash.

Upstream bug report: https://github.com/golang/go/issues/29689
also contains a minimal reproducer https://github.com/zyga/go-bug-29689
which crashes on latest golang binary release.

### DIFF
--- a/cmd/snap-update-ns/bootstrap.c
+++ b/cmd/snap-update-ns/bootstrap.c
@@ -271,7 +271,11 @@ static int instance_key_validate(const char *instance_key)
 	// engine.
 	int i = 0;
 	for (i = 0; instance_key[i] != '\0'; i++) {
-		if (islower(instance_key[i]) || isdigit(instance_key[i])) {
+		char c = instance_key[i];
+		/* NOTE: We are reimplementing islower() and isdigit()
+		 * here. For context see
+		 * https://github.com/golang/go/issues/29689 */
+		if ((c >= 'a' && c <= 'z') || (c >= '0' && c <= '9')) {
 			continue;
 		}
 		bootstrap_msg =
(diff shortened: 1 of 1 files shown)


zyga:
We have a reply on the bug report. This is the way things will misbehave if we use pre-init array. This is not a Go bug.

cmd/snap-update-ns/bootstrap.c
```diff
@@ -340,11 +340,14 @@ static int parse_arg_u(int argc, char * const *argv, int *optind, unsigned long
...
 	char *uid_text_end = NULL;
 	unsigned long parsed_uid = strtoul(uid_text, &uid_text_end, 10);
+	char c = *uid_text;
 	if (
 			/* Reject overflow in parsed representation */
 			(parsed_uid == ULONG_MAX && errno != 0)
 			/* Reject leading whitespace allowed by strtoul. */
-			|| (isspace(*uid_text))
+			/* NOTE: This is not using isspace, for some reason that is crashing
+			 * at runtime. Reported upstream as https://github.com/golang/go/issues/29689 */
```
bboozzoo: Comment needs an update now.

zyga: Done

zyga:
I will implement islower/isspace now.

PR #6368, opened 2019-01-11
tests: fix daemon-notify test checking denials considering all the log lines

sergiocazzolato:
This change fixes some errors that is reproduced in boards during beta validation.

See the error:
https://paste.ubuntu.com/p/QXgPpCPZ8J/

By using all the log lines we make sure the full log is checked to count
denials.

It also fix the scenario with repeat in google:
https://paste.ubuntu.com/p/7T8g8YGdvc/

### DIFF
--- a/tests/main/interfaces-daemon-notify/task.yaml
+++ b/tests/main/interfaces-daemon-notify/task.yaml
@@ -12,22 +12,27 @@ prepare: |
     install_local test-snapd-daemon-notify
 
 execute: |
+    # shellcheck source=tests/lib/journalctl.sh
+    . "$TESTSLIB/journalctl.sh"
+
     echo "The interface is not connected by default"
     snap interfaces -i daemon-notify | MATCH -- "- +test-snapd-daemon-notify:daemon-notify"
 
     echo "When the interface is connected"
     snap connect test-snapd-daemon-notify:daemon-notify
 
-    echo "Then after we restart the servive there is not denials"
-    denials_before="$(snap logs -n=all test-snapd-daemon-notify.notify | grep -c 'Permission denied' || true)"
+    echo "And the service is stopped"
     snap stop test-snapd-daemon-notify.notify
+
+    denials_before="$(get_journalctl_log -u snap.test-snapd-daemon-notify.notify.service | grep -c 'Permission denied' || true)"
+    echo "Then after we restart the servive there is not denials"
     for _ in $(seq 10); do
         if snap start test-snapd-daemon-notify.notify; then
             break
         fi
         sleep 1
     done
-    denials_after="$(snap logs -n=all test-snapd-daemon-notify.notify | grep -c 'Permission denied' || true)"
+    denials_after="$(get_journalctl_log -u snap.test-snapd-daemon-notify.notify.service | grep -c 'Permission denied' || true)"
 
     [ "$denials_before" -eq "$denials_after" ]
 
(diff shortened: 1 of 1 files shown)


tests/main/interfaces-daemon-notify/task.yaml
```diff
@@ -19,15 +19,15 @@ execute: |
     snap connect test-snapd-daemon-notify:daemon-notify
 
     echo "Then after we restart the servive there is not denials"
-    denials_before="$(snap logs -n=100 test-snapd-daemon-notify.notify | grep -c 'Permission denied' || true)"
+    denials_before="$(snap logs -n=all test-snapd-daemon-notify.notify | grep -c 'Permission denied' || true)"
```
bboozzoo: Maybe it's a mistake to use `snap logs` after all. Since you added `get_journalctl_logs`, we should use that to ensure we're only browsing the logs from this test run. So the line should be changed to:
```
# shellcheck source=tests/lib/journalctl.sh
. "$TESTSLIB/journalctl.sh"
...
denials_before="$(get_journalctl_log -u snap.test-snapd-daemon-notify.notify.service | grep -c 'Permission denied' || true)"
```

tests/main/interfaces-daemon-notify/task.yaml
```diff
@@ -39,14 +39,14 @@ execute: |
     snap disconnect test-snapd-daemon-notify:daemon-notify
 
     echo "Then the snap is not able to sending notification messages"
```
bboozzoo: While already pushing the changes, can you fix this typo too? s/sending/send/

tests/main/interfaces-daemon-notify/task.yaml
```diff
@@ -19,15 +19,15 @@ execute: |
     snap connect test-snapd-daemon-notify:daemon-notify
 
     echo "Then after we restart the servive there is not denials"
-    denials_before="$(snap logs -n=100 test-snapd-daemon-notify.notify | grep -c 'Permission denied' || true)"
+    denials_before="$(snap logs -n=all test-snapd-daemon-notify.notify | grep -c 'Permission denied' || true)"
     snap stop test-snapd-daemon-notify.notify
```
bboozzoo: Now looking at the test, this line should be moved just before we do the counting.

tests/main/interfaces-daemon-notify/task.yaml
```diff
@@ -39,14 +39,14 @@ execute: |
     snap disconnect test-snapd-daemon-notify:daemon-notify
 
     echo "Then the snap is not able to sending notification messages"
-    denials_before="$(snap logs -n=100 test-snapd-daemon-notify.notify | grep -c 'Permission denied' || true)"
+    denials_before="$(snap logs -n=all test-snapd-daemon-notify.notify | grep -c 'Permission denied' || true)"
     snap stop test-snapd-daemon-notify.notify
```
bboozzoo: This stop comand should also be moved up.

tests/main/interfaces-daemon-notify/task.yaml
```diff
@@ -12,22 +12,27 @@ prepare: |
...
     echo "When the interface is connected"
     snap connect test-snapd-daemon-notify:daemon-notify
 
-    echo "Then after we restart the servive there is not denials"
-    denials_before="$(snap logs -n=100 test-snapd-daemon-notify.notify | grep -c 'Permission denied' || true)"
+    echo "And the service is stopped"
     snap stop test-snapd-daemon-notify.notify
+
+    denials_before="$(get_journalctl_log -u snap.test-snapd-daemon-notify.notify.service | grep -c 'Permission denied' || true)"
+    echo "Then after we restart the servive there is not denials"
```
bboozzoo:
```suggestion
    echo "There are no additional denials after the service has been restarted"
```

PR #6369, opened 2019-01-11
Add check for snap binaries dir not being in path

liamg:
See https://bugs.launchpad.net/snappy/+bug/1593157

### DIFF
--- a/cmd/snap/cmd_snap_op.go
+++ b/cmd/snap/cmd_snap_op.go
@@ -27,6 +27,7 @@ import (
 	"sort"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"github.com/jessevdk/go-flags"
 
@@ -257,6 +258,10 @@ func (mx *channelMixin) setChannelFromCommandline() error {
 	return nil
 }
 
+// isSnapInPath checks whether the snap binaries dir (e.g. /snap/bin)
+// is in $PATH.
+//
+// TODO: consider symlinks
 func isSnapInPath() bool {
 	paths := filepath.SplitList(os.Getenv("PATH"))
 	for _, path := range paths {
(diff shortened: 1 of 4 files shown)


cmd/snap/cmd_snap_op.go
```diff
@@ -277,6 +288,9 @@ func showDone(cli *client.Client, names []string, op string, esc *escapes) error
 				// TRANSLATORS: the args are a snap name optionally followed by a channel, then a version (e.g. "some-snap (beta) 1.3 installed")
 				fmt.Fprintf(Stdout, i18n.G("%s%s %s installed\n"), snap.Name, channelStr, snap.Version)
 			}
+			if !isSnapInPath() {
```
bboozzoo: Wonder if it'd be better if the message was printed before the `<foo> .. installed`.

liamg:
Oops, you're right, thanks. It also needs a `\n` after the message.

ogra1:
> > Perhaps @chipaca or @degville have some suggestions about the warning message.
> 
> The warning message is clear to me. I'm not sure whether we should recommend a course of action though (_Please add it to ensure snaps run correctly_) if it has already been added and the user simply needs to logout and back in again.
> 
> Maybe something like `Check your $PATH and refresh your environment to ensure snaps can run correctly.`?

the current message as well as that suggestion above does not really cover the most seen use-case of users installing snapd (in a bash shell) and simply not knowing that they have to re-login to make the change take effect, we see this a lot on IRC and in the forum. this should definitely be added here so users that just miss the re-login don't start to fiddle with their configs even though this wouldn't be necessary.

ogra1:
> How about: `First try logging out and back in again to refresh your environment. If this doesn't work, check your $PATH.`

sounds definitely better ...

chipaca:
How about “If you've not restarted your session (logged out and back in, or rebooted) since you installed snapd, try doing that: if that does not work, please see [some forum url that explains the problem and what to do about it, so the user gets detailed instructions, and we can keep track of distro/shell/desktop combos that need tweaking]”?

mvo5:
Very nice, I love it!

PR #6373, opened 2019-01-14
overlord/ifacestate: helper API to obtain the state of connections

bboozzoo:
Add a helper method to query the state of connections. The API returns
connections that are currently established, as well as ones that were explicitly
disconnected. Connection state tells whether the connection was established
automtically, manually or triggered by the gadget.

The primary purpose of the change is to make the information accessible in the
daemon, to satisfy the API calls to `/v2/connections` endpoint.

### DIFF
--- a/overlord/ifacestate/ifacemgr.go
+++ b/overlord/ifacestate/ifacemgr.go
@@ -162,8 +162,8 @@ type ConnectionState struct {
 	ByGadget bool
 	// Interface name of the connection
 	Interface string
-	// Undesired indicates whether the connection was explicitly
-	// disconnected
+	// Undesired indicates whether the connection, otherwise established
+	// automatically, was explicitly disconnected
 	Undesired bool
 }
 
--- a/overlord/ifacestate/ifacemgr.go
+++ b/overlord/ifacestate/ifacemgr.go
@@ -155,6 +155,39 @@ func (m *InterfaceManager) Repository() *interfaces.Repository {
 	return m.repo
 }
 
+type ConnectionState struct {
+	// Auto indicates whether the connection was established automatically
+	Auto bool
+	// ByGadget indicates whether the connection was trigged by the gadget
+	ByGadget bool
+	// Interface name of the connection
+	Interface string
+	// Undesired indicates whether the connection was explicitly
+	// disconnected
+	Undesired bool
+}
+
+// ConnectionStates return the state of connections tracked by the manager
+func (m *InterfaceManager) ConnectionStates() (connStateByRef map[string]ConnectionState, err error) {
+	m.state.Lock()
+	defer m.state.Unlock()
+	states, err := getConns(m.state)
+	if err != nil {
+		return nil, err
+	}
+
+	connStateByRef = make(map[string]ConnectionState, len(states))
+	for cref, cstate := range states {
+		connStateByRef[cref] = ConnectionState{
+			Auto:      cstate.Auto,
+			ByGadget:  cstate.ByGadget,
+			Interface: cstate.Interface,
+			Undesired: cstate.Undesired,
+		}
+	}
+	return connStateByRef, nil
+}
+
 // DisableUDevMonitor disables the instantiation of udev monitor, but has no effect
 // if udev is already created; it should be called after creating InterfaceManager, before
 // first Ensure.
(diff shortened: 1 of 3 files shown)


overlord/ifacestate/ifacestate_test.go
```diff
@@ -4962,7 +4962,7 @@ func (s *interfaceManagerSuite) TestAttributesRestoredFromConns(c *C) {
 	c.Check(conn.Slot.Attr("number", &number), IsNil)
 	c.Check(number, Equals, int64(1))
 
-	ifacestate.UpdateConnectionInConnState(conns, conn, false, false)
+	ifacestate.UpdateConnectionInConnState(conns, conn, false, false, false)
```
mvo5: (unreleated to this PR) I think this indicates that the API for UpdateConnectionInConnState can be improved, in order to see what the three "false" means one must go to the declaration of UpdateConnectionInConnState() to see what it is doing. Given that this is only in tests its not so critical and in the new code it looks fine, so maybe just change to something like:
```
var isAuto, byGadget, isUndesired bool
ifacestate.UpdateConnectionInConnState(conns, conn,  isAuto, byGadget, isUndesired)
```
here?

zyga: Nice suggestion, could be used as an interim pattern for boolean values.

bboozzoo:
@mvo5 the helper will be used in `/v2/connections` endpoint. The idea is to make some details about the connection avaialble through the endpoint. Specifically, the returned data must include whether the connection was established automatically or manually or was triggered via the gadget snap. We'd also want to optionally indicate when the user disconnected a connection and thus it became undesired.

overlord/ifacestate/ifacemgr.go
```diff
@@ -155,6 +155,39 @@ func (m *InterfaceManager) Repository() *interfaces.Repository {
...
+	// Auto indicates whether the connection was established automatically
+	Auto bool
+	// ByGadget indicates whether the connection was trigged by the gadget
+	ByGadget bool
+	// Interface name of the connection
+	Interface string
+	// Undesired indicates whether the connection was explicitly
+	// disconnected
+	Undesired bool
+}
```
zyga: Some suggestion on the shape of this API:

```
// IsConnected returns true if a connection is established.
func (cst *ConnectionState) IsConnected() bool {
  return !cst.Undesired
}

// ConnectionInitiator returns the entity that caused a connection to be formed. 
func (cst *ConnectionState) ConnectionInitiator () string {
  if !cst.IsConnected() {
    return ""
  }
  if cst.Auto {
    return "snapd"
  }
  if cst.ByGadget {
    return "gadget"
  }
  return "user"
}
```

I think that the state we keep is fine but spreading the manipulation of those fields across the tree will end up badly as we evolve that state.

overlord/ifacestate/ifacemgr.go
```diff
@@ -155,6 +155,39 @@ func (m *InterfaceManager) Repository() *interfaces.Repository {
...
 
+type ConnectionState struct {
+	// Auto indicates whether the connection was established automatically
+	Auto bool
+	// ByGadget indicates whether the connection was trigged by the gadget
+	ByGadget bool
+	// Interface name of the connection
+	Interface string
+	// Undesired indicates whether the connection was explicitly
+	// disconnected
```
stolowski: Please add: "Only applicable to auto-connections" (or a similiar explanation).

overlord/ifacestate/ifacemgr.go
bboozzoo: Need to think about. I think we can come back to this once I update the daemon bits. It may happen that the snapd API will expose an aggregate string that indicates who trigged the connection (like you're suggesting now), instead of individual fields.

overlord/ifacestate/ifacestate_test.go
bboozzoo: Pushed

overlord/ifacestate/ifacemgr.go
stolowski: On a side note, we will soon have auto-connections initiated by hotplug, but I'm not sure we will want to differentiate them. In any case, a note to myself, mostly ;)

PR #6376, opened 2019-01-14
tests: split the test interfaces-many in 2 and remove snaps on restore

sergiocazzolato:
The idea of this change is to make the tests run faster and remove the
needed snaps on restore to avoid kill-timeout reached running the
reset.sh.

The original issue is reproduced when we run interfaces-many and then another test with kill-timeout set in 5 minutes or less. The second test fails bacause during the prepare runs the reset.sh and it gives timeout trying to remove the provider and consumer snaps installed in interfaces-many.

Reproduce:
spread -debug google:ubuntu-core-16-64:tests/main/interfaces-many google:ubuntu-core-16-64:tests/main/snap-service


tests/main/interfaces-many-core-provided/task.yaml
```diff
@@ -107,9 +81,10 @@ execute: |
         fi
         snap interfaces | MATCH "$CONNECTED_PATTERN"
 
-        echo "Then $slotcmd_bn should succeed"
-        "$slotcmd" | MATCH PASS
```
jdstrand: Ah, this was a bug in the previous tests. Nice :)

PR #6377, opened 2019-01-15
tests: auto-clean the test directory

sergiocazzolato:
The idea of this change is to clean automatically the test directory during the restore step. With this change, it is not needed anymore to clean any change done in the test directory as it is also not needed to remove any snap.

### DIFF
--- a/tests/lib/prepare-restore.sh
+++ b/tests/lib/prepare-restore.sh
@@ -411,6 +411,9 @@ prepare_suite() {
 }
 
 prepare_suite_each() {
+    # back test directory to be restored during the restore
+    tar cf "${PWD}.tar" "$PWD"
+
     # save the job which is going to be executed in the system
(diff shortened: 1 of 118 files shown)


tests/lib/prepare-restore.sh
```diff
@@ -434,6 +437,13 @@ prepare_suite_each() {
 
 restore_suite_each() {
     rm -f "$RUNTIME_STATE_PATH/audit-stamp"
+
+    # restore test directory saved during prepare
+    if [ -f "${PWD}.tar" ]; then
+        rm -rf "$PWD"
+        tar -C/ -xf "${PWD}.tar"
+        rm -rf "${PWD}.tar"
```
bboozzoo: Wonder if we should `cd "$PWD"` at the end to avoid the weirdness of the directory being restored under a different inode, and current shell's $PWD being seemingly empty.

mvo5: Could be a follow-up.

PR #6378, opened 2019-01-15
 cmd: fix snap-device-helper to deal correctly with hooks 

mvo5:
Our current snap-device-helper code does not correctly handle
hooks. The input for the snap-device-helper is an "APPANME"
that looks like "snap_$snapname_$appname". There are no "."
allowed in a udev tag. So this appname needs to be translated
back into the correct security tag, like: "snap.$snapname.$appname".

This works most of the time, however it does not for hooks. They
have the form "snap_$snapname_hook_$hookname" and the script
cannot transform them correctly. This PR fixes this in a simple
way.

One open question is what about instances that are called "hook".
Those will now be handled incorrectly.

This also adds a spread test for the bug that was mentioned in the
forum.

### DIFF
--- a/cmd/snap-confine/snap-device-helper
+++ b/cmd/snap-confine/snap-device-helper
@@ -25,7 +25,11 @@ NOSNAP="${APPNAME#snap_}"
 # we need to make sure we change this to:
 # - "$snap_hook.$hookname"
 # - "$snap_$instance_hook.$hookname"
-if [ -z "${NOSNAP##*_hook_*}" ]; then
+if [ -z "${NOSNAP##*_hook_hook_*}" ]; then
+    # $instance is 'hook'; $snap_hook_hook.$hookname -> $snap_hook_hook.$hookname
+    NOSNAP="${NOSNAP%_hook_*}_hook.${NOSNAP#*_hook_hook_}"
+elif [ -z "${NOSNAP##*_hook_*}" ]; then
+    # $snap_$instance_hook_$hookname -> $snap_$instance_hook.$hookname
     NOSNAP="${NOSNAP%_hook_*}_hook.${NOSNAP#*_hook_}"
 fi
 
--- a/cmd/snap-confine/snap-device-helper
+++ b/cmd/snap-confine/snap-device-helper
@@ -18,6 +18,18 @@ MAJMIN="$4"
 NOSNAP="${APPNAME#snap_}"
 [ "$NOSNAP" != "$APPNAME" ] || { echo "malformed appname $APPNAME" >&2; exit 1; }
 
+# FIXME: this will break for instances that are called "hook" :(
+# Handle hooks first, the the nosnap part looks like this:
+# - "$snap_hook_$hookname"
+# - "$snap_$instance_hook_$hookname
+# we need to make sure we change this to:
+# - "$snap_hook.$hookname"
+# - "$snap_$instance_hook.$hookname"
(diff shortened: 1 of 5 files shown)


cmd/snap-confine/snap-device-helper
```diff
@@ -18,6 +18,18 @@ MAJMIN="$4"
...
+# FIXME: this will break for instances that are called "hook" :(
+# Handle hooks first, the the nosnap part looks like this:
+# - "$snap_hook_$hookname"
+# - "$snap_$instance_hook_$hookname
+# we need to make sure we change this to:
+# - "$snap_hook.$hookname"
+# - "$snap_$instance_hook.$hookname"
+if [ -z "${NOSNAP##*_hook_*}" ]; then
+    NOSNAP="${NOSNAP%_hook_*}_hook.${NOSNAP#*_hook_}"
+fi
```
jdstrand: We can avoid the bug with:
```
if [ -z "${NOSNAP##*_hook_hook_*}" ]; then
    # $instance is 'hook'; $snap_hook_hook.$hookname -> $snap_hook_hook.$hookname
    NOSNAP="${NOSNAP%_hook_*}_hook.${NOSNAP#*_hook_hook_}"
else if [ -z "${NOSNAP##*_hook_*}" ]; then
    # $snap_$instance_hook_$hookname -> $snap_$instance_hook.$hookname
    NOSNAP="${NOSNAP%_hook_*}_hook.${NOSNAP#*_hook_}"
fi

# Now deal with ...
```

cmd/snap-confine/snap-device-helper-test.c
```diff
@@ -188,17 +191,21 @@ static void test_sdh_err(void)
...
 static struct sdh_test_data instance_change_data =
-    { "change", "snap.foo_bar.baz", "devices.allow", "devices.deny" };
+    { "change", "snap.foo_bar.baz", "snap_foo_bar_baz", "devices.allow", "devices.deny" };
 static struct sdh_test_data instance_remove_data =
-    { "remove", "snap.foo_bar.baz", "devices.deny", "devices.allow" };
+    { "remove", "snap.foo_bar.baz", "snap_foo_bar_baz", "devices.deny", "devices.allow" };
+static struct sdh_test_data add_hook_data =
+    { "add", "snap.foo.hook.configure", "snap_foo_hook_configure", "devices.allow", "devices.deny" };
+static struct sdh_test_data instance_add_hook_data =
+    { "add", "snap.foo_bar.hook.configure", "snap_foo_bar_hook_configure", "devices.allow", "devices.deny" };
```
jdstrand: Perhaps add a test for 'snap.foo_hook.hook.configure'

mvo5: I think the problem is with an instance "hook" and a app "configure": `snap.foo_hook_configure` - or am I missing something?

mvo5: I mean, YES to the suggestion to add a test :) and yet it seems like there is this case (above) where the current transformation is ambiguous - unless I miss something of course.

bboozzoo: IMO a hook on a `foo_hook` instance is more ambiguous. With a workaround proposed by @jdstrand SNAPAPP expands correctly to `snap.foo_hook.hook.configure`, without it I get `snap.foo_hook_hook.hook.hook.hook_configure`.

PR #6382, opened 2019-01-15
tests: use pc-kernel from stable instead of edge to build core-16 and core-18

sergiocazzolato:
Currently core-18 is failing to boot due to lack of entropy, this can
be workarounded by using kernel from stable while we found a solution
for that problem.


sergiocazzolato:
Another solution is provided in #6383

mvo5:
Thanks for looking into this! Lets close this one, it seems like the 4.15 kernel got fixed with the most recent upload and it no longer fails to boot.

PR #6383, opened 2019-01-15
tests: provide a fake random device to the core images

sergiocazzolato:
The idea is to do the same for core images what we do fo classic images


sergiocazzolato:
This change fixes the issue that #6382 tries to fix

mvo5:
I think this PR is fine and because we do this already on classic helps with the consistency. We also generate ssh hostkeys on boot so there it will help as well. 

I don't think it will help with the entropy-at-boot issue though. This happens very early in the initramfs and looking at the libuuid source the getrandom(..., GRND_NONBLOCK) syscall is used, not access to /dev/random.

PR #6384, opened 2019-01-16
snapd: fix race in TestSanityFailGoesIntoDegradedMode test

mvo5:
Ensure the MockSanityCheck code runs at least *twice* to
avoid a race in the TestSanityFailGoesIntoDegradedMode test:

If we close the channel and this wakes up the "select"
in the test immediately and stops this go-routine, then the
check that the logbuf contains the error will fail.
By running this at least twice we know the error made
it to the log.

### DIFF
--- a/cmd/snapd/main_test.go
+++ b/cmd/snapd/main_test.go
@@ -61,10 +61,16 @@ func (s *snapdSuite) TestSanityFailGoesIntoDegradedMode(c *C) {
 
 	sanityErr := fmt.Errorf("foo failed")
 	sanityCalled := make(chan bool)
-	sanityClosed := false
+	sanityRan := 0
 	restore = snapd.MockSanityCheck(func() error {
-		if !sanityClosed {
-			sanityClosed = true
+		sanityRan++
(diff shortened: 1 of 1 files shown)


cmd/snapd/main_test.go
```diff
@@ -61,10 +61,16 @@ func (s *snapdSuite) TestSanityFailGoesIntoDegradedMode(c *C) {
...
-		if !sanityClosed {
-			sanityClosed = true
+		sanityRan++
+		// Ensure this ran at least *twice* to avoid a race here:
+		// If we close the channel and this wakes up the "select"
+		// below immediately and stops this go-routine then the
+		// check that the logbuf contains the error will fail.
+		// By running this at least twice we know the error made
+		// it to the log.
+		if sanityRan >= 2 {
```
bboozzoo: `close(sanityCalled)` should probably be called only when `sanityRan == 2`, otherwise there's a chance we could try to close it twice and cause a panic.

mvo5: Thanks! Great catch. Force pushed for easier cherry-picking into 2.37

PR #6388, opened 2019-01-16
tests: fix install-snaps test by changing the snap info regex

sergiocazzolato:
This regex is used to match the channel of the snap

google:ubuntu-16.04-64 .../tests/main/install-snaps# snap info
--unicode=never ia | grep -E "stable:.*--.*"
  stable:    --

error:
https://travis-ci.org/snapcore/spread-cron/builds/480186970#L6124

### DIFF
--- a/tests/main/install-snaps/task.yaml
+++ b/tests/main/install-snaps/task.yaml
@@ -97,11 +97,11 @@ execute: |
     CHANNELS="stable candidate beta edge"
     for CHANNEL in $CHANNELS; do
         # shellcheck disable=SC2153
-        if ! CHANNEL_INFO="$(snap info "$SNAP" | grep " $CHANNEL: ")"; then
+        if ! CHANNEL_INFO="$(snap info --unicode=never "$SNAP" | grep " $CHANNEL: ")"; then
             echo "Snap $SNAP not found"
             exit
         fi
-        if echo "$CHANNEL_INFO" | MATCH "$CHANNEL:.*–"; then
+        if echo "$CHANNEL_INFO" | MATCH "$CHANNEL:.*--"; then
             continue
         fi
 


tests/main/install-snaps/task.yaml
```diff
@@ -97,11 +97,11 @@ execute: |
     CHANNELS="stable candidate beta edge"
     for CHANNEL in $CHANNELS; do
         # shellcheck disable=SC2153
-        if ! CHANNEL_INFO="$(snap info "$SNAP" | grep " $CHANNEL: ")"; then
+        if ! CHANNEL_INFO="$(snap info --unicode=never "$SNAP" | grep " $CHANNEL: ")"; then
             echo "Snap $SNAP not found"
             exit
         fi
-        if echo "$CHANNEL_INFO" | MATCH "$CHANNEL:.*–"; then
+        if echo "$CHANNEL_INFO" | MATCH "$CHANNEL:.*--"; then
```
bboozzoo: Not strictly necessary, but doesn't hurt.

bboozzoo:
Just to be clear, the real fix is `--unicode=never`. Maybe we should think about adding:
```
environment:
    LC_ALL: C
```
to relevant tests.

tests/main/install-snaps/task.yaml
chipaca: This is changing an en-dash to two hyphen-dashes, so yeah very necessary :-)
In fact the current MATCH would never apply because en dashes are no longer printed to a pipe.

tests/main/install-snaps/task.yaml
```diff
@@ -97,11 +97,11 @@ execute: |
     CHANNELS="stable candidate beta edge"
     for CHANNEL in $CHANNELS; do
         # shellcheck disable=SC2153
-        if ! CHANNEL_INFO="$(snap info "$SNAP" | grep " $CHANNEL: ")"; then
+        if ! CHANNEL_INFO="$(snap info --unicode=never "$SNAP" | grep " $CHANNEL: ")"; then
```
chipaca: As of #6339 this is no longer necessary (and that is the cause of the breakage)

chipaca:
> Just to be clear, the real fix is `--unicode=never`. Maybe we should think about adding:
> 
> ```
> environment:
>     LC_ALL: C
> ```
> to relevant tests.

nah, the fix is the regex change; `--unicode=never` is now the default when writing to a pipe :-)

PR #6394, opened 2019-01-17
tests: iterate getting journal logs to support delay on boards on daemon-notify test

sergiocazzolato:
The interfaces-daemon-notify test is sometimes failing due to in the boards like RPI it taks more time to write the journal logs and retrieve that info, so when we get the logs the "Permission denied" logs were not written down yet.

### DIFF
--- a/tests/main/interfaces-daemon-notify/task.yaml
+++ b/tests/main/interfaces-daemon-notify/task.yaml
@@ -29,14 +29,14 @@ execute: |
 
     denials_before="$(get_journalctl_log -u snap.test-snapd-daemon-notify.notify.service | grep -c 'Permission denied' || true)"
     echo "Then after we restart the servive there is not denials"
+    snap start test-snapd-daemon-notify.notify
     for _ in $(seq 10); do
-        if snap start test-snapd-daemon-notify.notify; then
+        denials_after="$(get_journalctl_log -u snap.test-snapd-daemon-notify.notify.service | grep -c 'Permission denied' || true)"
+        if [ "$denials_before" -eq "$denials_after" ]; then
             break
         fi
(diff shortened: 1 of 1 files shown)


tests/main/interfaces-daemon-notify/task.yaml
```diff
@@ -29,14 +29,14 @@ execute: |
 
     denials_before="$(get_journalctl_log -u snap.test-snapd-daemon-notify.notify.service | grep -c 'Permission denied' || true)"
     echo "Then after we restart the servive there is not denials"
```
bboozzoo: `/servive there is not/service there are no/`

tests/main/interfaces-daemon-notify/task.yaml
```diff
@@ -29,14 +29,14 @@ execute: |
...
+    snap start test-snapd-daemon-notify.notify
     for _ in $(seq 10); do
-        if snap start test-snapd-daemon-notify.notify; then
+        denials_after="$(get_journalctl_log -u snap.test-snapd-daemon-notify.notify.service | grep -c 'Permission denied' || true)"
+        if [ "$denials_before" -eq "$denials_after" ]; then
```
bboozzoo: If the problem is logs appearing after a longer/unexpected delay, then this check should be reversed to `"$denials_before" -ne "$denials_after"`. Otherwise, it only confirms that there are as many denials as before, which can be caused by the sheer lack of denials (and the functionality working correctly), or that denials have not been logged yet.

sergiocazzolato: @bboozzoo thanks for the review, the changes are already pushed

PR #6393, opened 2019-01-17
packaging: import debian salsa packaging work

mvo5:
This PR imports the debian salsa packaging work into our tree under packaging/debian-sid.

With that we create a `debian` branch in our upstream tree that just contains the following change:
```
git rm debian
git mv packaging/debian-sid debian
```

Once we have that in our upstream tree we should be able to release with:
```
git checkout debian  
gbp buildpackage 
```

and do snapshot test builds with:
```
git checkout debian  
gbp buildpackage  --git-upstream-tree=SLOPPY
```

This work can also be merged back into salsa then. The idea is that we keep and update the debian packaging just like we do for ubuntu-14.04 and ubuntu-16.04+

PR #6396, opened 2019-01-17
packaging: import debian salsa packaging work and use in spead

mvo5:
This PR imports the debian salsa packaging work into our tree under packaging/debian-sid *without* importing the git history. Importing the git history is done in https://github.com/snapcore/snapd/pull/6393 but this PR fails because we have an invalid email in the git commit history that breaks the CLA check. Once we have resolved that the history PR can be merged and we can merge back and forth from this github tree to salsa.

During the debian spread tests, the packaging/debian-sid packaging is then used for building the deb package that is used during the full spread test run against debian-sid.

With that we create a `debian` branch in our upstream tree that just contains the following change:
```
git rm debian
git mv packaging/debian-sid debian
git commit -m 'make debian a real directory instead of a symlink so that tooling like gbp works correctly'
```

Once we have that in our upstream tree we should be able to release with:
```
git checkout debian  
gbp buildpackage 
```

and do snapshot test builds with:
```
git checkout debian  
gbp buildpackage  --git-upstream-tree=SLOPPY
```

This work can also be merged back into salsa then. The idea is that we keep and update the debian packaging just like we do for ubuntu-14.04 and ubuntu-16.04+

There is more work to be done to unify the packaging more, I think we should use symlinks for some things like we do in 14.04.


packaging/debian-sid/patches/disable-i18n.patch
```diff
@@ -0,0 +1,289 @@
...
+     # the various langpacks.
+Index: snapd.mvo/i18n/i18n.go
+===================================================================
+--- snapd.mvo.orig/i18n/i18n.go
++++ snapd.mvo/i18n/i18n.go
+@@ -21,83 +21,9 @@ package i18n
+ 
+ //go:generate update-pot
+ 
+-import (
```
bboozzoo: Maybe in the long run we could use build tags to provide a noop implementation at build time?

packaging/debian-sid/patches/skip-tests-seccomp-backend.patch
```diff
@@ -0,0 +1,31 @@
+Index: snapd.mvo/interfaces/seccomp/backend_test.go
+===================================================================
+--- snapd.mvo.orig/interfaces/seccomp/backend_test.go
++++ snapd.mvo/interfaces/seccomp/backend_test.go
+@@ -355,6 +355,8 @@ func (s *backendSuite) TestSocketcallIsA
+ }
+ 
+ func (s *backendSuite) TestSocketcallIsNotAddedWhenNotRequired(c *C) {
++	c.Skip("skipped until debian has check.v1 from 2018. Not() is broken in the old version")
```
bboozzoo: I think we could just tweak the test to avoid using Not()

packaging/debian-sid/snapd.dirs
```diff
@@ -0,0 +1,12 @@
+snap
+var/cache/snapd
+usr/lib/snapd
+var/lib/snapd/apparmor/snap-confine.d
+var/lib/snapd/auto-import
+var/lib/snapd/desktop
+var/lib/snapd/environment
+var/lib/snapd/firstboot
+var/lib/snapd/lib/gl
```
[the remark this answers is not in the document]
zyga: We should cross-check this list with the list present in snapd.spec in the opensuse packaging PR. I spent a lot of time getting it just right.

packaging/debian-sid/patches/disable-i18n.patch
zyga: We discussed a way to use go generate for that. Build tags are another thing.

But in fairness, what we need to do is just package that one library and be done with it. We just don't want to block on that with 2.37.

packaging/debian-sid/patches/skip-tests-seccomp-backend.patch
zyga: This patch is gone in my downstream package. The corresponding library was fixed.

mvo5:
Closing, I will rework this based on the 2.37 upload.

PR #6405, opened 2019-01-21
run-checks: ensure we use go-1.10 if available

mvo5:
The run-checks script must use go1.10 when its available. If
not it will FTBFS in the launchpad builders.

### DIFF
--- a/run-checks
+++ b/run-checks
@@ -16,7 +16,7 @@ else
 fi
 COVERMODE=${COVERMODE:-atomic}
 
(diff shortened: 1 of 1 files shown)


run-checks
```diff
@@ -16,6 +16,9 @@ else
 fi
 COVERMODE=${COVERMODE:-atomic}
 
+# ensure we use go-1.0 by default
```
chipaca:
```suggestion
# ensure we use go-1.10 by default
```

PR #6408, opened 2019-01-21
tests: add spread test for system dbus interface

mvo5:
We currently only test the dbus interface on the session bus and
also not on Ubuntu Core devices. There was a recent bugreport by
a customer about this interface and while we sort out the details
this should fill the gap in the testing we have.

### DIFF
--- a/tests/main/interfaces-system-dbus/task.yaml
+++ b/tests/main/interfaces-system-dbus/task.yaml
@@ -1,0 +1,37 @@
+summary: Ensure that the dbus interface works on the system bus.
+
+execute: |
+    #shellcheck source=tests/lib/dirs.sh
+    . "$TESTSLIB/dirs.sh"
+
+    echo "Install dbus system test snaps"
+    snap install --edge test-snapd-dbus-provider
+    snap install --edge test-snapd-dbus-consumer
+
+    echo "Wait for dbus service to auto-start"
+    for _ in $(seq 60); do
+        if dbus-send --system --print-reply --dest=com.dbustest.HelloWorld /com/dbustest/HelloWorld com.dbustest.HelloWorld.SayHello | MATCH "hello world"; then
+            break
+        fi
+        sleep 1
+    done
+
+    echo "The dbus consumer plug is disconnected by default"
+    snap interfaces -i dbus | MATCH '^- +test-snapd-dbus-consumer:dbus-system-test'
+
+    if [ "$(snap debug confinement)" = partial ]; then
+        exit 0
+    fi
+
+    echo "And the consumer is not able to access the provided method"
+    if test-snapd-dbus-consumer.dbus-system-consumer 2> call.error; then
+        echo "Expected permission error calling dbus method with disconnected plug"
+        exit 1
+    fi
+    MATCH "Permission denied" < call.error
+
+    echo "When the plug is connected"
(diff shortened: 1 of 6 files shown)


tests/main/interfaces-system-dbus/task.yaml
```diff
@@ -0,0 +1,37 @@
+summary: Ensure that the dbus interface works on the system bus.
+
+execute: |
+    #shellcheck source=tests/lib/dirs.sh
+    . "$TESTSLIB/dirs.sh"
+
+    echo "Install dbus system test snaps"
+    snap install --edge test-snapd-dbus-provider
```
sergiocazzolato: you could install it from beta

tests/main/interfaces-system-dbus/task.yaml
```diff
@@ -0,0 +1,37 @@
...
+        if dbus-send --system --print-reply --dest=com.dbustest.HelloWorld /com/dbustest/HelloWorld com.dbustest.HelloWorld.SayHello | MATCH "hello world"; then
+            break
+        fi
+        sleep 1
+    done
+
+    echo "The dbus consumer plug is disconnected by default"
+    snap interfaces -i dbus | MATCH '^- +test-snapd-dbus-consumer:dbus-system-test'
+
+    if [ "$(snap debug confinement)" = partial ]; then
```
sergiocazzolato: I think we should exit after check the interface works when the plug is connected, otherwise we are not testing almost nothing on systems with no strict confinement.

sergiocazzolato: The stes we use are:
check default connection
connect if needed
test connected behabiour
check confinement and exit is needed
check disconnected behabiour

mvo5: TBH I just used the existing test (interfaces-dbus) as a blueprint. We should probably update this as well.

tests/main/interfaces-system-dbus/task.yaml
mvo5: Indeed, soon I will be able to do that :) right now the system dbus bits are only available in edge but once things are all green I will promote.

tests/main/interfaces-system-dbus/task.yaml
sergiocazzolato: @mvo5, interfaces-joystick is a nice example
Thanks

bboozzoo: > check confinement and exit is needed

We could execute a bunch of interface tests even if confinement is `partial`. Not this one in particular, as it requires apparmor dbus support, but a deal of other tests use apparmor features which are available outside of Ubuntu too, eg. on Arch and hopefully on openSUSE Leap 15.0 once it lands.

tests/main/interfaces-system-dbus/task.yaml
```diff
@@ -0,0 +1,46 @@
...
+    fi
+    MATCH "Permission denied" < call.error
+
+    echo "When the plug is connected"
+    snap connect test-snapd-dbus-consumer:dbus-system-test test-snapd-dbus-provider:dbus-system-test
+
+    echo "Then the consumer is able to call the provided method"
+    test-snapd-dbus-consumer.dbus-system-consumer | MATCH "hello world"
+
+
```
bboozzoo: Bunch of empty lines. Is it a placeholder for something?

PR #6409, opened 2019-01-22
interfaces/apparmor: mock presence of overlayfs root

zyga:
During the release of the snapd 2.37 we noticed that the Debian
builds performed in sbuild are failing on several unit tests. The same
source package would build file in pbuilder.

Investigation uncovered that sbuild is using overlayfs root internally.
This is picked up by the apparmor overlayfs detector and causes snapd to
generate an additional configuration file for snap-confine.

For reference, the offending entry from /proc/self/mountinfo:

	228 23 0:40 / / rw,relatime shared:119 - overlay sid-amd64-sbuild rw,lowerdir=/var/lib/schroot/union/underlay/sid-amd64-sbuild-85592074-da40-4faa-8b25-a354b207cdf2,upperdir=/var/lib/schroot/union/overlay/sid-amd64-sbuild-85592074-da40-4faa-8b25-a354b207cdf2/upper,workdir=/var/lib/schroot/union/overlay/sid-amd64-sbuild-85592074-da40-4faa-8b25-a354b207cdf2/work

The extra generated file was upsetting tests that looked at
/var/lib/snapd/apparmor/snap-confine.

### DIFF
--- a/interfaces/apparmor/backend_test.go
+++ b/interfaces/apparmor/backend_test.go
@@ -822,6 +822,10 @@ func (s *backendSuite) TestSetupSnapConfineGeneratedPolicyNoNFS(c *C) {
 	restore := apparmor.MockIsHomeUsingNFS(func() (bool, error) { return false, nil })
 	defer restore()
 
+	// Make it appear as if overlay was not used.
+	restore = apparmor.MockIsRootWritableOverlay(func() (string, error) { return "", nil })
+	defer restore()
+
 	// Intercept interaction with apparmor_parser
 	cmd := testutil.MockCommand(c, "apparmor_parser", "")
 	defer cmd.Restore()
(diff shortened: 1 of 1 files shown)


interfaces/apparmor/backend_test.go
```diff
@@ -822,6 +822,10 @@ func (s *backendSuite) TestSetupSnapConfineGeneratedPolicyNoNFS(c *C) {
 	restore := apparmor.MockIsHomeUsingNFS(func() (bool, error) { return false, nil })
 	defer restore()
 
+	// Make it appear as if overlay was not used.
+	restore = apparmor.MockIsRootWritableOverlay(func() (string, error) { return "", nil })
```
mvo5: This looks fine! I wonder if we should just add this mocking as part of the backendSuite.SetUp() just once?

zyga: I was looking at this, there are some parts that need the "other" behavior. I think that instead I would look at discarding all of the mocking (the problem with our mocking is that we never know we need it in advance unless tests fail) and passing an interface with those methods. Then tests always pass a fake implementation with appropriate answers and everything is explicit.

PR #6413, opened 2019-01-23
packaging: import debian salsa packaging work, add sbuild test and use in spead

mvo5:
This PR imports the debian salsa packaging work on 2.37 into our tree under packaging/debian-sid *without* importing the git history. Importing the git history was done in https://github.com/snapcore/snapd/pull/6393 but this PR fails because we have an invalid email in the git commit history that breaks the CLA check. 

During the debian spread tests, the packaging/debian-sid packaging is then used for building the deb package that is used during the full spread test run against debian-sid.

Once that is in we could a `debian` branch in our upstream tree that just contains the following change:
```
git rm debian
git mv packaging/debian-sid debian
git commit -m 'make debian a real directory instead of a symlink so that tooling like gbp works correctly'
```

Once we have that in our upstream tree we should be able to release with:
```
git checkout debian  
gbp buildpackage 
```

and do snapshot test builds with:
```
git checkout debian  
gbp buildpackage  --git-upstream-tree=SLOPPY
```

This work can also be merged back into salsa then. The idea is that we keep and update the debian packaging just like we do for ubuntu-14.04 and ubuntu-16.04+

There is more work to be done to unify the packaging more, I think we should use symlinks for some things like we do in 14.04.


mvo5:
From what I can tell this fails in check_journalctl_ready() in debian-sid - I wonder if a newer systemd in sid breaks things? Maybe @sergiocazzolato can have a look?

tests/lib/prepare-restore.sh
```diff
@@ -304,6 +304,37 @@ prepare_project() {
...
+        mv packaging/debian-sid debian
+
+        # get the build-deps
+        apt build-dep -y ./
+
+        # and ensure we don't take any of the vendor deps
+        rm -rf vendor/*/
+
+        # and create a fake upstream tarball
+        tar -c -z -f ../snapd_"$(dpkg-parsechangelog --show-field Version|cut -d- -f1)".orig.tar.gz --exclude=./debian --exclude=./.git .
```
zyga: I was thinking about this step lately. I think we can merge this as-is but I would like to promote the act of making a release tarball into something that we have proper tools for, right in the tree. That is, that the tarball that we make on release day is the same tarball we were making in CI all those days before that.

tests/main/sbuild/task.yaml
```diff
@@ -0,0 +1,21 @@
...
+    echo "Ensure we have sbuild"
+    apt install -y sbuild eatmydata
+
+    echo "Create a sid sbuild env"
+    eatmydata sbuild-createchroot --include=eatmydata,ccache,gnupg sid /srv/chroot/sid-amd64-sbuild http://deb.debian.org/debian
+
+    echo "Allow test user to run sbuild"
+    sbuild-adduser test
+
+    echo "And build it normally"
```
zyga: We should eventually document why this happens but I'm happy just having this so I don't even want to sound like I want to block it :)

tests/main/sbuild/task.yaml
```diff
@@ -0,0 +1,21 @@
...
+priority: 500
+
+systems: [debian-sid-*]
+
+execute: |
+    echo "Ensure we have sbuild"
+    apt install -y sbuild eatmydata
+
+    echo "Create a sid sbuild env"
+    eatmydata sbuild-createchroot --include=eatmydata,ccache,gnupg sid /srv/chroot/sid-amd64-sbuild http://deb.debian.org/debian
```
zyga: This is one of the places we could integrate the proxy information we set in spread.yaml. This would allow for efficient execution of this test in qemu.

packaging/debian-sid/tests/integrationtests
```diff
@@ -0,0 +1,51 @@
...
+systemctl daemon-reload
+
+# ensure we are not get killed too easily
+printf '%s\n' "-950" > /proc/$$/oom_score_adj
+
+# see what mem we have (for debugging)
+cat /proc/meminfo
+
+# ensure we can do a connect to localhost
+echo ubuntu:ubuntu|chpasswd
```
zyga: I didn't notice this before. Will it harm us eventually once Debian adt runner is updated and test suddenly start to execute?

mvo5: We use the following in debian/tests/control: `Restrictions: ..., breaks-testbed, isolation-machine` - ie this makes it clear that the tests will break the machine and a snapshot (or similar) needs to be restored once the tests are finished.

tests/main/sbuild/task.yaml
mvo5: Good point, the http{,s}_proxy env should be set already - what is needed in addition to this?

tests/lib/prepare-restore.sh
mvo5: Right, currently we just use the tarfile generated with "gbp buildpackage -S" from the release branch.

tests/main/sbuild/task.yaml
zyga: Hmm, I think if you have a generic proxy that works for anything this is sufficient. In practice I never do, I always have apt-specific proxy and I need to tell sbuild to use it explicitly by using something like:
```
sbuild-createchroot --include=eatmydata,ccache,gnupg sid /srv/chroot/sid-amd64-sbuild ${apt_proxy}/debian${apt_proxy_port}
 ```

PR #6417, opened 2019-01-23
snap: fix reexec from the snapd snap for classic snaps

mvo5:
This fixes a leftover path that was only looking at core when running snap-exec for classic snaps.


cmd/snap/cmd_run.go
```diff
@@ -448,13 +448,25 @@ func (x *cmdRun) snapRunTimer(snapApp, timer string, args []string) error {
...
-	return strings.HasPrefix(exe, dirs.SnapMountDir)
+	// no re-exec
+	if !strings.HasPrefix(exe, dirs.SnapMountDir) {
+		return filepath.Join(dirs.DistroLibExecDir, toolName), nil
+	}
+
+	// snapBase will be "/snap/{core,snapd}/$rev/" because
+	// the snap binary is always at $root/usr/bin/snap
+	snapBase := filepath.Clean(filepath.Join(filepath.Dir(exe), "..", ".."))
+	// Run snap-confine from the core/snapd snap. That
```
bboozzoo: The comment needs an update. Maybe something like:
```
// The tools in core/snapd snap are statically linked, or mostly statically,
// with the exception of libraries such as libudev and libc
```

cmd/snap/cmd_run.go
```diff
@@ -448,13 +448,26 @@ func (x *cmdRun) snapRunTimer(snapApp, timer string, args []string) error {
...
 	}
-	return strings.HasPrefix(exe, dirs.SnapMountDir)
+	// no re-exec
+	if !strings.HasPrefix(exe, dirs.SnapMountDir) {
+		return filepath.Join(dirs.DistroLibExecDir, toolName), nil
+	}
+
+	// snapBase will be "/snap/{core,snapd}/$rev/" because
+	// the snap binary is always at $root/usr/bin/snap
+	snapBase := filepath.Clean(filepath.Join(filepath.Dir(exe), "..", ".."))
```
zyga: Can we please check that the assumption you spell out holds?

```
// The logic below only works if the last two path components are /usr/bin
if !strings.HasSuffix(exe, "/usr/bin/" + filepath.Base(exe)) {
    return filepath.Join(dirs.DistroLibExecDir, toolName), nil
}
```

mvo5: Hm, we shold also probably warn about this, yes?

PR #6420, opened 2019-01-23
miscellaneous policy updates - 2.37

jdstrand:
https://github.com/snapcore/snapd/pull/6419 for 2.37

### DIFF
--- a/interfaces/apparmor/template.go
+++ b/interfaces/apparmor/template.go
@@ -638,6 +638,9 @@ profile snap-update-ns.###SNAP_INSTANCE_NAME### (attach_disconnected) {
   # Allow reading /proc/version. For release.go WSL detection.
   @{PROC}/version r,
 
+  # Allow reading somaxconn, required in newer distro releases
+  @{PROC}/sys/net/core/somaxconn r,
+
   # Allow reading the os-release file (possibly a symlink to /usr/lib).
   /{etc/,usr/lib/}os-release r,
 
--- a/interfaces/apparmor/template.go
+++ b/interfaces/apparmor/template.go
@@ -375,6 +375,9 @@ var defaultTemplate = `
   @{PROC}/net/dev r,
   @{PROC}/@{pid}/net/dev r,
 
+  # Read-only of this snap
+  /var/lib/snapd/snaps/@{SNAP_NAME}_*.snap r,
+
   # Read-only for the install directory
   # bind mount used here (see 'parallel installs', above)
   @{INSTALL_DIR}/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}/                   r,
--- a/interfaces/builtin/dbus.go
+++ b/interfaces/builtin/dbus.go
@@ -78,12 +78,16 @@ dbus (bind)
     bus=###DBUS_BUS###
     name=###DBUS_NAME###,
 
-# For KDE applications, also support alternation since they use org.kde.foo-PID
-# as their 'well-known' name. snapd does not allow declaring a 'well-known'
-# name that ends with '-[0-9]+', so this is ok.
+# For KDE applications and some other cases, also support alternation for:
+# - using org.kde.foo-PID as the 'well-known' name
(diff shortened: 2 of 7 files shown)


interfaces/apparmor/template.go
```diff
@@ -635,6 +638,9 @@ profile snap-update-ns.###SNAP_INSTANCE_NAME### (attach_disconnected) {
   # Allow reading /proc/version. For release.go WSL detection.
   @{PROC}/version r,
 
+  # Allow reading somaxconn, required in newer distro releases
+  @{PROC}/sys/net/core/somaxconn r,
```
bboozzoo: AFAIK Go runtime is notorious for this and triggers SELinux denials.

interfaces/builtin/home.go
```diff
@@ -82,6 +82,7 @@ audit deny @{HOME}/bin/{,**} wl,
 
 const homeConnectedPlugAppArmorWithAllRead = `
 # Allow non-owner read to non-hidden and non-snap files and directories
+capability dac_override,
```
bboozzoo: Won't this allow write too?

jdstrand: @bboozzoo - it does not override the apparmor 'owner' check and there are no non-owner writes in home, so, 'no', but if we allowed non-owner writes, then 'yes'. That said, looking at the forum topic again, I'm instead going to change this to dac_read_search.

bboozzoo: Thanks for explaining this in detail!

interfaces/apparmor/template.go
jdstrand: Yeah...

interfaces/apparmor/template.go
```diff
@@ -375,6 +375,9 @@ var defaultTemplate = `
   @{PROC}/net/dev r,
   @{PROC}/@{pid}/net/dev r,
 
+  # Read-only of this snap
+  /var/lib/snapd/snaps/@{SNAP_NAME}_*.snap r,
```
zyga: CC @bboozzoo is this ok or should we have SNAP_INSTANCE_NAME here?

interfaces/builtin/dbus.go
```diff
@@ -78,12 +78,16 @@ dbus (bind)
...
+# - using org.kde.foo-PID as the 'well-known' name
+# - using org.foo.cmd_<num>_<num> as the 'well-known' name
+# Note, snapd does not allow declaring a 'well-known' name that ends with
+# '-[0-9]+' or that contains '_'. Parallel installs of DBus services aren't
+# supported at this time, but if they were, this could allow a parallel
+# install'swell-known name to overlap with the normal install.
 dbus (bind)
     bus=###DBUS_BUS###
-    name=###DBUS_NAME###-[1-9]{,[0-9]}{,[0-9]}{,[0-9]}{,[0-9]}{,[0-9]},
+    name=###DBUS_NAME###{_,-}[1-9]{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9]}{,_[1-9]{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9]}},
```
zyga: Eh, that's a long one.

interfaces/apparmor/template.go
bboozzoo: Because of how snap files are named, the rule will match both snap name and snap instance name right now. So `@{SNAP_NAME}` snap will be able to access `@{SNAP_INSTANCE_NAME}` snap files too. We do not allow `_` in snap names, so there is no chance of having a similarly named, but unrelated snap file being accessed there.

We could make the rule more explicit, and include a comment:
```
# parallel installs: this leaks access to instance snap files to @{SNAP_NAME}
 /var/lib/snapd/snaps/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}_*.snap r,
```

PR #6425, opened 2019-01-24
interfaces: add yubikey interface and allow reading /run/udev/data/+power_supply:*

jdstrand:
* interfaces: add yubikey interface
    
    Yubikey devices either show up as hidraw devices or keyboard devices.
    This adds support for hidraw devices based on the list of hardware from:
    
    https://support.yubico.com/support/solutions/articles/15000006449-using-your-u2f-yubikey-with-linux
    
    Since there are many different kinds of hidraw devices, use an apparmor
    glob rule but selectively udev tag devices based on the list of upstream
    vendor and product IDs. In this manner, connected snaps only have access
    to the yubikey hidraw devices. A future PR may add support for yubikey
    keyboard devices (which simply show up as a evdev keyboards).

* interfaces/hardware-observe: allow reading /run/udev/data/+power_supply:*

Note the change to hardware-observe is the general rule for `+power_supply` whereas the the one in yubikey is a hid device rule for `+power_supply`. I could split the hardware-observe into a separate PR, but it was so small I just did it here.

### DIFF
--- a/interfaces/builtin/yubikey.go
+++ b/interfaces/builtin/yubikey.go
@@ -1,0 +1,123 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package builtin
+
+import (
+	"fmt"
+	"github.com/snapcore/snapd/interfaces"
+	"github.com/snapcore/snapd/interfaces/udev"
+)
+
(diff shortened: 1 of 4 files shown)


jdstrand:
@oSoMoN - this is the PR we discussed.

interfaces/builtin/yubikey.go
```diff
@@ -0,0 +1,123 @@
...
+        - core
+    deny-auto-connection: true
+`
+
+type yubikeyDevice struct {
+	VendorId, ProductId string
+}
+
+// https://github.com/Yubico/libu2f-host/blob/master/70-u2f.rules
+var yubikeyDevices = map[string]yubikeyDevice{
```
zyga: Is there a need for this to be a map? I would think that it is more natural to have a list of things like:
```
type yubikeyInfo struct {
   VendorIDPattern, ProductIDPattern string
   Name string
}
```

interfaces/builtin/yubikey.go
```diff
@@ -0,0 +1,123 @@
...
+
+package builtin
+
+import (
+	"fmt"
+	"github.com/snapcore/snapd/interfaces"
+	"github.com/snapcore/snapd/interfaces/udev"
+)
+
+const yubikeySummary = `allows access to yubikey devices`
```
mvo5: (bikesheed) Hm, this is slightly misleading, it looks like this supports access to a lot of security key U2F devices, not all yubikeys. Should the interface be something like fidoU2f or somesuch?

jdstrand: @mvo5 - renamed to u2f-devices. Thanks!

interfaces/builtin/yubikey.go
jdstrand: Done

PR #6426, opened 2019-01-24
cmd/snap-confine: refactor and cleanup of seccomp loading

zyga:
This branch contains a small refactor and cleanup of parts of the seccomp loading code.
I moved some chunks of existing code into seccomp-support-ext.c, which is using the less
annoying clang-format based indenting scheme.

### DIFF
--- a/cmd/snap-confine/seccomp-support.c
+++ b/cmd/snap-confine/seccomp-support.c
@@ -36,20 +36,7 @@
 #include "../libsnap-confine-private/string-utils.h"
 #include "../libsnap-confine-private/utils.h"
 
-#ifndef SECCOMP_FILTER_FLAG_LOG
-#define SECCOMP_FILTER_FLAG_LOG 2
-#endif
-
-#ifndef seccomp
-// prototype because we build with -Wstrict-prototypes
-int seccomp(unsigned int operation, unsigned int flags, void *args);
-
-int seccomp(unsigned int operation, unsigned int flags, void *args)
-{
-	errno = 0;
-	return syscall(__NR_seccomp, operation, flags, args);
-}
-#endif
+#include "seccomp-support-ext.h"
 
 static const char *filter_profile_dir = "/var/lib/snapd/seccomp/bpf/";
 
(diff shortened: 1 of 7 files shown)


mvo5:
Some of the failures look real, e.g. `cannot apply seccomp profile: Invalid argument` in the debian9 upgrade test.

zyga:
Indeed, sorry about this. I will solve this quickly.

zyga:
@mvo5 should be good now, void * is happy to take void ** pointer by accident and that code only happens on older kernels.

cmd/snap-confine/seccomp-support.h
```diff
@@ -18,11 +18,34 @@
...
+ * seconds may elapse before the function gives up. Unless
+ * $SNAP_CONFINE_MAX_PROFILE_WAIT environment variable dictates otherwise, the
+ * default wait time is 120 seconds.
+ *
+ * A profile may contain valid BPF program or the string "@unrestricted\n".  In
+ * the former case the profile is applied to the current process using
+ * sc_apply_seccomp_filter. In the latter case no action takes place.
+ *
+ * The return value indicates if the process uses confinement or runs under the
+ * special non-confining "@unrestricted" profile.
```
jdstrand: Nice comment! :)

cmd/snap-confine/seccomp-support.c
```diff
@@ -157,87 +144,44 @@ int sc_apply_seccomp_bpf(const char *filter_profile)
 		sleep(1);
 	}
 
+	/* TODO: replace this with descending open/openat with validation callback
+	 * that returns a path descriptor to avoid TOCTOU */
```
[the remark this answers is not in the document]
zyga: While this is true, once we validated the path we saw it correct at that point in time we could still do one step more. As we descend and validate we can keep the state (similar to how we do in snap-update-ns) and then we can pass the file descriptor to the validated entity back to the caller. This would allow us to perfectly open the thing that we checked.

I don't think there is a way to exploit this but I think it's okay to improve it at one point.

cmd/snap-confine/seccomp-support.c
```diff
@@ -36,20 +36,7 @@
 #include "../libsnap-confine-private/string-utils.h"
 #include "../libsnap-confine-private/utils.h"
 
-#ifndef SECCOMP_FILTER_FLAG_LOG
```
mvo5: Pardon my ignorance, why does this move into a new file? I mean, it seems seccomp-support.c is an appropriate place for this code. Whats the advantage of the new  `-ext.h`  file? If we rename, maybe to `-filter.h`?

zyga: It moved along with the place where it is needed.

I used a new file to avoid the crippled `intent` based formatter. New files use `clang-format` which produces much nicer output.

PR #6427, opened 2019-01-24
interfaces: add block-devices interface

jdstrand:
Some applications need access to raw disks (eg, smartctl and hdparm) or
    disks at the block level (eg, flashing tools). This interface provides
    this access. Since read access gives the snap access to all data on the
    device and write access allows changing the contents of the disk, this
    interface is considered a "super-privileged" interface since it grants
    device ownership to the snap.
    
    Tested with 'hdparm -t /dev/sda', 'smartctl --test=short /dev/sda' and
    'smartctl --all /dev/sda'.
    
    References:
    https://forum.snapcraft.io/t/request-for-classic-confinement-sas2ircu/9023
    https://forum.snapcraft.io/t/classic-confinement-review-request-for-fwup/7379
    https://forum.snapcraft.io/t/classic-confinement-request-for-the-wimlib-snap/7424

### DIFF
--- a/interfaces/builtin/block_devices.go
+++ b/interfaces/builtin/block_devices.go
@@ -1,0 +1,98 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package builtin
+
+// Only allow raw disk devices; not loop, ram, CDROM, generic SCSI, network,
+// tape, raid, etc devices or disk partitions
+const blockDevicesSummary = `allows access to disk block devices`
+
+const blockDevicesBaseDeclarationPlugs = `
+  block-devices:
+    allow-installation: false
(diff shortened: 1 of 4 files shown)


interfaces/builtin/block_devices.go
```diff
@@ -0,0 +1,98 @@
...
+const blockDevicesBaseDeclarationSlots = `
+  block-devices:
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
+`
+
+// https://www.kernel.org/doc/Documentation/admin-guide/devices.txt
+// For now, only list common devices and skip the following:
```
zyga: Good call!

interfaces/builtin/block_devices.go
```diff
@@ -0,0 +1,98 @@
...
+const blockDevicesConnectedPlugAppArmor = `
+# Description: Allow write access to raw disk block devices.
+
+@{PROC}/devices r,
+/run/udev/data/b[0-9]*:[0-9]* r,
+/sys/block/ r,
+/sys/devices/**/block/** r,
+
+# Access to raw devices, not individual partitions
+/dev/hd[a-t] rw,                         # IDE, MFM, RLL
```
zyga: It's curious that the range `a-z` is not universally used.

interfaces/builtin/block_devices.go
```diff
@@ -0,0 +1,98 @@
...
+/dev/sd{,[a-h]}[a-z] rw,                 # SCSI
+/dev/sdi[a-v] rw,                        # SCSI continued
```
zyga: Do you think this list should include `/dev/sr[[:digit:]]` as well? It's more of an optical thing but then again, it is a block device too.

interfaces/builtin/block_devices.go
```diff
@@ -0,0 +1,98 @@
...
+	`KERNEL=="mpt2ctl*"`,
+	`KERNEL=="megaraid_sas_ioctl_node"`,
+}
+
+type blockDevicesInterface struct {
+	commonInterface
+}
+
+func init() {
+	registerIface(&blockDevicesInterface{commonInterface{
```
zyga: This looks good.

PR #6428, opened 2019-01-24
interfaces: add display-control interface

jdstrand:
interfaces: add display-control interface
    
    References:
    https://forum.snapcraft.io/t/interface-to-change-display-brightness/8890

### DIFF
--- a/interfaces/builtin/common_test.go
+++ b/interfaces/builtin/common_test.go
@@ -21,6 +21,7 @@ package builtin
 
 import (
 	. "gopkg.in/check.v1"
+	"os"
 
 	"github.com/snapcore/snapd/interfaces/apparmor"
 	"github.com/snapcore/snapd/interfaces/udev"
@@ -81,6 +82,15 @@ func MockEvalSymlinks(test *testutil.BaseTest, fn func(string) (string, error)) 
 	evalSymlinks = fn
 	test.AddCleanup(func() {
 		evalSymlinks = orig
+	})
+}
+
+// MockReadDir replaces the io/ioutil.ReadDir function used inside the caps package.
+func MockReadDir(test *testutil.BaseTest, fn func(string) ([]os.FileInfo, error)) {
+	orig := readDir
+	readDir = fn
+	test.AddCleanup(func() {
+		readDir = orig
 	})
 }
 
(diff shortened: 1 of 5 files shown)


jdstrand:
I only had access to two devices with different backlights and created as general a rule as I could for them. I figure we'll keep adding rules as hardware is found. Looking into this more, I have a better way.

jdstrand:
@mvo5 and @zyga - I looked into how other software is doing this and found that they are looking at the symlinks in /sys/class/backlight, so I came up with something that will work well for primary screens (see the commit message). I also noticed some DBus services that could be in the interface and added them.

interfaces/builtin/common_test.go
```diff
@@ -84,6 +85,15 @@ func MockEvalSymlinks(test *testutil.BaseTest, fn func(string) (string, error))
 	})
 }
 
+// MockReadDir replaces the io/ioutil.ReadDir function used inside the caps package.
+func MockReadDir(test *testutil.BaseTest, fn func(string) ([]os.FileInfo, error)) {
```
zyga: Typically we return a function that restores the mocking operation. I don't oppose this but we can change that on a drive-by next time.

interfaces/builtin/display_control.go
```diff
@@ -0,0 +1,137 @@
...
+    peer=(label=unconfined),
+
+/sys/class/backlight/ r,
+`
+
+type displayControlInterface struct {
+	commonInterface
+}
+
+func (iface *displayControlInterface) dereferencedBacklightPaths() []string {
```
zyga: This looks good.

interfaces/builtin/display_control.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	return paths
+}
+
+func (iface *displayControlInterface) AppArmorConnectedPlug(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	// add the static rules
+	spec.AddSnippet(displayControlConnectedPlugAppArmor)
+
+	// add the detected rules
+	for _, p := range iface.dereferencedBacklightPaths() {
```
zyga: I think this is still some time away but this feels like a good candidate for hot plug transformation.

jdstrand: Yes, I agree (and mentioned that in the git commit ;)

interfaces/builtin/common_test.go
jdstrand: I was just following what we did for MockEvalSymlinks, directly above.

PR #6434, opened 2019-01-25
interfaces: add u2f-devices interface (2.37)

jdstrand:
https://github.com/snapcore/snapd/pull/6425 for 2.37

### DIFF
--- a/interfaces/builtin/hardware_observe.go
+++ b/interfaces/builtin/hardware_observe.go
@@ -57,6 +57,7 @@ capability sys_admin,
 
 # power information
 /sys/power/{,**} r,
+/run/udev/data/+power_supply:* r,
 
 # interrupts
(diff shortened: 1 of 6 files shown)


mvo5:
This looks fine in itself, however the way the backport is done looks like it will cause problems. It looks like its one squashed commit but AFAICT we merged master via a merge commit. So when this gets merged back into master there will be conflicts. By cherry-picking the invidial commits of #6425 this can be avoided mostly. Or am I missing something here? If not, I'm happy to open a new backport PR that cherry picks the individual commits.

jdstrand:
> I mark as "change request" - what I really need is "answer request" here :)

Ok, I created a new branch from release/2.37, cherry-picked the 5 commits and force pushed.

PR #6444, opened 2019-01-29
[RFC] snap-confine: revert "cmd/snap-confine,snap-update-ns: discard quirks" (2.37)

mvo5:
This reverts commit 1da9316e717d119852b827becb5a17b33713d032.


mvo5:
We may still need this if it turns out more software (puppet?) is affected by the removal of wirtable /var/lib/ when we removed the quirks.

PR #6445, opened 2019-01-29
[RFC] snap-confine: revert "cmd/snap-confine,snap-update-ns: discard quirks"

mvo5:
This reverts commit 1da9316e717d119852b827becb5a17b33713d032.

This will unbreak people using jenkins which uses /var/lib/jenkins as its HOME. The quirks system made this HOME work (by luck/accient) and removing it broke all jenkins users.


mvo5:
We may still need this if it turns out more software (puppet?) is affected by the removal of wirtable /var/lib/ when we removed the quirks.

mvo5:
Closing for now but but but I have a feeling this one might come back.

PR #6451, opened 2019-01-30
tests: update smoke/sandbox test for armhf

mvo5:
While running the beta validation on the armhf I noticed that the
behaviour of illegal syscalls is different from the other arches
we have. Most arches just give ENOSYS in errno. However on armhf
the process is killed with SIGILL.

This PR updates the test to account for that.

### DIFF
--- a/tests/smoke/sandbox/task.yaml
+++ b/tests/smoke/sandbox/task.yaml
@@ -62,7 +62,14 @@ execute: |
             print(errno.errorcode[get_errno()])
     EOF
     echo "Running random syscall gives ENOSYS normally"
-    python3 sec.py 22082007 | MATCH ENOSYS
+    if [ "$(uname -m)" = "armv7l" ]; then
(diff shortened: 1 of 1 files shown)


tests/smoke/sandbox/task.yaml
```diff
@@ -62,7 +62,14 @@ execute: |
             print(errno.errorcode[get_errno()])
     EOF
     echo "Running random syscall gives ENOSYS normally"
-    python3 sec.py 22082007 | MATCH ENOSYS
+    if [ "$(uname -m)" = "armv7l" ]; then
+        # armhf kills the process with si_signo=SIGKILL, si_code=ILL_ILLTRP
+        python3 sec.py 22082007 || ret="$?"
```
bboozzoo: Looks like ENOSYS or SIGILL depends on the syscall number you roll: https://elixir.bootlin.com/linux/v5.0-rc4/source/arch/arm/kernel/traps.c#L596

PR #6456, opened 2019-01-30
interfaces: add network-manager-observe interface

jdstrand:
NetworkManager offers a very large API that is wholly exposed via the
network-manager interface. Some applications (eg, QtNetworkBearer) use
NetworkManager to check for connectivity and therefore do not require
access to the entire API. Implement the network-manager-observe
interface that exposes the Get* and List methods required for verifying
connectivity. Since this grants access to things like MAC addresses and
previous networks, treat this like network-observe and require manual
connection. This interface does not grant access to NetworkManager's
GetSecrets() API.
    
On classic distro, this interface lets the application communicate with
the system NetworkManager using the specified APIs. On all snaps
systems, a providing snap interested in exposing this smaller API is
expected to slot this interface in addition to the fuller
network-manager interface, which contains all the permanent slot policy
needed to run as NetworkManager (ie, this interface is currently not
standalone as there are no known applications that only implement this
reduced NM API).

References:
https://forum.snapcraft.io/t/qt-network-bearer-networkmanager-access-extend-network-observe/8399

### DIFF
--- a/interfaces/builtin/network_manager_observe.go
+++ b/interfaces/builtin/network_manager_observe.go
@@ -1,0 +1,146 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package builtin
+
+import (
+	"strings"
+
+	"github.com/snapcore/snapd/interfaces"
+	"github.com/snapcore/snapd/interfaces/apparmor"
+	"github.com/snapcore/snapd/release"
+	"github.com/snapcore/snapd/snap"
+)
+
+const networkManagerObserveBaseDeclarationSlots = `
+  network-manager-observe:
+    allow-installation:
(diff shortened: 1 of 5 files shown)


interfaces/builtin/network_manager_observe.go
```diff
@@ -0,0 +1,146 @@
...
+        - app
+        - core
+    deny-auto-connection: true
+    deny-connection:
+      on-classic: false
+`
+
+const networkManagerObserveSummary = `allows observing NetworkManager settings`
+
+const networkManagerObserveConnectedSlotAppArmor = `
```
bboozzoo: NM sends a couple of signals that could be useful too: org.freedesktop.NetworkManger.PropertiesChanged and org.freedesktop.NetworkManager.StateChanged on /org/freedesktop/NetworkManger. Note that org.freedesktop.NetworkManger.PropertiesChanged  isn't exactly the same as freedeskto.DBus.Properties.PropertiesChanged, but while at it we may want to allow both.

There is also a couple of potentially useful signals in the Settings interface (/org/freedesktop/NetworkManager/Settings path)

bboozzoo: Looked at `gio`, the `GNetworkMonitor` backend for NetworkManager uses `org.freedesktop.NetworkManger.PropertiesChanged` signal to trigger the refresh of a bunch of variables.

jdstrand: Ok, I added the corresponding signal rules for allowed method rules. I'm told this is all that is required by QtNetworkBearer; we can iterate on the interface as needed (but the signals you mentioned should be covered).

jdstrand: > NM sends a couple of signals that could be useful too: org.freedesktop.NetworkManger.PropertiesChanged and org.freedesktop.NetworkManager.StateChanged on /org/freedesktop/NetworkManger. Note that org.freedesktop.NetworkManger.PropertiesChanged isn't exactly the same as freedeskto.DBus.Properties.PropertiesChanged, but while at it we may want to allow both.
> 
> There is also a couple of potentially useful signals in the Settings interface (/org/freedesktop/NetworkManager/Settings path)

@bboozzoo - this is done.

PR #6462, opened 2019-01-31
snap-confine: fix incorrect "sanity timeout 3s" message

mvo5:
The code has a log message that says that the timeout is
initialized to 3 seconds. However the code actually uses
6 seconds. This "refactors" things so that timeout and
message stay in sync.

### DIFF
--- a/cmd/libsnap-confine-private/locking.c
+++ b/cmd/libsnap-confine-private/locking.c
@@ -34,6 +34,10 @@
 #include "../libsnap-confine-private/string-utils.h"
 #include "../libsnap-confine-private/utils.h"
 
+// SANITY_TIMEOUT is the timeout in seconds that is used when
+// "sc_enable_sanity_timeout()" is called
+static const int SANITY_TIMEOUT=6;
+
 /**
  * Flag indicating that a sanity timeout has expired.
  **/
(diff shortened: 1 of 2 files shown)


cmd/libsnap-confine-private/locking.c
```diff
@@ -61,8 +65,8 @@ void sc_enable_sanity_timeout(void)
 	if (sigaction(SIGALRM, &act, NULL) < 0) {
 		die("cannot install signal handler for SIGALRM");
 	}
-	alarm(6);
-	debug("sanity timeout initialized and set for three seconds");
+	alarm(SANITY_TIMEOUT);
+	debug("sanity timeout initialized and set for %i seconds", SANITY_TIMEOUT);
```
zyga: D'oh :)

cmd/libsnap-confine-private/locking.c
```diff
@@ -34,6 +34,10 @@
 #include "../libsnap-confine-private/string-utils.h"
 #include "../libsnap-confine-private/utils.h"
 
+// SANITY_TIMEOUT is the timeout in seconds that is used when
+// "sc_enable_sanity_timeout()" is called
+static const int SANITY_TIMEOUT=6;
```
bboozzoo: `SANITY_TIMEOUT = 6` unless it's our formatter's doing.

PR #6464, opened 2019-02-01
[RFC] many: add simple performance measure tool mostly for firstboot

mvo5:
This PR adds a very simplisit "measure" module to the tree that
can be used to take measurements of things that take time.

The API is minimal, the intended use (for now) is:
```
func foo() {
   m := measure.New("things happening in foo")
   defer m.Done()
   ...
}
```
The PR adds some measures into the firstboot code but we need to
experiment a bit too see where else we need to measure.

With that there is a new `snap debug measures` command that can
be used to get the most recent measurements.

This will result in something like:
```
snap debug measures
Measures:
installing for core_x1.snap took 533.215754ms
installing for pc-kernel_184.snap took 1.076903215s
installing for pc_24.snap took 13.288668ms
populateStateFromSeed took 1.64349125s
```
Making the output more structed is probably a good idea too :)

### DIFF
--- a/cmd/snap/cmd_debug_measures.go
+++ b/cmd/snap/cmd_debug_measures.go
@@ -1,0 +1,59 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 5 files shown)


mvo5:
Superseded by the work from @bboozzoo

PR #6498, opened 2019-02-13
cmd/libsnap: add cgroup-pids-support module

zyga:
As a part of refresh app awareness work snapd needs to differentiate snap
services from other snap commands as well as snap hooks. Services can be
reliably stopped and restarted during a refresh process but apps have no such
capability. To know that a refresh can proceed, before stopping services, snapd
can compute the set of PIDs associated to a given snap and subtract the subset
of PIDs associated with each service. If any processes remain then some
non-service processes must be running (and one can determine which as well).

To track per-app and per-hook processes we can use the PID control group
without enabling any resource allocation constraints.

### DIFF
--- a/cmd/libsnap-confine-private/cgroup-freezer-support.c
+++ b/cmd/libsnap-confine-private/cgroup-freezer-support.c
@@ -1,3 +1,20 @@
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
 // For AT_EMPTY_PATH and O_PATH
 #define _GNU_SOURCE
 
--- a/cmd/libsnap-confine-private/cgroup-freezer-support.c
+++ b/cmd/libsnap-confine-private/cgroup-freezer-support.c
@@ -11,6 +11,7 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#include "cgroup-support.h"
 #include "cleanup-funcs.h"
 #include "string-utils.h"
 #include "utils.h"
@@ -19,51 +20,9 @@ static const char *freezer_cgroup_dir = "/sys/fs/cgroup/freezer";
 
 void sc_cgroup_freezer_join(const char *snap_name, pid_t pid)
 {
-	// Format the name of the cgroup hierarchy.
 	char buf[PATH_MAX] = { 0 };
 	sc_must_snprintf(buf, sizeof buf, "snap.%s", snap_name);
-
-	// Open the freezer cgroup directory.
-	int cgroup_fd SC_CLEANUP(sc_cleanup_close) = -1;
-	cgroup_fd = open(freezer_cgroup_dir,
-			 O_PATH | O_DIRECTORY | O_NOFOLLOW | O_CLOEXEC);
-	if (cgroup_fd < 0) {
-		die("cannot open freezer cgroup (%s)", freezer_cgroup_dir);
-	}
-	// Create the freezer hierarchy for the given snap.
-	if (mkdirat(cgroup_fd, buf, 0755) < 0 && errno != EEXIST) {
-		die("cannot create freezer cgroup hierarchy for snap %s",
-		    snap_name);
-	}
-	// Open the hierarchy directory for the given snap.
-	int hierarchy_fd SC_CLEANUP(sc_cleanup_close) = -1;
-	hierarchy_fd = openat(cgroup_fd, buf,
-			      O_PATH | O_DIRECTORY | O_NOFOLLOW | O_CLOEXEC);
-	if (hierarchy_fd < 0) {
-		die("cannot open freezer cgroup hierarchy for snap %s",
-		    snap_name);
-	}
-	// Since we may be running from a setuid but not setgid executable, ensure
-	// that the group and owner of the hierarchy directory is root.root.
-	if (fchownat(hierarchy_fd, "", 0, 0, AT_EMPTY_PATH) < 0) {
-		die("cannot change owner of freezer cgroup hierarchy for snap %s to root.root", snap_name);
-	}
-	// Open the tasks file.
-	int tasks_fd SC_CLEANUP(sc_cleanup_close) = -1;
-	tasks_fd = openat(hierarchy_fd, "tasks",
-			  O_WRONLY | O_NOFOLLOW | O_CLOEXEC);
-	if (tasks_fd < 0) {
-		die("cannot open tasks file for freezer cgroup hierarchy for snap %s", snap_name);
-	}
-	// Write the process (task) number to the tasks file. Linux task IDs are
-	// limited to 2^29 so a long int is enough to represent it.
-	// See include/linux/threads.h in the kernel source tree for details.
-	int n = sc_must_snprintf(buf, sizeof buf, "%ld", (long)pid);
-	if (write(tasks_fd, buf, n) < n) {
-		die("cannot move process %ld to freezer cgroup hierarchy for snap %s", (long)pid, snap_name);
-	}
-	debug("moved process %ld to freezer cgroup hierarchy for snap %s",
-	      (long)pid, snap_name);
+	sc_cgroup_create_and_join(freezer_cgroup_dir, buf, pid);
 }
 
 bool sc_cgroup_freezer_occupied(const char *snap_name)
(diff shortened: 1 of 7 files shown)


cmd/libsnap-confine-private/cgroup-pids-support.c
```diff
@@ -0,0 +1,57 @@
...
+    }
+    // Open the hierarchy directory for the given snap security tag.
+    int hierarchy_fd SC_CLEANUP(sc_cleanup_close) = -1;
+    hierarchy_fd = openat(cgroup_fd, snap_security_tag, O_PATH | O_DIRECTORY | O_NOFOLLOW | O_CLOEXEC);
+    if (hierarchy_fd < 0) {
+        die("cannot open pids cgroup hierarchy for snap security tag %s", snap_security_tag);
+    }
+    // Since we may be running from a setuid but not setgid executable, ensure
+    // that the group and owner of the hierarchy directory is root.root.
+    if (fchownat(hierarchy_fd, "", 0, 0, AT_EMPTY_PATH) < 0) {
```
mvo5: Between the mkdirat() and this fchownat() there is a small window where the owner will not be root:root - is that a cause for concern? If so, should we set permissions to 0700 and only after the chown to the real 0755?

cmd/libsnap-confine-private/cgroup-pids-support.c
```diff
@@ -0,0 +1,57 @@
...
+    // Open the tasks file.
+    int tasks_fd SC_CLEANUP(sc_cleanup_close) = -1;
+    tasks_fd = openat(hierarchy_fd, "tasks", O_WRONLY | O_NOFOLLOW | O_CLOEXEC);
+    if (tasks_fd < 0) {
+        die("cannot open tasks file for pids cgroup hierarchy for snap security tag %s", snap_security_tag);
+    }
+    // Write the process (task) number to the tasks file. Linux task IDs are
+    // limited to 2^29 so a long int is enough to represent it.
+    // See include/linux/threads.h in the kernel source tree for details.
+    char buf[PATH_MAX] = {0};
```
mvo5: Why PATH_MAX here? If we know we write max 2^29 we could select a smaller buffer maybe (mind you, still large enough but 4k seems excessive for a pid. Or am I missing something?

cmd/libsnap-confine-private/cgroup-pids-support.c
```diff
@@ -0,0 +1,57 @@
...
+#include <unistd.h>
+
+#include "cleanup-funcs.h"
+#include "string-utils.h"
+#include "utils.h"
+
+static const char *pids_cgroup_dir = "/sys/fs/cgroup/pids";
+
+void sc_cgroup_pids_join(const char *snap_security_tag, pid_t pid) {
+    // Open the pids cgroup directory.
```
mvo5: We have very similar code in sc_cgroup_freezer_join at leat the fist ~30 lines or so look quite similar - is there a chance that we can consolidate these two functions?

cmd/libsnap-confine-private/cgroup-pids-support.c
zyga: I was just lazy, yeah, we could use a smaller buffer but this was a safe default.

cmd/libsnap-confine-private/cgroup-pids-support.c
zyga: Once we get a third version, yes, before that I'd rather not.

cmd/libsnap-confine-private/cgroup-pids-support.c
zyga: There's no cause for concern since there is no g+w bit set. There is nothing inside that is of any sensitivity.

zyga:
The associated AppArmor changes are in https://github.com/snapcore/snapd/pull/6499

cmd/libsnap-confine-private/cgroup-pids-support.c
mvo5: I respectfully disagree, it seems something like  http://paste.ubuntu.com/p/SSg3qf5xtj/ is all that is needed to unify the two helpers and I don't think we want duplication like this in something we merge into master even (or maybe especially) if we know this will change anyway. It means the reviewers will have to review this carefully (this is suid root code afterall) and when this is changed again do same same. Or am I missing something here?

zyga: https://en.wikipedia.org/wiki/Rule_of_three

pedronis: I agree with @mvo5 here,  we don't need 3, it's the nature of cgroups to have special "tasks" files to which you write pid to add to them. Is the freezer cgroup "tasks" different from the "tasks" here?  apart being in different places. Notice that writing this again completely you are asking people to re-review the same code essentially again.

cmd/libsnap-confine-private/cgroup-pids-support.c
bboozzoo: Seems like 12 bytes should be more than enough. IIRC sc_must_sprintf will also `die()` if it cannot fit the thing into the buffer.

cmd/libsnap-confine-private/cgroup-pids-support.c
zyga: I'll refactor this and post it back.

zyga: Done now

cmd/libsnap-confine-private/cgroup-support.c
```diff
@@ -0,0 +1,59 @@
...
+    if (fchownat(hierarchy_fd, "", 0, 0, AT_EMPTY_PATH) < 0) {
+        die("cannot change owner of cgroup hierarchy %s/%s to root.root", parent, name);
+    }
+    // Open the tasks file.
+    int tasks_fd SC_CLEANUP(sc_cleanup_close) = -1;
+    tasks_fd = openat(hierarchy_fd, "tasks", O_WRONLY | O_NOFOLLOW | O_CLOEXEC);
+    if (tasks_fd < 0) {
+        die("cannot open file %s/%s/tasks", parent, name);
+    }
+    FILE *stream SC_CLEANUP(sc_cleanup_file) = NULL;
```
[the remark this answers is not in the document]
pedronis: I don't see a lot of fprintf usage on anything but stderr so far in the lib or snap-confine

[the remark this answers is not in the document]
zyga: We use `fdopen` in snap-confine already. This is just easier to do correctly than hand-coding the right buffer size.

cmd/libsnap-confine-private/cgroup-support.c
```diff
@@ -0,0 +1,59 @@
...
+    if (stream == NULL) {
+        die("cannot open stream of %s/%s/tasks", parent, name);
+    }
+    // stream now owns tasks_fd
+    tasks_fd = -1;
+    // Write the process (task) number to the tasks file. Linux task IDs are
+    // limited to 2^29 so a long int is enough to represent it.
+    // See include/linux/threads.h in the kernel source tree for details.
+    int n = fprintf(stream, "%ld", (long)pid);
+    if (n < 0) {
```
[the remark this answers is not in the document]
zyga: Indeed, I performed a quick test and it seems you are right. Damn, I think stdio is full of surprises.

cmd/libsnap-confine-private/cgroup-support.c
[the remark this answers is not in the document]
pedronis: I said fprintf

cmd/libsnap-confine-private/cgroup-support.c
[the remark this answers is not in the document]
pedronis: Indeed, still unclear that the convenience is enough here to justify the switch. considering the fflush as well.

cmd/libsnap-confine-private/cgroup-freezer-support.c
```diff
@@ -19,51 +37,9 @@ static const char *freezer_cgroup_dir = "/sys/fs/cgroup/freezer";
 
 void sc_cgroup_freezer_join(const char *snap_name, pid_t pid)
 {
-	// Format the name of the cgroup hierarchy.
 	char buf[PATH_MAX] = { 0 };
```
mvo5: Renaming to `snap_security_tag` increases readability now I thinkl.

cmd/libsnap-confine-private/cgroup-pids-support.c
```diff
@@ -0,0 +1,26 @@
...
+ *
+ */
+
+#include "cgroup-pids-support.h"
+
+#include "cgroup-support.h"
+
+static const char *pids_cgroup_dir = "/sys/fs/cgroup/pids";
+
+void sc_cgroup_pids_join(const char *snap_security_tag, pid_t pid) {
```
mvo5: The sc_cgroup_pids_join() takes a snap_security tag but sc_croup_freezer_join() takes a snap_name. Might be nice to unify this (unless there is a good reason for this to be different of course).

cmd/libsnap-confine-private/cgroup-freezer-support.c
zyga: This is not the security tag however. It is really the snap name alone.

cmd/libsnap-confine-private/cgroup-pids-support.c
zyga: It is different for a reason. We may be able to change this in v2 with a real hierarchy where there's a snapd hierarchy, populated by snap names and finally by security tags (apps and hooks) where actual processes reside.

PR #6499, opened 2019-02-13
cmd/snap-confine: allow moving tasks to pids cgroup

zyga:
To potentially allow snapd to track processes per app/hook we must allow
snap-confine to move the created task to a dedicated pid cgroup. This
patch contains the apparmor permissions necessary to do so.


zyga:
@jdstrand we are doing this so that we can associate processes with each snap app or hook. This is a part of the refresh-app-awareness work where we can figure out if any non-service processes are running.

PR #6503, opened 2019-02-13
tests: use snap which takes 15 seconds to install on retryable-error test

sergiocazzolato:
This change makes sure the change is still "doing" when the error code is retrieved. 

See error: https://paste.ubuntu.com/p/q9Sr23mDq8

### DIFF
--- a/tests/main/retryable-error/task.yaml
+++ b/tests/main/retryable-error/task.yaml
@@ -4,13 +4,15 @@ summary: Ensure exit code for retryable error works
 backends: [-autopkgtest]
 
 execute: |
-    echo "Install a snap"
-    snap install test-snapd-tools &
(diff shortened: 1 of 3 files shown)


tests/main/retryable-error/task.yaml
```diff
@@ -4,13 +4,15 @@ summary: Ensure exit code for retryable error works
 backends: [-autopkgtest]
 
 execute: |
-    echo "Install a snap"
-    snap install test-snapd-tools &
+    echo "Install a snap which takes some time to be installed"
+    snap pack "$TESTSLIB/snaps/test-snapd-sleep-install"
+    snap install --dangerous test-snapd-sleep-install_*.snap &
```
zyga: Would using  `install_local &` work?

PR #6504, opened 2019-02-13
tests: not checking 'tracking channel' after refresh core on nested execution

sergiocazzolato:
The test was failing but after a discussion the core is working as designed.

### DIFF
--- a/tests/lib/nested.sh
+++ b/tests/lib/nested.sh
@@ -200,7 +200,11 @@ del_device(){
     echo "device deleted"
 }
 
-get_nested_core_revision(){
-    ID=$1
-    execute_remote "snap info core" | awk "/${ID}: / {print(\$3)}" | sed -e 's/(\(.*\))/\1/'
+get_nested_core_revision_for_channel(){
+    local CHANNEL=$1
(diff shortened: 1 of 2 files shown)


tests/lib/nested.sh
```diff
@@ -200,7 +200,11 @@ del_device(){
     echo "device deleted"
 }
 
-get_nested_core_revision(){
-    ID=$1
-    execute_remote "snap info core" | awk "/${ID}: / {print(\$3)}" | sed -e 's/(\(.*\))/\1/'
+get_nested_core_revision_for_channel(){
+    local CHANNEL=$1
+    execute_remote "snap info core" | awk "/${CHANNEL}: / {print(\$4)}" | sed -e 's/(\(.*\))/\1/'
```
zyga: This is good

tests/lib/nested.sh
```diff
@@ -200,7 +200,11 @@ del_device(){
...
+}
+
+get_nested_core_revision_installed(){
+    execute_remote "snap info core" | awk "/installed: / {print(\$3)}" | sed -e 's/(\(.*\))/\1/'
```
zyga: This is good too

PR #6505, opened 2019-02-14
packaging: avoid race in snapd.postinst

mvo5:
I added Steve hoping there is a better way to influence the order of
the debhelper snippets.

We need to copy the debhelper apparmor snippet into our manually
written snapd.posint because what dehelper is doing here is wrong.

It automatically added bits which run in the *wrong* order:
- it (re)starts snapd first
- it reloads the snap-confine apparmor profile *after* that
which of course means that on an upgrade there is a time window
in which snapd is available for the system-key check but
snap-confine has an old profile and will die because it tries
to do things that it cannot do with that old profile. By moving
the apparmor load here we ensure that the profile for snap-confine
is loaded before snapd restarts which means that when snapd is
ready snap-confine also has the right profile.

We need to backport this to 2.34 in security even :/


mvo5:
Closing while looking for a better approach

PR #6507, opened 2019-02-14
cmd/snap-confine: allow writes to /var/lib/**

zyga:
This patch restores permissions that snap-confine lost during the
2.36->2.37 transition. This allows a specific, otherwise unsupported,
use case to remain operational.

When an LXD container running something like Jenkins, which keeps data
in /var/lib/<foo>, wishes to use classic snaps and redirect the output
from said classic snap to a log in /var/lob/<foo>, this use case would
work in 2.36 because of additional permissions necessary for the
so-called "quirk" system that used to be present then. The 2.37 release
dropped that, along the extra permissions.

This issue resurfaced during the 2.37.1, 2.37.2 and hopefully the
upcoming 2.37.3 will fix it once and for all. The added regression test
should ensure that things keep working in case we refactor this at some
time in the future.

The fix is unfortunate but in absence of more advanced apparmor features
(e.g. delegation) we have no other choice.


zyga:
Force pushed to remove trailing whitespace.

PR #6508, opened 2019-02-14
packaging: avoid race in snapd.postinst

mvo5:
We need to run dh_apparmor before dh_systemd_enable so that
the snapd.postinst is written in the right way.

The old code was doing things in the *wrong* order:
- it (re)starts snapd first
- it reloads the snap-confine apparmor profile *after* that
which of course means that on an upgrade there is a time window
in which snapd is available for the system-key check but
snap-confine has an old profile and will die because it tries
to do things that it cannot do with that old profile.

Together with the removal of "snapd.framework.target" this
race caused real world upgrades to stop snaps.

This bug can cause snap services to fail to start on upgrades. If snapd is
upgraded from a really old version of snapd that sill has snapd.framework.target
then the removal of snapd.framework.target will stop/restart all snap services.
However because of this race the service cannot start yet so they fail.

We need to put this in the security pocket.

This also adds a spread test to ensure the upgrade really works. 

When adding the spread test I noticed that the initial fix that just fixes the apparmor
rule loading is not sufficient, there is also a bug in snap-confine that needs a fixing.
This is the change included in snap-confine/mount-support.c - with that the spread
test works and the go-webserver is active again after the upgrade.

### DIFF
--- a/cmd/snap-confine/mount-support.c
+++ b/cmd/snap-confine/mount-support.c
@@ -581,7 +581,8 @@ void sc_populate_mount_ns(struct sc_apparmor *apparmor, int snap_update_ns_fd,
 					die("cannot locate the core or legacy core snap (current symlink missing?)");
 				}
 			}
-			die("cannot locate the base snap: %s", base_snap_name);
+			if (access(rootfs_dir, F_OK) != 0)
+			        die("cannot locate the base snap: %s", base_snap_name);
 		}
 		struct sc_mount_config normal_config = {
 			.rootfs_dir = rootfs_dir,
--- a/cmd/snap-confine/mount-support.c
+++ b/cmd/snap-confine/mount-support.c
@@ -581,6 +581,8 @@ void sc_populate_mount_ns(struct sc_apparmor *apparmor, int snap_update_ns_fd,
 					die("cannot locate the core or legacy core snap (current symlink missing?)");
 				}
 			}
+			// If after the special case handling above we are
+			// still not ok, die
 			if (access(rootfs_dir, F_OK) != 0)
 			        die("cannot locate the base snap: %s", base_snap_name);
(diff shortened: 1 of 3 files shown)


cmd/snap-confine/mount-support.c
```diff
@@ -581,7 +581,8 @@ void sc_populate_mount_ns(struct sc_apparmor *apparmor, int snap_update_ns_fd,
 					die("cannot locate the core or legacy core snap (current symlink missing?)");
 				}
 			}
-			die("cannot locate the base snap: %s", base_snap_name);
+			if (access(rootfs_dir, F_OK) != 0)
```
zyga: Can you explain this? I don't see how this is required.

mvo5: Sure, the code above may change rootfs_dir (the fallback code that switches to ubuntu-core) so we need to re-check here if things are still inaccessible .

jdstrand: Yes, what @zyga said. Note that your added conditional will always be true. Ie, with this change you end up with:
```
if (access(rootfs_dir, F_OK) != 0) {
   if (sc_streq(base_snap_name, "core")) {
        ... updates rootfs_dir ...
        if (access(rootfs_dir, F_OK) != 0) {
            die
        }
    }
    if (access(rootfs_dir, F_OK) != 0) { # your addition is always true
        die
    }
}
```

jdstrand:
> LGTM provided you remove the unneeded conditional.

Sorry, this PR is good to go as is (see my updated comment)

cmd/snap-confine/mount-support.c
zyga: Ah, I see this too now. Thank you!

PR #6509, opened 2019-02-14
overlord/snapstate: discard mount namespace when undoing 1st link snap

zyga:
When a snap is installed for the first time but the installation fails,
e.g. because the install hook fails or because a service fails to start,
the mount namespace would remain behind, possibly messing up subsequent
attempts to install the snap.

When the undo handler for "link-snap" detects that it runs for the 1st
revision is already removes snap cookies, this is the perfect place to
add the logic to discard the mount namespace.

Thanks to Samuele for pointing this out!

Fixes: https://bugs.launchpad.net/snapd/+bug/1815722

### DIFF
--- a/overlord/snapstate/handlers.go
+++ b/overlord/snapstate/handlers.go
@@ -1083,6 +1083,11 @@ func (m *SnapManager) undoLinkSnap(t *state.Task, _ *tomb.Tomb) error {
 
 	if len(snapst.Sequence) == 1 {
 		// XXX: shouldn't these two just log and carry on? this is an undo handler...
+		err = m.backend.DiscardSnapNamespace(snapsup.InstanceName())
+		if err != nil {
+			t.Errorf("cannot discard snap namespace %q, will retry in 3 mins: %s", snapsup.InstanceName(), err)
+			return &state.Retry{After: 3 * time.Minute}
+		}
 		if err := m.removeSnapCookie(st, snapsup.InstanceName()); err != nil {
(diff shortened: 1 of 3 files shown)


overlord/snapstate/handlers.go
```diff
@@ -1083,6 +1083,11 @@ func (m *SnapManager) undoLinkSnap(t *state.Task, _ *tomb.Tomb) error {
 
 	if len(snapst.Sequence) == 1 {
 		// XXX: shouldn't these two just log and carry on? this is an undo handler...
+		err = m.backend.DiscardSnapNamespace(snapsup.InstanceName())
+		if err != nil {
+			t.Errorf("cannot discard snap namespace %q, will retry in 3 mins: %s", snapsup.InstanceName(), err)
+			return &state.Retry{After: 3 * time.Minute}
```
stolowski: 3 minutes is a lot, is there a reason to wait that long?

Also two questions:
- shouldn't we bail out if it keeps failing? Unlink Retry in other places where we know a problem (e.g. a conflict) will eventually go away, in this case it might be a persistent issue (until rebooted)?
-  what happens if there is a reboot before we execute undo and the namespace is naturally gone? Will we hit an error here?

zyga: We also use 3 minutes in the natural case in discard-snap. I don't recall why we wait this long but I didn't want to invent a new time now.

- This doesn't fail in any way I know of
- If there's a reboot this gets cleaned up and snap-discard-ns just happily goes ahead

PR #6511, opened 2019-02-14
daemon/api: fix error case for disconnect conflict

stolowski:
The err variable in the "disconnect" handling is shadowed and in case of conflict error we end up with an empty Change in Hold state and no error is returned to the client. This PR fixes this.
Note, the important bit is below what's visible in the diff - the `ts, err = ifacestate.Disconnect(st, conn)` may return conflict error (but err is shadowed).


chipaca:
There's a separate shadowing detector, I looked into using it but we shadow things so much...

PR #6514, opened 2019-02-15
interfaces/apparmor: allow sending and receiving signals from ourselves

stolowski:
Allow signaling ourselves, to fix master.


interfaces/apparmor/template.go
```diff
@@ -715,6 +715,8 @@ profile snap-update-ns.###SNAP_INSTANCE_NAME### (attach_disconnected) {
 
   # Allow receiving signals from unconfined (eg, systemd)
   signal (receive) peer=unconfined,
+  # Allow sending and receiving signals from ourselves.
+  signal peer=snap-update-ns.###SNAP_INSTANCE_NAME###,
```
jdstrand: The rule is fine. You should use this equivalent rule though:
```
signal peer=@{profile_name},
```

stolowski: Thanks! Done.

PR #6516, opened 2019-02-15
interfaces/seccomp: generate global seccomp profile

zyga:
When initializing the seccomp backend, write the global profile to disk.
This is picked up by snap-confine and applied to all non-classic snaps.


interfaces/seccomp/backend.go
```diff
@@ -82,8 +83,57 @@ func seccompToBpfPath() string {
...
+	0x00, 0x06, 0x00, 0x00, 0x00, 0x05, 0x00, 0x01, 0x00, 0x06, 0x00, 0x00, 0x7f, 0xff, 0x00, 0x00,
+}
+
+func isBigEndian() bool {
+	switch runtime.GOARCH {
+	case "s390x":
+		return true
+	case "ppc":
+		return true
+		// TODO: perhaps more here?
```
jdstrand: What did you have in mind? (not a blocker)

zyga: A better way to detect big-endian vs little-endian architectures. This one feels like a hack.

PR #6519, opened 2019-02-19
squashfs: unset SOURCE_DATE_EPOCH in the TestBuildDate test

mvo5:
The SOURCE_DATE_EPOCH is used in reproducible builds and will
force mksquashfs to use a specific date. We need to unset it
for this specific test. It will be set on all debian/ubuntu
buildds.

This unbreaks debian and ubuntu/disco builds.


snap/squashfs/squashfs_test.go
```diff
@@ -590,6 +590,14 @@ func (s *SquashfsTestSuite) TestUnsquashfsStderrWriter(c *C) {
 }
 
 func (s *SquashfsTestSuite) TestBuildDate(c *C) {
+	// This env is used in reproducible builds and will force
+	// squashfs to use a specific date. We need to unset it
+	// for this specific test.
+	if oldEnv := os.Getenv("SOURCE_DATE_EPOCH"); oldEnv != "" {
```
zyga: Ah, nice catch mvo!

PR #6520, opened 2019-02-19
wrappers: Add an X-SnapInstanceName field to desktop files

robert-ancell:
Allow the snap that provides a .desktop file to be determined by putting the
snap name into the .desktop file (using the X-SnapInstanceName key).

This is useful for tools that scan the available .desktop files and want to get
additional snap information for that application. The particular motivation
for this change is gnome-control-center now exposes per-application
configuration and needs to determine if the .desktop file is from a Snap,
Flatpak (uses X-Flatpak key) or traditional packaging system.

### DIFF
--- a/wrappers/desktop.go
+++ b/wrappers/desktop.go
@@ -171,6 +171,11 @@ func sanitizeDesktopFile(s *snap.Info, desktopFile string, rawcontent []byte) []
 		newContent.Grow(len(bline) + 1)
 		newContent.Write(bline)
 		newContent.WriteByte('\n')
+
+		// insert snap name
+		if bytes.Equal(bline, []byte("[Desktop Entry]")) {
+			newContent.Write([]byte("X-SnapName=" + s.InstanceName() + "\n"))
+		}
 	}
 
 	return newContent.Bytes()
--- a/wrappers/desktop.go
+++ b/wrappers/desktop.go
@@ -174,7 +174,7 @@ func sanitizeDesktopFile(s *snap.Info, desktopFile string, rawcontent []byte) []
 
 		// insert snap name
 		if bytes.Equal(bline, []byte("[Desktop Entry]")) {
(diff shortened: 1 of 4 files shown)


wrappers/desktop.go
```diff
@@ -171,6 +171,11 @@ func sanitizeDesktopFile(s *snap.Info, desktopFile string, rawcontent []byte) []
 		newContent.Grow(len(bline) + 1)
 		newContent.Write(bline)
 		newContent.WriteByte('\n')
+
+		// insert snap name
+		if bytes.Equal(bline, []byte("[Desktop Entry]")) {
+			newContent.Write([]byte("X-SnapName=" + s.SnapName() + "\n"))
```
jhenstridge: This should probably be `InstanceName` rather than `SnapName`.  For the purpose of matching against the package ID xdg-desktop-portal determines from the AppArmor security label, I think the instance name is what we want.

robert-ancell:
I can't work out what is failing in the spread tests

chipaca:
@robert-ancell I'll pick this up. Thank you!

wrappers/desktop.go
```diff
@@ -171,6 +171,11 @@ func sanitizeDesktopFile(s *snap.Info, desktopFile string, rawcontent []byte) []
 		newContent.Grow(len(bline) + 1)
 		newContent.Write(bline)
 		newContent.WriteByte('\n')
+
+		// insert snap name
+		if bytes.Equal(bline, []byte("[Desktop Entry]")) {
+			newContent.Write([]byte("X-SnapName=" + s.InstanceName() + "\n"))
```
zyga: As long as people don't include `X-SnapName=` in their desktop files themselves.

robert-ancell: The desktop files are validated and this field is not allowed.

PR #6521, opened 2019-02-19
daemon: make ucrednetGet not loop

chipaca:
This is the second half of the belts-and-suspenders fix for
[lp:1813365], which was distro-patched.

The previous change to this already made us not be vulnerable; this
change makes it nicer, and makes it easier to not mess up this way
again (ever).

[lp:1813365]: https://bugs.launchpad.net/snapd/+bug/1813365

### DIFF
--- a/daemon/ucrednet.go
+++ b/daemon/ucrednet.go
@@ -23,8 +23,8 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"regexp"
 	"strconv"
-	"strings"
 	sys "syscall"
 )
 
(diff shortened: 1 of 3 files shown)


daemon/ucrednet.go
```diff
@@ -35,29 +35,23 @@ const (
 	ucrednetNobody    = uint32((1 << 32) - 1)
 )
 
+var raddrRegexp = regexp.MustCompile(`^pid=(\d+);uid=(\d+);socket=([^;]*);$`)
```
mvo5: :heart:

PR #6523, opened 2019-02-20
snap, wrappers: support StartTimeout

chipaca:
Before this change you could specify the StopTimeout of a snap's
daemons, but not their StartTimeout. This adds that support.

### DIFF
--- a/snap/info_snap_yaml.go
+++ b/snap/info_snap_yaml.go
@@ -78,6 +78,7 @@ type appYaml struct {
 	ReloadCommand   string          `yaml:"reload-command,omitempty"`
 	PostStopCommand string          `yaml:"post-stop-command,omitempty"`
 	StopTimeout     timeout.Timeout `yaml:"stop-timeout,omitempty"`
+	StartTimeout    timeout.Timeout `yaml:"start-timeout,omitempty"`
 	WatchdogTimeout timeout.Timeout `yaml:"watchdog-timeout,omitempty"`
 	Completer       string          `yaml:"completer,omitempty"`
 	RefreshMode     string          `yaml:"refresh-mode,omitempty"`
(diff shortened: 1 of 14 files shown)


snap/info_snap_yaml.go
```diff
@@ -78,6 +78,7 @@ type appYaml struct {
 	ReloadCommand   string          `yaml:"reload-command,omitempty"`
 	PostStopCommand string          `yaml:"post-stop-command,omitempty"`
 	StopTimeout     timeout.Timeout `yaml:"stop-timeout,omitempty"`
+	StartTimeout    timeout.Timeout `yaml:"start-timeout,omitempty"`
```
bboozzoo: A side note, we have validation for WatchdogTimeout (>0), if you feel like it, the PR could be extended to add validation for (Start|Stop)Timeout too.

tests/main/snap-service-start-timeout/task.yaml
```diff
@@ -0,0 +1,12 @@
+summary: Check that snaps can use start-timeout
+
+execute: |
+    dir="$TESTSLIB/snaps/test-snapd-service-start-timeout"
+
+    # with the 30s sleep, start-timeout stops the snap from working
+    ! snap try "$dir"
```
bboozzoo: I think this may break with the new SELinux profile. Perhaps we could limit the test to run on non-{Fedora,CentOS} systems?

tests/main/snap-service-start-timeout/task.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+    # drop the 'sleep 30'
+    sed -i -e '/@@@/d' "$dir/forking.sh"
```
bboozzoo: When another tests happens to use the same snap and ends up running after this one, the result may be a bit surprising.

tests/main/snap-service-start-timeout/task.yaml
bboozzoo: Discussed this offline and decided to go with a note in the test file for now.

PR #6525, opened 2019-02-20
interfaces/wayland: allow wayland server snaps function on classic too

gerboland:
Installing mir-kiosk on a classic system needs the same apparmor rules as if installed on core. This PR removes the branching on classic.

Fixes https://bugs.launchpad.net/snapd/+bug/1804869/

### DIFF
--- a/interfaces/builtin/wayland.go
+++ b/interfaces/builtin/wayland.go
@@ -73,6 +73,9 @@ network netlink raw,
 /run/udev/data/c13:[0-9]* r,
 /run/udev/data/+input:input[0-9]* r,
 /run/udev/data/+platform:* r,
+
+# MESA reads this dri config file
+/etc/drirc r,
 `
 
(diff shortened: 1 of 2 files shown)


gerboland:
@jdstrand Yep I noticed the /dev/drirc denials in the log, caused by mesa. I'm told most systems don't have that file, but tbh having read access to it won't hurt. I'll add to silence the denials.

Re: multiple wayland slots putting sockets in /run/user/something - yes it is something I thought of and needs some kind of fix. libwayland-server will create "wayland-0" by default, but if that exists, will try "wayland-1" and so on. So one could have both /run/user/1000/wayland-{0,1} sockets available. Clients will connect to"wayland-0", unless told otherwise. There's nothing in this interface to prevent client connected to server1:wayland actually connecting to server0's socket.

I believe snapd has means of doing that, I have to figure out how.

PR #6526, opened 2019-02-20
tests: run tests on opensuse leap 15.0 instead of 42.3

sergiocazzolato:
This is using the new image already created on google.


spread.yaml
```diff
@@ -84,6 +84,8 @@ backends:
                 workers: 4
```
mvo5: Should we remove  - opensuse-42.3-64: from above or may this come back eventually?

zyga: I would keep it until its EOL later this year.

zyga:
It would be good to run this with `--debug` and check what exactly failed here:
```
+ echo 'systemctl reports the system is in degraded mode'
systemctl reports the system is in degraded mode
+ systemctl --failed
  UNIT                           LOAD   ACTIVE SUB    DESCRIPTION          
â systemd-vconsole-setup.service loaded failed failed Setup Virtual Console
```

tests/main/interfaces-desktop-host-fonts/task.yaml
```diff
@@ -16,12 +16,7 @@ restore: |
     rm -f /usr/lib/fontconfig/cache/cache.txt
 
 execute: |
-    #shellcheck source=tests/lib/pkgdb.sh
-    . "$TESTSLIB"/pkgdb.sh
-    distro_install_package fontconfig
-
```
zyga: I'm curious about this part. There are changes with additions of fontconfig above and here removal of the installation of the fontconfig package. What is this about?

zyga: Ahh, I'm blind. Sorry :)

PR #6531, opened 2019-02-21
cmd/snap: fix error messages for snapshots commands if ID is not uint

stolowski:
Some snapshots-related commands expect set ID (a number). This is currently defined as uint64 in the definitions for go-flags, so when a string is provided instead, we end up with an ugly error message, e.g:
$ snap restore a
error: strconv.ParseUint: parsing "a": invalid syntax

The parsing in this particular case is unfortunately handled internally by go-flags based on the command definitions, so we cannot do anything about this in Execute(). Therefore I had to change the type of of snapshotID argument to string so we can do validation ourselves.


stolowski:
Seems the change affected snapshot-basic spread test, will investigate.

PR #6532, opened 2019-02-21
daemon: allow downloading snaps blobs via .../file

chipaca:
This change exposes a new endpoint, `/v2/snaps/<name>/file` that, if the
snap is installed, active, and not in try mode, allows downloading the
snap blob. As the data is the same as what could be re-created by
running `mksquashfs` on `/snap/<name>/current/`, there is no info leak.

### DIFF
--- a/daemon/api_blob.go
+++ b/daemon/api_blob.go
@@ -1,0 +1,67 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package daemon
+
+import (
+	"net/http"
+
+	"github.com/snapcore/snapd/overlord/auth"
+	"github.com/snapcore/snapd/overlord/snapstate"
+	"github.com/snapcore/snapd/overlord/state"
+	"github.com/snapcore/snapd/snap"
+)
+
+var snapBlobCmd = &Command{
+	Path:     "/v2/snaps/{name}/blob",
+	UserOK:   true,
+	PolkitOK: "io.snapcraft.snapd.manage",
+	GET:      getSnapBlob,
+}
+
+func getSnapBlob(c *Command, r *http.Request, user *auth.UserState) Response {
+	vars := muxVars(r)
+	name := vars["name"]
+
+	st := c.d.overlord.State()
+	st.Lock()
+	var snapst snapstate.SnapState
+	var info *snap.Info
+	err := snapstate.Get(st, name, &snapst)
+	if err == nil {
+		info, err = snapst.CurrentInfo()
+	}
+	st.Unlock()
+	switch err {
+	case nil:
+		// ok
+	case state.ErrNoState:
+		return SnapNotFound(name, err)
+	default:
+		return InternalError("cannot download blob for snap %q: %v", name, err)
+	}
+	if !snapst.Active {
+		return BadRequest("cannot download blob of inactive snap %q", name)
+	}
+	if snapst.TryMode {
+		return BadRequest("cannot download blob for try-mode snap %q", name)
+	}
+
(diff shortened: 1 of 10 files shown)


daemon/api_blob.go
```diff
@@ -0,0 +1,67 @@
...
+	st.Unlock()
+	switch err {
+	case nil:
+		// ok
+	case state.ErrNoState:
+		return SnapNotFound(name, err)
+	default:
+		return InternalError("cannot download blob for snap %q: %v", name, err)
+	}
+	if !snapst.Active {
```
zyga: Why cannot we get a blob of an inactive snap? Is there a technical reason or a design reason behind this?

chipaca: We don't want to expose more than the user could get by copying things that are already mounted

zyga: But inactive snaps _are_ mounted. They are not exposed as far as systemd units or launchers are concerned but they are definitely not invisible in SNAP_MOUNT_DIR.

chipaca: Yes, that bit is more about not exposing things that aren't needed that we're not sure will remain the way they are in the future.
That is: inactive snaps are mounted, but tomorrow we might stop doing that (we have people asking for that, with good reasons for it). And nothing inherently requires that we have them mounted. And the intended user of the api doesn't need them. So, safest to leave them out.

daemon/api_blob.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/snapcore/snapd/overlord/auth"
+	"github.com/snapcore/snapd/overlord/snapstate"
+	"github.com/snapcore/snapd/overlord/state"
+	"github.com/snapcore/snapd/snap"
+)
+
+var snapBlobCmd = &Command{
+	Path:     "/v2/snaps/{name}/blob",
+	UserOK:   true,
```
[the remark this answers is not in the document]
chipaca: The reasoning is that a user can already access the data because it's mounted, so this isn't an info leak.
The 0600 is because there could be sensitive data there, in a snap that's disabled (think a vendor core device where a user needs to pay to enable a premium snap, for example).

daemon/api_blob.go
zyga: I think we should treat "snap download" as a cache. If you have it on disk, just get that. Putting arbitrary restrictions won't stop anyone from "sudo cp" things out of /var/.

daemon/api_blob.go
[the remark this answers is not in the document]
pedronis: should we call it blob?   other options:
* file
* snapfile
* content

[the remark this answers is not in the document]
bboozzoo: Download help message says:
```
The download command downloads the given snap and its supporting assertions
```
Then install help message:
```
Use --name to set the instance name when installing from snap file.
```
Throughout the code there are some references to 'blob' throughout the code, but mostly in tests.

`content` sounds fine too, though I have a feeling it might be understood as a way to get uncompressed snap content.

daemon/api_blob_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+type blobSuite struct{}
+
+func (s *blobSuite) SetUpTest(c *check.C) {
+	dirs.SetRootDir(c.MkDir())
+}
+
+func (s *blobSuite) TestGetBlob(c *check.C) {
+	defer daemon.MockMuxVars(func(*http.Request) map[string]string {
+
```
zyga: Nitpick, stray newline?

daemon/api_blob_test.go
```diff
@@ -0,0 +1,101 @@
...
+	c.Check(daemon.SnapBlobCmd.Path, check.Equals, "/v2/snaps/{name}/blob")
+
+	o := overlord.Mock()
+	daemon.NewWithOverlord(o)
+	st := o.State()
+
+	type scenario struct {
+		status                   int
+		exists, active, try, wat bool
```
zyga: What does `wat` stand for?

bboozzoo: Wat?

daemon/api_blob.go
```diff
@@ -0,0 +1,67 @@
...
+
+	st := c.d.overlord.State()
+	st.Lock()
+	var snapst snapstate.SnapState
+	var info *snap.Info
+	err := snapstate.Get(st, name, &snapst)
+	if err == nil {
+		info, err = snapst.CurrentInfo()
+	}
+	st.Unlock()
```
pedronis: can we really unlock already here, or there is a chance that the file will be deleted by the time we try to create a FileResponse for it?

daemon/api_blob.go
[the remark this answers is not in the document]
pedronis: My preference ATM would be "file"   (we are under /snaps/ so "snapfile" seems a bit redundant)

daemon/api_blob.go
chipaca: good point; I'll change it to a defer

PR #6539, opened 2019-02-26
cmd, daemon: split out the common bits of mapLocal and mapRemote

chipaca:
daemon's mapLocal and mapRemote helpers had a lot of common code, that
if split out could also be reused in cmd/snap/cmd_info. This change
splits that common code out, and prepares things for a nice tweak and
simplifying refactor to info's handling of local files.


cmd/snapinfo.go
```diff
@@ -59,6 +110,19 @@ func ClientAppInfoNotes(app *client.AppInfo) string {
 	return strings.Join(notes, ",")
 }
 
+type BySnapApp []*snap.AppInfo
```
zyga:
```suggestion
// BySnapApp sorts apps by (snap name, app name)
type BySnapApp []*snap.AppInfo
```

chipaca: Will do.

daemon/snap.go
```diff
@@ -120,10 +118,18 @@ func allLocalSnapInfos(st *state.State, all bool, wanted map[string]bool) ([]abo
...
 		var info *snap.Info
-		var publisher *snap.StoreAccount
 		var err error
 		if all {
+			var publisher snap.StoreAccount
+			var gotPublisher bool
 			for _, seq := range snapst.Sequence {
+				if !gotPublisher {
+					gotPublisher = true
+					publisher, err = publisherAccount(st, seq.SnapID)
```
bboozzoo: Is it posible to have the publisher change across revisions?

chipaca: hmmm. That's the same question as “can a snap ID change between revisions”. And the answer is yes: you can have a local revision (snap ID would be `""`, publisher unknown) and then a store revision with a given snap ID and publisher.

chipaca: I've made it grab the publisher for every revision, and added a test for the simplest way this can occur (the local-vs-from-store I described). Note that it is also in theory possible for it to happen via some stranger mechanism, but it hasn't ever happened yet :-)

The simplest way of seeing this in action is having a from-the-store snap, installing something from a snap with the same name using `--dangerous`, and then doing `snap list --all`: you should see, e.g.,
```
$ snap list icdiff --all
Name    Version  Rev  Tracking  Publisher  Notes
icdiff  1.9.1-1  1    stable    chipaca    disabled
icdiff  1.9.4-1  x1   stable    -          -
```
if you see any publisher there for the local revision, it's wrong :-)

daemon/api_test.go
```diff
@@ -52,6 +52,7 @@ import (
 	"github.com/snapcore/snapd/asserts/assertstest"
 	"github.com/snapcore/snapd/asserts/sysdb"
 	"github.com/snapcore/snapd/client"
+	"github.com/snapcore/snapd/cmd"
```
mvo5: This is nothing new so not relevant for this PR - but it does feel slightly odd to me that `apinfo.go` (or now `snapinfo.go` lifes under `github.com/snapcore/snapd/cmd`). Seems like something we may want to look at at some point.

daemon/snap.go
```diff
@@ -268,74 +254,34 @@ func appInfosFor(st *state.State, names []string, opts appInfoOptions) ([]*snap.
...
-	}
-
-	snapapps := make([]*snap.AppInfo, 0, len(localSnap.Apps))
-	for _, app := range localSnap.Apps {
-		snapapps = append(snapapps, app)
-	}
-	sort.Sort(bySnapApp(snapapps))
-
-	apps, err := cmd.ClientAppInfosFromSnapAppInfos(snapapps)
+	result, err := cmd.ClientSnapFromSnapInfo(localSnap)
```
mvo5: This is very nice!

daemon/snap.go
```diff
@@ -348,43 +294,15 @@ func mapLocal(about aboutSnap) *client.Snap {
...
 func mapRemote(remoteSnap *snap.Info) *client.Snap {
-	status := "available"
-	if remoteSnap.MustBuy {
-		status = "priced"
-	}
-
-	confinement := remoteSnap.Confinement
-	if confinement == "" {
-		confinement = snap.StrictConfinement
+	result, err := cmd.ClientSnapFromSnapInfo(remoteSnap)
```
mvo5: This is very nice too :)

daemon/api_test.go
chipaca: It's there until it grows up a little more at which point we might know better where to put it. It was the result of a long conversation with @pedronis about exactly where to put it...

PR #6541, opened 2019-02-26
tests: change how xdg-desktop-portal is prepared and restored for desktop-portal-* tests

sergiocazzolato:
This is to avoid error caused for dir not umounted properly

error: https://paste.ubuntu.com/p/VFGvkXN6Wk/


tests/lib/desktop-portal.sh
```diff
@@ -51,10 +51,10 @@ teardown_portals() {
     distro_purge_package xdg-desktop-portal
     distro_auto_remove_packages
 
-    if [ -d "${USER_RUNTIME_DIR}" ]; then 
-        umount --lazy "${USER_RUNTIME_DIR}/doc" || :
-        rm -rf "${USER_RUNTIME_DIR:?}"/* "${USER_RUNTIME_DIR:?}"/.[!.]*
+    if mount | grep "${USER_RUNTIME_DIR}/doc"; then
```
bboozzoo: I'm thinking we should probabaly kill `xdg-desktop-portal` and `xdg-document-portal` too.

When debugging this yesterday, I tracked it down to this code in xdg-desktop-porta: https://github.com/flatpak/xdg-desktop-portal/blob/master/document-portal/document-portal-fuse.c#L718-L735 What I believe happens, is that python does `open(path, 'w')`, which will attempt to truncate the file if it exists. Then in fuse handlers inside document-portal, the code path for when the inode exists and the caller requested O_TRUNC, returns -ENOSYS, resulting in `OSError: [Errno 38] Function not implemented` on the Python side.

So, IMO the proper fix should be ensure that xdg-document-portal (and xdg-desktop-portal) is not running (kill when necessary) and unmount "$USER_RUNTIME_DIR/doc" if it has not been cleaned up.

bboozzoo:
Shellcheck failed:
```
In tests/lib/desktop-portal.sh line 48:
    if ps -ef | grep -E '^test .*/fakeportalui/.*'; then
       ^-- SC2009: Consider using pgrep instead of grepping ps output.
```

sergiocazzolato:
@bboozzoo it is already fixed.

zyga:
@bboozzoo can you please discuss with @sergiocazzolato tomorrow about the details? @sergiocazzolato let's wait until that happens with merging this please.

tests/lib/desktop-portal.sh
sergiocazzolato: @bboozzoo, thanks for reviewing, I checked that but the error still happened. I added a check for any process related to xdg-desktop and nothing was running but the error was reproduced. 

I also verified with @jhenstridge that there are some issues with the setup and restore and that's why initially the package xdg-desktop-portal was being installed and remove on each test. 

The solution I found based on the issue is to pre-create the directory as part of the test.

sergiocazzolato:
@zyga, We already agreed with @bboozzoo on the current solution. We need to see the tests pass after merge with master.

sergiocazzolato:
The error is not being reproduced anymore, possibly the issue was fixed on xdg portal package and installed in the test machines.

PR #6542, opened 2019-02-26
cmd/snap: fix `snap services` completion

chipaca:
`snap services` was not filtering the things it suggested for
completion. This fixes that, and adds a test.

### DIFF
--- a/cmd/snap/cmd_services_test.go
+++ b/cmd/snap/cmd_services_test.go
@@ -23,6 +23,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"sort"
+	"strings"
 	"time"
 
 	"gopkg.in/check.v1"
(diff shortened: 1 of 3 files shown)


cmd/snap/complete.go
```diff
@@ -439,17 +439,25 @@ func (s serviceName) Complete(match string) []flags.Completion {
...
+		if !strings.HasPrefix(name, match) {
+			continue
+		}
+		ret = append(ret, flags.Completion{Item: name})
+		if len(match) <= len(app.Snap) {
+			snaps[app.Snap]++
+		}
+	}
+	for snap, n := range snaps {
+		if n > 1 {
```
bboozzoo: Isn't this check unnecessary? It seems that the snap will only turn up in `snaps` if there is a partial match on the snap name.

chipaca: I'm having it complete straight to the service when there's only one of them

bboozzoo: Ahh ok, that's fine then.

cmd/snap/cmd_services_test.go
```diff
@@ -222,6 +224,51 @@ foo.zed  enabled  active    -
 	c.Check(n, check.Equals, 1)
 }
 
+func (s *appOpSuite) TestServiceCompletion(c *check.C) {
```
stolowski: Nice

PR #6546, opened 2019-02-27
packaging: build snapctl as a static binary

bboozzoo:
The snapctl binary can be called from inside the mount ns, and thus should be
built statically to ensure it operates regardless of the base that is used.

### DIFF
--- a/packaging/snapd.mk
+++ b/packaging/snapd.mk
@@ -60,7 +60,7 @@ all: $(go_binaries)
 snap snap-seccomp:
 	go build -buildmode=pie $(import_path)/cmd/$@
 
-# Those tree need to be built as static binaries. They run on the inside of a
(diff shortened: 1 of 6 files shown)


packaging/snapd.mk
```diff
@@ -57,13 +57,13 @@ go_binaries = snap snapctl snap-seccomp snap-update-ns snap-exec snapd
 .PHONY: all
 all: $(go_binaries) 
 
-snap snapctl snap-seccomp:
+snap snap-seccomp:
 	go build -buildmode=pie $(import_path)/cmd/$@
 
-# Those two need to be built as static binaries. They run on the inside of a
+# Those tree need to be built as static binaries. They run on the inside of a
```
zyga:
```suggestion
# Those three need to be built as static binaries. They run on the inside of a
```

PR #6561, opened 2019-03-04
cmd/snap-confine: chown private /tmp to root.root

zyga:
When snap-confine creates a private /tmp directory for a given snap it
first creates a temporary directory in /tmp/ named after the snap, along
with a random name. Inside that directory it creates a /tmp directory
with permissions appropriate for a future /tmp, namely 1777.

Up until recently the that directory was owned by the user who first
invoked snap-confine. Since the directory is reused by all the users on
the system this logic makes no sense.

This patch changes the related logic so that the private /tmp directory
is owned by root, just like the real one.

### DIFF
--- a/cmd/snap-confine/mount-support.c
+++ b/cmd/snap-confine/mount-support.c
@@ -62,8 +62,6 @@
 // TODO: fold this into bootstrap
 static void setup_private_mount(const char *snap_name)
 {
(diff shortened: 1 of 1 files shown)


zyga:
NOTE: This is stacked on top of https://github.com/snapcore/snapd/pull/6560

PR #6562, opened 2019-03-05
timings: base API for recording timings in state

stolowski:
API for measuring execution times and storing timings in state. 
- allows creating a Timings tree for given activity (e.g. task or execution of ensure).
- within the tree we can nest further measurements.
- for each measurement we record its duration, label and summary.
- after finishing the measurements, entire tree can be saved in state.
- on Save the tree is flattenned: all the nested measurements are flattened as a list, for every node its nest level is also stored.
- up to MaxTimings are kept in the state. Pruning is done on every Save.

Thanks to Maciej and Zygmunt for initial work on this area.


timings/timings.go
```diff
@@ -0,0 +1,106 @@
...
+
+// StartNested creates a new nested Measure and initiates performance measurement.
+// Nested measure needs to be stopped by calling Stop on it.
+func (m *Measure) StartNested(label, summary string) *Measure {
+	meas := &Measure{
+		label:   label,
+		summary: summary,
+		start:   time.Now(),
+	}
+	m.nested = append(m.nested, meas)
```
bboozzoo: We'll probably need to think about locking here in case this is used from multiple goroutines.

Adding the 'sample' here, means that forgetting to call Stop(), or processing the samples before stop is called will give unexpected timing results.

timings/state.go
```diff
@@ -0,0 +1,120 @@
...
+	return end.Sub(start)
+}
+
+// flatten flattens nested measurements into a single list within rootTimingJson.NestedTImings.
+func (t *Timings) flatten() interface{} {
+	data := &rootTimingJson{
+		timingJson: timingJson{
+			Level:    0,
+			Label:    t.m.label,
+			Duration: timeDuration(t.m.stop, t.m.start),
```
bboozzoo: Shouldn't the parameter order be reversed? WIth the default implementation of `timeDuration`, this will effectively be  `start.Sub(stop)`.

timings/timings.go
```diff
@@ -0,0 +1,106 @@
...
+//   timing := timings.New(timings.Ensure)
+//   measure := timing.Start("computation", "...")
+//   ....
+//   nested := measure.StartNested("sub-computation", "...")
+//   ....
+//   nested.Stop()
+//   measure.Stop()
+type Timings struct {
+	subject Subject
+	meta    map[string]string
```
bboozzoo: `map[string]interface{}` will probably be more versatile

timings/timings.go
pedronis: I'm doubtful for now that it makes sense to use this cross goroutines, so I think it doesn't need locks until proven.

timings/timings.go
pedronis: again unclear we need that, seems better to start simple

timings/timings.go
stolowski: Yes, the assumption discussed with Samuele was we will be using it while in same goroutine, so "no locking" is by design at the moment.

timings/timings.go
stolowski: We haven't discussed that, but the idea here is to pass some simple "attributes" such as task id, and they will be displayed by the client. This will be rather rigid, therefore I'm proposing string -> string here.

timings/state.go
stolowski: Thanks for spotting! Fixed.

timings/timings.go
```diff
@@ -0,0 +1,106 @@
...
+
+package timings
+
+import (
+	"fmt"
+	"time"
+)
+
+// Subject is a type of the activity being measured; available subjects are defined below.
+type Subject string
```
mvo5: I'm slightly wary of `Subject` - how will that be used in the future (e.g. will it grow more categories?). I guess I mostly wodner what value does it add to the data. As it is now (with just the two subjects) it feels to me like not much but maybe I just need some examples to see the value :)

mvo5:
The shape of this looks fine, thank you! One tiny question about "Subject" inline.

timings/timings.go
```diff
@@ -0,0 +1,106 @@
...
+		label:   label,
+		summary: summary,
+		start:   time.Now(),
+	}
+	return t.m
+}
+
+// StartNested creates a new nested Measure and initiates performance measurement.
+// Nested measure needs to be stopped by calling Stop on it.
+func (m *Measure) StartNested(label, summary string) *Measure {
```
pedronis: wondering if Nested in the name adds clarity  vs  the win of having the same signature/possibly being able to use an interface for both Timings.Start and Measure.StartNested

timings/timings.go
```diff
@@ -0,0 +1,106 @@
...
+		start:   time.Now(),
+	}
+	m.nested = append(m.nested, meas)
+	return meas
+}
+
+// Stops the measurement.
+func (m *Measure) Stop() {
+	if !m.stop.IsZero() {
+		panic(fmt.Sprintf("measure %q already stopped", m.label))
```
pedronis: mmh, a panic is bad, keeping Timings around to store an error on it seems a bit wasteful as well.  Wondering if we should just ignore this under the assumption that is easy to eye ball that Start Stop pair,  on the other end it might happen that we call Start twice by mistake because Timings need to be passed around

timings/timings.go
bboozzoo: Sounds fair, let's start simple then.

timings/timings.go
```diff
@@ -0,0 +1,106 @@
...
+func New(subject Subject, metaInfo map[string]string) *Timings {
+	return &Timings{
+		subject: subject,
+		meta:    metaInfo,
+	}
+}
+
+// Starts creates and returns root Measure and initiates performance measurement.
+// Measurement needs to be stopped by calling Stop on it.
+func (t *Timings) Start(label, summary string) *Measure {
```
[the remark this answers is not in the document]
pedronis: I'm also wondering what these (label,summary) will be for a task typically, just copy of task kind and summary?

timings/timings.go
stolowski: @pedronis yeah, I'm not sure either. FWTW that's the "what is it about" aspect we discussed, and that it describes entire tree.

timings/timings.go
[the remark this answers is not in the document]
stolowski: I wouldn't expect these to match task label/summary 1:1 - only in (rare) cases where a measurement encompasses entire activity of the task handler. In some cases I'd expect more than 1 measure for same task if it has a few activities, then summaries and/or labels would be more specific. WDYT?

timings/timings.go
pedronis: tags should be enough for that

timings/timings.go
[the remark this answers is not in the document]
pedronis: My expectation was more that one task goroutine <-> 1 Timings but we can see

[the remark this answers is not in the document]
pedronis: the problem also with having many Timings per tasks is that our naive pruning could leave with a partial subset of timings per one task run, that is confusing

timings/timings.go
```diff
@@ -0,0 +1,106 @@
...
+//   nested := measure.StartNested("sub-computation", "...")
+//   ....
+//   nested.Stop()
+//   measure.Stop()
+type Timings struct {
+	subject Subject
+	meta    map[string]string
+
+	// root measurement
+	m *Measure
```
pedronis: I'm coming to the conclusion that it would be easier to make things more symmetric by making this a list too!

timings/state.go
```diff
@@ -0,0 +1,120 @@
...
+
+	timings = append(timings, &entryJSON)
+	st.Set("timings", timings)
+	return nil
+}
+
+// Purge removes excess timings from the "timings" list in the state (starting from the oldest),
+// ensuring that up to maxTimings is kept.
+func Purge(st *state.State, maxTimings int) error {
+	st.Lock()
```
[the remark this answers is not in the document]
stolowski: Good point, removed.

timings/state.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/snapcore/snapd/overlord/state"
+)
+
+// timingJson and rootTimingJson aid in marshalling of flattened timings into state.
+type timingJson struct {
```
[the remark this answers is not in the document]
stolowski: I'd say having a separate one improves readability and intent is more clear, plus it helps keep fileds private.

timings/timings.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// Measure represents a single measure with optional nested measurements.
+type Measure struct {
```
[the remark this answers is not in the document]
stolowski: Renamed, although I've Timings and Timing now, so probably naming needs to change still.

timings/timings.go
```diff
@@ -0,0 +1,106 @@
...
+	label, summary string
+	start, stop    time.Time
+	nested         []*Measure
+}
+
+// New creates a Timings object.
+func New(subject Subject, metaInfo map[string]string) *Timings {
```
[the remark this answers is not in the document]
stolowski: NIce idea, done.

timings/timings.go
```diff
@@ -0,0 +1,106 @@
...
+		subject: subject,
+		meta:    metaInfo,
+	}
+}
+
+// Starts creates and returns root Measure and initiates performance measurement.
+// Measurement needs to be stopped by calling Stop on it.
+func (t *Timings) Start(label, summary string) *Measure {
+	if t.m != nil {
+		panic(fmt.Sprintf("timing %q already started", label))
```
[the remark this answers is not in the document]
stolowski: Per later comment and discussions, got rid of that, Start is no problem anymore as it always creates new Timing, now also for the root Timings object. For Stop it's ignored.

timings/timings.go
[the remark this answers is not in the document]
stolowski: I'm not quite sure about the pruning concern, the way it's implemented right now is: we keep a list of Timings trees (flattened) and always prune entire trees. So what remains after pruning are complete trees for some executions.

timings/timings.go
stolowski: Removed, we have Start method now with same siganture for Timings and Timing.

timings/timings.go
stolowski: Removed, see earlier comment.

timings/timings.go
stolowski: Done. Note it has a consequence, you can now create Timings object and have sibling timings already on root level, so a duration for the entire tree needs to calculated, which I'm now doing on Save, total duration is calculated from first start time and most recent stop time of any Timing in the tree.

timings/state.go
```diff
@@ -0,0 +1,104 @@
...
+
+	serialized, err := json.Marshal(t.flatten())
+	if err != nil {
+		return fmt.Errorf("internal error: could not marshal value: %v", err)
+	}
+	entryJSON := json.RawMessage(serialized)
+
+	stateTimings = append(stateTimings, &entryJSON)
+	if len(stateTimings) > MaxTimings {
+		stateTimings = stateTimings[MaxTimings:]
```
pedronis: this should be  stateTimings[len(state)-MaxTimings:]  I think,  missing test?

timings/timings.go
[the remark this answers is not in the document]
pedronis: I'm personally fine with the name, they say what is in the thin, also users should not really deal often with single Timing itself.  If people find then confusing, we could rename Timing to TimingMeasurement.

timings/state.go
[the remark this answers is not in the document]
pedronis: we could make Timing an interface to keep the encapsulation (and keep an internal timing)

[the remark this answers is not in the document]
pedronis: it might be unfounded, my unmeasured worry is the extra copying going on in Save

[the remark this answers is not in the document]
stolowski: > we could make Timing an interface to keep the encapsulation (and keep an internal timing)

Hmm, I'm not following the idea you have in mind here, note that `Stop()` is only present on the Timing, the only common method on Timings and Timing is `Start`. Could you elaborate a bit or draft a quick/rough example of what do you mean by this? Thanks

timings/timings.go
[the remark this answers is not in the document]
stolowski: Clarified in a hangout.

timings/state.go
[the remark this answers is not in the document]
stolowski: Clarified in the hangout, most likely will explore this in a followup.

timings/state.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	var maxStopTime time.Time
+	if len(t.timings) > 0 {
+		flattenRecursive(data, t.timings, 0, &maxStopTime)
+		data.TotalDuration = timeDuration(t.timings[0].start, maxStopTime)
+	}
+	return data
+}
+
+func flattenRecursive(data *rootTimingJson, measures []*Timing, nestLevel int, maxStopTime *time.Time) {
```
pedronis: s/measures/timings/

timings/state.go
```diff
@@ -0,0 +1,104 @@
...
+	for _, m := range measures {
```
pedronis: s/m/tm/  or s/m/tmeas/  or s/m/tim/

timings/timings.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// Timings represents a tree of Timing measurements for a single execution of measured activity.
+// A Timings tree object should be created at the beginning of the activity,
+// followed by starting at least one Timing measurement, and then saved at the end of the activity.
+//
+// Calling Start on the Timings objects creates a Timing and starts new
+// performance measurement. Measurement needs to be finished by calling Stop
+// function on the Timing object.
+// Nested measures may be collected by calling Start on Timing objects. Similar
```
pedronis: s/measures/measurements/

timings/timings.go
```diff
@@ -0,0 +1,98 @@
...
+// to the above, nested measurements need to be finished by calling Stop on them.
+//
+// Typical usagage:
+//   timingsTree := timings.New(timings.Ensure)
+//   timing := timing.Start("computation", "...")
+//   ....
+//   nestedTiming := measure.Start("sub-computation", "...")
```
pedronis: s/measure/timing/ ?

timings/timings.go
```diff
@@ -0,0 +1,98 @@
...
+// followed by starting at least one Timing measurement, and then saved at the end of the activity.
+//
+// Calling Start on the Timings objects creates a Timing and starts new
+// performance measurement. Measurement needs to be finished by calling Stop
+// function on the Timing object.
+// Nested measures may be collected by calling Start on Timing objects. Similar
+// to the above, nested measurements need to be finished by calling Stop on them.
+//
+// Typical usagage:
+//   timingsTree := timings.New(timings.Ensure)
```
pedronis: New(nil) ?  or New(map[string]string{"tag": "tag-value"}) ?

timings/timings.go
```diff
@@ -0,0 +1,98 @@
...
+// Starts creates a Timing and initiates performance measurement.
+// Measurement needs to be stopped by calling Stop on it.
+func (t *Timings) Start(label, summary string) *Timing {
+	meas := start(label, summary)
+	t.timings = append(t.timings, meas)
+	return meas
+}
+
+// Start creates a new nested Timing and initiates performance measurement.
+// Nested measure needs to be stopped by calling Stop on it.
```
pedronis: s/measure/measurements/

timings/timings.go
```diff
@@ -0,0 +1,98 @@
...
+
+// New creates a Timings object.
+func New(tags map[string]string) *Timings {
+	return &Timings{
+		tags: tags,
+	}
+}
+
+func start(label, summary string) *Timing {
+	meas := &Timing{
```
pedronis: s/meas/tmeas/  and also everywhere else in the file ?

timings/timings.go
```diff
@@ -0,0 +1,98 @@
...
+// A Timings tree object should be created at the beginning of the activity,
+// followed by starting at least one Timing measurement, and then saved at the end of the activity.
+//
+// Calling Start on the Timings objects creates a Timing and starts new
+// performance measurement. Measurement needs to be finished by calling Stop
+// function on the Timing object.
+// Nested measurements may be collected by calling Start on Timing objects. Similar
+// to the above, nested measurements need to be finished by calling Stop on them.
+//
+// Typical usagage:
```
mvo5: Type in `usage`

timings/timings.go
```diff
@@ -0,0 +1,98 @@
...
+//   timingsTree := timings.New(map[string]string{"tag": "tag-value"})
+//   timing := timing.Start("computation", "...")
+//   ....
+//   nestedTiming := timing.Start("sub-computation", "...")
+//   ....
+//   nestedTiming.Stop()
+//   timing.Stop()
+//   timingsTree.Save()
+type Timings struct {
```
mvo5: Sorry - I know this is bike-shed land again but I find "Timing**s**" and "Timing" a bit too subtle. Maybe going to the suggestion of "Timings" and TimingMeasure" or "TimingsCollection" and "Timing"  (open for more ideas of course).

timings/timings_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+	s.st = state.New(nil)
+	s.duration = 0
+}
+
+func (s *timingsSuite) TearDownTest(c *C) {
+	s.BaseTest.TearDownTest(c)
+}
+
+func (s *timingsSuite) mockDuration(c *C) {
```
mvo5: Nice mocks (this and the next one) for the testing!

timings/timings.go
pedronis: I'm fine with a little bit of bikeshedding on Monday, but Timings is the one that will be referred to more often so should have the shorter or about same length name,  so not a fan of TimingsCollection

pedronis: Or maybe that's not true.  Maybe best is to land as is, and decide on first usage what are the best names.

pedronis: Spent a bit of time thinking further on this, we have actually two issues:

* a Timing is a tree of timing intervals too, that's lost with in the "Timing" name
* our Start methods don't start the object they are on, so Start/Stop is not a pair

I think:

* Timings is fine, it's bunch of timings
* s/Timing/Span/  using the 2nd def of span "the length of time for which something lasts."
* s/Start()/StartSpan() *Span/

timings/state.go
```diff
@@ -0,0 +1,104 @@
...
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/snapcore/snapd/overlord/state"
+)
+
+// timingJson and rootTimingJson aid in marshalling of flattened timings into state.
+type timingJson struct {
+	Level    int           `json:"level"`
```
pedronis: I would omitempty this one too

timings/state.go
```diff
@@ -0,0 +1,104 @@
...
+	Label    string        `json:"label,omitempty"`
+	Summary  string        `json:"summary,omitempty"`
+	Duration time.Duration `json:"duration"`
+}
+
+type rootTimingJson struct {
+	Tags          map[string]string `json:"tags,omitempty"`
+	NestedTimings []*timingJson     `json:"timings,omitempty"`
+	// the duration between start time of the first timing and the most recent stop of any timing within the tree
+	TotalDuration time.Duration `json:"total-duration"`
```
pedronis: I think at the top level  we should capture a start time and maxStopTime instead of just a duration, tasks keep that themselves but that is not be true for ensure subfunctions. Some minimal timeline information is though likely  useful.

timings/state.go
```diff
@@ -0,0 +1,104 @@
...
+// flatten flattens nested measurements into a single list within rootTimingJson.NestedTimings
+// and calculates total duration.
+func (t *Timings) flatten() interface{} {
+	data := &rootTimingJson{
+		Tags: t.tags,
+	}
+	var maxStopTime time.Time
+	if len(t.timings) > 0 {
+		flattenRecursive(data, t.timings, 0, &maxStopTime)
+		data.TotalDuration = timeDuration(t.timings[0].start, maxStopTime)
```
pedronis: see comment about keeping start/stop time info for the top level

timings/timings.go
stolowski: Thanks for thinking that through, sounds very good to me, done.

stolowski:
Did the renaming and addressed other comments, I hope I didn't miss anything.

timings/timings.go
```diff
@@ -0,0 +1,98 @@
...
+// followed by starting at least one Timing measurement, and then saved at the end of the activity.
+//
+// Calling Start on the Timings objects creates a Timing and starts new
+// performance measurement. Measurement needs to be finished by calling Stop
+// function on the Timing object.
+// Nested measurements may be collected by calling Start on Timing objects. Similar
+// to the above, nested measurements need to be finished by calling Stop on them.
+//
+// Typical usagage:
+//   timingsTree := timings.New(map[string]string{"tag": "tag-value"})
```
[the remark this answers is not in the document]
stolowski: Thanks for the suggestions, done. I think this doc comment will need updating again after the first PR using this API lands, as it will grow some helpers (such as a helper to create Timings from a task with proper tags defined automatically).

PR #6564, opened 2019-03-05
cmd/snap, tests: refactor info to unify handling of 'direct' snaps

chipaca:
Before this change, info's handling of 'direct' snaps (i.e. `snap info
foo.snap` and `snap info ./foo`) used a separate codepath from the
handling of locally-installed snaps and remote snaps. This led to the
output for direct snaps to diverge more than necessary from the output
of snaps that snapd knew about, as well as to some code duplication.

This change fixes that: before,
```
~$ snap info basic-desktop_1.0_all.snap
path:    "basic-desktop_1.0_all.snap"
name:    basic-desktop
summary: ""
version: 1.0 -
```
and now,
```
~$ snap info basic-desktop_1.0_all.snap
path:       "basic-desktop_1.0_all.snap"
name:       basic-desktop
summary:    ""
version:    1.0 -
build-date: today at 11:24 GMT
license:    GPL-3.0
description: |
  A basic snap with desktop apps
commands:
  - basic-desktop.echo
  - basic-desktop.echoecho
```


tests/main/snap-info/check.py
```diff
@@ -57,20 +57,26 @@ def verRelRevNotesRx(s):
 
 res = list(yaml.load_all(sys.stdin))
 
-equals("number of entries", len(res), 7)
+# equals("number of entries", len(res), 7)
```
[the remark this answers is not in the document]
chipaca: I forgot to un-comment the check :-)

cmd/snap/cmd_info.go
```diff
@@ -119,52 +121,22 @@ func maybePrintBase(w io.Writer, base string, verbose bool) {
...
-		notes = NotesFromInfo(info)
+		return nil, err
 	}
-	fmt.Fprintf(w, "version:\t%s %s\n", info.Version, notes)
-	maybePrintType(w, string(info.Type))
-	maybePrintBase(w, info.Base, verbose)
-	if sha3_384 != "" {
-		fmt.Fprintf(w, "sha3-384:\t%s\n", sha3_384)
+
+	direct, err := cmd.ClientSnapFromSnapInfo(info)
```
mvo5: (not a concern of this PR) - feels a bit strange that `ClientSnapFromSnapInfo()` is under `cmd` …

cmd/snap/cmd_info.go
```diff
@@ -119,52 +121,22 @@ func maybePrintBase(w io.Writer, base string, verbose bool) {
 	}
 }
 
-func tryDirect(w io.Writer, path string, verbose bool, termWidth int) bool {
-	path = norm(path)
-
+func loadDirect(path string) (*client.Snap, error) {
```
pedronis: load seems like the wrong verb

cmd/snap/cmd_info.go
```diff
@@ -462,17 +434,32 @@ func (x *infoCmd) Execute([]string) error {
...
+			remote, resInfo, remoteErr = x.client.FindOne(snapName)
+			local, _, localErr = x.client.Snap(snapName)
+		}
+		// note direct == nil, or local == nil and remote == nil
+		theSnap := coalesce(direct, local, remote)
 
-		if both == nil {
+		if theSnap == nil {
+			if x.Verbose {
+				if directErr != nil {
```
pedronis: if we get here, we will always print the 3 no? not a fan of the stacked error error error output

cmd/snap/cmd_info.go
```diff
@@ -481,27 +468,41 @@ func (x *infoCmd) Execute([]string) error {
...
+			fmt.Fprintf(w, "path:\t%q\n", snapName)
+		}
+		fmt.Fprintf(w, "name:\t%s\n", theSnap.Name)
+		printSummary(w, theSnap.Summary, termWidth)
+		if direct == nil {
+			fmt.Fprintf(w, "publisher:\t%s\n", longPublisher(esc, theSnap.Publisher))
+		} else {
+			// NotesFromRemote might be better called NotesFromNotInstalled but that's nasty
+			fmt.Fprintf(w, "version:\t%s %s\n", theSnap.Version, NotesFromRemote(theSnap, nil))
+			if !osutil.IsDirectory(snapName) {
```
pedronis: I don't like that we have direct checks like this but also maybe* helpers. I think slightly would prefer more maybe helpers

cmd/snap/cmd_info.go
```diff
@@ -522,35 +523,42 @@ func (x *infoCmd) Execute([]string) error {
...
+		maybePrintType(w, theSnap.Type)
+		maybePrintBase(w, theSnap.Base, x.Verbose)
+		if direct != nil && x.Verbose && !osutil.IsDirectory(snapName) {
+			sha3_384, _, _ := asserts.SnapFileSHA3_384(snapName)
+			if sha3_384 != "" {
+				fmt.Fprintf(w, "sha3-384:\t%s\n", sha3_384)
 			}
-			if !local.InstallDate.IsZero() {
-				fmt.Fprintf(w, "refresh-date:\t%s\n", x.fmtTime(local.InstallDate))
+		} else {
```
pedronis: this else seems wrong, shouldn't be just on direct == nil

cmd/snap/cmd_info.go
pedronis: examine?

cmd/snap/cmd_info.go
chipaca: yeah, i'm open to ideas on this front

cmd/snap/cmd_info.go
chipaca: I'm not sure what's wrong with `load`, but `examine` is probably wrong

cmd/snap/cmd_info.go
chipaca: yeah, looks like i goofed (undetectably)

cmd/snap/cmd_info.go
pedronis: as we discussed  "load" sounds like ti will load the entire snap in memory, and we tend not to use it.

clientSnapFromSnapFile is probably the most verbose name I can think for this?

cmd/snap/cmd_info.go
pedronis: what are we trying to address here, are the error content usually meaningful here?  is the problem that not found is too generic?  I'm also not sure that the first impulse if something doesn't work is to switch to verbose? Maybe we should always pack more info in the actual error we return

cmd/snap/cmd_info.go
pedronis: does it mean we are missing a test?

cmd/snap/cmd_info.go
```diff
@@ -462,17 +434,32 @@ func (x *infoCmd) Execute([]string) error {
 			continue
 		}
 
-		if tryDirect(w, snapName, x.Verbose, termWidth) {
-			noneOK = false
-			continue
-		}
-		remote, resInfo, _ := x.client.FindOne(snapName)
-		local, _, _ := x.client.Snap(snapName)
+		var direct, local, remote *client.Snap
```
pedronis: direct sounds too boolean too me,  though probably they all have the problem already?

should we bit the bullet and rename  to  diskSnap (or fsSnap),  localSnap and remoteSnap? could be a follow up (the diff is already painful)

cmd/snap/cmd_info.go
chipaca: going with `clientSnapFromPath` (because `SnapFile` is arguably what `snap.Open` gives you).

cmd/snap/cmd_info.go
chipaca: I'll leave out this bit of the change for now. The intention is that as we 'hide' the error, people will get 'not found' for info of a remote snap when the network is out for example, which is suboptimal. But in those cases this is too verbose also. So it needs more thought anyway.

cmd/snap/cmd_info.go
chipaca: it's undetectable because every print inside the `else` is a maybe so nothing is printed in the cases where it's done a bit of additional work

cmd/snap/cmd_info.go
chipaca: maybe another one for later though

pedronis: that was my main pain point reviewing this though, if it cannot be addressed now (after so long) then probably we should close this until we are less busy

cmd/snap/cmd_info.go
pedronis: this needs addressing as well

cmd/snap/cmd_info.go
pedronis: if the goal is to land it quickly we should undo this bit for now then

cmd/snap/cmd_info.go
pedronis: so this seems to be a dep for the health check work, I still think the new even bigger mix of if and helpers is painful, we do need to do something

chipaca: ok, I'll give it a go in this. I didn't want to make it bigger :-) but, fair enough

chipaca:
@pedronis that was a biggie, but I think it does what we wanted :-)

pedronis:
thanks for the refactoring, seems some area are not covered much by unit tests,  this was the case already before, so not  a blocker per se, but given I'm looking over worth to point those out

cmd/snap/cmd_info.go
```diff
@@ -305,37 +226,217 @@ func printDescr(w io.Writer, descr string, termWidth int) error {
...
+	// (and os->core rename happens, etc)
+	t := iw.theSnap.Type
+	switch t {
+	case "", "app", "application":
+		return
+	case "os":
+		t = "core"
+	}
+
+	fmt.Fprintf(iw, "type:\t%s\n", t)
```
pedronis: this bit seems not tested, only the "app" case

cmd/snap/cmd_info.go
```diff
@@ -305,37 +226,217 @@ func printDescr(w io.Writer, descr string, termWidth int) error {
...
+
+func (iw *infoWriter) maybePrintPublisher() {
+	if iw.diskSnap != nil {
+		// snaps read from disk won't have a publisher
+		return
+	}
+	fmt.Fprintf(iw, "publisher:\t%s\n", longPublisher(iw.esc, iw.theSnap.Publisher))
+}
+
+func (iw *infoWriter) maybePrintVersion() {
```
pedronis: should we call it maybePrintStandaloneVersion because is only for the disk snap case?  or not worth?

cmd/snap/cmd_info.go
```diff
@@ -305,37 +226,217 @@ func printDescr(w io.Writer, descr string, termWidth int) error {
...
+	if iw.diskSnap == nil {
+		// snaps not read from disk will have version information shown elsewhere
+		return
+	}
+	// NotesFromRemote might be better called NotesFromNotInstalled but that's nasty
+	fmt.Fprintf(iw, "version:\t%s %s\n", iw.diskSnap.Version, NotesFromRemote(iw.diskSnap, nil))
```
pedronis: this does not seem to be cover by unit tests

cmd/snap/cmd_info.go
```diff
@@ -351,18 +452,62 @@ func maybePrintServices(w io.Writer, snapName string, allApps []client.AppInfo,
...
-		fmt.Fprintln(w, svc)
+		fmt.Fprintln(iw, svc)
 	}
 }
 
+func (iw *infoWriter) maybePrintNotes() {
+	if !iw.verbose {
+		return
+	}
+	fmt.Fprintln(iw, "notes:\t")
```
pedronis: seems the verbose case is not unit tested

cmd/snap/cmd_info.go
```diff
@@ -351,18 +452,62 @@ func maybePrintServices(w io.Writer, snapName string, allApps []client.AppInfo,
...
+
+	fmt.Fprintf(iw, "  ignore-validation:\t%t\n", iw.localSnap.IgnoreValidation)
+	return
+}
+
+func (iw *infoWriter) maybePrintSum() {
+	if !iw.verbose {
+		return
+	}
+	if iw.diskSnap == nil {
```
pedronis: seems the whole diskSnap != nil is not unit tested

cmd/snap/cmd_info.go
```diff
@@ -469,17 +622,23 @@ func (x *infoCmd) Execute([]string) error {
 			continue
 		}
 
-		if tryDirect(w, snapName, x.Verbose, termWidth) {
-			noneOK = false
-			continue
-		}
-		remote, resInfo, _ := x.client.FindOne(snapName)
-		local, _, _ := x.client.Snap(snapName)
+		iw.arg = snapName
```
pedronis: do we ever use this value of arg?  it seems arg is only used where diskSnap != nil or am I missing something?  if that's the case maybe we should call it path perhaps?

cmd/snap/cmd_info.go
```diff
@@ -469,17 +622,23 @@ func (x *infoCmd) Execute([]string) error {
...
+		var err error
 
-		both := coalesce(local, remote)
+		iw.diskSnap, err = clientSnapFromPath(snapName)
+		if err == nil {
+			iw.localSnap, iw.remoteSnap, iw.resInfo = nil, nil, nil
```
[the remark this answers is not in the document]
pedronis: infoWriter  could grow

setupDiskSnap(diskSnap)

and 

setupSnap(localSnap, remoteSnap)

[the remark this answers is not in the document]
chipaca: or indeed `setSnap(snapName)`

cmd/snap/cmd_info.go
chipaca: fair point, I'll rename it "path" and only set it if it's a disk snap

cmd/snap/cmd_info.go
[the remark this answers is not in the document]
pedronis: I thought about that, the issue with the latter, is that then it blurs whether is just a infoWriter or more, also we will not reuse it so unclear that moving all the client etc class from Execute to it is worth it

cmd/snap/cmd_info.go
```diff
@@ -457,10 +602,18 @@ func (x *infoCmd) Execute([]string) error {
 
 	esc := x.getEscapes()
 	w := tabwriter.NewWriter(Stdout, 2, 2, 1, ' ', 0)
+	iw := &infoWriter{
```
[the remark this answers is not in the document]
chipaca: but then it'd just be a function with a bunch of flags ... :-|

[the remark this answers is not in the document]
pedronis: yes, there is only this place building it, might be easier to instead split the fields in infoWriter definition into two groups with comments (this is fixed, this per snap)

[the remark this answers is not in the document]
mvo5: > but then it'd just be a function with a bunch of flags ... :-|

Indeed - my concern was mostly to make it easier to figure out what to set, the comment suggestion from Samuele is probably the best approach.

cmd/snap/cmd_info.go
```diff
@@ -305,37 +217,240 @@ func printDescr(w io.Writer, descr string, termWidth int) error {
...
 
-func maybePrintCommands(w io.Writer, snapName string, allApps []client.AppInfo, n int) {
-	if len(allApps) == 0 {
+type writeflusher interface {
+	io.Writer
+	Flush() error
+}
+
+type infoWriter struct {
+	// these fields are set every iteration:
```
pedronis: nitpick but the ':' is superfluous, unusual

cmd/snap/cmd_info.go
```diff
@@ -305,37 +217,240 @@ func printDescr(w io.Writer, descr string, termWidth int) error {
...
+	theSnap    *client.Snap
+	diskSnap   *client.Snap
+	localSnap  *client.Snap
+	remoteSnap *client.Snap
+	resInfo    *client.ResultInfo
+	path       string
+	// these fields don't change and so can be set once:
```
pedronis: nitpick but the ':' is superfluous, unusual

cmd/snap/cmd_info.go
```diff
@@ -305,37 +217,240 @@ func printDescr(w io.Writer, descr string, termWidth int) error {
...
+	writeflusher
+	esc       *escapes
+	termWidth int
+	fmtTime   func(time.Time) string
+	absTime   bool
+	verbose   bool
+}
+
+func (iw *infoWriter) setupDiskSnap(path string, diskSnap *client.Snap) {
```
pedronis: afaict this is not exercised by unit tests

cmd/snap/cmd_info.go
```diff
@@ -74,106 +76,31 @@ func init() {
...
-func maybePrintBase(w io.Writer, base string, verbose bool) {
-	if verbose && base != "" {
-		fmt.Fprintf(w, "base:\t%s\n", base)
-	}
-}
-
-func tryDirect(w io.Writer, path string, verbose bool, termWidth int) bool {
-	path = norm(path)
-
+func clientSnapFromPath(path string) (*client.Snap, error) {
```
pedronis: this is also not exercised afaict for the case Open succeeds

cmd/snap/export_test.go
```diff
@@ -81,6 +79,100 @@ var (
 	InterfacesDeprecationNotice = interfacesDeprecationNotice
 )
 
+func MaybePrintServices(w writeflusher, theSnap *client.Snap) {
```
pedronis: not completely a fan of these exported helpers that both build the infoWriter and call the methods.  We can keep the type opaque to tests but still have a bit more orthogonality,  for example something like, just as an example:

``` 
-func MaybePrintType(w writeflusher, theSnap *client.Snap) {
-       iw := &infoWriter{writeflusher: w, theSnap: theSnap}
+func IW(w writeflusher, theSnap *client.Snap) *infoWriter {
+       return &infoWriter{writeflusher: w, theSnap: theSnap}
+}
+
+func MaybePrintType(iw *infoWriter) {
        iw.maybePrintType()
 }
```

`IW` not the final name(!), we would have a couple of those infoWriter builders,  some could also use the setup*Snap methods.

cmd/snap/cmd_info_test.go
```diff
@@ -64,9 +65,12 @@ type flushBuffer struct{ bytes.Buffer }
 func (*flushBuffer) Flush() error { return nil }
 
 func (s *infoSuite) TestMaybePrintServices(c *check.C) {
+	var buf flushBuffer
+	iw := snap.NewInfoWriter(&buf)
 	for _, infos := range [][]client.AppInfo{svcAppInfos, mixedAppInfos} {
-		var buf flushBuffer
-		snap.MaybePrintServices(&buf, &client.Snap{Name: "foo", Apps: infos})
+		buf.Reset()
+		snap.SetupDiskSnap(iw, "", &client.Snap{Name: "foo", Apps: infos})
```
pedronis: any particular reason to use SetupDiskSnap a bit more here and other places vs SetupSnap,  the latter is more the common case in a sense?

chipaca: no good reason. I preferred writing `""` over `nil, nil` ?

pedronis: I like all the `nil`s :)

PR #6568, opened 2019-03-05
overlord/snapstate: fix restoring of "old-current" revision config in undoLinkSnap

stolowski:
When undoing snap update, the undoLinkSnap handler doesn't correctly restore snap configuration for the "old current" revision that we are returning to. This PR fixes it. There is some code needed for the new unit test that was already implemented in other test, so it's extracted into a helper. The new test fails without the fix in handlers.go.

Thanks for @chipaca for XXX in the code.


zyga:
Thank you for the updates!

PR #6570, opened 2019-03-06
snapstate: only keep 2 snaps on classic

mvo5:
We had a recent discussion with the foundations team about the number
of snaps that should be kept on the system for rollback. Because
classic does not have a kernel snap the number of 2 snaps (one active
and one as a rollback target) is sufficient there. This PR changes
the default on classic from 3 to 2 and updates the tests.


chipaca:
@zyga it will not act instantly, just per-snap on next refresh, and it will not change the setting of somebody that set it to something; it affects the default.

degville:
@pedronis good point - I think *System options* is the only place that mentions the (old) default.

PR #6572, opened 2019-03-06
cmd/snap-confine: populate enter_non_classic_execution_environment

zyga:
Move all of the code that differentiates strict and classic confinement
into the dedicated functions. There are no other changes, apart from
indent reduction and comment re-flow for readability.

Signed-off-by: Zygmunt Krynicki <zygmunt.krynicki@canonical.com>

PR #6573, opened 2019-03-06
cmd/snap-confine: call sc_should_use_normal_mode once

zyga:
Historically snap-confine used two modes of operation. Initially on core
systems snap-confine would unshare the mount namespace and preform some
alterations but would otherwise run with the contents of the initial
mount namesapce. On classic systems snap-confine would pivot into the
base snap (core) and perform some more alterations.

With the introduction of multiple boot bases, per-snap bases some of
that logic seems rusty. To make it less confusing what is going on
reduce the number of places that determine if legacy (sans-pivot) or
normal mode (with pivot) to exactly one.

There is more cleanup that can be done but it will require making the
invocation structure something we can pass around to more functions.
This will be attempted in subsequent patches.

### DIFF
--- a/cmd/snap-confine/mount-support.c
+++ b/cmd/snap-confine/mount-support.c
@@ -516,7 +516,8 @@ static bool __attribute__ ((used))
 }
 
 void sc_populate_mount_ns(struct sc_apparmor *apparmor, int snap_update_ns_fd,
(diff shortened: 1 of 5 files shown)


zyga:
This is based on https://github.com/snapcore/snapd/pull/6572 and https://github.com/snapcore/snapd/pull/6571

PR #6575, opened 2019-03-07
cmd/snap-confine: pass sc_invocation instead of numerous args around

zyga:
Several places use a combination of base_snap_name, snap_name (or
snap_instance, those are used interchangeably) and normal_mode flag and pass
them around as separate arguments.

With the recent introduction of sc_invocation as a private feature of
snap-confine, we can now take advantage of the fact that it can tidy up much of
this complexity. This is done without impacting behaviour in any way.

I highlighted the special case of ubuntu-core fallback logic modifying the base
snap name behind the back of the rest of the code. This is not a new property
so I kept it intact for now


zyga:
This is based on #6571 #6572 and #6573

cmd/snap-confine/snap-confine-invocation.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <stdbool.h>
+
+/**
+ * sc_invocation contains information about how snap-confine was invoked.
+ *
+ * All of the pointer fields have the life-cycle bound to the main process.
+ **/
+typedef struct sc_invocation {
+	/* Things declared by the system. */
```
pedronis: are classic_confinement and executable going to be used soon?

zyga: I would like to use them instead of the various locals in main(). The reason they are added here is that we can define sc_invocation as the external data about invoking snap-confine to run some program. It's the complete set.

zyga: I have prepared tweak/refactor-sc-main-5 for that.

cmd/snap-confine/ns-support.h
```diff
@@ -91,9 +92,8 @@ void sc_close_mount_ns(struct sc_mount_ns *group);
  * function returns zero.
  **/
 int sc_join_preserved_ns(struct sc_mount_ns *group, struct sc_apparmor
-			 *apparmor, const char *base_snap_name,
-			 const char *snap_name, int snap_discard_ns_fd,
-			 bool is_normal_mode);
+			 *apparmor, sc_invocation * inv,
```
pedronis: here it could be a "const" already?

cmd/snap-confine/snap-confine-invocation.h
```diff
@@ -0,0 +1,39 @@
...
+	const char *base_snap_name;
+	const char *executable;
+	const char *security_tag;
+	const char *snap_instance;
+	bool classic_confinement;
```
pedronis: I would sort them:

executable
snap_instance
base_snap_name
security_tag
classic_confinement

or same but exeutable last

cmd/snap-confine/ns-support.h
zyga: Yes, I'll switch that over.

cmd/Makefile.am
```diff
@@ -226,6 +228,9 @@ snap_confine_snap_confine_SOURCES = \
 	snap-confine/ns-support.h \
 	snap-confine/snap-confine-args.c \
 	snap-confine/snap-confine-args.h \
+	snap-confine/snap-confine-invocation.c \
+	snap-confine/snap-confine-invocation.h \
+	snap-confine/snap-confine-invocation.h \
```
zyga: Opps, let me fix that first.

zyga:
I accidentally pushed bits from -5  here, sorry about that (I meant to split it off for easier review before).

cmd/snap-confine/snap-confine-invocation.c
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+#include "snap-confine-invocation.h"
+
+#include <string.h>
+
+#include "../libsnap-confine-private/snap.h"
+#include "../libsnap-confine-private/utils.h"
+
+void sc_init_invocation(sc_invocation *inv, const struct sc_args *args, const char *snap_instance) {
```
pedronis: it's a bit unclear if this merit a new file instead of being a static helper in snap-confine.c,  also "invocation" as a concept doesn't tell anything much to people new to this so while understand the need for the .h file I'm sceptical it's useful to have a dedicated module about it,  the fallback code in the other PR put here doesn't help the case. Overall the snap-confine code is already a bit spread out with sometimes uneven layering depth.

cmd/snap-confine/snap-confine-invocation.h
```diff
@@ -0,0 +1,50 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
```
pedronis: this is so fundamental/passed around that it might as well be "snap-confine.h"

cmd/snap-confine/ns-support.c
```diff
@@ -383,6 +383,7 @@ static int sc_inspect_and_maybe_discard_stale_ns(int mnt_fd,
 		// systemd. This makes us end up in a situation where the outer base
 		// snap will never match the rootfs inside the mount namespace.
 		bool should_discard =
+		    inv->
 		    is_normal_mode ? should_discard_current_ns(base_snap_dev) :
```
jdstrand: Breaking up 'inv->' and 'is_normal_mode' seems like strange formatting; can't we format it so 'inv->is_normal_mode' is together?

zyga: No, it's indent. We would have to reformat the whole file or ignore indent from now on.

cmd/snap-confine/snap-confine-invocation.c
```diff
@@ -0,0 +1,68 @@
...
+    const char *base_snap_name = sc_args_base_snap(args) ?: "core";
+    sc_snap_name_validate(base_snap_name, NULL);
+
+    /* The executable is conveyed via untrusted command lne. It must be set
+     * but cannot be validated further than that at this time. It might be
+     * arguable to validate it to be snap-exec in one of the well-known
+     * locations or one of the special-cases like strace / gdb but this is
+     * not done at this time. */
+    const char *executable = sc_args_executable(args);
+    /* TODO: validate NULL */
```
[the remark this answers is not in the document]
zyga: I know, this is a refactor. I kept this property and intend to address it separately.

cmd/snap-confine/snap-confine-invocation.c
```diff
@@ -0,0 +1,68 @@
...
+     * tag. */
+    const char *security_tag = sc_args_security_tag(args);
+    if (!verify_security_tag(security_tag, snap_instance)) {
+        die("security tag %s not allowed", security_tag);
+    }
+
+    /* The base snap name is conveyed via untrusted, optional, command line
+     * argument. It may be omitted where it implies the "core" snap is the
+     * base. */
+    const char *base_snap_name = sc_args_base_snap(args) ?: "core";
```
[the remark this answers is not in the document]
zyga: Yes this is a gcc. extension but it is convenient. I don't see why we should avoid using it.

Memory management is such that the values here are not freed since they are a bound the the lifecycle of `args`. I'm happy to explore changing this but this is done consistently across sc_invocation.

cmd/snap-confine/snap-confine-invocation.c
```diff
@@ -0,0 +1,68 @@
...
+     * not done at this time. */
+    const char *executable = sc_args_executable(args);
+    /* TODO: validate NULL */
+
+    /* Invocation helps to pass relevant data to various parts of snap-confine. */
+    memset(inv, 0, sizeof *inv);
+    inv->base_snap_name = base_snap_name;
+    inv->executable = executable;
+    inv->security_tag = security_tag;
+    inv->snap_instance = snap_instance;
```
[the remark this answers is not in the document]
zyga: As I replied above this is done consistently but I'm happy to change that.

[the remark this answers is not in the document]
mvo5: Just my 2¢ but I'm in favour of changing it to use sc_strdup() to ensure the ownership of this memory is clear and we don't have unexpected side-effects in the future.

cmd/snap-confine/snap-confine-invocation.c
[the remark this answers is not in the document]
zyga: I changed it though I hold my position that this is not an issue.

cmd/snap-confine/snap-confine-invocation.c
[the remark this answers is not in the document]
zyga: Done now.

cmd/snap-confine/snap-confine-invocation.h
zyga: Done now.

cmd/snap-confine/snap-confine-invocation.c
zyga: Merged into snap-confine.[ch] now

zyga:
@pedronis I had to undo the meld into `snap-confine.c`, we need `snap-confine-invocation.c` for linking with unit tests.

cmd/snap-confine/snap-confine-invocation.c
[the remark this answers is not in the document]
zyga: I changed it due to memory management changes.

pedronis:
@zyga which unit tests?  3343a13 was green

zyga:
@pedronis in the -5 branch.

zyga:
Hmm, travis doesn't seem to see this branch https://travis-ci.org/snapcore/snapd/builds/507790331?utm_source=github_status&utm_medium=notification

pedronis:
> @pedronis in the -5 branch.

@zyga do you have a link to the failure?

zyga:
@pedronis no, because I pushed the fixed version after merging locally. The problem is simple, mount-support depends on some of the functions defined in snap-confine.c and that cannot be merged into unit tests due to clashes with main. We could use ifdef tricks but I think that's less desirable.

cmd/snap-confine/snap-confine-invocation.h
mvo5: I'm confused, I thought this comment was moving `sc_invocation` into `snap-confine.h` - am I misunderstanding the comment? I mean, its marked as "Done now" and "resolved" and yet we still have `sc_invocation` in snap-confine-invocation.h - what am I missing :)

zyga: This was reverted as indicated here https://github.com/snapcore/snapd/pull/6575#issuecomment-473858513

cmd/snap-confine/snap-confine-invocation.h
```diff
@@ -0,0 +1,65 @@
...
+ * Invocation is constructed based on command line arguments as well as
+ * environment value (SNAP_INSTANCE_NAME). All input is untrusted and is
+ * validated internally.
+ **/
+void sc_init_invocation(sc_invocation *inv, const struct sc_args *args, const char *snap_instance);
+
+/**
+ * sc_fini_invocation releases memory used by its constituents.
+ **/
+void sc_fini_invocation(sc_invocation *inv);
```
mvo5: See above: why do we have two very similar cleanup functions  here?

cmd/snap-confine/snap-confine-invocation.c
```diff
@@ -0,0 +1,88 @@
...
+#include "../libsnap-confine-private/snap.h"
+#include "../libsnap-confine-private/string-utils.h"
+#include "../libsnap-confine-private/utils.h"
+
+void sc_init_invocation(sc_invocation *inv, const struct sc_args *args, const char *snap_instance) {
+    /* Snap instance name is conveyed via untrusted environment. It may be
+     * unset (typically when experimenting with snap-confine by hand). It
+     * must also be a valid snap instance name. */
+    if (snap_instance == NULL) {
+        die("SNAP_INSTANCE_NAME is not set");
```
mvo5: This feels like the error message is at the wrong place now, I mean: `sc_init_invocation` has no idea where snap_instance comes from. 

I think we should die() here but with a error message local to the function (like "cannot call sc_init_invocation with snap_instance == NULL" or something) and also check for NULL in the getenv("SNAP_INSTANCE_NAME") and die there with the old error message (and maybe cleaup the error message to say "environemtn SNAP_INSTANCE_NAME is not set" or something like this.

Alternatively call getenv() inside sc_init_invocation() and not pass as parameter.

cmd/snap-confine/snap-confine-invocation.c
```diff
@@ -0,0 +1,88 @@
...
+    inv->snap_instance = sc_strdup(snap_instance);
+    inv->classic_confinement = sc_args_is_classic_confinement(args);
+
+    debug("security tag: %s", inv->security_tag);
+    debug("executable:   %s", inv->executable);
+    debug("confinement:  %s", inv->classic_confinement ? "classic" : "non-classic");
+    debug("base snap:    %s", inv->base_snap_name);
+}
+
+void sc_fini_invocation(sc_invocation *inv) {
```
mvo5: Why is the cleanup split into two helpers? AFAICT we have no other `fini` helper and both sc_fini_invocation and sc_cleanup_invocation can be used with SC_CLEANUP so whats the advantage?

cmd/snap-confine/snap-confine-invocation.c
```diff
@@ -0,0 +1,88 @@
...
+    sc_cleanup_string(&inv->snap_instance);
```
mvo5: Should we set `inv->snap_instance` (and frieds) to NULL here? As this function is written now it looks slightly dangerous to me, if sc_fini_invocation() is called twice on the same struct we will get a double free segfault, no?

cmd/snap-confine/snap-confine.c
```diff
@@ -124,29 +117,11 @@ int main(int argc, char **argv)
...
-	debug("security tag: %s", security_tag);
-	debug("executable:   %s", executable);
-	debug("confinement:  %s",
-	      classic_confinement ? "classic" : "non-classic");
-	debug("base snap:    %s", base_snap_name);
+	/* Collect all invocation parameters. This gives us authoritative
+	 * information about what needs to be invoked and how. The data comes
+	 * from either the environment or from command line arguments */
+	sc_invocation SC_CLEANUP(sc_cleanup_invocation) invocation;
+	sc_init_invocation(&invocation, args, getenv("SNAP_INSTANCE_NAME"));
```
mvo5: See above, I would move getenv() to a different line with a extra check for NULL or move it inside sc_init_invocation().

cmd/snap-confine/snap-confine-invocation.c
zyga: Since this error message is the only thing displayed when snap-confine is invoked without setting the environment variable I would rather ensure the message is useful. We can move the exact same check to `snap-confine.c`'s main function and just ensure that it is not NULL here.

cmd/snap-confine/snap-confine-invocation.c
zyga: We don't do it in any of the cleanup functions so far. Since the purpose of the cleanup function is to be used with the GCC extension that ensures it is called only once. I can certainly add appropriate code to all the cleanup functions. Should I perform this in a separate branch, since adding it there will instantly benefit all the consumers.

cmd/snap-confine/snap-confine-invocation.h
zyga: Partially historic, partially intentional. Cleanup functions, usually, have different signature (they take a pointer to an object being cleaned up, which is usually a pointer in itself). Here the actual structure is not heap-allocated so the functions have identical signature.

We don't have a consistent approach, some structures are heap allocated with private definition. Others use a public definition and are sometimes stack allocated, with only the constituent fields needing to be managed.

cmd/snap-confine/snap-confine-invocation.c
zyga: I can probably just make this private to advance the discussion.

cmd/snap-confine/snap-confine.c
zyga: Done now.

cmd/snap-confine/snap-confine-invocation.c
```diff
@@ -0,0 +1,88 @@
...
+static void sc_fini_invocation(sc_invocation *inv) {
+    sc_cleanup_string(&inv->snap_instance);
+    sc_cleanup_string(&inv->base_snap_name);
+    sc_cleanup_string(&inv->security_tag);
+    sc_cleanup_string(&inv->executable);
+}
+
+void sc_cleanup_invocation(sc_invocation *inv) {
+    if (inv != NULL) {
+        sc_fini_invocation(inv);
```
mvo5: I don't see why we do not just fold sc_fini_invocation() here? I mean, this helper is only used in a single place, its private and trivial. Whats the advantage of the split? Will sc_fini_invocation be used in more places in the future? Or is it just a code-style preference?

zyga: It's a code style preference. I know it's a hit-and-miss so far (mainly because it is a while to change something in that codebase). I would like to have init/fini for the contents of the structure but not for the structure itself.

Given that we are getting into a flow of refactoring snap-confine I would like to keep it, so that it can be in line with other structures that get the same treatment.

mvo5: What is the difference between `sc_cleanup_...` and `sc_fini_...` ? I still don't see why we need two when they do the same? Maybe I need an example where the two are actually different? If we want symetry, why not "init" and "cleanup"? Sorry for being so thick here but I have this YAGNI feeling and I really would like to understand if its real or just me not knowing the details.

zyga: Cleanup is specifically the cleanup handler that takes a pointer to the object being cleaned up (e.g. a `int *` for sc_cleanup_close`). The finaliser runs on the object with releasing the memory of the object itself (e.g. `close` as an analogy). In this case however those are identical because the structure is not heap allocated.

zyga: After discussing this I folded the function into one and decided to make a separate pass to make memory management more consistent around the tree.

PR #6577, opened 2019-03-08
many: make Remodel() download everything first before installing

mvo5:
For remodel we cannot assume that we have network access during
the operation. Some remodeling may switch to a different way of
managing the network or similar. This means we need to order the
wait chains in the remodel task so that all downloads and checks
for assertions happen first and the actual install operations
(that do not require network) happen after this.

This PR implements this by taking the TaskSet from Install() or
Update() and marking the "check-assertions" task with a new
TaskEdge marker. This way devicestate can split the taskset on
this marker and adjust the wait chains accordingly.

### DIFF
--- a/overlord/devicestate/devicestate.go
+++ b/overlord/devicestate/devicestate.go
@@ -293,24 +293,25 @@ func CanManageRefreshes(st *state.State) bool {
 	return false
 }
 
-// splitTsAfterEdge takes a taskset ts and splits it into two tasksets
-// after the first occurrence of the given task kind
-func splitTsAfterEdge(edge state.TaskSetEdge, ts *state.TaskSet) (*state.TaskSet, *state.TaskSet) {
-	tsBefore := state.NewTaskSet()
-	tsAfter := state.NewTaskSet()
-	edgeTask := ts.Edge(edge)
-	found := false
+// extractDownloadInstallEdgesFromTs is a helper that extract the first, last
+// download and install tasks from a TaskSet
+func extractDownloadInstallEdgesFromTs(ts *state.TaskSet) (firstDl, lastDl, firstInst, lastInst *state.Task) {
+	edgeTask := ts.Edge(snapstate.DownloadAndChecksDoneEdge)
 	for _, t := range ts.Tasks() {
-		if !found {
-			tsBefore.AddTask(t)
-		} else {
-			tsAfter.AddTask(t)
+		if firstDl == nil {
+			firstDl = t
+		}
+		if firstInst == nil && lastDl != nil {
+			firstInst = t
 		}
 		if t == edgeTask {
-			found = true
+			lastDl = t
+		}
+		if firstInst != nil {
+			lastInst = t
 		}
 	}
-	return tsBefore, tsAfter
+	return firstDl, lastDl, firstInst, lastInst
 }
 
 // Remodel takes a new model assertion and generates a change that
@@ -331,6 +332,10 @@ func Remodel(st *state.State, new *asserts.Model) ([]*state.TaskSet, error) {
 	if current.Series() != new.Series() {
 		return nil, fmt.Errorf("cannot remodel to different series yet")
 	}
+	// FIXME: we need language in the model assertion to declare
+	// what transitions are ok across device services (aka serial
+	// vaults) before we allow remodel like this.
+	//
 	// Right now we only allow "remodel" to a different revision of
 	// the same model.
 	if current.BrandID() != new.BrandID() {
@@ -396,7 +401,8 @@ func Remodel(st *state.State, new *asserts.Model) ([]*state.TaskSet, error) {
 	// Ensure all download/check tasks are run *before* the install
 	// tasks. During a remodel the network may not be available so
 	// we need to ensure we have everything local.
-	var prevDownload, prevInstall *state.TaskSet
+	var lastDownloadInChain, firstInstallInChain *state.Task
+	var prevDownload, prevInstall *state.Task
 	for _, ts := range tss {
 		// make sure all things happen sequentially:
 		// - all tasks inside ts{Download,Install} already wait for
@@ -411,24 +417,25 @@ func Remodel(st *state.State, new *asserts.Model) ([]*state.TaskSet, error) {
 		//     download2 <- verify1
 		//     install1 <- verify1 <- download1
 		//     install2 <- verify2 <- install1
-		tsDownload, tsInstall := splitTsAfterEdge(snapstate.DownloadAndChecksDoneEdge, ts)
+		firstDownload, lastDownload, firstInstall, lastInstall := extractDownloadInstallEdgesFromTs(ts)
 		if prevDownload != nil {
-			tsDownload.Tasks()[0].WaitFor(prevDownload.Tasks()[len(prevDownload.Tasks())-1])
+			firstDownload.WaitFor(prevDownload)
 		}
 		if prevInstall != nil {
-			tsInstall.Tasks()[0].WaitFor(prevInstall.Tasks()[len(prevInstall.Tasks())-1])
+			firstInstall.WaitFor(prevInstall)
 		}
-		prevDownload = tsDownload
-		prevInstall = tsInstall
+		prevDownload = lastDownload
+		prevInstall = lastInstall
+		// update global state
+		lastDownloadInChain = lastDownload
+		if firstInstallInChain == nil {
+			firstInstallInChain = firstInstall
+		}
 	}
 	// Make sure the first install waits for the last download. With this
 	// our (simplified) wait chain looks like:
 	// download1 <- verify1 <- download2 <- verify2 <- install1 <- install2
-	lastTsDownload, _ := splitTsAfterEdge(snapstate.DownloadAndChecksDoneEdge, tss[len(tss)-1])
-	tLastDownload := lastTsDownload.Tasks()[len(lastTsDownload.Tasks())-1]
-	_, firstTsInstall := splitTsAfterEdge(snapstate.DownloadAndChecksDoneEdge, tss[0])
-	tFirstInstall := firstTsInstall.Tasks()[0]
-	tFirstInstall.WaitFor(tLastDownload)
+	firstInstallInChain.WaitFor(lastDownloadInChain)
 
 	// Set the new model assertion - this *must* be the last thing done
 	// by the change.
--- a/overlord/devicestate/devicestate.go
+++ b/overlord/devicestate/devicestate.go
@@ -293,6 +293,26 @@ func CanManageRefreshes(st *state.State) bool {
 	return false
 }
 
+// splitTsAfterEdge takes a taskset ts and splits it into two tasksets
+// after the first occurrence of the given task kind
+func splitTsAfterEdge(edge state.TaskSetEdge, ts *state.TaskSet) (*state.TaskSet, *state.TaskSet) {
+	tsBefore := state.NewTaskSet()
+	tsAfter := state.NewTaskSet()
+	edgeTask := ts.Edge(edge)
+	found := false
+	for _, t := range ts.Tasks() {
+		if !found {
+			tsBefore.AddTask(t)
+		} else {
+			tsAfter.AddTask(t)
+		}
+		if t == edgeTask {
+			found = true
+		}
+	}
+	return tsBefore, tsAfter
+}
+
 // Remodel takes a new model assertion and generates a change that
 // takes the device from the old to the new model or an error if the
 // transition is not possible.
@@ -303,9 +323,6 @@ func CanManageRefreshes(st *state.State) bool {
 // - Need new session/serial if changing store or model
 // - Check all relevant snaps exist in new store
 //   (need to check that even unchanged snaps are accessible)
-// - Download everything in a first phase of the change and "pin" cache
-//   files (also get assertions), which means also dealing with new bases
-//   and content providers
 func Remodel(st *state.State, new *asserts.Model) ([]*state.TaskSet, error) {
 	current, err := Model(st)
 	if err != nil {
@@ -314,9 +331,6 @@ func Remodel(st *state.State, new *asserts.Model) ([]*state.TaskSet, error) {
 	if current.Series() != new.Series() {
 		return nil, fmt.Errorf("cannot remodel to different series yet")
 	}
-	// FIXME: we need language in the model assertion to declare what
-	// transitions are ok before we allow remodel like this.
-	//
 	// Right now we only allow "remodel" to a different revision of
 	// the same model.
 	if current.BrandID() != new.BrandID() {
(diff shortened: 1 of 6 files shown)


overlord/devicestate/devicestate.go
```diff
@@ -314,9 +331,6 @@ func Remodel(st *state.State, new *asserts.Model) ([]*state.TaskSet, error) {
 	if current.Series() != new.Series() {
 		return nil, fmt.Errorf("cannot remodel to different series yet")
 	}
-	// FIXME: we need language in the model assertion to declare what
-	// transitions are ok before we allow remodel like this.
-	//
```
pedronis: this should stay but with "across device services (aka serial vaults)"

overlord/devicestate/devicestate.go
```diff
@@ -293,6 +293,26 @@ func CanManageRefreshes(st *state.State) bool {
 	return false
 }
 
+// splitTsAfterEdge takes a taskset ts and splits it into two tasksets
+// after the first occurrence of the given task kind
+func splitTsAfterEdge(edge state.TaskSetEdge, ts *state.TaskSet) (*state.TaskSet, *state.TaskSet) {
```
pedronis: this seems doesn't need to be a split anymore but could return just the relevant 4 tasks,  otoh unclear we need it at all if the original code marked more edges about starts and ends etc

overlord/devicestate/devicestate.go
```diff
@@ -374,11 +382,53 @@ func Remodel(st *state.State, new *asserts.Model) ([]*state.TaskSet, error) {
...
+		// - add wait of each first ts{Download,Install} task for
+		//   the last previous ts{Download,Install} task
+		//   Our chains now looks like:
+		//     download1 (no waits)
+		//     download2 <- verify1
+		//     install1 <- verify1 <- download1
+		//     install2 <- verify2 <- install1
+		tsDownload, tsInstall := splitTsAfterEdge(snapstate.DownloadAndChecksDoneEdge, ts)
+		if prevDownload != nil {
+			tsDownload.Tasks()[0].WaitFor(prevDownload.Tasks()[len(prevDownload.Tasks())-1])
```
pedronis: see my comment,  it seems we could maybe not need all this indexing

overlord/devicestate/devicestate.go
mvo5: Indeed, I reworked the code to not use indexing anymore

overlord/devicestate/devicestate.go
```diff
@@ -293,6 +293,27 @@ func CanManageRefreshes(st *state.State) bool {
 	return false
 }
 
+// extractDownloadInstallEdgesFromTs is a helper that extract the first, last
+// download and install tasks from a TaskSet
```
pedronis: "download phase" and "install phase" perhaps?

overlord/devicestate/devicestate.go
```diff
@@ -293,6 +293,27 @@ func CanManageRefreshes(st *state.State) bool {
...
+		}
+		if firstInst == nil && lastDl != nil {
+			firstInst = t
+		}
+		if t == edgeTask {
+			lastDl = t
+		}
+		if firstInst != nil {
+			lastInst = t
+		}
```
pedronis: firstDl and lastInst  are just  tasks[0] and tasks[-1], no?  it seems slightly silly to use the loop for that

the loop is only needed to find firstInst, which comes after lastDl (which is just edgeTask)

(I still have a  slight preference for doInstall to mark all these but it could be done later, it's not a blocker per se)

overlord/devicestate/devicestate.go
```diff
@@ -320,8 +338,9 @@ func Remodel(st *state.State, new *asserts.Model) ([]*state.TaskSet, error) {
 	if current.Series() != new.Series() {
 		return nil, fmt.Errorf("cannot remodel to different series yet")
 	}
-	// FIXME: we need language in the model assertion to declare what
-	// transitions are ok before we allow remodel like this.
+	// FIXME: we need language in the model assertion to declare
```
pedronis: // TODO: we need dedicated assertion language to permit for model transitions  before we allow that cross vault

overlord/devicestate/devicestate.go
```diff
@@ -380,11 +393,55 @@ func Remodel(st *state.State, new *asserts.Model) ([]*state.TaskSet, error) {
...
+		//     install1 <- verify1 <- download1
+		//     install2 <- verify2 <- download2
+		// - add wait of each first ts{Download,Install} task for
+		//   the last previous ts{Download,Install} task
+		//   Our chains now looks like:
+		//     download1 (no waits)
+		//     download2 <- verify1
+		//     install1 <- verify1 <- download1
+		//     install2 <- verify2 <- install1
+		firstDownload, lastDownload, firstInstall, lastInstall := extractDownloadInstallEdgesFromTs(ts)
```
pedronis: these variable names are a bit confusing,  either

firstOfDowloand    or downloadStart    and   lastOfDownload   or downloadLast   etc

overlord/devicestate/devicestate.go
```diff
@@ -380,11 +393,55 @@ func Remodel(st *state.State, new *asserts.Model) ([]*state.TaskSet, error) {
...
+		//   Out chains look like this:
+		//     install1 <- verify1 <- download1
+		//     install2 <- verify2 <- download2
+		// - add wait of each first ts{Download,Install} task for
+		//   the last previous ts{Download,Install} task
+		//   Our chains now looks like:
+		//     download1 (no waits)
+		//     download2 <- verify1
+		//     install1 <- verify1 <- download1
+		//     install2 <- verify2 <- install1
```
pedronis: I don't understand this comment,  I would expect left to right to be clearer:

download1 <- verify1 <- install1...

That's the style used in the comment below BTW.

I don't understand or I'm not sure the lines from 422 down are corrects.

We should also explicitly start explaining whether   "A <- B"   means  B waits for A

overlord/devicestate/devicestate.go
```diff
@@ -380,11 +393,55 @@ func Remodel(st *state.State, new *asserts.Model) ([]*state.TaskSet, error) {
...
+		prevInstall = lastInstall
+		// update global state
+		lastDownloadInChain = lastDownload
+		if firstInstallInChain == nil {
+			firstInstallInChain = firstInstall
+		}
+	}
+	// Make sure the first install waits for the last download. With this
+	// our (simplified) wait chain looks like:
+	// download1 <- verify1 <- download2 <- verify2 <- install1 <- install2
```
pedronis: this one looks correct.

overlord/state/task.go
```diff
@@ -465,14 +465,24 @@ func (t *Task) At(when time.Time) {
 	}
 }
 
+// TaskSetEdge allows marking tasks inside a TaskSet
```
pedronis: // TaskSetEdge designates tasks inside a TaskSet for outside reference.

overlord/state/task.go
```diff
@@ -501,6 +511,14 @@ func (ts *TaskSet) AddTask(task *Task) {
 	ts.tasks = append(ts.tasks, task)
 }
 
+// MarkEdge marks the given task as a specific edge
```
pedronis: missing period.

overlord/state/task.go
```diff
@@ -465,14 +465,24 @@ func (t *Task) At(when time.Time) {
...
+	edges map[TaskSetEdge]*Task
 }
 
 // NewTaskSet returns a new TaskSet comprising the given tasks.
 func NewTaskSet(tasks ...*Task) *TaskSet {
-	return &TaskSet{tasks}
+	return &TaskSet{tasks, nil}
+}
+
+// Edge returns the task marked with the given TaskSetEdge
```
pedronis: missing period.

overlord/devicestate/devicestate.go
mvo5: Indeed, that is confusing. I updated it now to have a legend at the top and to use the A <- B style (from below) consistently.

overlord/state/task.go
mvo5: There was one more missing period (outside of my PR) that I fixed opportunistly.

overlord/devicestate/devicestate.go
```diff
@@ -294,22 +294,21 @@ func CanManageRefreshes(st *state.State) bool {
...
+	// we know we always start with downloads
+	firstDl = tasks[0]
+	// and always end with installs
+	lastInst = tasks[len(tasks)-1]
+	for _, t := range tasks {
 		if firstInst == nil && lastDl != nil {
 			firstInst = t
 		}
 		if t == edgeTask {
 			lastDl = t
```
pedronis: nitpick: this could be a bool flag, and then we hard code edgeTask in the `return`

overlord/devicestate/devicestate.go
```diff
@@ -410,32 +409,40 @@ func Remodel(st *state.State, new *asserts.Model) ([]*state.TaskSet, error) {
...
+		//     download2 <- verify2 <- install2 (as before)
+		//     download3 <- verify3 <- install3 (as before)
+		//     verify1 <- download2 (added)
+		//     verify2 <- download3 (added)
+		//     install1  <- install2 (added)
+		//     install2  <- install3 (added)
+		downloadStart, downloadLast, installFirst, installLast := extractDownloadInstallEdgesFromTs(ts)
 		if prevDownload != nil {
-			firstDownload.WaitFor(prevDownload)
+			downloadStart.WaitFor(prevDownload)
```
pedronis: mostly a remark, but we don't strictly need to serialize the download, just the install phases,  we could have the first install wait for all the downloadLasts

overlord/devicestate/devicestate.go
```diff
@@ -410,32 +409,40 @@ func Remodel(st *state.State, new *asserts.Model) ([]*state.TaskSet, error) {
...
+		prevInstall = installLast
 		// update global state
-		lastDownloadInChain = lastDownload
+		lastDownloadInChain = downloadLast
 		if firstInstallInChain == nil {
-			firstInstallInChain = firstInstall
+			firstInstallInChain = installFirst
 		}
 	}
 	// Make sure the first install waits for the last download. With this
```
pedronis: should this refer to download3, verify3, etc  for consistency with the previous comments?

overlord/devicestate/devicestate.go
```diff
@@ -293,6 +293,26 @@ func CanManageRefreshes(st *state.State) bool {
 	return false
 }
 
+// extractDownloadInstallEdgesFromTs is a helper that extract the first, last
+// download phase and install phase tasks from a TaskSet
+func extractDownloadInstallEdgesFromTs(ts *state.TaskSet) (firstDl, lastDl, firstInst, lastInst *state.Task) {
+	edgeTask := ts.Edge(snapstate.DownloadAndChecksDoneEdge)
```
zyga: Would it be safer to return nil if the edge is unset or if the ts is empty?

overlord/devicestate/devicestate.go
```diff
@@ -293,6 +293,26 @@ func CanManageRefreshes(st *state.State) bool {
...
+	tasks := ts.Tasks()
+	// we know we always start with downloads
+	firstDl = tasks[0]
+	// and always end with installs
+	lastInst = tasks[len(tasks)-1]
+	for _, t := range tasks {
```
zyga: Now that, I believe, I understand this loop, would it be more direct to code this as follows:
```
tasks := ts.Tasks()
var edgeTaskIndex int
for i, task := range tasks {
  if task == edgeTask {
    edgeTaskIndex = i
    break
  }
}
return tasks[0], tasks[edgeTaskIndex], tasks[edgeTaskIndex+1], tasks[len(tasks)-1]
```

overlord/devicestate/devicestate.go
```diff
@@ -356,22 +374,16 @@ func Remodel(st *state.State, new *asserts.Model) ([]*state.TaskSet, error) {
 	}
 	userID := 0
 
+	// adjust kernel track
 	var tss []*state.TaskSet
-	addTss := func(ts *state.TaskSet) {
```
zyga: Why don't we need this anymore? Specifically the waiting?

overlord/devicestate/devicestate.go
zyga: While I agree with Samuele on this I would remark that unless we do something smart with network congestion (e.g. serialize downloads elsewhere) this will have the positive effect of working effectively on slower links that are easily congested by parallel downloads. In some cases downloading multiple files in parallel will cause downloads to fail whereas the same set processed sequentially would work.

overlord/state/task.go
```diff
@@ -465,14 +465,24 @@ func (t *Task) At(when time.Time) {
...
 }
 
+// TaskSetEdge designates tasks inside a TaskSet for outside reference.
+type TaskSetEdge string
+
 // A TaskSet holds a set of tasks.
 type TaskSet struct {
 	tasks []*Task
+
+	edges map[TaskSetEdge]*Task
```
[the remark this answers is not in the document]
pedronis: I would like to understand what your concerns are?  the name?  he fact we don't check that in MarkEdge that the task belong to the taskset?

I'm ok to switch to a plainer name,  not ok to use a very long name though.

The doc says above:

// TaskSetEdge designates tasks inside a TaskSet for outside reference.

overlord/devicestate/devicestate.go
pedronis: it probably makes sense at some point to introduce some kind of max-parallel-downloads option

overlord/state/task.go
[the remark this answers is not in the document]
pedronis: @mvo5 @zyga  As I mentioned in the standup we shouldn't block the landing of this PR specifically on agreeing on a new name/if a new name is needed here.  The other various comments from the last review passes are valid to consider though.

overlord/devicestate/devicestate.go
mvo5: I add a comment for now, serializing it is nice for testing but I agree its not really needed.

overlord/state/task.go
```diff
@@ -465,14 +465,24 @@ func (t *Task) At(when time.Time) {
...
 }
 
 // NewTaskSet returns a new TaskSet comprising the given tasks.
 func NewTaskSet(tasks ...*Task) *TaskSet {
-	return &TaskSet{tasks}
+	return &TaskSet{tasks, nil}
```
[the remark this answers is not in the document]
mvo5: Yes, `go vet` will complain otherwise.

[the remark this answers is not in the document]
mvo5: Yes, `go vet` will complain otherwise.

overlord/state/task.go
```diff
@@ -465,14 +465,24 @@ func (t *Task) At(when time.Time) {
...
+}
+
+// Edge returns the task marked with the given TaskSetEdge.
```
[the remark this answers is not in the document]
mvo5: I extended TaskSetEdge a bit with some example use.

overlord/devicestate/devicestate.go
mvo5: Sure, this works as well.

zyga:
Thank you for adding the extra error checks and for simplifying the essential loop. I don't have any other remarks that would warrant a blocker at this time.

PR #6580, opened 2019-03-11
cmd/snap-confine: drop unused dependency on libseccomp

zyga:
Historically snap-confine would use libseccomp to compile and attach a
seccomp profile on startup of each new snap application or hook.

For a number or releases however this has not been the case. Instead the
seccomp profile is compiled with snap-seccomp (which does link to
libseccomp) and is stored on disk in /var/lib/snapd/seccomp/bpf.  Since
the change all that snap-confine has do to is to load the compiled
profile from disk and pass it to a system call for attachment. Therefore
the whole dependency as well as the associated autoconf/automake
boilerplate can be discarded.

With this change we do exactly what we did before but in there is one
less variant of snap-confine that may be built that would require
testing.

packaging/ubuntu: stop passing --enable-static-libseccomp

### DIFF
--- a/cmd/Makefile.am
+++ b/cmd/Makefile.am
@@ -224,6 +224,10 @@ snap_confine_snap_confine_SOURCES = \
 	snap-confine/mount-support.h \
 	snap-confine/ns-support.c \
 	snap-confine/ns-support.h \
+	snap-confine/seccomp-support-ext.c \
+	snap-confine/seccomp-support-ext.h \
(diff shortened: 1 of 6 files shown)


mvo5:
We also need to update BUILT_USING_PACKAGES in the packaging

zyga:
@mvo5 since we still bundle libseccomp in snap-seccomp there is nothing more to do there.

PR #6582, opened 2019-03-11
daemon, snap: screenshots _only_ shows the deprecation notice, from 2.39

chipaca:
As per https://forum.snapcraft.io/t/screenshots-are-going-away/8086
`"screenshots"` should no longer be in responses from 2.39.

This keeps the entry, but only with the note (the deprecation note has
been there since 2.36).

2.42 would drop the entry entirely.


snap/info.go
```diff
@@ -753,7 +753,7 @@ type AppInfo struct {
 
 // ScreenshotInfo provides information about a screenshot.
 type ScreenshotInfo struct {
-	URL    string `json:"url"`
+	URL    string `json:"url,omitempty"`
```
robert-ancell: Note this is technically an API break. This isn't an issue in snapd-glib (we treat missing fields as empty) but I don't know if any other clients will care.

robert-ancell:
Still in progress [SRUing the snapd-glib/gnome-software changes](https://bugs.launchpad.net/ubuntu/bionic/+source/gnome-software/+bug/1799614).

chipaca:
@robert-ancell would you rather we kicked this back to 2.40?

robert-ancell:
@chipaca yes please.

chipaca:
Updated the forum topic with this.

chipaca:
@robert-ancell and now?

robert-ancell:
Current status:
19.04 - Supported in gnome-software 3.30.6-2ubuntu2
18.10 - Supported in gnome-software 3.30.2-0ubuntu11 
18.04 LTS - Supported in gnome-software 3.28.1-0ubuntu4.18.04.10, 
16.04 LTS - Supported in gnome-software 3.20.5-0ubuntu0.16.04.12 which is currently completing phased updates (at 90%).
snap-store - Fixed in candidate channel, expected to migrated to stable in the next few days.

In short in a few days this should be available in all supported versions. We can't be sure that people have downloaded the updates, but hopefully they will at the time they update snapd? In terms of other distros we can't make any assumptions, but the master branches all use the media API.

So we're as prepared as we can be :)

PR #6585, opened 2019-03-11
tests: add undo test with hanging stop command

mvo5:
While debugging a hanging change I noticed that we lack some tests
related to the undo handling of bad services. This PR adds two
new tests:
1. A snap where the new snap service is broken and the refresh breaks
   the old snap service as well
2. A snap where the stop command takes forever

### DIFF
--- a/tests/main/refresh-undo-haning-stop-cmd/task.yaml
+++ b/tests/main/refresh-undo-haning-stop-cmd/task.yaml
@@ -1,0 +1,37 @@
+summary: Check that undo works even with a hanging stop command
+
+systems: [-ubuntu-14.04-*]
+
+environment:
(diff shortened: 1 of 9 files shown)


tests/main/refresh-undo-haning-stop-cmd/task.yaml
```diff
@@ -0,0 +1,37 @@
+summary: Check that undo works even with a hanging stop command
```
bboozzoo: There is a typo in the test directory name `s/refresh-undo-haning-stop-cmd/refresh-undo-hanging-stop-cmd/`

mvo5:
This will need some work, so closing it for now (the real reason for the hangs was now identified as systemd PR#8083).

PR #6590, opened 2019-03-12
daemon/api: filter connections with hotplug-gone=true

stolowski:
Filter out connections ("snap connections") that have hotplug-gone flag set - these are connections remembered in the state (but not in the repository), so that they can be restored if same device is plugged back in the future. This aspect was overlooked as critical bits of "snap connections" and hotplug landed at about the same time, so we didn't have proper test. The problem was exposed by #6491.

### DIFF
--- a/daemon/api_connections.go
+++ b/daemon/api_connections.go
@@ -122,7 +122,7 @@ func collectConnections(ifaceMgr *ifacestate.InterfaceManager, filter collectFil
 	connsjson.Slots = make([]*slotJSON, 0, len(ifaces.Slots))
 
 	for crefStr, cstate := range connStates {
-		if cstate.Undesired && filter.connected {
+		if (cstate.Undesired && filter.connected) || cstate.HotplugGone {
 			continue
 		}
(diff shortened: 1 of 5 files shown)


stolowski:
> Looks good, is there a plan to show this somehow (eventually?)

@zyga Yes, we discussed this briefly on the standup, we may want to show them somewhere some day, but it's unclear atm.

daemon/api_connections.go
```diff
@@ -122,7 +122,7 @@ func collectConnections(ifaceMgr *ifacestate.InterfaceManager, filter collectFil
 	connsjson.Slots = make([]*slotJSON, 0, len(ifaces.Slots))
 
 	for crefStr, cstate := range connStates {
-		if cstate.Undesired && filter.connected {
+		if (cstate.Undesired && filter.connected) || cstate.HotplugGone {
```
bboozzoo: I think I'd make this a separate `if` with a note, eg.
```
if cstate.HotplugGone {
    // XXX: hotplug connection for which the device with the slot is gone
    continue
}
```

PR #6595, opened 2019-03-13
tests: add regression test for systemctl race fix

mvo5:
The systemd in UC16 and UC18 has a race that can cause
systemctl to hang forever when operations like "start"
are intereleaved with "systemctl daemon-reload"s.

This got fixed upstream in the systemd PR#8803

    https://github.com/systemd/systemd/pull/8803

And we backported this fix via the SRU

    https://bugs.launchpad.net/bugs/1819728

This PR adds a regression test that ensures the fix
works correctly.


mvo5:
Fwiw, I tested this with:
```
SPREAD_CORE_CHANNEL=beta $GOPATH/bin/spread -shell-after -v  qemu:ubuntu-core-16-64:tests/main/systemctl-race-8803
```
and beta (which does not has the fix) hangs as expected.

mvo5:
Marked blocked as we currently have no fix for UC16 that is stable.

mvo5:
Marked as blocked for now because we really only want to merge this once this once the systemd SRU with the fix is in xenial proper. But useful for testing because we have a patched systemd in our ppa:snappy-dev/image PPA.

mvo5:
The systemd fix went into -updates so this can land now.

tests/regression/lp-1819728/task.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+    echo "Then the system will eventually recover and no systemctl are left"
+    for _ in $(seq 10); do
+        if ! pgrep systemctl; then
+            break
+        fi
+        sleep 6
+    done
+    echo "No systemctl left"
+    ! pgrep systemctl
```
zyga: Perhaps add something like:
```suggestion
    ! pgrep systemctl
debug: |
    ps aux | grep systemctl
```

tests/regression/lp-1819728/task.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+description: |
+    See https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1819728
+    for details about this issue.
+
+# TODO: enable on ubuntu-16.04-64, ubuntu-18.04-64 once the SRU
+#       for systemd is in the "-updates" pocket.
+systems: [ubuntu-core-16-64, ubuntu-core-18-64]
+
+kill-timeout: 2m
```
zyga: @sergiocazzolato suggested to use a 5 minute timeout to avoid issues on slower devices.

```suggestion
kill-timeout: 5m
```

PR #6604, opened 2019-03-15
tests: tests add check to detect a broken snap on reset

sergiocazzolato:
This will help to detect when the reset left a broken snap on core systems.


mvo5:
The shellcheck tests are failing - also the git commits for this PR look a bit stange, easiest might be a branch cleanly from upstream/master again and just cherry-pick e785e00 (or we could squash-merge too).

tests/lib/reset.sh
```diff
@@ -145,6 +145,13 @@ reset_all_snap() {
     if [ "$1" != "--keep-stopped" ]; then
         systemctl start snapd.service snapd.socket
     fi
+
+    # Exit in case there is a snap in broken state after restoring the snapd state
+    if snap list | grep -E "broken$"; then
+        echo "snap in broken state"
+        exit 1
```
stolowski: Do we have enough context if this happens? Should we e.g. list all snaps?

sergiocazzolato: I think that's enough to see which snap is not being reset correctly.

PR #6606, opened 2019-03-15
selinux, systemd: support mount contexts for snap images

bboozzoo:
SELinux policy we ship allows snaps under targeted policy to run on a filesystem that is mounted with `snappy_snap_t` type. This can be applied to snaps by passing a  `context=<selinux-context>` option to mount.

The PR hardcodes the mount context directly, but leaves a TODO note about ways of extending this following the method used by projects using `go-selinux` (eg. `podman`).


systemd/systemd_test.go
```diff
@@ -560,6 +566,38 @@ WantedBy=multi-user.target
...
+	c.Assert(filepath.Join(dirs.SnapServicesDir, mountUnitName), testutil.FileEquals, fmt.Sprintf(`
+[Unit]
+Description=Mount unit for foo, revision 42
+Before=snapd.service
+
+[Mount]
+What=%s
+Where=/snap/snapname/123
+Type=squashfs
+Options=nodev,ro,x-gdu.hide,context=%s
```
stolowski: Is it ok to pass empty `...context=` on darwin (no unexpected errors from mount?)?

systemd/systemd_test.go
```diff
@@ -560,6 +566,38 @@ WantedBy=multi-user.target
...
+
+[Install]
+WantedBy=multi-user.target
+`[1:], mockSnapPath, selinux.SnapMountContext()))
```
stolowski: Would it be possible to explicitily check the expected context string here, or is too annoying because of darwin vs linux build?

systemd/systemd_test.go
bboozzoo: I have no clue. We do not reach this code on darwin FWIW.

selinux/label_linux.go
```diff
@@ -77,3 +77,12 @@ func RestoreContext(aPath string, mode RestoreMode) error {
...
 	return exec.Command("restorecon", args...).Run()
 }
+
+// SnapMountContext finds out the right context for mounting snaps
+func SnapMountContext() string {
+	// TODO: consider reading this from an external configuration file, such
+	// as per app contexts, from
+	// /etc/selinux/targeted/contexts/snapd_contexts like go-selinux and
+	// podman do for container volumes.
+	return "system_u:object_r:snappy_snap_t:s0"
```
zyga: What is the possible evolution path of this string? When/if we wish to change it (e.g. to `snapd_snap_t` instead of `snappy_snap_t`, what would have to happen? Can you please add the answer as a comment here.

systemd/systemd.go
```diff
@@ -548,6 +550,11 @@ func (s *systemd) AddMountUnitFile(snapName, revision, what, where, fstype strin
 		}
 		options = append(options, newOptions...)
 		fstype = newFsType
+		if release.SELinuxLevel() != release.NoSELinux {
+			if mountCtx := selinux.SnapMountContext(); mountCtx != "" {
+				options = append(options, fmt.Sprintf("context=%s", mountCtx))
```
zyga: Nitpick :)

```suggestion
				options = append(options, "context="+mountCtx)
```

selinux/label_linux.go
bboozzoo: This is going to involve quite some changes. First, it's the same problem as now we had with mount units. Those are generated during install. Any later policy changes, that could involve typ change are not tracked by snapd, so we do not know when to regenerate the units. On top of this, remounting a block device with changed security contest, while it's used/mounted already will be stopped by the kernel. That's why we'll need to recommend reboot to transition to the new policy and it'll probably take the same for any future transition.

Conan-Kudo: @zyga It is very likely that the policy prefix will not change for a very long, because of the complexity of the type change. I would probably resist it myself, simply because this policy is not just about `snapd`, it's about the entire snappy system.

zyga: Can this please be added to the code as a code comment.

systemd/systemd_test.go
```diff
@@ -560,6 +565,38 @@ WantedBy=multi-user.target
...
+
+[Mount]
+What=%s
+Where=/snap/snapname/123
+Type=squashfs
+Options=nodev,ro,x-gdu.hide,context=%s
+
+[Install]
+WantedBy=multi-user.target
+`[1:], mockSnapPath, "system_u:object_r:snappy_snap_t:s0"))
```
mvo5: We could include this in the ``` text above, couldn't we? I guess its done this way to highlight the differences to a normal mount unit(?)

systemd/systemd_test.go
mvo5: We only pass "context=" if context is != "" (AFAICT from the code above).

PR #6607, opened 2019-03-15
cmd: typedef mountinfo structures

zyga:
I'm not sure why I didn't initially. Use of typedef helps to avoids
overly long lines whenever mountinfo is used.


zyga:
@jdstrand those changes are due to indent, sadly. I'm happy that over time there will be less indent-based files and more clang-format based ones.

PR #6610, opened 2019-03-15
interfaces/builtin: add add exec "/" to docker-support

anonymouse64:
This access is needed to allow recent versions of runC inside docker
18.06.3 and 18.09.3 to run properly. Note that the snap is not affected
by CVE-2019-5736 directly, but this access is needed so that the
mitigation works properly.

### DIFF
--- a/interfaces/builtin/docker_support.go
+++ b/interfaces/builtin/docker_support.go
@@ -139,6 +139,10 @@ ptrace (read, trace) peer=docker-default,
 
 #cf bug 1502785
 / r,
+
+# needed by runc for mitigation of CVE-2019-5736
+# see apparmor bug 1820344
+/ ix,
 `
 
 const dockerSupportConnectedPlugSecComp = `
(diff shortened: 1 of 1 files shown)


jdstrand:
> +1, agree with jdstrand about the bug report.
> 
> @jdstrand is there a kernel bug tracking this? Should we file one?

With @anonymouse64 filing a bug against apparmor, that is enough for jj to look into and escalate if it is a larger kernel/LSM issue as needed (it might be apparmor-specific, don't know yet).

anonymouse64:
I filed apparmor bug https://bugs.launchpad.net/apparmor/+bug/1820344 for this

interfaces/builtin/docker_support.go
```diff
@@ -139,6 +139,10 @@ ptrace (read, trace) peer=docker-default,
 
 #cf bug 1502785
 / r,
+
+# needed by runc for mitigation of CVE-2019-5736
+# see apparmor bug 1820344
```
zyga:
```suggestion
# For details see https://bugs.launchpad.net/apparmor/+bug/1820344
```

PR #6614, opened 2019-03-16
cmd/snap-confine: use fixed private tmp directory

zyga:
When snap-confine prepares a fresh per-snap mount namespace
it constructs a private /tmp directory that is not shared with other
snaps.

As a part of this work snap-confine would create a randomly named
directory in /tmp/snap.$SNAP_NAME_xxxxx with the x characters replaced
by random unique name.

Those directories would never be recycled and would be re-created when
the mount namespace was to be discarded. This was accumulating data in
/tmp that would not be normally removed unless the machine had rebooted.
In addition whenever multiple temporary directories existed for a single
snap the user would be unable to easily determine which one is the right
one, presenting usability issues for applications that use data in /tmp
for whatever reason (e.g. temporary log files).

This patch changes the path to be fixed, always
/tmp/snap.$SNAP_INSTANCE_NAME/tmp. This opens the avenue towards snapd
managing that directory (e.g. to remove it on snap removal).

There is only one user-visible consequence: discarding the mount
namespace and re-running a snap application no longer gives the
application fresh /tmp directory. It is unlikely anything apart from
snapd test suite will observe this change though.

The test that was checking if the base snap for a given snap has changed
has now been updated not to rely on the re-creation of /tmp.

### DIFF
--- a/cmd/snap-confine/mount-support.c
+++ b/cmd/snap-confine/mount-support.c
@@ -62,47 +62,82 @@
 // TODO: fold this into bootstrap
 static void setup_private_mount(const char *snap_name)
 {
-	char tmpdir[MAX_BUF] = { 0 };
-
-	// Create a 0700 base directory, this is the base dir that is
-	// protected from other users.
-	//
-	// Under that basedir, we put a 1777 /tmp dir that is then bind
-	// mounted for the applications to use
-	sc_must_snprintf(tmpdir, sizeof(tmpdir), "/tmp/snap.%s_XXXXXX", snap_name);
-	if (mkdtemp(tmpdir) == NULL) {
-		die("cannot create temporary directory essential for private /tmp");
-	}
-	// now we create a 1777 /tmp inside our private dir
+	// TODO: remove after switching to global umask handling.
 	mode_t old_mask = umask(0);
-	char *d = sc_strdup(tmpdir);
-	sc_must_snprintf(tmpdir, sizeof(tmpdir), "%s/tmp", d);
-	free(d);
 
-	if (mkdir(tmpdir, 01777) != 0) {
-		die("cannot create temporary directory for private /tmp");
+	// Create a 0700 base directory, this is the base dir that is protected
+	// from other users. This directory name is NOT randomly generated. This
+	// has several properties:
+	//
+	// Users can relate to the name and can find the temporary directory as
+	// visible from within the snap. If this directory was random it would be
+	// harder to find because there may be situations in which multiple
+	// directories related to the same snap name would exist.
+	//
+	// Snapd can partially manage the directory. Specifically on snap remove
+	// snapd could remove the directory and everything in it, potentially
+	// avoiding runaway disk use on a machine that either never reboots or uses
+	// persistent /tmp directory.
+	//
+	char base_dir[MAX_BUF] = { 0 };
+	char tmp_dir[MAX_BUF] = { 0 };
+	int base_dir_fd SC_CLEANUP(sc_cleanup_close) = -1;
+	int tmp_dir_fd SC_CLEANUP(sc_cleanup_close) = -1;
+	sc_must_snprintf(tmp_dir, sizeof(tmp_dir), "/tmp/snap.%s/tmp",
+			 snap_name);
+	sc_must_snprintf(base_dir, sizeof(base_dir), "/tmp/snap.%s", snap_name);
+
+	// Create /tmp/snap.$SNAP_NAME/ 0700 root.root
+	if (mkdir(base_dir, 0700) < 0 && errno != EEXIST) {
+		die("cannot create base directory %s", base_dir);
 	}
-	umask(old_mask);
-
+	base_dir_fd = open(base_dir,
+			   O_RDONLY | O_DIRECTORY | O_CLOEXEC | O_NOFOLLOW);
+	if (base_dir_fd < 0) {
+		die("cannot open base directory %s", base_dir);
+	}
+	if (fchmod(base_dir_fd, 0700) < 0) {
+		die("cannot chmod base directory %s to 0700", base_dir);
+	}
+	if (fchown(base_dir_fd, 0, 0) < 0) {
+		die("cannot chown base directory %s to root.root", base_dir);
+	}
+	// Create /tmp/snap.$SNAP_NAME/tmp 01777 root.root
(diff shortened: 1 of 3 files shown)


cmd/snap-confine/mount-support.c
```diff
@@ -62,47 +62,82 @@
...
-	// MS_PRIVATE needs linux > 2.6.11
+	}
+	if (chdir("/") < 0) {
+		die("cannot change directory to \"/\"");
+	}
+	sc_do_mount(tmp_dir, "/tmp", NULL, MS_BIND, NULL);
 	sc_do_mount("none", "/tmp", NULL, MS_PRIVATE, NULL);
-	// do the chown after the bind mount to avoid potential shenanigans
-	if (chown("/tmp/", 0, 0) < 0) {
-		die("cannot change ownership of /tmp");
```
jdstrand: FYI, the chown was done after the mounts due to @tyhicks recommendation as part of 3f40f354757e6cb9dcfa37db57090796f4bdb6c2. It seems the history of the PR that introduced that is lost, but the idea is that before bdbfeebef03245176ae0dc323392bb0522a339b1, these directories could be created as non-root depending on the invoking user and moving the chown after making the dir private guarded against non-root doing things to the dir prior to the bind mount (and thus making the snap's /tmp have weird perms thereafter).

Today, we no longer chown to non-root so there is no potential for non-root to change the tmpdir to something else, so moving this chown earlier (and dropping the shenanigans comment) is safe.

cmd/snap-confine/mount-support.c
```diff
@@ -62,47 +62,82 @@
...
 {
-	char tmpdir[MAX_BUF] = { 0 };
+	// TODO: remove after switching to global umask handling.
+	mode_t old_mask = umask(0);
 
-	// Create a 0700 base directory, this is the base dir that is
-	// protected from other users.
+	// Create a 0700 base directory, this is the base dir that is protected
+	// from other users. This directory name is NOT randomly generated. This
+	// has several properties:
```
jdstrand: This comment is incomplete. You both create a 0700 basedir and a 1777 tmpdir. Also, on every invocation you unconditionally chmod and chown them (which is fine in terms of security).

cmd/snap-confine/mount-support.c
```diff
@@ -62,47 +62,82 @@
...
-	sc_must_snprintf(tmpdir, sizeof(tmpdir), "/tmp/snap.%s_XXXXXX", snap_name);
-	if (mkdtemp(tmpdir) == NULL) {
-		die("cannot create temporary directory essential for private /tmp");
+	char base_dir[MAX_BUF] = { 0 };
+	char tmp_dir[MAX_BUF] = { 0 };
+	int base_dir_fd SC_CLEANUP(sc_cleanup_close) = -1;
+	int tmp_dir_fd SC_CLEANUP(sc_cleanup_close) = -1;
+	sc_must_snprintf(tmp_dir, sizeof(tmp_dir), "/tmp/snap.%s/tmp",
+			 snap_name);
+	sc_must_snprintf(base_dir, sizeof(base_dir), "/tmp/snap.%s", snap_name);
```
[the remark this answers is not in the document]
zyga: If it wasn't for the bind mount I wanted to avoid the `tmp_dir` altogether. I will look at simplifying this more.

cmd/snap-confine/mount-support.c
```diff
@@ -62,47 +62,82 @@
...
+			   O_RDONLY | O_DIRECTORY | O_CLOEXEC | O_NOFOLLOW);
+	if (base_dir_fd < 0) {
+		die("cannot open base directory %s", base_dir);
 	}
-	umask(old_mask);
-
+	if (fchmod(base_dir_fd, 0700) < 0) {
+		die("cannot chmod base directory %s to 0700", base_dir);
+	}
+	if (fchown(base_dir_fd, 0, 0) < 0) {
```
mgerstner: Wouldn't it be more careful to abort execution if the `base_dir` is not owned by `root:root` in the first place? Since the directory now has a fixed name an attacker can prepare dirs there that already contain prepared content, or even mounted file systems (like FUSE).

Also there is a race condition between the `open()` and the `fchmod()` / `fchown()`. If the directory already exists and is owned by a regular user then this regular user might be able to rename the directory before ownership is actually transferred to root.

zyga: To mount a filesystem the attacker would have to be root and we are not trying to defend against that.

As for the race condition, indeed it might be better to abort if the directory, as opened, is owned by non-root user. I will look at a way to avoid the race condition if I can and if not, just give up early as you suggested.

jdstrand: Re the race, this isn't a problem because the parent directory is owned by root, not the user, so normal DAC prevents a non-root users from renaming while we've still got the fd. We aren't protecting from unconfined root cause if you have that you can just replace snapd.

mgerstner: > Re the race, this isn't a problem because the parent directory is owned by root, not the user, so normal DAC prevents a non-root users from renaming while we've still got the fd. We aren't protecting from unconfined root cause if you have that you can just replace snapd.

I beg to differ. `base_dir` is `/tmp/snap.%s` and `base_dir_fd` refers to that directory. So we can have the following situation:

<pre>
attacker $ mkdir /tmp/snap.mysnap
# snap-confine opens /tmp/snap.mysnap as `base_dir_fd`
attacker $ mv /tmp/snap.mysnap /tmp/my.special.dir
# snap-confine calls fchown() on `base_dir_fd`
</pre>

So as a result `/tmp/my.special.dir` would now be owned by root. This is giving an attacker the possibility to bring arbitrarily named directories into existence that are owned by root (could even be moved to a home directory if not located on a different file system).

Also the actual `/tmp/snap.mysnap` will not be the expected root-owned directory, leading to unspecified results in follow up logic in snap-confine.

zyga: I will be changing this in a follow-up fix today.

jdstrand: > I beg to differ. `base_dir` is `/tmp/snap.%s` and `base_dir_fd` refers to that directory.

Yes, but I was thinking this happened after we shared the mount namespace, where in the parent dir, /tmp, is not 1777. I may have been wrong in thinking that, but it doesn't matter, @zyga is looking to address this.

zyga: This took a while to be something I can propose: https://github.com/snapcore/snapd/pull/6714

PR #6615, opened 2019-03-18
spread: some debug info to catch /var/snap/.. not being empty

bboozzoo:
The following problem seems to come up quite often:
```
+ snap info test-snapd-tools
+ snap remove test-snapd-tools
- Remove data for snap "test-snapd-tools" (x1) (failed to remove snap "test-snapd-tools" base directory: remove /var/snap/test-snapd-tools: directory not empty)
```
Let's try to collect some debug information.


bboozzoo:
Can we close it now?

chipaca:
Yes.

PR #6617, opened 2019-03-18
cmd/snap: fix regression of snap saved command

stolowski:
The recent fix for more readable error messages if set id argument is ommited for snapshot commands caused unexpected regression in `snap saved` command - it errors out if set ID is omitted (previously, before the fix, we would get integer set id = 0 if omitted, which means all sets). This PR fixes it.

### DIFF
--- a/tests/main/snapshot-basic/task.yaml
+++ b/tests/main/snapshot-basic/task.yaml
@@ -17,8 +17,8 @@ execute: |
     SET_ID=$( snap save test-snapd-tools test-snapd-rsync | cut -d\  -f1 | tail -n1 )
 
     # check it includes both snaps
-    snap saved | grep test-snapd-tools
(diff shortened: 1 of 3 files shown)


tests/main/snapshot-basic/task.yaml
```diff
@@ -17,6 +17,8 @@ execute: |
     SET_ID=$( snap save test-snapd-tools test-snapd-rsync | cut -d\  -f1 | tail -n1 )
 
     # check it includes both snaps
+    snap saved | grep test-snapd-tools
```
zyga: Why not `MATCHES`?

stolowski: Right.. Because I've just copied and tweaked the existing checks of this test. Updated, but I'm not touching existing greps in this test to keep the diff at minimum.

PR #6618, opened 2019-03-19
tests: validates snapd from ppa

sergiocazzolato:
This initially is being used to validate snapd from ppa:ubuntu-security-proposed/ppa


spread.yaml
```diff
@@ -42,6 +42,7 @@ environment:
     NO_PROXY: "127.0.0.1"
     NEW_CORE_CHANNEL: "$(HOST: echo $SPREAD_NEW_CORE_CHANNEL)"
     SRU_VALIDATION: "$(HOST: echo ${SPREAD_SRU_VALIDATION:-0})"
+    PPA_VALIDATION: "$(HOST: echo ${SPREAD_PPA_VALIDATION:-})"
```
zyga: SRU_VALIDATION defaults to 0 and seems to be a boolean. PPA_VALIDATION is named alike but defaults to an empty string and seems to be used as a PPA name. I think this warrants a different name. Perhaps `PPA_VALIDATION_NAME`?

In addition, can you please add a comment about the purpose of the variable here (in the yaml).

sergiocazzolato: Done, thanks for reviewing!!!

tests/lib/pkgdb.sh
```diff
@@ -438,6 +438,13 @@ distro_install_build_snapd(){
         if [[ "$SPREAD_SYSTEM" = ubuntu-14.04-* ]] && [ "$SPREAD_REBOOT" = 0 ]; then
             REBOOT
         fi
+    elif [ -n "$PPA_VALIDATION_NAME" ]; then
+        apt install -y snapd
+        add-apt-repository -y "$PPA_VALIDATION_NAME"
+        apt update
+        apt install -y --only-upgrade snapd
+        add-apt-repository --remove "$PPA_VALIDATION_NAME"
+        apt update
```
mvo5: This is not strictly needed, apt will use its sources.list{,.d}. But its still fine to have it here.

tests/lib/pkgdb.sh
```diff
@@ -438,6 +438,13 @@ distro_install_build_snapd(){
         if [[ "$SPREAD_SYSTEM" = ubuntu-14.04-* ]] && [ "$SPREAD_REBOOT" = 0 ]; then
             REBOOT
         fi
+    elif [ -n "$PPA_VALIDATION_NAME" ]; then
+        apt install -y snapd
+        add-apt-repository -y "$PPA_VALIDATION_NAME"
+        apt update
+        apt install -y --only-upgrade snapd
```
mvo5: If the PPA does not have an updated snapd this call will not error and we only test the distro snapd:
```
$ sudo apt install -y --only-upgrade snapd ; echo $?
Reading package lists... Done
Building dependency tree       
Reading state information... Done
snapd is already the newest version (2.37.4+18.10.1).
0 upgraded, 0 newly installed, 0 to remove and 221 not upgraded.
0
```
Adding something like:
```
apt show snapd | grep "APT-Sources: http.*ppa.launchpad.net"
```
after the install will double check that it really comes from the PPA.

sergiocazzolato: done

tests/lib/pkgdb.sh
sergiocazzolato: Nice, Thanks for the review!!

tests/lib/pkgdb.sh
```diff
@@ -438,6 +438,16 @@ distro_install_build_snapd(){
...
+        add-apt-repository --remove "$PPA_VALIDATION_NAME"
+        apt update
+
+        # Double check that it really comes from the PPA
+        apt show snapd | grep "APT-Sources: http.*ppa.launchpad.net"
```
chipaca: you could make it a `MATCH` instead of a `grep` to help debug it if it breaks :-)

PR #6620, opened 2019-03-19
tests/main/remodel: clean up before reverting the state

chipaca:
While it's cleanup, it's also things that should work, so I've added
them to the test itself and not to the restore section.

### DIFF
--- a/tests/main/remodel/task.yaml
+++ b/tests/main/remodel/task.yaml
@@ -69,3 +69,12 @@ execute: |
     # FIXME: check that this is the active model assertion
     echo "and we got the new model assertion"
     snap debug model|MATCH "revision: 2"
+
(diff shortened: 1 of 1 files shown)


tests/main/remodel/task.yaml
```diff
@@ -69,3 +69,12 @@ execute: |
     # FIXME: check that this is the active model assertion
     echo "and we got the new model assertion"
     snap debug model|MATCH "revision: 2"
+
+    # FIXME: this fails (meaning, the remove does not fail)
+    # echo "We can't remove a required snap
+    # ! snap remove test-snapd-tools
+
+    # when the above no longer fails, this will fail, so you'll have
```
mvo5: Maybe this can mention "so we have to remodel or cleanup manually in restore to remove from the required snaps" or something?

PR #6626, opened 2019-03-20
strutil: make SplitUnit public, allow negative numbers

bboozzoo:
Make the helper public so that it can be reused in parsing of gadget size/offset values that have the following format: `<bytes> | <bytes/2^20>M | <bytes/2^30>G`.

### DIFF
--- a/strutil/strutil.go
+++ b/strutil/strutil.go
@@ -114,26 +114,27 @@ func TruncateOutput(data []byte, maxLines, maxBytes int) []byte {
 	return data
 }
 
-// splitUnit takes a string of the form "123unit" and splits
+// SplitUnit takes a string of the form "123unit" and splits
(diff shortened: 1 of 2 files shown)


strutil/strutil.go
```diff
@@ -114,26 +114,27 @@ func TruncateOutput(data []byte, maxLines, maxBytes int) []byte {
...
-	var nonDigit int
+	lastDigit := -1
 	for i, c := range inp {
-		if !unicode.IsDigit(c) {
-			number, err = strconv.ParseInt(inp[0:i], 10, 64)
-			if err != nil {
-				return 0, "", err
-			}
-			nonDigit = i
+		if unicode.IsDigit(c) || c == '-' {
```
zyga: As discussed on IRC, this will catch various interesting runes, such as `৬`.

PR #6628, opened 2019-03-20
overlord/snapshotstate: support auto flag (automatic snapshots 1/N)

stolowski:
Introduce`auto` flag for snapshots, it will be set by automatic snapshots in the followup PRs. On the client side, the "auto" flag will be shown in Notes section of `snap saved` output.


overlord/snapshotstate/backend/backend.go
```diff
@@ -152,7 +152,7 @@ func Filename(snapshot *client.Snapshot) string {
 }
 
 // Save a snapshot
-func Save(ctx context.Context, id uint64, si *snap.Info, cfg map[string]interface{}, usernames []string) (*client.Snapshot, error) {
+func Save(ctx context.Context, id uint64, si *snap.Info, cfg map[string]interface{}, usernames []string, auto bool) (*client.Snapshot, error) {
```
mvo5: I dislike bools in APIs because they don't tell about the "why", e.g. in something like "Save(ctx, uid, si, nil, false)" its not really obvious what the "false" (or "true") actually mean. I would prefer a flag or a typed boolean maybe? 

I would prefer something like:
```
type Flags struct {
    Auto bool
}
 func Save(ctx context.Context, id uint64, si *snap.Info, cfg map[string]interface{}, usernames []string, flags *Flags) (*client.Snapshot, error) {
```
And "nil" would mean empty flags (just like in e.g. snapstate or devicestate).

Or something slightly more unusual like:
```
type AutoFlag bool

const Auto AutoFlag = true
const NoAuto AutoFlag = false
func Save(ctx context.Context, id uint64, si *snap.Info, cfg map[string]interface{}, usernames []string, auto AutoFlag) (*client.Snapshot, error) {
```
I leave this to you and Samuele :) We could also keep it but I think its an easy win if we change it.

stolowski: Thanks, this is a good idea. I like Flags struct more because it's extensible (if we ever need that) and it's consistent with what we do in ifacemgr with connect/disconnect methods and their auto flags. Done.

PR #6631, opened 2019-03-20
tests: split travis spread execution in 2 jobs for ubuntu and non ubuntu systems

sergiocazzolato:
It also includes to run together the main suite with the upgrade suite
to improve the jobs time and avoid lack of jobs.

In case of lack of jobs on travis, we can add more vms to reduce the execution time.

### DIFF
--- a/client/packages.go
+++ b/client/packages.go
@@ -117,12 +117,13 @@ type ResultInfo struct {
 // - Private: return snaps that are private
 // - Query: only return snaps that match the query string
 type FindOptions struct {
-	Refresh bool
-	Private bool
-	Prefix  bool
-	Query   string
-	Section string
-	Scope   string
+	Query    string
+	CommonID string
+	Section  string
+	Scope    string
+	Private  bool
+	Prefix   bool
+	Refresh  bool
 }
 
 var ErrNoSnapsInstalled = errors.New("no snaps installed")
@@ -179,8 +180,14 @@ func (client *Client) Find(opts *FindOptions) ([]*Snap, *ResultInfo, error) {
 	if opts.Prefix {
 		q.Set("name", opts.Query+"*")
 	} else {
-		q.Set("q", opts.Query)
+		if opts.CommonID != "" {
+			q.Set("common-id", opts.CommonID)
+		}
+		if opts.Query != "" {
+			q.Set("q", opts.Query)
+		}
 	}
+
 	switch {
 	case opts.Refresh && opts.Private:
 		return nil, nil, fmt.Errorf("cannot specify refresh and private together")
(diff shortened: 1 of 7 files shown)


run-checks
```diff
@@ -55,6 +55,12 @@ case "${1:-all}" in
     --spread)
         SPREAD=1
```
mvo5: Would it make sense to have `SPREAD=full` here?

run-checks
```diff
@@ -55,6 +55,12 @@ case "${1:-all}" in
...
         ;;
+    --spread-ubuntu)
+        SPREAD=2
```
mvo5: And SPREAD=ubuntu-only here?

run-checks
```diff
@@ -55,6 +55,12 @@ case "${1:-all}" in
...
+        ;;
+    --spread-no-ubuntu)
+        SPREAD=3
```
mvo5: And SPREAD=no-ubuntu here? This way the checks below will convey a bit more context.

run-checks
zyga: I agree with mvo. There's no reason to use integers for this.

mvo5:
There is something funny going on - from the log:
```
./run-checks: 267: [: Illegal number: ubuntu-only
```
whats strange about this is that the script does not exit with an error :/

run-checks
```diff
@@ -258,14 +264,23 @@ if [ "$UNIT" = 1 ]; then
...
-if [ "$SPREAD" = 1 ]; then
+if [ "$SPREAD" -gt 0 ]; then
     TMP_SPREAD="$(mktemp -d)"
     addtrap "rm -rf \"$TMP_SPREAD\""
 
     export PATH=$TMP_SPREAD:$PATH
     ( cd "$TMP_SPREAD" && curl -s -O https://niemeyer.s3.amazonaws.com/spread-amd64.tar.gz && tar xzvf spread-amd64.tar.gz )
 
-    spread google: google-upgrade:tests/upgrade/
+    if [ "$SPREAD" = full ]; then
```
zyga: Perhaps this should be a switch statement?

sergiocazzolato: Yes, thanks for the review!!

spread.yaml
```diff
@@ -687,7 +657,6 @@ suites:
         systems: [-ubuntu-core-*, -opensuse-*]
         # autopkgtest runs against localhost which causes problems with
```
mvo5: I think we also need to remove this comment now - autopkgtest does no longer run the "upgrade" test.

sergiocazzolato: done

PR #6634, opened 2019-03-21
snap: add validation of gadget.yaml

bboozzoo:
Add validation of gadget definition based on what ubuntu-image does. Try to match error message where possible.

The validation only considers the structure of gadget.yaml, internal references and sanity of defined entries.

Things like validation of structure overlap, offset sizes are best done in a separate step, when the structure is 'resolved' for writing. Meaning, that the information from gadget.yaml is conbined with the contents of the gadget snap (eg. when `size` is left unspecified) and platform requirements (eg. default sector size). The `ubuntu-image` tries to do that upfront, but makes some assumptions along the way.

The PR contains #6609


zyga:
Related to this work: what is the future interaction with ubuntu-image? I would expect that there is only one implementation of certain things. Will ubuntu-image depend on "snap" to pack/validate gadgets? Will we have to live with two implementations?

bboozzoo:
As discussed, we'd like to eventually grow the functionality to be able to assemble the image ourselves. At which time ubuntu-image could be updated to use `snap`.

zyga:
@bboozzoo that is a good plan. Thanks

snap/gadget.go
```diff
@@ -249,3 +490,101 @@ func (e *editionNumber) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	*s, err = ParseGadgetSize(gs)
+	if err != nil {
+		return fmt.Errorf("cannot parse size %q: %v", gs, err)
+	}
+	return err
+}
+
+// ParseGadgetSize parses a string expressing size in gadget declaration. The
+// accepted format is one of: <bytes> | <bytes/2^20>M | <bytes/2^30>G.
+func ParseGadgetSize(gs string) (GadgetSize, error) {
```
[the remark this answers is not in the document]
bboozzoo: Yeah, I wanted to use the helper we have, but the formats accepted in gadget.yaml are only these 3:
- `<number>` (bytes)
- `<number>M`
- `<number>G`

snap/gadget.go
```diff
@@ -214,7 +231,11 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
 
 	// basic validation
 	var bootloadersFound int
-	for _, v := range gi.Volumes {
+	for name, v := range gi.Volumes {
+		if err := validateVolume(name, &v); err != nil {
+			return nil, fmt.Errorf("invalid volume %q: %v", name, err)
```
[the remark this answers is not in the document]
mvo5: Ha! I was thinking exactly the same and then wondered if we have a pattern and then wondered if go has a pattern and then got lost in the weeds. But I still wonder if there is a pattern for this in the go community in cases like this where it seems like both choices make sense. Maybe @pedronis has an opinion/thoughts?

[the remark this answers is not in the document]
zyga: There's this https://github.com/pkg/errors that has very nicely defined `Wrap` function: https://github.com/pkg/errors/blob/master/errors.go#L181 for the case of taking one error and wrapping it with some more messages.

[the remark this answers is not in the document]
bboozzoo: You can have multiple volumes in gadget.yaml. Returning error as is would imply the `validate*` has to include all necessary information to be able to locate the invalid entity.

FWIW we have a similar problem with code in `snap/validate.go`.

snap/gadget.go
```diff
@@ -25,10 +25,19 @@ import (
...
 	"strconv"
 	"strings"
 
 	"gopkg.in/yaml.v2"
+
+	"github.com/snapcore/snapd/strutil"
+)
+
+var (
+	validVolumeName = regexp.MustCompile("^[a-z-]+$")
```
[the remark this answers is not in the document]
bboozzoo: Looks like ubuntu-image is bad too, this is regexp they use: `^[-a-zA-Z0-9]+$`, and this is what the docs page states: `[a-z-]+`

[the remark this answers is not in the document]
bboozzoo: Updated to match what ubuntu-image does.

snap/gadget.go
```diff
@@ -249,3 +490,101 @@ func (e *editionNumber) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+// named structure.
+type GadgetRelativeOffset struct {
+	RelativeTo string
+	Offset     GadgetSize
+}
+
+func ParseGadgetRelativeOffset(grs string) (*GadgetRelativeOffset, error) {
+	toWhat := ""
+	sizeSpec := grs
+	idx := strings.IndexRune(grs, '+')
```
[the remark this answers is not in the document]
bboozzoo: Did the change locally, looks very much the same.

snap/gadget.go
```diff
@@ -249,3 +490,101 @@ func (e *editionNumber) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	*e = editionNumber(u)
 	return nil
 }
+
+// GadgetSize describes the size of a structure item or an offset within the
+// structure.
+type GadgetSize uint64
```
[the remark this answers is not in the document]
bboozzoo: It's cluttering the `snap` namespace a bit now, but once I introduce the `gadget` package, it would become `gadget.Size`, where it could probably stay exported. What do you think?

snap/gadget.go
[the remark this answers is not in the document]
pedronis: I think this if fine, we just need to be careful not to use "volume" in the errors below too unnecessarily

snap/gadget.go
```diff
@@ -234,6 +256,225 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+func indexAndName(idx int, name string) string {
+	if name != "" {
+		return fmt.Sprintf("#%v (%q)", idx, name)
+	}
+	return fmt.Sprintf("#%v", idx)
+}
+
+func validateVolume(name string, vol *GadgetVolume) error {
+	if !validVolumeName.MatchString(name) {
+		return errors.New("invalid volume name")
```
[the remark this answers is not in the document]
pedronis: do we need to mention "volume" in that error?

snap/gadget.go
```diff
@@ -234,6 +256,221 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
 	return &gi, nil
 }
 
+func fmtIndexAndName(idx int, name string) string {
+	if name != "" {
+		return fmt.Sprintf("#%v (%q)", idx, name)
+	}
+	return fmt.Sprintf("#%v", idx)
```
pedronis: this path seems not to be exercised in the tests

snap/gadget.go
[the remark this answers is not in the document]
pedronis: it's fine (in that world)

snap/gadget.go
```diff
@@ -234,6 +256,221 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+
+func validateStructureUpdate(up *VolumeUpdate, vs *VolumeStructure) error {
+	if vs.IsBare() && len(vs.Update.Preserve) > 0 {
+		return errors.New("preserving files during update is not supported for non-filesystem structures")
+	}
+
+	names := make(map[string]bool, len(vs.Update.Preserve))
+	for _, n := range vs.Update.Preserve {
+		if names[n] {
+			return fmt.Errorf("duplicate preserve entry %q", n)
```
pedronis: this reads strangely,  maybe quotes help:

   duplicate "preserve" entry %q

snap/gadget.go
```diff
@@ -234,6 +256,221 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+		return fmt.Errorf("GUID structure ID %q with non-GPT schema %q", s, vol.Schema)
+	}
+	if schema != "mbr" && isMBR {
+		return fmt.Errorf("MBR structure ID %q with non-MBR schema %q", s, vol.Schema)
+	}
+
+	return nil
+}
+
+func validateRole(vs *VolumeStructure, vol *GadgetVolume) error {
```
pedronis: this should go after validateVolumeStructure

snap/gadget.go
```diff
@@ -234,6 +256,221 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+			if c.OffsetWrite.RelativeTo != "" && !structureNames[c.OffsetWrite.RelativeTo] {
+				return fmt.Errorf("structure %v, content %v refers to an unknown structure %q",
+					fmtIndexAndName(idx, s.Name), fmtIndexAndName(cidx, c.Image), c.OffsetWrite.RelativeTo)
+			}
+		}
+	}
+	return nil
+}
+
+func validateStructureType(s string, vol *GadgetVolume) error {
```
pedronis: this should go after validateVolumeStructure

snap/gadget.go
```diff
@@ -234,6 +256,221 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+		return fmt.Errorf("MBR structure ID %q with non-MBR schema %q", s, vol.Schema)
+	}
+
+	return nil
+}
+
+func validateRole(vs *VolumeStructure, vol *GadgetVolume) error {
+	if vs.Type == "bare" {
+		if vs.Role != "" && vs.Role != "mbr" {
+			return fmt.Errorf("confclicting role/type: %q/%q", vs.Role, vs.Type)
```
pedronis: typo "conflicting", also this error seems not to be exercised

snap/gadget.go
```diff
@@ -234,6 +256,221 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+	switch vsRole {
+	case "system-data":
+		if vs.Label != "" && vs.Label != "writable" {
+			return fmt.Errorf(`role %q must have an implicit label or "writable", not %q`, vs.Role, vs.Label)
+		}
+	case "mbr":
+		if vs.Size > 446 {
+			return errors.New("mbr structures cannot be larger than 446 bytes")
+		}
+		if vs.Offset != 0 {
```
pedronis: do I understand correctly that OffsetWrite can be set in this case?

snap/gadget.go
```diff
@@ -234,6 +256,221 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+	}
+	return nil
+}
+
+func validateStructureType(s string, vol *GadgetVolume) error {
+	// Type can be one of:
+	// - "mbr" (backwards compatible)
+	// - "bare"
+	// - [0-9A-Z]{2} - MBR type
+	// - hybrid ID
```
pedronis: this list is missing just UUID for the GPT case?

snap/gadget.go
```diff
@@ -234,6 +256,221 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+		code := s[:idx]
+		guid := s[idx+1:]
+		if len(code) != 2 || len(guid) != 36 || !validTypeID.MatchString(code) || !validGUUID.MatchString(guid) {
+			return fmt.Errorf("invalid format of hybrid type ID")
+		}
+	}
+
+	if schema != "gpt" && isGPT {
+		// type: <uuid> is only valid for GPT volumes
+		return fmt.Errorf("GUID structure ID %q with non-GPT schema %q", s, vol.Schema)
```
pedronis: this error is not exercised

snap/gadget.go
```diff
@@ -234,6 +256,221 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+		return fmt.Errorf("invalid volume schema %q", vol.Schema)
+	}
+
+	structureNames := make(map[string]bool, len(vol.Structure))
+	for idx, s := range vol.Structure {
+		if err := validateVolumeStructure(&s, vol); err != nil {
+			return fmt.Errorf("invalid structure %s: %v", fmtIndexAndName(idx, s.Name), err)
+		}
+		if s.Name != "" {
+			structureNames[s.Name] = true
```
pedronis: can structures repeat the same name?

snap/gadget.go
```diff
@@ -234,6 +256,225 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+		return fmt.Errorf("invalid role %q: %v", vs.Role, err)
+	}
+	if vs.Filesystem != "" && !strutil.ListContains([]string{"ext4", "vfat", "none"}, vs.Filesystem) {
+		return fmt.Errorf("invalid filesystem %q", vs.Filesystem)
+	}
+
+	bare := vs.IsBare()
+	for i, c := range vs.Content {
+		var err error
+		if bare {
```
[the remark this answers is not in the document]
pedronis: but we could simply  set a checker outside and not have the if here:

    var contentChecker func(*VolumeContent) error

snap/gadget.go
```diff
@@ -234,6 +256,221 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+		return errors.New("invalid volume name")
+	}
+	if vol.Schema != "" && vol.Schema != "gpt" && vol.Schema != "mbr" {
+		return fmt.Errorf("invalid volume schema %q", vol.Schema)
+	}
+
+	structureNames := make(map[string]bool, len(vol.Structure))
+	for idx, s := range vol.Structure {
+		if err := validateVolumeStructure(&s, vol); err != nil {
+			return fmt.Errorf("invalid structure %s: %v", fmtIndexAndName(idx, s.Name), err)
```
pedronis: seems tests never get to this error

snap/gadget.go
```diff
@@ -234,6 +256,221 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+	// Hybrid ID is 2 hex digits of MBR type, followed by 36 GUUID
+	// example: EF,C12A7328-F81F-11D2-BA4B-00A0C93EC93B
+
+	schema := vol.Schema
+	if schema == "" {
+		schema = "gpt"
+	}
+
+	if s == "" {
+		return errors.New(`type is not specified`)
```
pedronis: this error is not exercised

snap/gadget.go
```diff
@@ -234,6 +256,221 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+		// noop
+	default:
+		return fmt.Errorf("invalid role %q", vs.Role)
+	}
+	return nil
+}
+
+func validateVolumeStructure(vs *VolumeStructure, vol *GadgetVolume) error {
+	if err := validateStructureType(vs.Type, vol); err != nil {
+		return fmt.Errorf("invalid type %q: %v", vs.Type, err)
```
pedronis: tests don't get to this error

snap/gadget.go
```diff
@@ -234,6 +256,221 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+	}
+	if err := validateRole(vs, vol); err != nil {
+		return fmt.Errorf("invalid role %q: %v", vs.Role, err)
```
pedronis: tests don't get to this error

snap/gadget.go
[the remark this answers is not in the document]
bboozzoo: Updated. The error just below needs the same treatment.

snap/gadget.go
bboozzoo: I don't see anything in ubuntu-image that would prevent that. OTOH, that wouldn't be particularly useful either.

snap/gadget.go
bboozzoo: Another piece that AFAIK is not handled by ubuntu-image. Nonetheless, we should check for it.

bboozzoo: Opened a PR to ubuntu-image with the fix too.

snap/gadget.go
```diff
@@ -52,25 +61,34 @@ type GadgetVolume struct {
...
-	Label       string          `yaml:"filesystem-label"`
-	Offset      string          `yaml:"offset"`
-	OffsetWrite string          `yaml:"offset-write"`
-	Size        string          `yaml:"size"`
-	Type        string          `yaml:"type"`
-	ID          string          `yaml:"id"`
-	Filesystem  string          `yaml:"filesystem"`
-	Content     []VolumeContent `yaml:"content"`
-	Update      VolumeUpdate    `yaml:"update"`
+	Name        string               `yaml:"name"`
```
zyga: What is the difference between the label and the name?

bboozzoo: Label is just a filesystem label. Name written as partition name if GPT schema is in use, otherwise it's informative.

snap/gadget.go
```diff
@@ -234,6 +256,231 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
 }
 
+func fmtIndexAndName(idx int, name string) string {
+	if name != "" {
+		return fmt.Sprintf("#%v (%q)", idx, name)
+	}
+	return fmt.Sprintf("#%v", idx)
+}
+
+func validateVolume(name string, vol *GadgetVolume) error {
```
zyga: +1

snap/gadget.go
```diff
@@ -234,6 +256,231 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+			if c.OffsetWrite.RelativeTo != "" && !structureNames[c.OffsetWrite.RelativeTo] {
+				return fmt.Errorf("structure %v, content %v refers to an unknown structure %q",
+					fmtIndexAndName(idx, s.Name), fmtIndexAndName(cidx, c.Image), c.OffsetWrite.RelativeTo)
+			}
+		}
+	}
+	return nil
+}
+
+func validateVolumeStructure(vs *VolumeStructure, vol *GadgetVolume) error {
```
zyga: +1

snap/gadget.go
```diff
@@ -234,6 +256,231 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+	return nil
+}
+
+func validateStructureType(s string, vol *GadgetVolume) error {
+	// Type can be one of:
+	// - "mbr" (backwards compatible)
+	// - "bare"
+	// - [0-9A-Z]{2} - MBR type
+	// - GPT UUID
+	// - hybrid ID
```
zyga: Perhaps add a wikipedia reference to explain what hybrid IDs are for.

snap/gadget.go
```diff
@@ -234,6 +256,231 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+		return err
+	}
+
+	// TODO: validate structure size against sector-size; ubuntu-image uses
+	// a tmp file to find out the default sector size of the device the tmp
+	// file is created on
+	return nil
+}
+
+func validateStructureType(s string, vol *GadgetVolume) error {
```
zyga: +1

snap/gadget.go
```diff
@@ -234,6 +256,231 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+		return fmt.Errorf("MBR structure type with non-MBR schema %q", vol.Schema)
+	}
+
+	return nil
+}
+
+func validateRole(vs *VolumeStructure, vol *GadgetVolume) error {
+	if vs.Type == "bare" {
+		if vs.Role != "" && vs.Role != "mbr" {
+			return fmt.Errorf("conflicting type: %q", vs.Type)
```
zyga: Should this tell the user about the role?

snap/gadget.go
```diff
@@ -234,6 +256,231 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+		vsRole = "mbr"
+	}
+
+	switch vsRole {
+	case "system-data":
+		if vs.Label != "" && vs.Label != "writable" {
+			return fmt.Errorf(`role of this kind must have an implicit label or "writable", not %q`, vs.Label)
+		}
+	case "mbr":
+		if vs.Size > 446 {
```
zyga: Can you add a reference that documents the size limit.

snap/gadget.go
```diff
@@ -234,6 +256,231 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+		return fmt.Errorf("GUID structure type with non-GPT schema %q", vol.Schema)
+	}
+	if schema != "mbr" && isMBR {
+		return fmt.Errorf("MBR structure type with non-MBR schema %q", vol.Schema)
+	}
+
+	return nil
+}
+
+func validateRole(vs *VolumeStructure, vol *GadgetVolume) error {
```
zyga: +1

snap/gadget.go
```diff
@@ -234,6 +256,231 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+		}
+	case "system-boot", "":
+		// noop
+	default:
+		return fmt.Errorf("unsupported role")
+	}
+	return nil
+}
+
+func validateBareContent(vc *VolumeContent) error {
```
zyga: +1

snap/gadget.go
```diff
@@ -234,6 +256,231 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+		return fmt.Errorf("cannot use non-image content for bare file system")
+	}
+	if vc.Image == "" {
+		return fmt.Errorf("missing image file name")
+	}
+	return nil
+}
+
+func validateFilesystemContent(vc *VolumeContent) error {
+	if vc.Image != "" || vc.Offset != 0 || vc.OffsetWrite != (GadgetRelativeOffset{}) || vc.Size != 0 {
```
zyga: Perhaps add an IsZero method to GadgetRelativeOffset?

snap/gadget.go
```diff
@@ -234,6 +256,231 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+	return nil
+}
+
+func validateStructureUpdate(up *VolumeUpdate, vs *VolumeStructure) error {
+	if vs.IsBare() && len(vs.Update.Preserve) > 0 {
+		return errors.New("preserving files during update is not supported for non-filesystem structures")
+	}
+
+	names := make(map[string]bool, len(vs.Update.Preserve))
+	for _, n := range vs.Update.Preserve {
```
zyga: This pattern repeats, perhaps factor out a helper?

```
validateUniqueStrings(items []string) error  {
    seen. := make(map[string]bool, len(items))
    for _, item := rangeg items {
      if seen(item) {
          return fmt.Errorf("duplicate element %q",  item)
      }
      seen[item] = true
    }
    return nil
}
```

snap/gadget.go
```diff
@@ -234,6 +256,231 @@ func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
...
+	if vc.Image != "" || vc.Offset != 0 || vc.OffsetWrite != (GadgetRelativeOffset{}) || vc.Size != 0 {
+		return fmt.Errorf("cannot use image content for non-bare file system")
+	}
+	if vc.Source == "" || vc.Target == "" {
+		return fmt.Errorf("missing source or target")
+	}
+	return nil
+}
+
+func validateStructureUpdate(up *VolumeUpdate, vs *VolumeStructure) error {
```
zyga: +1

snap/gadget.go
```diff
@@ -249,3 +496,105 @@ func (e *editionNumber) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+
+const (
+	SizeMiB = GadgetSize(2 << 20)
+	SizeGiB = GadgetSize(2 << 30)
+)
+
+func (s *GadgetSize) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var gs string
+	if err := unmarshal(&gs); err != nil {
+		return errors.New(`failed to unmarshal`)
```
zyga:
```suggestion
		return errors.New(`cannot to unmarshal gadget size`)
```

snap/gadget.go
```diff
@@ -249,3 +496,105 @@ func (e *editionNumber) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	*s, err = ParseGadgetSize(gs)
+	if err != nil {
+		return fmt.Errorf("cannot parse size %q: %v", gs, err)
+	}
+	return err
+}
+
+// ParseGadgetSize parses a string expressing size in gadget declaration. The
+// accepted format is one of: <bytes> | <bytes/2^20>M | <bytes/2^30>G.
+func ParseGadgetSize(gs string) (GadgetSize, error) {
```
zyga: +1

snap/gadget.go
```diff
@@ -249,3 +496,105 @@ func (e *editionNumber) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	}
+	if sizeSpec == "" {
+		return nil, errors.New("missing offset")
+	}
+
+	size, err := ParseGadgetSize(sizeSpec)
+	if err != nil {
+		return nil, fmt.Errorf("cannot parse offset %q: %v", sizeSpec, err)
+	}
+	if size > 4*SizeGiB {
```
zyga: The comment below is redundant. Can you please add a reference to the origin of the limit?

snap/gadget.go
```diff
@@ -249,3 +496,105 @@ func (e *editionNumber) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	return &GadgetRelativeOffset{
+		RelativeTo: toWhat,
+		Offset:     size,
+	}, nil
+}
+
+func (s *GadgetRelativeOffset) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var grs string
+	if err := unmarshal(&grs); err != nil {
+		return errors.New(`failed to unmarshal`)
```
zyga:
```suggestion
		return errors.New(`cannot to unmarshal gadget relative offset`)
```

snap/gadget.go
zyga: Ah, can you document that in the structure please?

snap/gadget.go
pedronis: I'm not sure the other cases involve a []string directly, so it would get a bit awkward, also we would need to send in fragments to get a good error message

snap/gadget.go
```diff
@@ -52,25 +61,34 @@ type GadgetVolume struct {
...
+	// filesystem content
 	Source string `yaml:"source"`
 	Target string `yaml:"target"`
 
-	Image       string `yaml:"image"`
-	Offset      string `yaml:"offset"`
-	OffsetWrite string `yaml:"offset-write"`
-	Size        string `yaml:"size"`
+	// bare content
+	Image       string               `yaml:"image"`
```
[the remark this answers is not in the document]
bboozzoo: All locations (`image` and `source`) are relative to the gadget base directory.

snap/gadget.go
bboozzoo: I'll drop it in the description of structure fields

snap/gadget.go
bboozzoo: There will be some more code added with actual gadget data write support. I'll add the helper there if the check comes up more than once.

bboozzoo:
Pushed a patch adding comments to volume, structure and content, based on what's there in https://docs.snapcraft.io/the-gadget-snap/696.

PR #6641, opened 2019-03-22
snap-gdb-shim: switch to the SUDO_UID when available

mvo5:
Right now `snap run --gdb` will run everything as root.
This is a problem because most developers expect to debug
their programs as user. However we need to run as root
because the setuid snap-confine will not run with root
permissions when run under gdb.

The fix is to drop root again after snap-confine was run
but before the actual snap application starts.


cmd/snap-gdb-shim/snap-gdb-shim.c
```diff
@@ -29,6 +29,13 @@ int main(int argc, char **argv)
 			printf("-%s-\n", argv[i]);
 		}
 	}
+        // check if we run as SUDO and if so switch to a normal user
+        const char *sudo_uid_env = getenv("SUDO_UID");
+        if (sudo_uid_env != NULL) {
+           int sudo_uid = atoi(sudo_uid_env);
+           if (sudo_uid != 0)
+              setuid(sudo_uid);
```
zyga: Do you need to do something about the group as well?

cmd/snap-gdb-shim/snap-gdb-shim.c
```diff
@@ -29,6 +29,13 @@ int main(int argc, char **argv)
 			printf("-%s-\n", argv[i]);
 		}
 	}
+        // check if we run as SUDO and if so switch to a normal user
+        const char *sudo_uid_env = getenv("SUDO_UID");
+        if (sudo_uid_env != NULL) {
+           int sudo_uid = atoi(sudo_uid_env);
```
zyga: Could you please use `strtol` instead of `atoi` and check for errors?
We have an example in the tree as quick reference:

```
        char *endptr = NULL;
        errno = 0;
        long env_max_wait = strtol(MAX_PROFILE_WAIT, &endptr, 10);
        if (errno != 0 || MAX_PROFILE_WAIT == endptr || *endptr != '\0'
            || env_max_wait <= 0) {
            die("SNAP_CONFINE_MAX_PROFILE_WAIT invalid");
        }
```

cmd/snap-gdb-shim/snap-gdb-shim.c
```diff
@@ -29,6 +29,13 @@ int main(int argc, char **argv)
...
+           if (sudo_uid != 0)
```
zyga: Style nitpick, could you please use the if (...) { ... } form, even in a trivial case like this.

cmd/snap-gdb-shim/snap-gdb-shim.c
jdstrand: You'll want to check return codes as well. Granted, failure to drop when the snap command is called via sudo is not catastrophic since the application is running under confinement, but best practice is, well, best. :)

cmd/snap-gdb-shim/snap-gdb-shim.c
bboozzoo: `strtoul` even?

cmd/snap-gdb-shim/snap-gdb-shim.c
```diff
@@ -15,20 +15,64 @@
...
 
 #include "../libsnap-confine-private/utils.h"
 
+int must_parse_int(const char *s)
+{
+	char *endptr = NULL;
+
+	errno = 0;
+	long i = strtol(s, &endptr, 10);
+	if (errno != 0 || s == endptr || *endptr != '\0' || i < 0) {
```
zyga: Given that we have https://github.com/snapcore/snapd/blob/379f29abd21494b24a35e3234692a59df9a2cef4/cmd/snap-confine/seccomp-support.c#L130 perhaps this could be extracted into one of the utilities and used in both places?

cmd/snap-gdb-shim/snap-gdb-shim.c
```diff
@@ -15,20 +15,64 @@
...
 		}
 	}
+
+	if (getuid() == 0) {
+		// check if we run as SUDO and if so switch to a normal user
+		const char *sudo_uid_env = getenv("SUDO_UID");
+		if (sudo_uid_env != NULL) {
+			int sudo_uid = must_parse_int(sudo_uid_env);
+			if (sudo_uid != 0) {
+				if (setuid(sudo_uid) != 0) {
```
zyga: Nitpick: C code in snapd has a style of `< 0` comparisons, unless the used function returns a non-standard error indicator.

cmd/snap-gdb-shim/snap-gdb-shim.c
```diff
@@ -15,20 +15,64 @@
...
+			}
+		}
+	}
+	if (getgid() == 0) {
+		// check if we run as SUDO and if so switch to a normal user
+		const char *sudo_gid_env = getenv("SUDO_GID");
+		if (sudo_gid_env != NULL) {
+			int sudo_gid = must_parse_int(sudo_gid_env);
+			if (sudo_gid != 0) {
+				if (setgid(sudo_gid) != 0) {
```
zyga: You may no longer be able to change the group. You should probably swap the two paragraphs around, drop the group first, then drop the user.

cmd/snap-gdb-shim/snap-gdb-shim.c
```diff
@@ -15,20 +15,64 @@
...
+}
+
 int main(int argc, char **argv)
 {
 	if (sc_is_debug_enabled()) {
 		for (int i = 0; i < argc; i++) {
 			printf("-%s-\n", argv[i]);
 		}
 	}
+
```
zyga: As a style suggestion I would unconditionally load the environment variable here:
```
const char *sudo_uid_env = getenv("SUDO_UID");
```

And then merge the if condition into one, to avoid one level of nesting:
```
if (sudo_uid_env != NULL && getuid() == 0) {
   ...
}
```

cmd/libsnap-confine-private/utils.c
```diff
@@ -217,3 +217,15 @@ int sc_nonfatal_mkpath(const char *const path, mode_t mode)
...
+
+int sc_must_parse_int(const char *s)
+{
+	char *endptr = NULL;
+
+	errno = 0;
+	long i = strtol(s, &endptr, 10);
+	if (errno != 0 || s == endptr || *endptr != '\0' || i < 0) {
+		die("cannot parse number in '%s'", s);
+	}
```
jdstrand: This function should be renamed sc_must_parse_positive_int() if you are going to error on 'i < 0'.

cmd/libsnap-confine-private/utils.c
```diff
@@ -217,3 +217,15 @@ int sc_nonfatal_mkpath(const char *const path, mode_t mode)
...
+	return i;
```
jdstrand: You read in a long and return an int.

cmd/snap-confine/seccomp-support.c
```diff
@@ -125,11 +125,8 @@ bool sc_apply_seccomp_profile_for_security_tag(const char *security_tag)
 	long max_wait = 120;
 	const char *MAX_PROFILE_WAIT = getenv("SNAP_CONFINE_MAX_PROFILE_WAIT");
 	if (MAX_PROFILE_WAIT != NULL) {
-		char *endptr = NULL;
-		errno = 0;
-		long env_max_wait = strtol(MAX_PROFILE_WAIT, &endptr, 10);
-		if (errno != 0 || MAX_PROFILE_WAIT == endptr || *endptr != '\0'
-		    || env_max_wait <= 0) {
+		long env_max_wait = sc_must_parse_int(MAX_PROFILE_WAIT);
+                if(env_max_wait <= 0) {
```
jdstrand: As written, sc_must_parse_int() will die() before it returns < 0.

cmd/snap-gdb-shim/snap-gdb-shim.c
```diff
@@ -29,6 +30,30 @@ int main(int argc, char **argv)
...
+		int sudo_uid = sc_must_parse_int(sudo_uid_env);
+		if (sudo_uid != 0) {
+			if (setuid(sudo_uid) < 0) {
+				die("cannot switch to uid %d", sudo_uid);
+			}
+		}
+	}
+	// Ideally we would also call setgroups() now but seccomp will
+	// prevent this. At this point we are inside the confinement
+	// of the snap already.
```
jdstrand: This comment should be above the setgid() code cause setgroups() cannot be run after dropping privs. Note that https://github.com/snapcore/snapd/pull/6681 introduces allowing `setgroups(0, NULL)` so perhaps add to your comment you can do this or remove the comment, add that change to this PR and do `setgroups(0, NULL)` here.

cmd/libsnap-confine-private/utils.c
mvo5: Thanks, indeed! I will just make it parse any int and then the caller can check for< 0 if needed.

cmd/libsnap-confine-private/utils.c
mvo5: Indeed, I changed that to long now.

cmd/snap-confine/seccomp-support.c
mvo5: This is fixed now, sc_must_parse_int() will also parse negatives now.

cmd/snap-gdb-shim/snap-gdb-shim.c
mvo5: Good point, I updated the comment. However while looking at this I noticed this PR apparently has a fatal flaw - the setuid/setgid calls happen inside the confinement (they need to run after snap-confine did its job) but at this point they get denied. I added a spread test and indeed:
```
[  561.127570] audit: type=1400 audit(1554797174.738:44): apparmor="DENIED" operation="capable" profile="snap.test-snapd-tools.env" pid=17870 comm="snap-gdb-shim" capability=6  capname="setgid"
```
This seems to be tricky to fix :/

jdstrand: In https://github.com/snapcore/snapd/pull/6681 I found myself in the position of either adding implied seccomp rules that allowed the drop or to reorganize the launcher slightly to not need any implied rules. I opted for the latter.

I mention this because you seem faced with a similar dilemma: either add some capability rules the apparmor profile on the fly (I *highly* dislike this) or to rethink the invocation so that we've already dropped by the time we exec (or to somehow support --gdb without sudo). I don't have specification recommendations on this, but recommend the rethinking route.

mvo5:
Closing to work on a new approach

PR #6643, opened 2019-03-22
tests: deny ioctl - TIOCSTI with garbage in high bits

zyga:
We were alerted to the fact that seccomp filter matching is misbehaving
when the kernel uses 32bit value for an argument that is 64bit. In other
words where the high bits are simply ignored.

The seccomp profile generated by snapd for regular seccomp profiles for
ioctl has code similar to this one:

    # filter for arch x86_64 (3221225534)
    if ($arch == 3221225534)
      # filter for syscall "ioctl" (16) [priority: 65531]
      if ($syscall == 16)
	if ($a1.hi32 == 0)
	  if ($a1.lo32 == 21522)
	  else
	    action ALLOW;
	else
	  action ALLOW;
      # default action
      action ERRNO(1);

Note that the comparison is such so that if a1.hi32 is not zero, the
action is ALLOW.

A simple solution would be to perform arithmetics fix the generated
bytecode. Alas our libseccomp and golang-seccomp bindings cannot do that
at this time.

To resolve the problem we can take advantage of the recently introduced
global profile. A global profile that rejects any attempts to use ioctl
TIOCSTI, without regard for the contents of the high word, is rejected.
The global profile contains the equivalent of the following profile, for
all architectures:

    if ($arch == 3221225534)
      # filter for syscall "ioctl" (16) [priority: 65533]
      if ($syscall == 16)
	if ($a1.hi32 & 0x00000000 == 0)
	  if ($a1.lo32 & 0xffffffff == 21522)
	    # TIOCSTI is 21522
	    action ERRNO(1);
      # default action
      action ALLOW;

This patch contains a C program that generates the blacklist as well as
a regression test for the fix that was silently landed earlier.

Architectures of a given word order are handled with one BPF program.
Specifically for little endian that is x86_64, aarch64 and ppc64le and
for big-endian only s390x. The issue does not affect any 32bit
architecture.


zyga:
Rebased on master to see if this fixes the build.

cmd/snap-seccomp-blacklist/Makefile
```diff
@@ -0,0 +1,40 @@
...
+
+.PHONY: clean
+clean:
+	rm -f snap-seccomp-blacklist snap-seccomp-blacklist.o *.pfc *.bpf
+
+.PHONY: fmt
+fmt: snap-seccomp-blacklist.c
+	clang-format -style='{BasedOnStyle: Google, IndentWidth: 4, ColumnLimit: 120}' -i $^
+
+DESTDIR ?=
```
stolowski: Mostly curious, this looks a bit unusual; if it's not set why set it to empty, will make complain if unset?

stolowski:
A bit of a meta-question (and probably a silly one as I don't know all the background nor secomp): is it expected this issue will ever be addressed upstream, or it's not possible? If so, can we have a test (possibly as a separate .c binary) that exhibits the issue and acts a canary that alerts us (fails) when it's fixed upstream (so we know we can phase out the workaround)?

tests/regression/lp-1812973/lp-1812973.c
```diff
@@ -0,0 +1,63 @@
...
+        res = ioctl64(fd, TCFLSH, TCIOFLUSH);
+        saved_errno = errno;
+        printf("unrelated TCFLSH: %d (%m) (expect ENOTTY)\n", res);
+        if (res < 0 && saved_errno == ENOTTY) {
+            rc = EXIT_SUCCESS;
+        }
+    } else if (argc == 2 && strcmp(argv[1], "--good-high") == 0) {
+        res = ioctl64(fd, TCFLSH | (1UL << 32UL), TCIOFLUSH);
+        saved_errno = errno;
+        printf("unrelated TCFLSH: %d (%m) (expect ENOTTY) \n", res);
```
stolowski: Nitpick: extra space before `\n`.

tests/regression/lp-1812973/Makefile
```diff
@@ -0,0 +1,21 @@
...
+DESTDIR ?=
+
+.PHONY: install
+install: $(name)
+	install -m 755 -d $(DESTDIR)/usr/bin/
+	install -m 755 $(name) $(DESTDIR)/usr/bin/
+
+.PHONY: fmt
+fmt: $(name).c
+	clang-format -style='{BasedOnStyle: Google, IndentWidth: 4, ColumnLimit: 120}' -i $^
```
stolowski: Nitpick: perhaps we should move style definition to a central file (clang-format supports style def from external file), it's already used in cmd/Makefile.am.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+    vfprintf(stderr, fmt, va);
+    fputc('\n', stderr);
+    va_end(va);
+}
+
+static int populate_filter(scmp_filter_ctx ctx, const uint32_t *arch_tags, size_t num_arch_tags) {
+    int sc_err;
+
+    /* If the native architecture is not one of the supported 64bit
+     * architectures listed above then remove it.
```
stolowski: I think with "listed above" you mean listed in `arch_tags`? Can you clarify this comment please?

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+        SCMP_ARCH_PPC64LE,
+        SCMP_ARCH_S390X,
+    };
+    const uint32_t be_arch_tags[] = {
+        SCMP_ARCH_S390X,
+    };
+    const arch_set arch_sets[] = {
+        {"LE", le_arch_tags, sizeof le_arch_tags / sizeof *le_arch_tags},
+        {"BE", be_arch_tags, sizeof be_arch_tags / sizeof *be_arch_tags},
+    };
```
stolowski: Nicely structured!

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+
+    for (size_t i = 0; i < sizeof arch_sets / sizeof *arch_sets; ++i) {
+        const arch_set *as = &arch_sets[i];
+        int sc_err;
+        int fd = -1;
+        int fname_len;
+        char fname[PATH_MAX];
+
+        scmp_filter_ctx ctx = NULL;
+        ctx = seccomp_init(SCMP_ACT_ALLOW);
```
stolowski: This looks fine. We create separate contexts for LE and BE because we save separate pfcs for them below.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+        if (sc_err < 0) {
+            showerr("cannot export BPF program");
+            seccomp_release(ctx);
+            close(fd);
+            return -rc;
+        }
+
+        close(fd);
+        seccomp_release(ctx);
+    }
```
stolowski: Question: we may succesfully generate bpf for the first set and for some reason fail on the second set; is this ok to have just one bpf file or should it be all-or-nothing?

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+        sc_err = populate_filter(ctx, as->arch_tags, as->num_arch_tags);
+        if (sc_err < 0) {
+            seccomp_release(ctx);
+            return -rc;
+        }
+
+        /* Save pseudo-code program */
+        fname_len = snprintf(fname, sizeof fname, "%s-blacklist.pfc", as->name);
+        if (fname_len < 0 || fname_len >= sizeof fname) {
+            showerr("cannot format file name");
```
stolowski: Should this error have more context (file name, arch...)?

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+            close(fd);
+            return -rc;
+        }
+
+        close(fd);
+
+        /* Save binary program. */
+        fname_len = snprintf(fname, sizeof fname, "%s-blacklist.bpf", as->name);
+        if (fname_len < 0 || fname_len >= sizeof fname) {
+            showerr("cannot format file name");
```
stolowski: Same (about error message).

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+        }
+        fd = open(fname, O_CREAT | O_TRUNC | O_WRONLY | O_NOFOLLOW, 0644);
+        if (fd < 0) {
+            showerr("cannot open %s", fname);
+            seccomp_release(ctx);
+            return -rc;
+        }
+        sc_err = seccomp_export_bpf(ctx, fd);
+        if (sc_err < 0) {
+            showerr("cannot export BPF program");
```
stolowski: Same (about error message).

tests/regression/lp-1812973/task.yaml
```diff
@@ -0,0 +1,18 @@
...
+restore: |
+    make clean
+    snap remove test-snapd-lp-1812973
+    rm -f ./test-snapd-lp-1812973_1_all.snap
+    rm -f ./test-snapd-lp-1812973/usr/bin/lp-1812973
+execute: |
+    test-snapd-lp-1812973 --good
+    test-snapd-lp-1812973 --good-high
+    test-snapd-lp-1812973 --evil
+    test-snapd-lp-1812973 --evil-high
```
stolowski: Nice test ❤️

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+         * bits set in the more significant 32bit word of the 64 bit double
+         * word. While this kernel behavior remains suspect and presumably
+         * undesired it is unlikely to change for backwards compatibility
+         * reasons. As such we cannot block all requests with high-bits set. */
+        .arg = 1,
+        .op = SCMP_CMP_MASKED_EQ,
+        .datum_a = 0xffffffffUL,
+        .datum_b = TIOCSTI,
+    };
+    sc_err = seccomp_rule_add(ctx, SCMP_ACT_ERRNO(EPERM), sys_ioctl_nr, 1, no_tty_inject);
```
stolowski: Any reason we are not we using SCMP_SYS(..) to resolve ioctl number? FWTW SCMP_SYS seems to be suggested by man page of seccomp_rule_add.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+     * filtering there. */
+    sc_err = seccomp_attr_set(ctx, SCMP_FLTATR_ACT_BADARCH, SCMP_ACT_ALLOW);
+    if (sc_err < 0) {
+        showerr("cannot set action for unknown architectures");
+        return sc_err;
+    }
+
+    /* Resolve the name of "ioctl" on this architecture. */
+    int sys_ioctl_nr;
+    sys_ioctl_nr = seccomp_syscall_resolve_name("ioctl");
```
stolowski: I think this could go away if we use SCMP_SYS, see comment below.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
stolowski: Ah, I suppose it's fine since those are inputs during the build, and we will simply fail in the Makefile.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+     * In other words, As such arguments to scmp_rule_add() must always use
+     * native system call numbers. Translation for the correct architecture
+     * will be performed internally. This is not documented in libseccomp. But
+     * correct operation was confirmed using the pseudo-code program and the
+     * bpf_dbg tool from the kernel tools/bpf directory.
+     *
+     * NOTE: I'm not using scmp_rule_add_exact as that was not doing anything
+     * at all (presumably due to having all the architectures defined). */
+
+    const struct scmp_arg_cmp no_tty_inject = {
```
stolowski: Why not using `SCMP_CMP(...)` macro to define this structure?

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+         * 1) it blocks ioctl(fd, TIOCSTI, ptr).
+         * 1) it also blocks ioctl(fd, (1UL<<32) | TIOCSTI, ptr).
+         * 2) it doesn't block ioctl(fd, (1UL<<32) | (request not equal to TIOCSTI), ptr);
+         *
+         * We learned that existing programs make legitimate requests with all
+         * bits set in the more significant 32bit word of the 64 bit double
+         * word. While this kernel behavior remains suspect and presumably
+         * undesired it is unlikely to change for backwards compatibility
+         * reasons. As such we cannot block all requests with high-bits set. */
+        .arg = 1,
```
stolowski: What is 1? Is there a handy constant/definition in seccomp API to use instead?

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+         *
+         * 1) it blocks ioctl(fd, TIOCSTI, ptr).
+         * 1) it also blocks ioctl(fd, (1UL<<32) | TIOCSTI, ptr).
+         * 2) it doesn't block ioctl(fd, (1UL<<32) | (request not equal to TIOCSTI), ptr);
+         *
+         * We learned that existing programs make legitimate requests with all
+         * bits set in the more significant 32bit word of the 64 bit double
+         * word. While this kernel behavior remains suspect and presumably
+         * undesired it is unlikely to change for backwards compatibility
+         * reasons. As such we cannot block all requests with high-bits set. */
```
stolowski: Thanks for extensive comments/explanations, this helps a lot in reviewing this!

cmd/snap-seccomp-blacklist/Makefile
zyga: Yes, make does complain if you ask it to via `--warn-undefined-variables`. I do it regularly to avoid silly mistakes.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
zyga: Given that this is a single-trick-pony I don't think this is required.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
zyga: Because this makes this thing portable. It allows us to ask seccomp's opinion of the system call number even if the system call does not exist on the current architecture. This is untrue for ioctl but the general rule applies.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
zyga: We cannot, as explained above.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
stolowski: Ah, I see, makes sense. Perhaps deserves a short comment above where sys_ioctl_nr is resolved.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+        showerr("cannot resolve ioctl system call number");
+        return -ESRCH;
+    }
+
+    /* All of the rules must be added for the native architecture (using native
+     * system call numbers). When the final program is generated the set of
+     * architectures added earlier will be used to determine the correct system
+     * call number for each architecture.
+     *
+     * In other words, As such arguments to scmp_rule_add() must always use
```
bboozzoo: `In other words, arguments to scmp_rule_add() must always use...`

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+     * native system call numbers. Translation for the correct architecture
+     * will be performed internally. This is not documented in libseccomp. But
```
bboozzoo: `This is not documented in libseccomp, but...`

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+     * correct operation was confirmed using the pseudo-code program and the
+     * bpf_dbg tool from the kernel tools/bpf directory.
+     *
+     * NOTE: I'm not using scmp_rule_add_exact as that was not doing anything
```
bboozzoo: `NOTE: not using seccomp_rule_add_exact ...`

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+     * at all (presumably due to having all the architectures defined). */
+
+    const struct scmp_arg_cmp no_tty_inject = {
+        /* When faced with ioctl(fd, request); refuse to proceed when
+         * request&0xffffffff == TIOCSTI. This specific way to encode the
+         * filter has the following important properties:
+         *
+         * 1) it blocks ioctl(fd, TIOCSTI, ptr).
```
bboozzoo: I think I'd drop the list, seems unnecessary and quite confusing.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+         * 1) it also blocks ioctl(fd, (1UL<<32) | TIOCSTI, ptr).
+         * 2) it doesn't block ioctl(fd, (1UL<<32) | (request not equal to TIOCSTI), ptr);
+         *
+         * We learned that existing programs make legitimate requests with all
```
bboozzoo: And I'd swap this paragraph with the first one.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
```diff
@@ -0,0 +1,215 @@
...
+    return 0;
+}
+
+typedef struct arch_set {
+    const char *name;
+    const uint32_t *arch_tags;
+    size_t num_arch_tags;
+} arch_set;
+
+int main(int argc, char **argv) {
```
bboozzoo: Out of curiosity, couldn't this have been written in Go?

cmd/snap-seccomp-blacklist/Makefile
```diff
@@ -0,0 +1,40 @@
...
+
+.PHONY: install
+install:: blacklist.bpf | $(DESTDIR)/var/lib/snapd/seccomp/bpf
```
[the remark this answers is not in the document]
zyga: It's my personal preference that I think makes sense for a target such as `install`. It does so because the install target is fulfilled by many separate installations that have their own rules. Since we are not representing the installed file as the goal of the rule using the double colon allows us to have separate entries freely, and not clump them in one complex install rule.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
zyga: Only in the go that imports a preamble of C headers. I strongly desire to stop using C for this code but using it as a super thin layer on top of C is not a solution IMO.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
zyga: In addition make is smart enough to unlink partially created target files.

tests/regression/lp-1812973/Makefile
zyga: Great idea, done!

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
zyga: I added some context to error messages here and elsewhere.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
zyga: `SCMP_CMP` doesn't add any value in my eyes. I looked at this and at the `SCMP_A1` macro as well. The former is an inline structure definition, the latter is literally `1, ...`. I chose to make this explicit to avoid a layer of magic that is not helping in the general understanding.

My approach has the added benefit of naming field names that readers may not be familiar with.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
zyga: I changed the style of the list slightly but I would prefer to retain it unless you can point out what the confusion is about. The list captures the essence of what the program is trying to achieve so in my eyes it is central to what this patch does.

.clang-format
```diff
@@ -0,0 +1,3 @@
+BasedOnStyle: Google
+IndentWidth: 4
+ColumnLimit: 120
```
stolowski: Nice, thanks for this change!

zyga: This has one nice property, it allows one to call "clang-format" anywhere in the tree, and just get something consistent across the three and, from my point of view, nice looking.

cmd/snap-seccomp-blacklist/snap-seccomp-blacklist.c
stolowski: I slightly disagree (but won't insist or block on this) - such macros make things slightly more future-proof, and also more compact, improving radability; the struct is not even documented in man page of seccomp_rule_add (whereas the macros are), so having the struct explicitly defined doesn't make it more readable in my eyes (but maybe it does for seccomp hackers, therefore I'm not going to insist 😄 ).

zyga:
> A bit of a meta-question (and probably a silly one as I don't know all the background nor secomp): is it expected this issue will ever be addressed upstream, or it's not possible? If so, can we have a test (possibly as a separate .c binary) that exhibits the issue and acts a canary that alerts us (fails) when it's fixed upstream (so we know we can phase out the workaround)?

@stolowski we are not expecting this behaviour to change, while unexpected and perhaps undesired it **is** documented to behave this way.

zyga:
@stolowski this was pre-reviewed by @jdstrand during the embargo period.

jdstrand:
We talked about this and unless someone feels differently, I wasn't planning on reviewing this: I looked at it before, it is just build code and we have tests for all the 64 bit architectures for the resulting binary to catch if its output is wrong.

PR #6646, opened 2019-03-25
interfaces/adb-support: account for hubs on sysfs path

zyga:
The adb-support interface meant to allow accessing the serial number of
all USB devices. The apparmor regular expression was constructed based
on what I saw on my particular system. What I failed to take into
account were that USB hubs are represented in this path so the interface
would fail when additional paths are on the connection chain.

This patch fixes this by allowing to read serial number from any nesting
depth.

Fixes: https://bugs.launchpad.net/snapd/+bug/1821469


mvo5:
Looks good, needs also a second review from @jdstrand (but fortunately small).

PR #6648, opened 2019-03-25
tests: fix core revert test

sergiocazzolato:
2 fixes for the core revert test:
 1. Add wait until backend can be reached  after reboot
 2. Now the snap log contains debug info

To check the test run: > SPREAD_CORE_CHANNEL=stable SPREAD_CORE_REFRESH_CHANNEL=beta spread google-nested:ubuntu-16.04-64:tests/nested/core/core-revert

### DIFF
--- a/tests/lib/nested.sh
+++ b/tests/lib/nested.sh
@@ -123,14 +123,12 @@ add_debug_output_on_nested_core_vm(){
 StartLimitInterval=0
 [Service]
 Environment=SNAPD_DEBUG_HTTP=7 SNAPD_DEBUG=1
-ExecStartPre=/bin/touch /dev/iio:device0
 EOF
     copy_remote local.conf
     execute_remote "sudo mkdir -p /etc/systemd/system/snapd.service.d"
     execute_remote "sudo mv local.conf /etc/systemd/system/snapd.service.d"
     execute_remote "sudo systemctl daemon-reload"
-    execute_remote "sudo systemctl stop snapd.socket snapd.service"
-    execute_remote "sudo systemctl start snapd.socket snapd.service"
+    execute_remote "sudo systemctl restart snapd.socket snapd.service"
     rm -f local.conf
 }
 
--- a/tests/lib/nested.sh
+++ b/tests/lib/nested.sh
@@ -100,6 +100,7 @@ create_nested_core_vm(){
         echo "ssh not stablished, exiting..."
         exit 1
     fi
+    add_debug_output_on_nested_core_vm
 }
 
 start_nested_core_vm(){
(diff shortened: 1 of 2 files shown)


tests/lib/nested.sh
```diff
@@ -116,6 +117,23 @@ start_nested_core_vm(){
...
+StartLimitInterval=0
+[Service]
+Environment=SNAPD_DEBUG_HTTP=7 SNAPD_DEBUG=1
+ExecStartPre=/bin/touch /dev/iio:device0
+EOF
+    copy_remote local.conf
+    execute_remote "sudo mkdir -p /etc/systemd/system/snapd.service.d"
+    execute_remote "sudo mv local.conf /etc/systemd/system/snapd.service.d"
+    execute_remote "sudo systemctl daemon-reload"
+    execute_remote "sudo systemctl stop snapd.socket snapd.service"
```
bboozzoo: Why not restart instead of stop && start?

tests/lib/nested.sh
```diff
@@ -116,6 +117,23 @@ start_nested_core_vm(){
     fi
 }
 
+add_debug_output_on_nested_core_vm(){
+    cat <<EOF > local.conf
+[Unit]
+StartLimitInterval=0
+[Service]
+Environment=SNAPD_DEBUG_HTTP=7 SNAPD_DEBUG=1
+ExecStartPre=/bin/touch /dev/iio:device0
```
bboozzoo: Why do we need this?

zyga: Apparently some tests use it.

I wonder if this should be done in the corresponding test rather than here where it raises eyebrows.

bboozzoo: IIRC, `/dev/iio*` is industrial I/O, it is a bit surprising to see it here.

sergiocazzolato: Well I copied that from the configuration we have for the snapd test suite because we could run any test on the nested vm but not sure if it is gonna happen. I could remove that and add it in case it is needed in the future.

tests/lib/nested.sh
sergiocazzolato: done

sergiocazzolato:
test log: https://paste.ubuntu.com/p/8cwdR3DT27/

tests/nested/core/core-revert/task.yaml
```diff
@@ -33,6 +33,13 @@ execute: |
     #shellcheck source=tests/lib/nested.sh
     . "$TESTSLIB/nested.sh"
 
+    wait_connection_with_store(){
+        while ! execute_remote "sudo snap find test-snapd-tools | grep 'Tools for testing the snapd application'"; do
```
chipaca: you could instead `while ! snap debug connectivity`

mvo5: @sergiocazzolato WDYT about the suggestion from john?

sergiocazzolato:
As the connectino error is not being reproduced anymore I'll close this PR and in the future I'll open a new one to add debug info to snapd.

PR #6650, opened 2019-03-26
cmd/libsnap: neuter variables in cleanup functions

zyga:
The facility of cleanup functions, a gcc extension that is similar to
registering a C++ destructor to be called when a variable goes out of
scope, is very useful and allows to simplify error handling and
guarantee correctness.

During a recent code review it was noted that while useful, the cleanup
functions will misbehave if somehow invoked twice on the same object.
This normally never happens because the compiler is generating the calls
for us so manual use is unlikely.

To harden the code against issue like that the patch ensures that all
cleanup functions neuter the affected variable with a proper value,
either a NULL pointer or a -1 file descriptor, as appropriate.

New tests and documentation explain and illustrate the proper use of all
the common cleanup functions defined to date..


cmd/libsnap-confine-private/cleanup-funcs.c
```diff
@@ -24,33 +24,38 @@ void sc_cleanup_string(char **ptr)
 {
 	if (ptr != NULL) {
```
bboozzoo: `if (ptr != NULL && *ptr != NULL)`

mvo5: Sleeping over this I think strictly this is not needed because free(NULL) is fine. But having it is fine as well :)

PR #6656, opened 2019-03-27
tests: split travis spread execution in 2 jobs for ubuntu and non ubuntu systems

sergiocazzolato:
This is the same change than #6631 which was deleted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ jobs:
       script:
         - ./tests/lib/cla_check.py
     - stage: integration
-      name: spread
+      name: spread on ubuntu systems
       dist: xenial
       addons:
         apt:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -56,7 +56,19 @@ jobs:
         # override the default install for language:go
         - true
       script:
-        - ./run-checks --spread
+        - ./run-checks --spread-ubuntu
+    - stage: integration
+      name: spread on other systems than ubuntu
```
bboozzoo: And this could be named `Fedora, CentOS, OpenSUSE, Debian, Arch`

.travis.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       script:
         - ./tests/lib/cla_check.py
     - stage: integration
-      name: spread
+      name: spread on ubuntu systems
```
bboozzoo: Can we name it `Ubuntu 14.04, 16.04, 18.04, 18.10`

bboozzoo:
Let's discuss the plan at the standup.

.travis.yml
sergiocazzolato: thanks for reviewing! It is ok for me, lets discuss the names on the standup

PR #6658, opened 2019-03-28
testutil: run mocked commands through shellcheck

bboozzoo:
Shell is deceivingly easy to write. It's best that we run the test snippets
through shellcheck.

This adds ~2s to whole suite execution time on my box.

### DIFF
--- a/overlord/ifacestate/udevmonitor/udevmon_test.go
+++ b/overlord/ifacestate/udevmonitor/udevmon_test.go
@@ -90,6 +90,7 @@ E: MINOR=3
 E: MAJOR=0
 E: DEVNAME=ghi
 E: DEVTYPE=bzz
+__END__
 `)
 	defer cmd.Restore()
 
--- a/userd/settings_test.go
+++ b/userd/settings_test.go
@@ -59,7 +59,7 @@ elif [ "$1" = "set" ] && [ "$2" = "default-web-browser" ]; then
   # nothing to do
   exit 0
 else
-  echo "mock called with unsupported arguments $@"
+  echo "mock called with unsupported arguments $*"
   exit 1
 fi
 `)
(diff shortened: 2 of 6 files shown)


overlord/ifacestate/udevmonitor/udevmon_test.go
```diff
@@ -90,6 +90,7 @@ E: MINOR=3
 E: MAJOR=0
 E: DEVNAME=ghi
 E: DEVTYPE=bzz
+__END__
```
zyga: Ha :-)

userd/settings_test.go
```diff
@@ -59,7 +59,7 @@ elif [ "$1" = "set" ] && [ "$2" = "default-web-browser" ]; then
   # nothing to do
   exit 0
 else
-  echo "mock called with unsupported arguments $@"
+  echo "mock called with unsupported arguments $*"
```
zyga: Another ha :-)

overlord/ifacestate/udevmonitor/udevmon_test.go
stolowski: Hmm, interesting!

zyga: I wonder if shell was happy because end-of-input or if this shows the test was really not using it at all and something else was masked?

mvo5: I played with this and it looks like indeed EOF is fine and no error - a very interessting find :)

mvo5:
Thanks for this PR! 

While playing with the PR I noticed that shellcheck apparently does not check is if "set -e" is used :/ Would be nice to have this too - but totally unrelated to this PR.

mvo5:
There is a bit of discussion about enforcing of `set -e` in https://github.com/koalaman/shellcheck/issues/409 and some people apparently really don't like it - reading the rationals makes my dislike for shell grow. Oh well.

PR #6661, opened 2019-03-28
data/selinux, tests/main/selinux-clean: fine tune the policy, make sure that no denials are raised

bboozzoo:
Final touches to the policy. Patch describes why those were needed: https://github.com/snapcore/snapd/commit/261f16fc6b7815838854a1d6bca9ae8bde0de545

Add a test that performs basic snap operations with snaps that are typically run by the user (thus unconfined_t) and snaps that provide services, thus execute the whole targeted policy chain.


data/selinux/snappy.te
```diff
@@ -156,6 +159,10 @@ systemd_config_all_services(snappy_t)
 systemd_manage_all_unit_files(snappy_t)
 systemd_manage_all_unit_lnk_files(snappy_t)
 systemd_exec_systemctl(snappy_t)
+systemd_reload_all_services(snappy_t)
```
stolowski: Should we also have `systemd_start_all_services`?

tests/main/selinux-clean/task.yaml
```diff
@@ -0,0 +1,52 @@
...
+    snap install test-snapd-tools
+    test-snapd-tools.cmd echo 'hello world'
+    su -c 'test-snapd-tools.cmd echo hello world' test
+    snap remove test-snapd-tools
+    ausearch --checkpoint stamp --start checkpoint -m AVC 2>&1 | MATCH 'no matches'
+
+    #shellcheck source=tests/lib/snaps.sh
+    . "$TESTSLIB/snaps.sh"
+
+    install_local test-snapd-service
```
stolowski: Can we also test `snap start ...` and `snap stop ...` here for this snap, or are these covered already elsewhere?

data/selinux/snappy.te
bboozzoo: From inspecting the policy, this is likely covered by systemd_config_all_services(). This is what sesearch tells me:
```
$ sesearch -A -c service -s snappy_t
Found 2 semantic av rules:
   allow snappy_t init_script_file_type : service { start stop status reload enable disable kill load } ; 
   allow snappy_t systemd_unit_file_type : service { start stop status reload enable disable kill load } ; 
```

tests/main/selinux-clean/task.yaml
bboozzoo: Added

bboozzoo:
@Conan-Kudo  can you take a look?

tests/main/selinux-clean/task.yaml
stolowski: Thanks!

bboozzoo:
@zyga do you think you could take a look at this PR too?

tests/main/selinux-clean/task.yaml
```diff
@@ -0,0 +1,54 @@
...
+    distro_install_package fontconfig dejavu-sans-fonts
+
+    # Enable enforcing mode, our policy is already marked as permissive, so we
+    # will get audit entries but the program will not be stopped by SELinux
+    setenforce 1
+    ausearch --checkpoint stamp -m AVC || true
+
+restore: |
+    rm -f stamp
+    setenforce 0
```
zyga: Nitpick: I'd be cleaner if this test had restored the state of selinux rather than disabling it and assuming the initial state was the same.

PR #6663, opened 2019-03-28
cmd: make fmt

zyga:
I was putting this off because indent had changed once in the past
but now working with C code is just so damn annoying when my edits are
intermixed with the noise about how attributes are formatted or how
the pointer start spacing is supposed to look like.

### DIFF
--- a/cmd/snap-update-ns/bootstrap.c
+++ b/cmd/snap-update-ns/bootstrap.c
@@ -333,7 +333,8 @@ int validate_instance_name(const char *instance_name)
 }
 
 // parse the -u argument, returns -1 on failure or 0 on success.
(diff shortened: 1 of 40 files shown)


cmd/snap-update-ns/bootstrap.c
```diff
@@ -371,14 +373,15 @@ static int parse_arg_u(int argc, char * const *argv, int *optind, unsigned long
 	if (uid_out != NULL) {
 		*uid_out = parsed_uid;
 	}
-	*optind += 1; // Account for the argument to -u.
+	*optind += 1;		// Account for the argument to -u.
```
stolowski: Hmmm this looks peculiar, I'd expect a single space.

zyga: The mysteries of indent.

PR #6667, opened 2019-04-01
tests: enable tests that write /etc/{hostname,timezone} on core18

mvo5:
The systemd in Ubuntu 18.04 dropped some patches to make the
"writable-path" magic in core18 work. These patches got re-added
in bionic-proposed and in ppa:canonical-foundations/ubuntu-image.

This PR enables the tests on core18 to validate that the bug
is actually fixed.

The SRU is tracked via LP: 1778936.

### DIFF
--- a/tests/main/interfaces-hostname-control/task.yaml
+++ b/tests/main/interfaces-hostname-control/task.yaml
@@ -3,9 +3,6 @@ summary: Ensure that the hostname-control interface works
 details: |
     The hostname-control interface allows to configure the system hostname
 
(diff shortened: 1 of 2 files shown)


mvo5:
We have bionic-proposed disabled now in the core18 snap. However it is copied into ppa:canonical-foundations/ubuntu-image and this PPA is used during the package builds.

mvo5:
Marking as blocked until the SRU is officially accepted.

mvo5:
The systemd fix went into -updates, this can land now.

PR #6671, opened 2019-04-01

zyga:


zyga:
I found two issues in this and I'm working on fixing them, stay tuned.

zyga:
I will reopen this after rebasing.

zyga:
Re-opened as https://github.com/snapcore/snapd/pull/6673

PR #6674, opened 2019-04-01
tests: use apt via eatmydata

zyga:
On Debian and Ubuntu we have eatmydata pre-installed and thus can use it
to speed up package installation operations.

On my local test, performed on a 1st gen core i7 running Fedora 30, with
a low-end SSD, invoked as follows:

    $ time SPREAD_HTTPS_PROXY=... SPREAD_HTTP_PROXY=...
    ~/go/bin/spread -debug -v qemu:ubuntu-16.04-64:tests/main/snap-run

The times for three consecutive runs are:

    real    11m27,168s
    user    0m14,345s
    sys     0m1,216s

    real    11m35,334s
    user    0m15,294s
    sys     0m1,145s

    real    11m38,274s
    user    0m14,240s
    sys     0m1,140s

On the same machine, with the same hot proxy, the times for vanilla
master were:

    real    13m21,811s
    user    0m13,850s
    sys     0m1,088s

    real    13m16,955s
    user    0m14,451s
    sys     0m1,130s

    real    12m52,010s
    user    0m14,443s
    sys     0m1,155s

I didn't measure any other test but the savings are expected due to
package installation overhead that is incurred when dpkg synchronizes
particular file-system operations.

### DIFF
--- a/cmd/snap-confine/spread-tests/spread-prepare.sh
+++ b/cmd/snap-confine/spread-tests/spread-prepare.sh
@@ -56,7 +56,7 @@ build_debian_or_ubuntu_package() {
     git clone -b "$distro_packaging_git_branch" "$distro_packaging_git" distro-packaging
 
     # Install all the build dependencies declared by the package.
-    apt-get install --quiet -y gdebi-core
+    eatmydata apt-get install --quiet -y gdebi-core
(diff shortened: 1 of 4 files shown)


zyga:
Force pushed because I obviously did not commit eatmydata changes the first time.

mvo5:
Yeah, nice work zyga!

PR #6675, opened 2019-04-01
cmd/snap-confine: allow using tools from snapd snap

zyga:
The snap-confine snap uses tools, such as snap-update-ns or
snap-discard-ns from a tools directory. Due to complexity of the
re-execution system and the cross-distribution support the tools
directory can be present in one of many possible locations.

For a while we did support using the tools from core but we apparently
missed using tools from the snapd snap. This patch fixes that.

Fixes: https://bugs.launchpad.net/snapd/+bug/1822660


zyga:
The spread test of this is done indirectly via https://github.com/snapcore/snapd/pull/6673

PR #6682, opened 2019-04-03
snap, gadget: move gadget read/validation into separate package, tweak naming

bboozzoo:
As discussed, I'm moving gadget related bits from `snap` to the `gadget` package. While at it, a bunch of structs have been renamed, eg `snap.GadgetSize` -> `gadget.Size` to fit into the new namespace.


snap/gadget.go
```diff
@@ -20,614 +20,24 @@
...
-		return false
-	}
-	return true
-}
-
 // ReadGadgetInfo reads the gadget specific metadata from gadget.yaml
 // in the snap. classic set to true means classic rules apply,
 // i.e. content/presence of gadget.yaml is fully optional.
-func ReadGadgetInfo(info *Info, classic bool) (*GadgetInfo, error) {
+func ReadGadgetInfo(info *Info, classic bool) (*gadget.Info, error) {
```
mvo5: Do we want/need this? Or we will get rid of this eventually? Feels slightly redundant to have "snap.ReadGadgetInfo()" and "gadget.ReadGadgetInfo()" but then one takes "snap.Info" the other just a dir. No strong opinion, just curious.

gadget/gadget.go
```diff
@@ -0,0 +1,630 @@
...
+	if s != "system" && len(s) != validSnapIDLength {
+		return false
+	}
+	return true
+}
+
+// ReadGadgetInfo reads the gadget specific metadata from gadget.yaml
+// in the snap. classic set to true means classic rules apply,
+// i.e. content/presence of gadget.yaml is fully optional.
+func ReadGadgetInfo(gadgetRootDir string, classic bool) (*Info, error) {
```
mvo5: I wonder if we can rename this now? `gadget.ReadInfo()` or just `gadget.Read()` or something?

bboozzoo: Ah, `ReadInfo()` sounds about right.

snap/gadget.go
bboozzoo: I think it may be useful when we reach the stage when we'll have a command to assemble the image. At this point we'll either drop the redundant piece or refactor it. No strong feelings though.

gadget/gadget.go
```diff
@@ -0,0 +1,630 @@
...
+	if s != "system" && len(s) != validSnapIDLength {
+		return false
+	}
+	return true
+}
+
+// ReadInfo reads the gadget specific metadata from gadget.yaml
+// in the snap. classic set to true means classic rules apply,
+// i.e. content/presence of gadget.yaml is fully optional.
+func ReadInfo(gadgetRootDir string, classic bool) (*Info, error) {
```
stolowski: Some bikeshedding... Wouldn't snapMountDir be more accurate than gadgetRootDir?

bboozzoo: `gadgetSnapRootDir` ?

stolowski: Works for me

PR #6684, opened 2019-04-03
overlord,tests: perform soft refresh check in doInstall

zyga:
When a snap is installed or refreshed, perform a soft check. That is,
see if there are any applications running that would interfere with the
refresh process.

The soft check is the first half of the feature: the hard check is next,
and verifies that the applications are still not running just before
unlinking the old snap revision.

Note that the information about the snap to be refreshed needs to be
loaded from the state because we need to know the current set of apps
and hooks that are there.

This patch also contains a simple test for the soft refresh check of the
app-refresh-awareness feature. When the feature is enabled refreshing an
application while it is running is rejected.

Two tests that were not locking the state were amended to do so since we
now need to inspect the experimental flags in doInstall.

### DIFF
--- a/tests/main/refresh-app-awareness/task.yaml
+++ b/tests/main/refresh-app-awareness/task.yaml
@@ -23,8 +23,8 @@ execute: |
     test-snapd-refresh.sh -c 'exec sleep 1h' &
     pid=$!
     trap 'kill '"$pid"' || true' EXIT
-    # Try to install v2, it should fail because v1 is running.
-    snap install --dangerous test-snapd-refresh_2_all.snap || true
+    # Try to install v2, it should fail because v1 is running. Snapd is even kind enough to tell us what is preventing the install from working.
+    ! snap install --dangerous test-snapd-refresh_2_all.snap | MATCH 'error: cannot install snap file: snap "test-snapd-refresh" has running apps \(sh\)'
     test-snapd-refresh.version | MATCH v1
     # Kill the app process running from v1.
     kill "$pid"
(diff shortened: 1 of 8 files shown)


tests/main/refresh-app-awareness/task.yaml
```diff
@@ -0,0 +1,34 @@
...
+    # Once this is fixed please uncomment this line:
+    # snap unset core experimental.refresh-app-awareness
+execute: |
+    # Install v1 and see that it runs as expected.
+    snap install --dangerous test-snapd-refresh_1_all.snap
+    test-snapd-refresh.version | MATCH v1
+    # Run a sleeper app to keep the snap busy
+    test-snapd-refresh.sh -c 'exec sleep 1h' &
+    pid=$!
+    trap 'kill '"$pid"' || true' EXIT
```
chipaca: why not `trap "kill $pid || true" EXIT`?

zyga: Because then shellcheck says you gotta quote "$pid" inside the already quoted trap command argument I and don't like escaping double quotes :)

chipaca: `trap "kill '$pid' || true" EXIT` would still read better than your current one tho

zyga: It would require a shellcheck directive though.

zyga:
I'll craft a unit test. I actually didn't notice I didn't write any, sorry.

PR #6688, opened 2019-04-04
gadget: add validation of cross structure overlap and offset writes

bboozzoo:
Validate structure size, structure overlap and offset-write positions.


gadget/gadget.go
```diff
@@ -299,41 +317,99 @@ func validateVolume(name string, vol *Volume) error {
...
+	// - relative offsets that reference other structures by name
+	// - positioned structure overlap
+	// use structures positioned within the volume
+	for pidx, ps := range structures {
+		if ps.OffsetWrite.RelativeTo != "" {
+			// offset-write using a named structure
+			other := knownStructures[ps.OffsetWrite.RelativeTo]
+			if other == nil {
+				return fmt.Errorf("structure %v refers to an unknown structure %q",
+					fmtIndexAndName(ps.index, ps.Name), ps.OffsetWrite.RelativeTo)
```
cmatsuoka: We could use "#index (name)" as default format for `VolumeStructure` so a call to `fmtIndexAndName()` woudn't be necessary. Or perhaps we have a recommendation to use explicit formatting? I don't know, I'm new here.

gadget/gadget_test.go
```diff
@@ -639,13 +639,16 @@ func mustParseStructure(c *C, s string) *gadget.VolumeStructure {
 func (s *gadgetYamlTestSuite) TestValidateRole(c *C) {
 	uuidType := `
 type: 21686148-6449-6E6F-744E-656564454649
+size: 1023
 `
 	bareType := `
 type: mbr
+size: 446
```
cmatsuoka: It could be interesting to also test mbr with a size of 447, to make sure our limit works.

gadget/gadget.go
```diff
@@ -299,41 +317,99 @@ func validateVolume(name string, vol *Volume) error {
...
 	}
 
-	structureNames := make(map[string]bool, len(vol.Structure))
+	// named structures, for cross-referencing relative offset-write names
+	knownStructures := make(map[string]*PositionedStructure, len(vol.Structure))
+	// for validating structure overlap
+	structures := make([]PositionedStructure, len(vol.Structure))
+
+	lastOffset := Size(0)
+	farthestEnd := Size(0)
```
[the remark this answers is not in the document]
bboozzoo: Thanks for noticing. I went a bit further ahead with some other changes for positioning. This is not explicitly needed in validation, so I'll drop that.

gadget/gadget.go
bboozzoo: Structure name is optional. The helper was introduced in the previous drop of validation code, specifically to handle the case when name was not provided in gadget.yaml

cmatsuoka: I meant using the same logic of `fmtIndexAndName()` in `VolumeStructure`'s `String()` function, but that would be mostly an embellishment, and I don't know if it fits our current practices.

gadget/gadget.go
```diff
@@ -299,41 +317,99 @@ func validateVolume(name string, vol *Volume) error {
...
+		if end > farthestEnd {
+			farthestEnd = end
 		}
+		lastOffset = end
+	}
+
+	// sort by starting offset
+	sort.Sort(byStartOffset(structures))
+
+	lastEnd := Size(0)
```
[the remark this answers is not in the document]
cmatsuoka: I see you changed `lastOffset` in line 325, which is consistent with the proposed idea. There's also `lastEnd` in line 359 (that actually motivated the comment above) which could receive the same treatment.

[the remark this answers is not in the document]
bboozzoo: I've split out cross structure validation to a separate function and renamed that variable

gadget/gadget.go
bboozzoo: Actually `PositionedStructure` has all the info to make a proper string that contains the position of the structure within the original file. I've added a Stringer support to that type.

gadget/gadget.go
```diff
@@ -532,6 +620,9 @@ type Size uint64
 const (
 	SizeMiB = Size(2 << 20)
 	SizeGiB = Size(2 << 30)
+
+	SizeMBR          = Size(446)
```
zyga: Could you please add a comment for this. Even something as simple as `SizeMBR is the maximum size of the master boot record`.

gadget/gadget.go
```diff
@@ -454,7 +542,7 @@ func validateRole(vs *VolumeStructure, vol *Volume) error {
 			return fmt.Errorf(`role of this kind must have an implicit label or "writable", not %q`, vs.Label)
 		}
 	case "mbr":
-		if vs.Size > 446 {
+		if vs.Size > SizeMBR {
```
zyga: Can the MBR be smaller? Is there a smallest MBR?

gadget/gadget.go
```diff
@@ -532,6 +620,9 @@ type Size uint64
 const (
 	SizeMiB = Size(2 << 20)
 	SizeGiB = Size(2 << 30)
+
+	SizeMBR          = Size(446)
+	SizeLBA48Pointer = Size(4)
```
zyga: Same request here.

gadget/gadget.go
```diff
@@ -299,41 +329,99 @@ func validateVolume(name string, vol *Volume) error {
...
+	previousEnd := Size(0)
 	for idx, s := range vol.Structure {
 		if err := validateVolumeStructure(&s, vol); err != nil {
-			return fmt.Errorf("invalid structure %s: %v", fmtIndexAndName(idx, s.Name), err)
+			return fmt.Errorf("invalid structure %v: %v", fmtIndexAndName(idx, s.Name), err)
+		}
+		start := s.Offset
+		if start == 0 {
+			start = previousEnd
+		}
```
zyga: Since structures are a sequence I believe this is not correct.

Consider this example.
- structure A, size 10, offset 5
- structure B, size 5, offset 0

Reading this literally I would expect the structures to be back to back, with B being first, followed by A.

The code will make the structures be arranged A, B, with B starting at offset 15.

gadget/gadget.go
bboozzoo: Well, the truth it it's not the actual size of MBR (as in Master Boot Record), which has 512 bytes (1 sector), but more of the space available for bootstrap code in a classic MBR setup, which is 446. It also (incorrectly?) is referred to as `'mbr'` throughout ubuntu-image parser code and gadget.yaml.

I suppose, a description like `SizeMBR is the maximum size of a structure of 'mbr' role` is closest to reality/use.

gadget/gadget.go
bboozzoo: See my previous comment. The structure can be smaller than 446 bytes, though it's probably not useful.

gadget/gadget.go
bboozzoo: Pushed an update.

gadget/gadget.go
bboozzoo: Added a note.

gadget/gadget.go
bboozzoo: Added a note here as well.

gadget/gadget.go
```diff
@@ -299,41 +329,106 @@ func validateVolume(name string, vol *Volume) error {
...
+	// for validating structure overlap
+	structures := make([]PositionedStructure, len(vol.Structure))
+
+	previousEnd := Size(0)
 	for idx, s := range vol.Structure {
 		if err := validateVolumeStructure(&s, vol); err != nil {
-			return fmt.Errorf("invalid structure %s: %v", fmtIndexAndName(idx, s.Name), err)
+			return fmt.Errorf("invalid structure %v: %v", fmtIndexAndName(idx, s.Name), err)
 		}
+		var start Size
```
zyga: Thanks for fixing this!

gadget/gadget.go
```diff
@@ -299,41 +329,106 @@ func validateVolume(name string, vol *Volume) error {
...
+	// use structures positioned within the volume
+	for pidx, ps := range structures {
+		if ps.Role == "mbr" || ps.Type == "mbr" {
+			if ps.StartOffset != 0 {
+				return fmt.Errorf(`structure %v has "mbr" role and must start at offset 0`, ps)
+			}
+		}
+		if ps.OffsetWrite != nil && ps.OffsetWrite.RelativeTo != "" {
+			// offset-write using a named structure
+			other := knownStructures[ps.OffsetWrite.RelativeTo]
```
zyga: What happens if the structure refers to itself? I think this should be disallowed.

gadget/gadget.go
```diff
@@ -299,41 +329,106 @@ func validateVolume(name string, vol *Volume) error {
...
+			if other == nil {
+				return fmt.Errorf("structure %v refers to an unknown structure %q",
+					ps, ps.OffsetWrite.RelativeTo)
+			}
+			// offset is written as a 4-byte pointer value at offset-write address
+			if ps.OffsetWrite.Offset > (other.Size - SizeLBA48Pointer) {
```
zyga: I don't understand this part. Can you expand on the comment please?

gadget/gadget.go
```diff
@@ -299,41 +329,106 @@ func validateVolume(name string, vol *Volume) error {
...
 			// content relative offset only possible if it's a bare structure
 			continue
 		}
-		for cidx, c := range s.Content {
-			if c.OffsetWrite.RelativeTo != "" && !structureNames[c.OffsetWrite.RelativeTo] {
+		for cidx, c := range ps.Content {
+			if c.OffsetWrite == nil || c.OffsetWrite.RelativeTo == "" {
+				continue
+			}
+			relativeToStructure := knownStructures[c.OffsetWrite.RelativeTo]
```
zyga: Same remark about self-relative structures.

gadget/gadget.go
```diff
@@ -299,41 +329,106 @@ func validateVolume(name string, vol *Volume) error {
...
+					ps, fmtIndexAndName(cidx, c.Image), c.OffsetWrite.RelativeTo)
 			}
 		}
 	}
 	return nil
 }
 
 func validateVolumeStructure(vs *VolumeStructure, vol *Volume) error {
+	if vs.Size == 0 {
+		return errors.New("missing size")
```
zyga: Does this look OK in the final composed error?

bboozzoo: It's pretty specific: `invalid volume "foo": invalid structure #<idx> (<name>): missing size`

gadget/gadget.go
bboozzoo: Why? I don't see such check being enforced in `ubuntu-image`.

gadget/gadget.go
bboozzoo: Pushed a tweak.

gadget/gadget.go
```diff
@@ -388,8 +388,8 @@ func validateCrossVolumeStructure(structures []PositionedStructure, knownStructu
 				return fmt.Errorf("structure %v refers to an unknown structure %q",
 					ps, ps.OffsetWrite.RelativeTo)
 			}
-			// offset is written as a 4-byte pointer value at offset-write address
-			if ps.OffsetWrite.Offset > (other.Size - SizeLBA48Pointer) {
+			// make sure there is enough room left in the structure to write the pointer
```
zyga: So we actually write the pointer into the disk? I didn't anticipate that. Let's catch up tomorrow :)

zyga:
I think I now understand this better. Thank you for the explanation.

PR #6694, opened 2019-04-05
tests: improve how snaps are cached

sergiocazzolato:
This contains a mechanism to cache correctly the snaps to be used during test executions. The mechanism downloads the snap ans leaves it as a partial download to be resumed.
Also it is included an inprovement (speedup) for how the snapd state is restored.


zyga:
Quick sanity check question: what are the before/after timings for this branch? Does it actually speed things up now?

sergiocazzolato:
@zyga, thanks for your question:
1. running tests for ubuntu core 18 on external backend it improves between 1 and 2 minutes for each test which installs a core base snap.
2. based on travis, on google, it goes from 36 minutes (current average) to 33 minutes (with the inprovement)

tests/lib/get_snap_sha.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python3
+
+from urllib import request
+from pprint import pprint
+import json
+import sys
+
+
+if len(sys.argv) != 4:
+	print("Number of parameters extected is 2: snap, channel and architecture")
```
zyga: Two or three?

How about:
```
import argparse

parser = argparse.ArgumentParser()
parser.add_argument("snap")
parser.add_argument("channel")
parser.add_argument("arch")
opts = parser.parse_args()
```

Then you can use `opts.channel` instead of `channel`. It's just as long but the error messages are nicer and parsing is more complete.

tests/lib/get_snap_sha.py
```diff
@@ -0,0 +1,22 @@
...
+
+snap = sys.argv[1]
+channel = sys.argv[2]
+arch = sys.argv[3]
+
+info_request='https://api.snapcraft.io/v2/snaps/info/{}'.format(snap)
+snap_info=json.loads(request.urlopen(request.Request(info_request, headers={'Snap-Device-Series': '16'})).read().decode('utf-8'))
+
+for channel_info in snap_info.get('channel-map'):
+	if channel_info.get('channel').get('risk') == channel and channel_info.get('channel').get('architecture') == arch:
```
zyga: Why are you using `get` here? If the key is absent you will get `None` which doesn't have a `get` method anyway. Am I missing something?

tests/lib/prepare.sh
```diff
@@ -643,10 +648,24 @@ cache_snaps(){
...
-        snap install "$snap_name" || true
-        # The snap may fail during removal such as core snap on classic
-        snap remove "$snap_name" || true
+        snap download "$snap_name"
+        local snap_sha="$(python3 $TESTSLIB/get_snap_sha.py $snap_name stable $curr_arch)"
+        test -n "$snap_sha"
+
+        # Copy all of the snaps back to the spool directory. From there we
+        # will reuse them during subsequent `snap install` operations.
+        cp "${snap_name}"_*.snap /var/lib/snapd/snaps/
```
zyga: Can we hardline them instead? Each hardlink saves time.

tests/lib/systems.sh
```diff
@@ -27,3 +27,34 @@ is_ubuntu_14_system(){
...
+    x86_64)
+        echo 'amd64'
+        ;;
+    i386)
+        echo 'i386'
+        ;;
+    armv7l)
+        echo 'armhf'
+        ;;
+    aarch64*)
```
zyga: What is the * for?

tests/lib/systems.sh
```diff
@@ -27,3 +27,34 @@ is_ubuntu_14_system(){
...
+        echo 'arm64'
+        ;;
+    ppc64*)
+        echo 'ppc64el'
+        ;;
+    s390*)
```
zyga: This is probably incorrect, it is either s390 or s390x, those are different architectures.

tests/lib/systems.sh
```diff
@@ -27,3 +27,34 @@ is_ubuntu_14_system(){
...
+    i386)
+        echo 'i386'
+        ;;
+    armv7l)
+        echo 'armhf'
+        ;;
+    aarch64*)
+        echo 'arm64'
+        ;;
+    ppc64*)
```
zyga: What is the * for?

tests/lib/get_snap_sha.py
sergiocazzolato: This is not needed anymore, this file has been removed. Thanks for reviewing that :)

tests/lib/prepare.sh
sergiocazzolato: This is not being copied anymore.

tests/lib/systems.sh
sergiocazzolato: This function has been removed.

tests/lib/systems.sh
sergiocazzolato: This function has been removed.

tests/lib/get_snap_sha.py
sergiocazzolato: This file has been removed. There is a new implementation using .partial files which works much better.

tests/lib/systems.sh
sergiocazzolato: This function has been removed.

sergiocazzolato:
@zyga Thanks for reviewing this. Finally I implemented a method to download the snap and leave it as a .partial file in /var/lib/snapd/snaps, so when snapd needs a snap finds there the .partial ans resumes the download which is complete.
The only problem is that when we install with --dangerous it is downloading the base snap independently if it has a partial fil for hte base snap.

tests/lib/prepare.sh
```diff
@@ -621,10 +621,16 @@ prepare_ubuntu_core() {
     fi
 
     echo "Ensure the core snap is cached"
+    # Cache snaps
     if is_core18_system; then
         if ! snap list core; then
             cache_snaps core
         fi
+        if snap list core; then
```
bboozzoo: This `if` and the one above cannot be true at the same time. Maybe refactor this to:
```
if is_core18_system; then
    if snap list core;
        echo "core snap on core18 should not be installed yet"
        snap list
        exit 1
    fi
    cache_snaps core
fi
```

sergiocazzolato: done

sergiocazzolato: Thanks for the review

sergiocazzolato:
@chipaca thanks for the review, I'll try to address that once it is available.

tests/lib/prepare.sh
```diff
@@ -637,10 +637,14 @@ prepare_ubuntu_core() {
     fi
 
     echo "Ensure the core snap is cached"
+    # Cache snaps
     if is_core18_system; then
-        if ! snap list core; then
-            cache_snaps core
+        if snap list core; then
```
zyga: This will display information about core. Perhaps you want to redirect the output to /dev/null?

tests/lib/prepare.sh
```diff
@@ -658,19 +662,18 @@ prepare_ubuntu_core() {
...
+        snap download "$snap_name"
+
         # Copy all of the snaps back to the spool directory. From there we
         # will reuse them during subsequent `snap install` operations.
-        cp -- *.snap /var/lib/snapd/snaps/
-        set +x
-    )
+        snap_file=$(ls "${snap_name}"_*.snap)
+        mv "${snap_file}" /var/lib/snapd/snaps/"${snap_file}".partial
+        rm -f "${snap_name}"_*
```
zyga: If this is removing the assert file then please adjust the glob. If this is doing something else can you please add a comment explaining it?

tests/lib/prepare.sh
sergiocazzolato: Done

tests/lib/prepare.sh
sergiocazzolato: done

sergiocazzolato:
@zyga, comments addressed, thanks for reviewing!!

PR #6695, opened 2019-04-06
interfaces: add jack1 implicit classic interface

ymauray:
This is a simple interface to allow `jack` client application (built with `libjack`) to connect to the `jack` server running outside the snap environment, using shared memory.

`libjack` builds the shared memory endpoint like this : `jack-<userid>/<servername>/jack-<serverid>`

`jackd` needs to be running before a client can connect. In order to run `jackd` (I'm using `jack1`), first determine what soundcards are available (here with the results on my machine) :
```
$ aplay -L 2>/dev/null | grep ^hw:CARD | grep DEV=0 | sed 's/.*=\([^,]*\),.*/\1/g
PCH
CODEC
Headset
```
Then, `jack` is started like this (I've chosen the `Headset` sound card) : 
```
/usr/bin/jackd -dalsa -r44100 -p128 -n4 -D -Chw:Headset -Phw:Headset
```
I have made [a very simple client](https://github.com/frenchguych/jack-test) to check that this interface works.


interfaces/builtin/jack.go
```diff
@@ -0,0 +1,52 @@
...
+
+/*
+ * libjack builds the shared memory endpoint like this :
+ * jack-<userid>/<server name>/jack-<server id>
+ *
+ * see libjack/shm.c in jack1's source tree,
+ * and common/shm.c in jack2's source tree.
+ */
+const jackConnectedPlugAppArmor = `
+/dev/shm/jack-*/*/* mrwk,
```
jdstrand: Not a full review, but I believe this can be changed to:
```
owner /dev/shm/jack-[0-9]*/*/* mrwk,
```

but I also wonder why you need 'm' and 'k'. What are the policy violations you see in devmode when this interface is not connected?

Do you have a test snap and instructions on how to setup jack so that I may poke at this a little bit?

ymauray: To be honest I took thise from the pulse interface. I'm happy to change that and retest, then I'll update the PR. 

As for the test snap and the configuration for jackd, I believe you have everything necessary in the PR description. Let me know if you need anything else.

ymauray: I've changed the profile as per your suggestions.
I've also updated [my test snap](https://github.com/frenchguych/jack-test) so it now plays a 880 Hz tone @ 60 bpm. Everything sounds good on my machine.

interfaces/builtin/jack.go
```diff
@@ -0,0 +1,52 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2016 Canonical Ltd
```
zyga:
```suggestion
 * Copyright (C) 2019 Canonical Ltd
```

interfaces/builtin/jack.go
```diff
@@ -0,0 +1,52 @@
...
+/*
+ * libjack builds the shared memory endpoint like this :
+ * jack-<userid>/<server name>/jack-<server id>
+ *
+ * see libjack/shm.c in jack1's source tree,
+ * and common/shm.c in jack2's source tree.
+ */
+const jackConnectedPlugAppArmor = `
+owner /dev/shm/jack-[0-9]*/*/* rw,
+`
```
zyga: Should the jack interface carry system calls for connecting to the socket? I understand those are easily granted by network but perhaps they should be explicitly allowed here as well?

interfaces/builtin/jack_test.go
```diff
@@ -0,0 +1,93 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2016 Canonical Ltd
```
zyga:
```suggestion
 * Copyright (C) 2019 Canonical Ltd
```

interfaces/builtin/jack.go
ymauray: I will need help on this, as I'm really not an expert. Do you mean I should add "shmctl", as in pulseaudio.go ?

So far in my test snap I only have the `jack` plug, and it works fine :

```
apps:
  jack-client-test:
    command: jack-client-test
    plugs:
      - jack
```

interfaces/builtin/jack.go
ymauray: Fixed

interfaces/builtin/jack_test.go
ymauray: Fixed

interfaces/builtin/jack.go
zyga: I think we should have enough permissions to open the socket and talk to it. This includes things like calling `socket` and `connect` with appropriate arguments. Perhaps one thing that would help to flesh this out would be a simple test that shows one can indeed talk to the Jack socket?

A trivial C program that opens the socket and asks for, say, protocol version, or something equivalent, would be beneficial to the confidence in completeness of this pull request.

ymauray: I believe you have everything you need in the pull request description. [This is the snap I use](https://github.com/frenchguych/jack-test/) to test my changes on my machine. It connects to the jack server and emits a beep every second.
If that is not clear or not enougth, I'd be happy to provide another test case.

zyga: Ah, I missed that, sorry. I'm somewhat used to having terse description where all the interesting things are in the pull request diff. I will re-process this with the test snap in place. Thanks!

ymauray:
I have signed the CLA and I am a member of the contributor-agreement-canonical group, so that check should pass on next build.

interfaces/builtin/jack.go
```diff
@@ -0,0 +1,52 @@
...
+
+/*
+ * libjack builds the shared memory endpoint like this :
+ * jack-<userid>/<server name>/jack-<server id>
+ *
+ * see libjack/shm.c in jack1's source tree,
+ * and common/shm.c in jack2's source tree.
+ */
+const jackConnectedPlugAppArmor = `
+owner /dev/shm/jack-[0-9]*/*/* rw,
```
jdstrand: I am not familiar with jack, so I googled and came across https://help.ubuntu.com/community/UbuntuStudioPreparation#Jack_-_jackd which suggests installing qjackctl, so in and Ubuntu 18.04 Desktop VM, I did, launched it and pressed 'start' and found the following in /dev/shm:
```
jack-1000-0          jack_sem.1000_default_dbusapi    jack_sem.1000_default_system
jack-1000-1          jack_sem.1000_default_freewheel  jack-shm-registry
jack_default_1000_0  jack_sem.1000_default_qjackctl
```

As you can see, none of these paths works with the specified rule and it isn't clear to me which of these files the snap should have access to. Additionally, jack-client-test couldn't find the files in the first place:
```
connect(2) call to /dev/shm/jack-1000/default/jack_0 failed (err=No such file or directory)
attempt to connect to server failed
jack_client_open() failed, status = 0x11
Unable to connect to JACK server
```

jdstrand: This makes it more clear what all those files are:
```
$ ls -l /dev/shm
total 23060
-rw-r--r-- 1 jamie jamie 82280346 Apr 11 17:00 jack-1000-0
-rw-r--r-- 1 jamie jamie     1187 Apr 11 17:00 jack-1000-1
srwxr-xr-x 1 jamie jamie        0 Apr 11 17:00 jack_default_1000_0
-rwxr-xr-x 1 jamie jamie       12 Apr 11 17:00 jack_sem.1000_default_dbusapi
-rwxr-xr-x 1 jamie jamie       12 Apr 11 17:00 jack_sem.1000_default_freewheel
-rwxr-xr-x 1 jamie jamie       12 Apr 11 17:00 jack_sem.1000_default_qjackctl
-rwxr-xr-x 1 jamie jamie       12 Apr 11 17:00 jack_sem.1000_default_system
-rw-rw-rw- 1 jamie jamie    70744 Apr 11 16:47 jack-shm-registry
```

ymauray: Interesting. I don't have that at all, even though I'm running jack through QJackCtl too :
```
yannick@heimdall:~$ ls -l /dev/shm/
total 2124
drwx------ 3 yannick yannick      60 avr 12 00:07 jack-1000
-rwxrwxr-x 1 yannick yannick    1040 avr  8 22:11 u1000-Shm_125376fc
-rwxrwxr-x 1 yannick yannick   32784 avr  8 22:11 u1000-Shm_61df3bf5
-rwxrwxr-x 1 yannick yannick   32784 avr  8 22:11 u1000-Shm_78b2a20b
-rwxrwxr-x 1 yannick yannick 2097152 avr  8 22:11 u1000-ValveIPCSharedObj-Steam
```
Only `jack-1000` is relevant here, and matches what can be found in `libjack`'s source code.

ymauray: Oh ! I bet you are using jack2. I'm using jack1. 
The naming of the shm endpoint apparently changed between those 2 versions.
jack1 : `libjack/shm.c`, function `jack_shmalloc` : 
```c
snprintf (name, sizeof(name), "/jack-%d", registry->index);
```

jack2 : `common/shm.c`, function `jack_shmalloc` : 
```c
snprintf (name, sizeof (name), "jack-%d-%d", GetUID(), registry->index);
```

So I guess this interface is only for jack1.

ymauray: Yep, confirmed : here's what I have when running jack2 :
```
$ ls -l /dev/shm
-rw-r--r-- 1 yannick yannick 82280346 avr 12 00:32 jack-1000-0
-rw-r--r-- 1 yannick yannick     1187 avr 12 00:32 jack-1000-1
srwxr-xr-x 1 yannick yannick        0 avr 12 00:32 jack_default_1000_0
-rwxr-xr-x 1 yannick yannick       12 avr 12 00:32 jack_sem.1000_default_freewheel
-rwxr-xr-x 1 yannick yannick       12 avr 12 00:32 jack_sem.1000_default_system
-rw-rw-rw- 1 yannick yannick    70744 avr 12 00:32 jack-shm-registry
....
```

FYI, I've tested the following applications :
- iDJC (streamer/recorder)
- Ardour5 (DAW)
- Catia (patchbay)

None of them work with jack2. They work wonderfully with jack1

ymauray:
I'm going to need help to figure out what went wrong with the Travis job. For some reasons, I can't run the spread tests on my machine.

chipaca:
The problem in the spread test is
```
+ snap connect test-snapd-policy-app-consumer:jack1 test-snapd-policy-app-provider-core:jack1
error: snap "test-snapd-policy-app-provider-core" has no slot named "jack1"
```
you probably need to update the `test-snapd-policy-app-provider-core` snap (in `tests/lib/snaps/test-snapd-policy-app-provider-core`), or perhaps the test in `tests/main/interfaces-many-snap-provided/` which is the one that failed. HTH.

chipaca:
To run the spread test, the easiest way would be, IMHO,
```
sudo apt install qemu-kvm autopkgtest
mkdir -p ~/.spread/qemu
cd ~/.spread/qemu
adt-buildvm-ubuntu-cloud -r xenial
mv adt-xenial-amd64-cloud.img ubuntu-16.04.img
cd ~/bin/
wget -O- -nv https://niemeyer.s3.amazonaws.com/spread-amd64.tar.gz  | tar xzvv
spread -v -debug qemu:ubuntu-core-16-64:tests/main/interfaces-many-snap-provided
```
HTH

ymauray:
Hi ! Thanks, I will try that this evening. Hopefully I'll figure out where I made a mistake and I'll re-push on my branch.

ymauray:
It seems adt-buildvm-ubuntu-cloud doesn't exist on 18.04...

chipaca:
Ah, in 18.04 it's called `autopkgtest-buildvm-ubuntu-cloud`.

Hm, looks like in xenial the `adt-` one is just a symlink to the `autopkgtest-` one. I should update the HACKING.md doc.

ymauray:
Ah, yes indeed ! I've downloaded the image. I'll get back to it this evening. thanks for the help !!

interfaces/builtin/jack.go
jdstrand: I've been wanting to look at this and see if we could support both jack1 and jack2 in the same interface. While I've not been able to do this yet, I just wanted to let others know what I was thinking. Hopefully I can get to this soon (else, @ymauray, feel free to push on jack2 in the meantime if you're so inclined).

ymauray: Interesting, I guess I could try and see if I can make that work for both jack 1 and jack 2.
I was, however, under the impression that jack2 was not as popular as jack1.
It is definitely not compatible.
Therefore, is it wise to have one interface for 2 essentially different things ?

jdstrand: This is a good point especially when considering a future slot implementation that provides the jack server slot for consumers (plugs). From a pure policy POV for an implicit classic interface where we don't know what the server is (since it is provided as a deb), I was thinking it was ok to treat it as more of an apparmor abstraction. However interfaces are not just collections of rules and interfaces were designed with the idea that different api-breaking slot implementations would need different interface since interfaces form a contract between the consumer and the provider. In this case, if we did as I suggested (where in the future we implement the non-implicit interface code) then a user might install a jack2 provider and a jack1 consumer, use 'snap connect' to connect them and things are broken.

As such, I'll review this PR with only jack1 in mind, and if you or someone else is so inclined, afterward someone can submit a jack2 PR.

jdstrand: > I am not familiar with jack, so I googled and came across https://help.ubuntu.com/community/UbuntuStudioPreparation#Jack_-_jackd which suggests installing qjackctl, so in an Ubuntu 18.04 Desktop VM, I did, launched it and pressed 'start' and found the following in /dev/shm:

To verify JACK1, I installed jackd1, launched qjackctl, went into Settings, unchecked 'realtime', then pressed 'Start' and found the following in /dev/shm:
```
$ ls -lR jack-1000/
jack-1000/:
total 0
drwx------ 2 jamie jamie 140 May 24 08:44 default

jack-1000/default:
total 0
srwxr-xr-x 1 jamie jamie 0 May 24 08:44 jack_0
srwxr-xr-x 1 jamie jamie 0 May 24 08:44 jack_ack_0
prw-r--r-- 1 jamie jamie 0 May 24 08:44 jack-ack-fifo-8641-0
prw-r--r-- 1 jamie jamie 0 May 24 08:44 jack-ack-fifo-8641-1
prw-r--r-- 1 jamie jamie 0 May 24 08:44 jack-ack-fifo-8641-2
```

I then installed jack-client-test, which has no plugs, and added the above apparmor rules and was able to run the client. Therefore, all required seccomp rules are in the default template.

interfaces/builtin/jack.go
jdstrand: > Should the jack interface carry system calls for connecting to the socket? I understand those are easily granted by network but perhaps they should be explicitly allowed here as well?

See my comment above. Nothing else is needed beyond what is already granted in the default template.

interfaces/builtin/jack_test.go
```diff
@@ -0,0 +1,93 @@
...
+	s.slotInfo = &snap.SlotInfo{
+		Snap:      &snap.Info{SuggestedName: "core", Type: snap.TypeOS},
+		Name:      "jack",
+		Interface: "jack1",
+	}
+	s.slot = interfaces.NewConnectedSlot(s.slotInfo, nil, nil)
+	snapInfo := snaptest.MockInfo(c, mockPlugSnapInfoYaml, nil)
+	s.plugInfo = snapInfo.Plugs["jack1"]
+	s.plug = interfaces.NewConnectedPlug(s.plugInfo, nil, nil)
+}
```
jdstrand: All of the above can be simplified with the more modern (within the codebase; based on u2f_devices_test.go):
```
type jack1InterfaceSuite struct {
	testutil.BaseTest

	iface    interfaces.Interface
	slotInfo *snap.SlotInfo
	slot     *interfaces.ConnectedSlot
	plugInfo *snap.PlugInfo
	plug     *interfaces.ConnectedPlug
}

var _ = Suite(&jack1InterfaceSuite{
	iface: builtin.MustInterface("jack1"),
})

const jack1ConsumerYaml = `name: consumer
version: 0
apps:
 app:
  plugs: [jack1]
`

const jack1CoreYaml = `name: core
version: 0
type: os
slots:
  jack1:
`

func (s *jack1InterfaceSuite) SetUpTest(c *C) {
	s.plug, s.plugInfo = MockConnectedPlug(c, jack1ConsumerYaml, nil, "jack1")
        s.slot, s.slotInfo = MockConnectedSlot(c, jack1CoreYaml, nil, "jack1")
}
```

interfaces/builtin/jack_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (s *JackInterfaceSuite) TestName(c *C) {
+	c.Assert(s.iface.Name(), Equals, "jack1")
+}
+
+func (s *JackInterfaceSuite) TestSanitizeSlot(c *C) {
+	c.Assert(interfaces.BeforePrepareSlot(s.iface, s.slotInfo), IsNil)
+	slot := &snap.SlotInfo{
+		Snap:      &snap.Info{SuggestedName: "some-snap"},
+		Name:      "jack",
```
jdstrand: s/jack/jack1/

interfaces/builtin/jack_test.go
```diff
@@ -0,0 +1,93 @@
...
+		Interface: "jack1",
+	}
+	c.Assert(interfaces.BeforePrepareSlot(s.iface, slot), ErrorMatches,
+		"jack1 slots are reserved for the core snap")
```
jdstrand: I think that this can be all on one line, unless `go fmt` did this for you.

interfaces/builtin/jack_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"github.com/snapcore/snapd/interfaces"
+	"github.com/snapcore/snapd/interfaces/apparmor"
+	"github.com/snapcore/snapd/interfaces/builtin"
+	"github.com/snapcore/snapd/snap"
+	"github.com/snapcore/snapd/snap/snaptest"
+	"github.com/snapcore/snapd/testutil"
+)
+
+type JackInterfaceSuite struct {
```
jdstrand: Note, there is no reason for this to be public. Use jack1InterfaceSuite here and below instead.

interfaces/builtin/jack_test.go
```diff
@@ -0,0 +1,93 @@
...
+	c.Assert(interfaces.BeforePreparePlug(s.iface, s.plugInfo), IsNil)
+}
+
+func (s *JackInterfaceSuite) TestUsedSecuritySystems(c *C) {
+	apparmorSpec := &apparmor.Specification{}
+	err := apparmorSpec.AddConnectedPlug(s.iface, s.plug, s.slot)
+	c.Assert(err, IsNil)
+	c.Assert(apparmorSpec.SecurityTags(), DeepEquals, []string{"snap.other.app"})
+	c.Assert(apparmorSpec.SnippetForTag("snap.other.app"), testutil.Contains, "owner /dev/shm/jack-[0-9]*/*/* rw")
+}
```
jdstrand: This is fine but is using the older style. I suggest:
```
func (s *jack1InterfaceSuite) TestAppArmorSpec(c *C) {
        spec := &apparmor.Specification{}
        c.Assert(spec.AddConnectedPlug(s.iface, s.plug, s.slot), IsNil)
        c.Assert(spec.SecurityTags(), DeepEquals, []string{"snap.consumer.app"})
        c.Assert(spec.SnippetForTag("snap.consumer.app"), testutil.Contains, `owner /dev/shm/jack-[0-9]*/*/* rw,`)
}
```

interfaces/builtin/jack_test.go
```diff
@@ -0,0 +1,93 @@
...
+
```
jdstrand: Please add:
```
func (s *jack1InterfaceSuite) TestStaticInfo(c *C) {
	si := interfaces.StaticInfoOf(s.iface)
	c.Assert(si.ImplicitOnCore, Equals, false)
	c.Assert(si.ImplicitOnClassic, Equals, true)
	c.Assert(si.Summary, Equals, `allows interacting with a JACK1 server`)
	c.Assert(si.BaseDeclarationSlots, testutil.Contains, "jack1")
}
```

ymauray:
> Thanks again for the PR. Just a few more changes and then we can finalize the name and get others to review.

Thanks !! I will make the changes over the week-end.

interfaces/builtin/jack.go
```diff
@@ -0,0 +1,51 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package builtin
+
+const jackSummary = `allows access to the jack1 socket`
+
+const jackBaseDeclarationSlots = `
+  jack1:
```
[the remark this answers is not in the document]
ymauray: http://www.jackaudio.org/downloads/ uses jack1 and jack2.
Given that there will probably be 2 interfaces, one for each version, I'd go with jack1 and jack2

ymauray:
> you probably need to update the `test-snapd-policy-app-provider-core` snap (in `tests/lib/snaps/test-snapd-policy-app-provider-core`), or perhaps the test in `tests/main/interfaces-many-snap-provided/` which is the one that failed. HTH.

I just noticed that I did update that file and it is part of the pull request. 
Running the spread tests on my machine gives the same error, but I'm lost... I don't know how to fix this.

interfaces/builtin/jack.go
```diff
@@ -0,0 +1,51 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package builtin
+
+const jackSummary = `allows access to the jack1 socket`
```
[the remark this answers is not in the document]
ymauray: Fixed in 07c970d546a87f000834aa9c6efaa89feb6c6e60

interfaces/builtin/jack.go
```diff
@@ -0,0 +1,51 @@
...
+    deny-auto-connection: true
+`
+
+/*
+ * libjack builds the shared memory endpoint like this :
+ * jack-<userid>/<server name>/jack-<server id>
+ *
+ * see libjack/shm.c in jack's source tree
+ */
+const jackConnectedPlugAppArmor = `
```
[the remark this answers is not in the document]
ymauray: Fixed in 07c970d546a87f000834aa9c6efaa89feb6c6e60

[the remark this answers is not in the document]
ymauray: Fixed in 07c970d546a87f000834aa9c6efaa89feb6c6e60

interfaces/builtin/jack_test.go
ymauray: Fixed in 07c970d546a87f000834aa9c6efaa89feb6c6e60

interfaces/builtin/jack_test.go
ymauray: Fixed in 07c970d546a87f000834aa9c6efaa89feb6c6e60

interfaces/builtin/jack_test.go
ymauray: Fixed in 07c970d546a87f000834aa9c6efaa89feb6c6e60

interfaces/builtin/jack_test.go
ymauray: Fixed in 07c970d546a87f000834aa9c6efaa89feb6c6e60

interfaces/builtin/jack_test.go
ymauray: Fixed in 07c970d546a87f000834aa9c6efaa89feb6c6e60

interfaces/builtin/jack.go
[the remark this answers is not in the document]
pedronis: I think jack1 and jack2 are fine names.

interfaces/builtin/jack.go
```diff
@@ -0,0 +1,47 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
```
jdstrand: Please rename this file jack1.go

interfaces/builtin/jack.go
```diff
@@ -0,0 +1,47 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package builtin
+
+const jackSummary = `allows interacting with a JACK1 server`
```
jdstrand: Adjust to be jack1Summary

interfaces/builtin/jack.go
```diff
@@ -0,0 +1,47 @@
...
+
+const jackBaseDeclarationSlots = `
```
jdstrand: Adjust to be jack1BaseDeclarationSlots

interfaces/builtin/jack.go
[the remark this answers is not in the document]
jdstrand: Ok, jack1 it is. Please rename the files accordingly and update everything to use s/jack/jack1/ (I'll comment where).

interfaces/builtin/jack.go
```diff
@@ -0,0 +1,47 @@
...
+  jack1:
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
+`
+
+const jackConnectedPlugAppArmor = `
```
jdstrand: Please adjust to be jack1ConnectedPlugAppArmor

interfaces/builtin/jack.go
```diff
@@ -0,0 +1,47 @@
...
+owner /dev/shm/jack-[0-9]*/*/* rw,
+`
+
+func init() {
+	registerIface(&commonInterface{
+		name:                  "jack1",
+		summary:               jackSummary,
+		implicitOnClassic:     true,
+		baseDeclarationSlots:  jackBaseDeclarationSlots,
+		connectedPlugAppArmor: jackConnectedPlugAppArmor,
```
jdstrand: Also adjust these to be jack1Summary, jack1BaseDeclarationSlots and jack1ConnectedPlugAppArmor.

interfaces/builtin/jack_test.go
```diff
@@ -0,0 +1,99 @@
...
+	var mockPlugSnapInfoYaml = `name: other
+version: 1.0
+apps:
+ app:
+  command: foo
+  plugs: [jack1]
+`
+	s.slotInfo = &snap.SlotInfo{
+		Snap:      &snap.Info{SuggestedName: "core", Type: snap.TypeOS},
+		Name:      "jack",
```
jdstrand: Please adjust to be "jack1".

interfaces/builtin/jack_test.go
jdstrand: I won't block on this, but do recommend you make this change.

interfaces/builtin/jack_test.go
```diff
@@ -0,0 +1,99 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
```
jdstrand: Please rename to jack1_test.go

interfaces/builtin/jack.go
ymauray: Done.

interfaces/builtin/jack.go
ymauray: Done.

interfaces/builtin/jack.go
ymauray: Done.

interfaces/builtin/jack.go
ymauray: Done.

interfaces/builtin/jack.go
ymauray: Done.

interfaces/builtin/jack_test.go
ymauray: Done.

interfaces/builtin/jack_test.go
ymauray: Done.

ymauray:
> Thanks for making the changes! I missed that you did the latest batch so sorry for the delay. LGTM

No problem !

> The spread test failures look unrelated. I just now retried them.

It seems they are still failling. I had another look but still can't find a way to fix them.

jdstrand:
You could fetch/merge from trunk and see if that helps. If not, we can keep mashing retry.

jdstrand:
This is a legitimate failure:

```
++ echo test-snapd-policy-app-consumer.jack1
+ plug_iface=test-snapd-policy-app-consumer:jack1
+ CONNECTED_PATTERN='test-snapd-policy-app-provider-core:jack1 +test-snapd-policy-app-consumer'
+ echo 'When slot test-snapd-policy-app-provider-core:jack1 is connected'
When slot test-snapd-policy-app-provider-core:jack1 is connected
+ snap connect test-snapd-policy-app-consumer:jack1 test-snapd-policy-app-provider-core:jack1
error: snap "test-snapd-policy-app-provider-core" has no slot named "jack1"
```

Remove your changes to tests/lib/snaps/test-snapd-policy-app-provider-core/meta/snap.yaml and this should pass (you only add interfaces here if they can be provided by a slot implementation, but jack1 is an implicit interface). This should've been caught in previous reviews, sorry.

ymauray:
Done, and the spread tests passed !! \o/

ymauray:
So what's the next move ? Do I have anything to do on my side ?

ymauray:
Thanks ! Am I allowed to brag about this ? :-)
I guess I shall start working on the jack2 interface now !

chipaca:
> Am I allowed to brag about this ?

bragging is mandatory.

PR #6697, opened 2019-04-08
interfaces/daemon_notify: add {net,sys}_admin capabilities, update spread test

anonymouse64:
Fixes LP bug [1823580](https://bugs.launchpad.net/snapd/+bug/1823580)

The `net_admin` and `sys_admin` capabilities are necessary for `systemd-notify` to actually send a message to the notify socket with the appropriate PID, such that a `Type=notify` daemon will actually be marked as active after doing so. 

I'm reasonably certain (?) that the spread test is now functionally exercising the `daemon-notify` interface, however it's quite likely I have my "regex inside shell string inside yaml string" quoting incorrect, so feel free to adjust that as necessary. 

Also, it would be really nice to get this into a 2.38.X release, but it's not critical at the moment.


interfaces/builtin/daemon_notify.go
```diff
@@ -44,6 +44,9 @@ const daemonNotifyConnectedPlugAppArmorTemplate = `
 
 # Allow using systemd-notify in shell scripts.
 /{,usr/}bin/systemd-notify ixr,
+
+capability sys_admin,
```
zyga: I'm sceptical about those being actually required. Can you justify this somehow? What was the denial you had otherwise observed?

tests/lib/snaps/test-snapd-daemon-notify/bin/notify
```diff
@@ -1,3 +1,5 @@
 #!/bin/bash
 
-systemd-notify --ready --status="Test service ready" || true
+systemd-notify --ready --status="Test service ready" --pid="$BASHPID" || true
```
zyga: Could we use `$$` instead of `$BASHPID` here?

tests/lib/snaps/test-snapd-daemon-notify/meta/hooks/install
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -e
+
+# stop and disable the service so that the install can continue
+snapctl stop --disable "$SNAP_NAME.notify"
```
zyga: Isn't it sufficient to disable the service? It won't be started yet anyway, right?

tests/lib/snaps/test-snapd-daemon-notify/meta/hooks/install
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -e
```
zyga: Nitpick: this could be a plain shell script.

tests/lib/snaps/test-snapd-daemon-notify/meta/snap.yaml
```diff
@@ -3,8 +3,12 @@ version: 1.0
 summary: Basic daemon-notify snap
 description: A basic snap which allow sending notification messages to systemd through the notify socket
 
+hooks:
+  install:
+    plugs: [network, network-bind]
```
zyga: Why do we need to use both network and network-bind here?

tests/main/interfaces-daemon-notify/task.yaml
```diff
@@ -21,14 +19,15 @@ execute: |
     echo "The interface is not connected by default"
     snap interfaces -i daemon-notify | MATCH -- "- +test-snapd-daemon-notify:daemon-notify"
 
+    echo "The service cannot be started with the interface disconnected"
+    snap start --enable test-snapd-daemon-notify.notify
```
zyga: Nitpick: I think `start` is sufficient, `--enable` is not needed since we are not trying to enable the unit to be active across reboots.

anonymouse64:
@zyga Yes see these do happen directly from the `systemd-notfiy` binary (from the LP bug):
```
Apr 07 19:38:25 audit[28390]: AVC apparmor="DENIED" operation="capable" profile="snap.edgexfoundry.consul" pid=28390 comm="systemd-notify" capability=12 capname="net_admin"
```
(note that there's an additional `/proc/1/environ` access from the LP bug that I don't think is necessary for systemd-notify to work properly) 

As explained in the linked LP bug, after providing `net_admin`, it then proceeds to fail on `sys_admin`.

From a justification perspective, I think it's necessary because systemd-notify is running as a different process than the one that systemd started as the daemon, and so systemd-notify needs these permissions in order to prove to systemd that it has permission to update the status of the specified PID. Without these permissions it wouldn't be clear to systemd if this request came from somewhere else. As it is now, the only way to use `systemd-notify` is to `exec systemd-notify --ready` or to write a C program using `sd_notify` directly, but this severely limits the usefulness of this interface. For example, the customer use case we have is an upstream daemon written in Go, and so we would have to write a C program wrapper that forks off the Go daemon, then calls `sd_notify` directly, then waits on the original fork to exit... instead of a much simpler shell script that calls `systemd-notify --ready --pid-$BASHPID` like we have now. Perhaps this was intentional? 

As an aside, the other way we could make this work is by changing the `NotifyAccess` setting in the generated systemd unit to be something like `NotifyAccess=all`. I don't think we want to do this however as then snaps that use `daemon-notify` would be able to send updates about any other snap that was `daemon: notify`. Currently since we don't specify `NotifyAccess`, it defaults to `NotifyAccess=main`, which expects that the PID of the process that systemd started for the process is the one sending the status update message. Lastly, I have not checked, but it could be the case that with these new capabilities, a snap can already send updates on behalf of another snap daemon that is `daemon: notify`. I will do a test to see if that's the case and report back

tests/lib/snaps/test-snapd-daemon-notify/meta/snap.yaml
anonymouse64: Eh, whenever I use `network` I always also add `network-bind` as a matter of taste, but it's not strictly necessary

tests/lib/snaps/test-snapd-daemon-notify/meta/hooks/install
anonymouse64: No, we need to specifically *disable* the service here in the install hook because if we don't then the installation of the snap will be aborted when the service is started up and it doesn't have permissions to use daemon-notify and exits.

/me remembers that @zyga was the one who first told me about this feature :-)

tests/lib/snaps/test-snapd-daemon-notify/bin/notify
anonymouse64: No, as `$$` will be the PID of the `systemd-notify` process, not of the parent bash process that systemd started. We need the pid to specifically be whatever systemd started when it used `Exec=` from the service unit file

interfaces/builtin/daemon_notify.go
anonymouse64: Yes see my other comment or the LP bug but these denials come from the `systemd-notify` binary directly.

anonymouse64:
So after some testing, when I add `net_admin` and `sys_admin` capabilities to `daemon-notify`, then a snap is able to notify systemd about the status of a different snap's services. To test this I did:

1. Build and install a new snapd binary with `net_admin` and `sys_admin` added to the daemon-notify AppArmor snippet.
2. Make a duplicate test-snapd-daemon-notify snap (called `test-snapd-daemon-notify2`) with the following changes:
* Add an empty configure hook (so we can use `snap set test-snapd-daemon-notify2`)
* In the `notify` script use `snapctl get pid` to determine what pid to use with `--pid` when the service is started
3. Modify the original test-snapd-daemon-notify to have the `notify` binary do `sleep 100000` before calling `systemd-notify`. This is to ensure that this snap won't ever successfully mark itself as active in order to allow the other snap to mark it as active
4. Install both snaps
5. Start the original snap with `snap start test-snapd-daemon-notify.notify` (it will now sit waiting forever
6. Using ps, find the PID of the `test-snapd-daemon-notify.notify` snap and provide this to `test-snapd-daemon-notify2` with `snap set test-snapd-daemon-notify2 pid=$THE_OTHER_SNAP_SVC_PID`
7. Start the duplicate snap with `snap start test-snapd-daemon-notify2.notify`
8. Note that the original `test-snapd-daemon-notify.notify` daemon is now immediately marked by systemd as ready by the duplicate snap.

See https://pastebin.ubuntu.com/p/R5ktwryN8V/ for the full diff between the two snaps.

If I run the above test without `net_admin` and without `sys_admin`, then the duplicate snap isn't allowed to notify the first one, failing with:

```
Apr 08 08:51:09 audit[17763]: AVC apparmor="DENIED" operation="open" profile="snap.test-snapd-daemon-notify2.notify" name="/proc/1/environ" pid=17763 comm="systemd-notify" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
Apr 08 08:51:09 audit[17763]: AVC apparmor="DENIED" operation="capable" profile="snap.test-snapd-daemon-notify2.notify" pid=17763 comm="systemd-notify" capability=12  capname="net_admin"
Apr 08 08:51:09 audit[17763]: AVC apparmor="DENIED" operation="capable" profile="snap.test-snapd-daemon-notify2.notify" pid=17763 comm="systemd-notify" capability=21  capname="sys_admin"
Apr 08 08:51:09 systemd[1]: snap.test-snapd-daemon-notify2.notify.service: Got notification message from PID 17763, but reception only permitted for main PID 17737
Apr 08 08:51:09 kernel: audit: type=1400 audit(1554731469.180:568): apparmor="DENIED" operation="open" profile="snap.test-snapd-daemon-notify2.notify" name="/proc/1/environ" pid=17763 comm="systemd-notify" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
Apr 08 08:51:09 kernel: audit: type=1400 audit(1554731469.180:569): apparmor="DENIED" operation="capable" profile="snap.test-snapd-daemon-notify2.notify" pid=17763 comm="systemd-notify" capability=12  capname="net_admin"
Apr 08 08:51:09 kernel: audit: type=1400 audit(1554731469.180:570): apparmor="DENIED" operation="capable" profile="snap.test-snapd-daemon-notify2.notify" pid=17763 comm="systemd-notify" capability=21  capname="sys_admin"
```

The test above was done on Ubuntu 18.04 desktop.

I will point out that this leak was specifically crafted to find this vulnerability and in practice I think the impact of allowing one snap to send status updates for is likely low (but there could be some kind of attack that re-ordered services that exposes some other vulnerability), but I'll let others (such as @jdstrand ) decide on whether it's acceptable or not. It is more likely to me that a malicious snap could use `sys_admin` and `net_admin` to do other bad things to the system though.

My personal opinion as a user who wants to use `daemon-notify` is that it's much more useful if we can use `systemd-notify --ready --pid=$BASHPID` directly as it's a useful mechanism for proper service ordering and dependency management when you're not in full control of the service source code (if I was in charge of the source code I would just have the service use `sd_notify` directly). Additionally, the fact that the interface already doesn't auto-connect implies to me that even without these new capabilities, the `daemon-notify` interface is privileged.

tests/main/interfaces-daemon-notify/task.yaml
anonymouse64: Done

tests/lib/snaps/test-snapd-daemon-notify/meta/snap.yaml
anonymouse64: Removed

tests/lib/snaps/test-snapd-daemon-notify/meta/hooks/install
anonymouse64: Done

tests/lib/snaps/test-snapd-daemon-notify/meta/hooks/install
anonymouse64: Hmm although if you meant why we are calling stop, it's because there's no `snaptl disable ...`, so we always have to stop and disable it.

chipaca:
@zyga could you re-review this?
I suspect @jdstrand  also needs to take a look.

mvo5:
@zyga can you please have a look at this one? It seems your feedback got addressed. Fwiw @anonymouse64  the test failures look real but probably not deep(?).

anonymouse64:
@mvo5 @chipaca is it possible somehow to easily run the spread tests locally on these other OS's? The HACKING.md only explains how to setup an ubuntu VM to run from spread and all of the failures are on systems I don't have immediate access to:
```
    - google:amazon-linux-2-64:tests/main/interfaces-daemon-notify
    - google:arch-linux-64:tests/main/interfaces-daemon-notify
    - google:centos-7-64:tests/main/interfaces-daemon-notify
    - google:debian-9-64:tests/main/interfaces-daemon-notify
    - google:debian-sid-64:tests/main/interfaces-daemon-notify
    - google:fedora-29-64:tests/main/interfaces-daemon-notify
    - google:opensuse-15.0-64:tests/main/interfaces-daemon-notify
    - google:opensuse-tumbleweed-64:tests/main/interfaces-daemon-notify
```

jdstrand:
> @mvo5 @chipaca is it possible somehow to easily run the spread tests locally on these other OS's? The HACKING.md only explains how to setup an ubuntu VM to run from spread and all of the failures are on systems I don't have immediate access to:
> 
> ```
>     - google:amazon-linux-2-64:tests/main/interfaces-daemon-notify
>     - google:arch-linux-64:tests/main/interfaces-daemon-notify
>     - google:centos-7-64:tests/main/interfaces-daemon-notify
>     - google:debian-9-64:tests/main/interfaces-daemon-notify
>     - google:debian-sid-64:tests/main/interfaces-daemon-notify
>     - google:fedora-29-64:tests/main/interfaces-daemon-notify
>     - google:opensuse-15.0-64:tests/main/interfaces-daemon-notify
>     - google:opensuse-tumbleweed-64:tests/main/interfaces-daemon-notify
> ```

You might be interested in https://blog.strandboge.com/2019/04/16/cloud-images-qemu-cloud-init-and-snapd-spread-tests/. It doesn't cover everything you mentioned here, but hopefully it will get you started.

anonymouse64:
Thanks @jdstrand that's super helpful thanks

chipaca:
What's the status of this?

@zyga third time asking you to look at this again

tests/lib/snaps/test-snapd-daemon-notify/meta/snap.yaml
```diff
@@ -3,8 +3,12 @@ version: 1.0
 summary: Basic daemon-notify snap
 description: A basic snap which allow sending notification messages to systemd through the notify socket
 
+hooks:
+  install:
+    plugs: [network]
```
zyga: Why is the network plug necessary?

anonymouse64: hmm it might not be actually, I thought it was but it's been awhile since I've looked at this

anonymouse64: the network plug didn't end up being necessary and it was for the install hook anyways so it wouldn't have affected the main daemon being tested here

anonymouse64:
Alright I updated this by simplifying the snap a little bit and removed an unnecessary plug for the install hook, but I will cease work on this branch until @jdstrand gets time to investigate what is a better approach for this. As such, I'm going to close this PR after the tests run to see if the errant spread failures got resolved with my new commits. I believe they were due to older versions of systemd acting weirdly but didn't ever get around to debugging that anymore.

zyga:
@anonymouse64 there are multiple spread failures. It seems that parsing output of systemctl is too fragile.
```
2019-09-04 08:21:13 Error executing google:debian-9-64:tests/main/interfaces-daemon-notify : 
-----
+ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/journalctl.sh
++ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/state.sh
+++ SNAPD_STATE_PATH=/home/gopath/src/github.com/snapcore/snapd/tests/snapd-state
+++ SNAPD_STATE_FILE=/home/gopath/src/github.com/snapcore/snapd/tests/snapd-state/snapd-state.tar
+++ RUNTIME_STATE_PATH=/home/gopath/src/github.com/snapcore/snapd/tests/runtime-state
+++ SNAPD_ACTIVE_UNITS=/home/gopath/src/github.com/snapcore/snapd/tests/runtime-state/snapd-active-units
+++ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/dirs.sh
++++ export SNAP_MOUNT_DIR=/snap
++++ SNAP_MOUNT_DIR=/snap
++++ export LIBEXECDIR=/usr/lib
++++ LIBEXECDIR=/usr/lib
++++ export MEDIA_DIR=/media
++++ MEDIA_DIR=/media
++++ case "$SPREAD_SYSTEM" in
+++ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/boot.sh
++++ GRUB_EDITENV=grub-editenv
++++ case "$SPREAD_SYSTEM" in
+++ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/systemd.sh
+++ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/systems.sh
++ JOURNALCTL_CURSOR_FILE=/home/gopath/src/github.com/snapcore/snapd/tests/runtime-state/journalctl_cursor
+ echo 'The interface is not connected by default'
The interface is not connected by default
+ MATCH -- '- +test-snapd-daemon-notify:daemon-notify'
+ snap interfaces -i daemon-notify

'snap interfaces' is deprecated; use 'snap connections'.
+ echo 'The service cannot be started with the interface disconnected'
The service cannot be started with the interface disconnected
+ snap start test-snapd-daemon-notify.notify
Started.
+ MATCH 'disabled\s*inactive'
+ snap services test-snapd-daemon-notify.notify
Service                          Startup   Current  Notes
test-snapd-daemon-notify.notify  disabled  active   -
-----
travis_time:end:1445ab23:start=1567585272932444963,finish=1567585273238672708,duration=306227745
```

Can you please look at making those something more robust, e.g. by using one of the systemctl helpers, if one is broadly available, instead of pattern matching?

anonymouse64:
@zyga I will hold off on fixing any spread tests until @jdstrand has a chance to look at this which won't be for a while

jdstrand:
@anonymouse64 - I though we agreed this PR should be closed for now due to higher priority items?

anonymouse64:
As long as it's on your internal "rainy day list" I'm fine to close this

jdstrand:
> As long as it's on your internal "rainy day list" I'm fine to close this

Actually, I had some notes on this from my in progress investigation:

'_Looking at `man sd_notify` it has sd_pid_notify and sd_pid_notifyf which allow giving the PID of another process so that one can update status of other PIDs "provided the appropriate privileges are available". Because apparmor only allows access to the socket and systemd doesn't query LSMs for application isolation, this must be in a separate interface since root daemons in snaps would be able to modify the status of PIDs not ownd by the snap._'

Recalling that, NAK on this PR. If you want to be able to use `--pid`, you should plug another interface. As it happens, network-control gives both net_admin and sys_admin. A more suitable interface would likely be a new daemon-notify-control that is daemon-notify plus these capabilities. 

The best solution would be either snap integration in systemd or snapd providing a new 'systemd-notify' utility in the core/base runtimes that talks to snapd which can then perform the requested operation on behalf of the snap after verifying access (similar to xdg-open, xdg-settings, etc). This would not require a new interface and would make 'daemon-notify' auto-connectable (or even part of the default template). Direct systemd integration would allow systemd-notify as well as the sd_pid_notify() and sd_pid_notifyf() APIs to work, whereas replacing the systemd-notify binary wouldn't.

(The pending investigation where those notes came from have to do with the request to make it auto-connected/in the default template, which I've been reluctant to allow due to a CVE history in this area of systemd. The investigation has taken so long because for me to reverse my opinion I need to do a deep audit of this part of systemd which has fallen behind other work)

PR #6698, opened 2019-04-08
overlord/ifacestate: introduce HotplugKey type use short key in change summaries

stolowski:
Introduce HotplugKey type based on string and use it everywhere in hotplug code instead of plain string. Provide ShortString method (similiar to recently landed ShortString method of HotplugDeviceInfo) and use it for task and change summaries; debug messages will still use full hotplug key string. Also fixed a few inconsistencies with "%s" vs "%q" in the messages.

Per one of the review comments suggesting use of unicode … for "...", I've also updated HotplugDeviceKey the same way.

### DIFF
--- a/interfaces/hotplug/deviceinfo.go
+++ b/interfaces/hotplug/deviceinfo.go
@@ -120,7 +120,7 @@ func (h *HotplugDeviceInfo) str(maxModelOrVendorLen int) string {
 
 	modelOrVendor := h.firstAttrValueOf("ID_MODEL_FROM_DATABASE", "ID_MODEL", "ID_MODEL_ID", "ID_VENDOR_FROM_DATABASE", "ID_VENDOR", "ID_VENDOR_ID")
 	if len(modelOrVendor) > maxModelOrVendorLen {
-		modelOrVendor = modelOrVendor[0:maxModelOrVendorLen] + "..."
+		modelOrVendor = modelOrVendor[0:maxModelOrVendorLen] + "…"
 	}
 
 	var serial string
--- a/interfaces/repo.go
+++ b/interfaces/repo.go
@@ -817,7 +817,7 @@ func (r *Repository) UpdateHotplugSlotAttrs(ifaceName string, hotplugKey snap.Ho
 	}
 
 	for _, slotInfo := range r.slots[snapName] {
-		if slotInfo.Interface == ifaceName && (slotInfo.HotplugKey == hotplugKey) {
+		if slotInfo.Interface == ifaceName && slotInfo.HotplugKey == hotplugKey {
 			if len(r.slotPlugs[slotInfo]) > 0 {
 				// slots should be updated when disconnected, and reconnected back after updating.
 				return nil, fmt.Errorf("internal error: cannot update slot %s while connected", slotInfo.Name)
--- a/interfaces/repo.go
+++ b/interfaces/repo.go
@@ -765,7 +765,7 @@ func (r *Repository) connected(snapName, plugOrSlotName string) ([]*ConnRef, err
 }
 
 // ConnectionsForHotplugKey returns all hotplug connections for given interface name and hotplug key.
-func (r *Repository) ConnectionsForHotplugKey(ifaceName, hotplugKey string) ([]*ConnRef, error) {
+func (r *Repository) ConnectionsForHotplugKey(ifaceName string, hotplugKey snap.HotplugKey) ([]*ConnRef, error) {
 	r.m.Lock()
 	defer r.m.Unlock()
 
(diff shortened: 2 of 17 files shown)


interfaces/repo.go
```diff
@@ -817,7 +817,7 @@ func (r *Repository) UpdateHotplugSlotAttrs(ifaceName, hotplugKey string, static
 	}
 
 	for _, slotInfo := range r.slots[snapName] {
-		if slotInfo.Interface == ifaceName && slotInfo.HotplugKey == hotplugKey {
+		if slotInfo.Interface == ifaceName && (slotInfo.HotplugKey == hotplugKey) {
```
zyga: Is this change relevant?

overlord/ifacestate/hotplug_test.go
```diff
@@ -1120,14 +1120,14 @@ func (s *hotplugSuite) TestRemoveDeviceTasks(c *C) {
...
-	c.Assert(key, Equals, "key")
+	c.Assert(key, DeepEquals, snap.HotplugKey("key"))
 
 	task2 := tss.Tasks()[1]
 	c.Assert(task2.Kind(), Equals, "hotplug-remove-slot")
 	iface, key, err = ifacestate.GetHotplugAttrs(task2)
 	c.Assert(err, IsNil)
 	c.Assert(iface, Equals, "interface")
-	c.Assert(key, Equals, "key")
+	c.Assert(key, DeepEquals, snap.HotplugKey("key"))
```
zyga: Why does this now require deep equals? Is this due to the fact that the string is wrapped in a new type?

overlord/ifacestate/ifacestate_test.go
```diff
@@ -5707,7 +5707,7 @@ func (s *interfaceManagerSuite) TestHotplugAddGoneSlot(c *C) {
 	slot := repo.Slot("core", "hotplugslot-old-name")
 	c.Assert(slot, NotNil)
 	c.Check(slot.Attrs, DeepEquals, map[string]interface{}{"foo": "bar"})
-	c.Check(slot.HotplugKey, Equals, "1234")
+	c.Check(slot.HotplugKey, DeepEquals, snap.HotplugKey("1234"))
```
zyga: Why is this the case that just above this changed fragment we can use `Equals` and here the test was updated to `DeepEquals`?

snap/hotplug_key.go
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package snap
+
+// HotplugKey is a string key of a hotplugged device
+type HotplugKey string
+
+// ShortString returns a truncated string representation of the hotplug key
+func (h HotplugKey) ShortString() string {
+	str := string(h)
```
zyga: I have not looked around the tree but perhaps this should be a common helper that is in `strutil`?

snap/hotplug_key.go
```diff
@@ -0,0 +1,33 @@
...
+	// this is normally never the case, but makes testing easier
+	if len(str) < 12 {
```
zyga: I don't doubt this value was picked carefully, could we have the reasoning of the value added as a code comment please?

snap/hotplug_key.go
```diff
@@ -0,0 +1,33 @@
...
+		return str
+	}
+	return str[0:12] + "..."
```
zyga: Should this be using the unicode horizontal ellipsis character U+2026? CC @chipaca

interfaces/repo.go
stolowski: Ah, not really, restored, thanks for spotting!

overlord/ifacestate/hotplug_test.go
stolowski: It doesn't require DeepEquals, I just felt it's better for custom types (in particular if they ever grow up, probably unlikely here). I can revert back to Equals if you think that's confusing?

overlord/ifacestate/ifacestate_test.go
stolowski: Not really needed, but see earlier explanation.

snap/hotplug_key.go
stolowski: I'm not sure it deserves a helper atm, we have two very specific use cases (HotplugKey and HotplugDeviceInfo - the latter truncates only vendor/model as part of the final string).

snap/hotplug_key.go
stolowski: Done (also updated it for HotplugDeviceInfo).

interfaces/hotplug/deviceinfo.go
```diff
@@ -120,7 +120,7 @@ func (h *HotplugDeviceInfo) str(maxModelOrVendorLen int) string {
 
 	modelOrVendor := h.firstAttrValueOf("ID_MODEL_FROM_DATABASE", "ID_MODEL", "ID_MODEL_ID", "ID_VENDOR_FROM_DATABASE", "ID_VENDOR", "ID_VENDOR_ID")
 	if len(modelOrVendor) > maxModelOrVendorLen {
-		modelOrVendor = modelOrVendor[0:maxModelOrVendorLen] + "..."
+		modelOrVendor = modelOrVendor[0:maxModelOrVendorLen] + "…"
```
bboozzoo: This should display correctly even in non-utf capable terminal, right?

snap/hotplug_key.go
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package snap
+
+import (
+	"fmt"
+)
+
+// HotplugKey is a string key of a hotplugged device
+type HotplugKey string
```
bboozzoo: Likie it way more than the bare string that carries no information about the purpose or format.

interfaces/hotplug/deviceinfo.go
stolowski: No really, it will appear as a rhombus, but as discussed on IRC that's fine.

PR #6699, opened 2019-04-08
daemon: add RootOnly flag to commands

chipaca:
Without this change, endpoints that needed the user to be actual root
(as opposed to gaining admin via login or polkit) would check ucrednet
directly.

This change adds a RootOnly flag to commands, and uses it for those
commands that need it, moving the checks to canAccess together with
the rest of access checks.


zyga:
As a separate comment, is there a spread test for root vs non-privileged user authentication?

daemon/daemon.go
```diff
@@ -169,6 +171,10 @@ func (c *Command) canAccess(r *http.Request, user *auth.UserState) accessResult
 		return accessOK
 	}
 
+	if c.RootOnly {
```
jdstrand: It seems from above that /v2/users falls under GET, but there is no `!c.RootOnly` check under the GET conditional above. This seems to not be needed because we aren't setting c.GuestOK and c.RootOnly or c.UserOK and c.RootOnly at the same time, but as a hardening measure perhaps do:
```
if r.Method == "GET" && !c.RootOnly {
...
}
```
In this manner we default closed if c.RootOnly is set for the GETs.

daemon/api.go
```diff
@@ -220,8 +221,9 @@ var (
 	}
 
 	usersCmd = &Command{
-		Path: "/v2/users",
-		GET:  getUsers,
+		Path:     "/v2/users",
+		GET:      getUsers,
+		RootOnly: true,
```
mvo5: Do we have a test for this? When I remove this line (RootOnly: true) it seems no test fails(?).

chipaca: I've added a test for this, at least to catch it if we remove it by mistake

daemon/daemon.go
chipaca: I've added that, and also added a panic if `RootOnly` is set together with any `*OK` property

daemon/daemon.go
```diff
@@ -124,8 +126,8 @@ var polkitCheckAuthorization = polkit.CheckAuthorization
 // - UserOK: any uid on the local system can access GET
 // - SnapOK: a snap can access this via `snapctl`
 func (c *Command) canAccess(r *http.Request, user *auth.UserState) accessResult {
-	if user != nil {
-		// Authenticated users do anything for now.
+	if user != nil && !c.RootOnly {
```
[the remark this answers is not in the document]
chipaca: done

chipaca:
@jdstrand note this is a pure refactor, the commands that now have RootOnly were checking for uid=0 'by hand' before. So there weren't (AFAIK) situations where a non-root process was able to access these APIs, to write the spread test you'd like to see.

PR #6705, opened 2019-04-09
bootloader: little kernel support

kubiko:
Enabling support for little kernel aka lk boot loaders
Little Kernel(lk) is popular bootloader used by many arm SoCs
Reference source: [https://github.com/littlekernel/lk](https://github.com/littlekernel/lk)
As it is often used by Android devices, almost every Qualcomm and MTK SoCs are supporting lk.
lk is using boot image format, which packs kernel, initrd, dtbs. This is then flashed to dedicated boot partition.
As lk is minimalistic bootloader it does not support advanced scripting as u-boot does.


zyga:
Hey @kubiko - thank you for the PR

For a branch that adds 1K lines of code it would be nice to have an introduction in the pull request message coupled with proper commit messages for subsequent changes. Would you mind extending the messages a little to explain what little kernel is, how it is it being added to the tree and other kind of useful review information please?

Note that we use a particular commit message style that uses a single or a pair of path components to highlight the area being changed (e.g. apparmor, snap-confine or other). We also try to use specific grammatical structure that describe the changes, for instance "add little kernel support" rather than "adding or added". While this is not something we enforce strongly we had recently discussed this topic and there's a recommendation to craft our commit messages with some extra effort and clarity.

From my POV you can rework the history to include the information and force-push to include it.

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,278 @@
...
+	if(osutil.FileExists(l.pathbak)) {
+		errr := l.SaveEnv(l.pathbak, w)
+		if errr != nil {
+			logger.Debugf("Save: failed to save backup environment")
+		}
+	}
+	return err;
+}
+
+func (l *Env) SaveEnv(path string, buf *bytes.Buffer) error {
```
bboozzoo: io.Reader instead of bytes.Buffer will probably be fine here

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,278 @@
...
+	f, err := os.OpenFile(path, os.O_WRONLY, 0660)
+	if err != nil {
+		return fmt.Errorf("error opening lk env file for env storing %v", err)
+	}
+	defer f.Close()
+
+	if _, err := f.Write(buf.Bytes()); err != nil {
```
bboozzoo: and then io.Copy(f, buf)

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,278 @@
...
+	w := bytes.NewBuffer(nil)
+	ss := binary.Size(l.env)
+	w.Grow(ss)
+	err = binary.Write(w, binary.LittleEndian, &l.env)
+	if err != nil {
+		return fmt.Errorf("error writting lk environment to buffer for validation %v", err)
+	}
+	crc := crc32.ChecksumIEEE(w.Bytes()[:ss-4]) // size of crc32 itself at the end of the structure
+	if crc != l.env.Crc32 {
+		return fmt.Errorf("lkenv:load: broken crc32 [0x%X]vs[0x%X]\n", crc, l.env.Crc32)
```
bboozzoo: fmt.Errorf("invalid environment checksum, got 0x%X expected 0x%X\n", crc, l.env.Crc32)

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,278 @@
...
+	if (err != nil) {
+		return l.LoadEnv(l.pathbak)
+	}
+	return nil
+}
+
+func (l *Env) LoadEnv(path string) error {
+	f, err := os.Open(path)
+	if err != nil {
+		return fmt.Errorf("error loading lk environment file: %v", err)
```
bboozzoo: fmt.Errorf("cannot open LK environment file: %v", err)

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	defer f.Close()
+	if err := binary.Read(f, binary.LittleEndian, &l.env); err != nil {
+		return fmt.Errorf("error reading lk environment from file %v", err)
```
bboozzoo: fmt.Errorf("cannot read environment file: %v", err)

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+func (l *Env) LoadEnv(path string) error {
+	f, err := os.Open(path)
+	if err != nil {
+		return fmt.Errorf("error loading lk environment file: %v", err)
+	}
+
+	defer f.Close()
+	if err := binary.Read(f, binary.LittleEndian, &l.env); err != nil {
```
bboozzoo: This could probably be rewritten as:
```
b := &bytes.Buffer{}
size := binary.Size(l.env)
if err := io.CopyN(b, f, size); err != nil {
    ....
}
if err := binary.Write(b, binary.LittleEndian, &l.len); err != nil {
    ....
}
expected := crc32.ChecksumIEEE(b.Bytes()[:size-4])
if expected != l.Crc32 {
    ....
}
````

File: bootloader/lkenv/lkenv.go [hunk 7]
```diff
@@ -0,0 +1,278 @@
...
+func (l *Env) Load() error {
+	err := l.LoadEnv(l.path)
+	if (err != nil) {
+		return l.LoadEnv(l.pathbak)
+	}
+	return nil
+}
+
+func (l *Env) LoadEnv(path string) error {
+	f, err := os.Open(path)
```
bboozzoo: We write to a backup environment file, but ignore it when reading?

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,278 @@
...
+	logger.Debugf("Load: validated crc32 (0x%X)", l.env.Crc32)
+	return nil
+}
+
+func (l *Env) Save() error {
+	logger.Debugf("Save")
+	w := bytes.NewBuffer(nil)
+	ss := binary.Size(l.env)
+	w.Grow(ss)
+	err := binary.Write(w, binary.LittleEndian, &l.env)
```
bboozzoo: Given that we write the thing once, calcualte the checksum, and then write it again with updated checksum, perhaps it would make sense to split the env struct into:
```
type SnapBootSelectImage struct {
    SnapBootSelect // sans Crc32 field
    /* crc32 value for structure */
    Crc32 uint32
}
```
Provided binary.Write/Read can handle embedded struct correctly.

bootloader/lk.go
```diff
@@ -0,0 +1,202 @@
...
+		bpart := filepath.Join("/dev/disk/by-partlabel/",bootPartition)
+
+		bpf, err := os.OpenFile(bpart, os.O_WRONLY, 0660)
+		if err != nil {
+			return fmt.Errorf("Failed to open boot partition [%s] %v", bpart, err)
+		}
+		defer bpf.Close()
+
+		buf := make([]byte, 1024)
+		for {
```
bboozzoo: IDK, it feels like this should be a part of the upcoming gadget update. I assume boot.img is the actual image of the partition where `snapbootsel[.bin]` are in. Is this a workaround for deploying the actual thing to the device?

bootloader/lk.go
```diff
@@ -0,0 +1,202 @@
...
+			return fmt.Errorf("Failed to unpack boot.img %v", err)
+		}
+		// write boot.img to free boot partition
+		bootimgName := filepath.Join(tmpdir, "boot.img")
+		bif, err := os.Open(bootimgName)
+		if err != nil {
+			return fmt.Errorf("Failed to open unpacked boot.img %v", err)
+		}
+		defer bif.Close()
+		bpart := filepath.Join("/dev/disk/by-partlabel/",bootPartition)
```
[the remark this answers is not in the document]
kubiko: well spotted!

bootloader/lkenv/lkenv.go
kubiko: explained in previous why this does not work

bootloader/lkenv/lkenv.go
kubiko: Fixed

bootloader/lkenv/lkenv.go
kubiko: Fixed

bootloader/lkenv/lkenv.go
kubiko: I did play with this option, but it somehow breaks structure extraction.
I'd prefer option when we can read to buffer, calculate crc, even validate crc on buffer and then write it to the structure. A bit what I did in write method

bootloader/lkenv/lkenv.go
kubiko: Fixed

bootloader/lkenv/lkenv.go
kubiko: We read it only if we need to, if crc check fails on main env, then backup is used. This is actually tested by `TestFailedCRCFallBack`
That is also why bak environment is optional

bootloader/lkenv/lkenv.go
kubiko: I updated implementation to avoid double write.
Reason this is kept as simple structure is to make it simple to read it in c on bootloader side.
I have played with idea to double wrap it win extra structure as you suggested but that would still mean two writes, first for inner structures to calculate crc and then write of wrapping structure. Now I actually rewind buffer by 4 bytes and re-write crc only.

bootloader/lk.go
kubiko: No, none of this is really related to gadget assets. Only relation to gadget is gadget.yaml, where bootloader is defined (though not actually used by snapd to determine bootlloader type! :) ) And then added extra supported roles for partitions. This is to make it more dynamic rather than hardcore expected partition labels.
`snapbootsel` is direct equivalent of `uboot.env`, just stored as partition rather than file. So it's changed every time new core/kernel is installed/tried.
boot.img is then asset extracted from kernel snap. boot.img is actually kernel + initrd + dtbs packed into one file, and flashed into dedicated partition. So we need two partitions for boot.img which we will alter based on installed kernel snap revision.

So none is workaround, for future improvement I'd like bootloader implementation to more actively use info from gadget snap, to determine bootloader type for example.

bootloader/lkenv/lkenv.go
kubiko: io.Reader does not work, as we need to use this buffer twice (to write backup), and io.Reader is not good for this. It works well if we do not use backup

kubiko:
> Thanks for the PR. Looks interesting, but feels a bit hard to follow at this point.
> 
> I would suggest splitting it up a bit. Namely, the patch that moves ExtractKernelAssets and RemoveKernelAssets could be a separate PR. I wonder whether there would be a way of avoiding importing "github.com/snapcore/snapd/snap" in the "bootloader" package. Maybe the `Bootloader` interface could be extended further.
> 
> I think a patch adding lk and lkEnv (sans {Remove,Extract}KernelAssets), could be another PR and that could land quickly.

Funny you mention another PR, I did actually try to land this change before in #6691, but it did not get any attention, here it did though :) 

About importing github.com/snapcore/snapd/snap, I believe bootloader is correct place to implement kernel asset extraction, as it's very much bootloader related, so I'm not sure how can we avoid this
Also having extract kernel assets as single function handling all bootloaders through bunch of `if` statements doesn't seems like best way. It was OK when u-boot was only bootloader needing this treatment.
Otherwise thanks for good review comments 👍

kubiko:
> Hey @kubiko - thank you for the PR
> 
> For a branch that adds 1K lines of code it would be nice to have an introduction in the pull request message coupled with proper commit messages for subsequent changes. Would you mind extending the messages a little to explain what little kernel is, how it is it being added to the tree and other kind of useful review information please?
> 
> Note that we use a particular commit message style that uses a single or a pair of path components to highlight the area being changed (e.g. apparmor, snap-confine or other). We also try to use specific grammatical structure that describe the changes, for instance "add little kernel support" rather than "adding or added". While this is not something we enforce strongly we had recently discussed this topic and there's a recommendation to craft our commit messages with some extra effort and clarity.
> 
> From my POV you can rework the history to include the information and force-push to include it.

I reworked history now with more info added.
I see there are two test failures related to mocked bootloader tests, which I will need to fix once I'm back from holidays

kubiko:
@zyga @bboozzoo all tests are now fixed, same for #6691

bootloader/lkenv/lkenv_test.go
```diff
@@ -0,0 +1,183 @@
...
+	env.Set("snap_try_kernel", "kernel-2")
+	env.Set("snap_core", "core-1")
+	env.Set("snap_try_core", "core-2")
+	env.Set("snap_gadget", "gadget-1")
+	env.Set("snap_try_gadget", "gadget-2")
+
+	err = env.Save()
+	c.Assert(err, IsNil)
+
+	env2 := lkenv.NewEnv(l.envPath)
```
mvo5: I think in addition to this test we want something that stores the expected bytes data in some variable and compares the written structure byte-to-byte. This test here is nice but if e.g. alignment breaks we won't notice because the same code that reads and writes is used in the test. To fix this I think we need an extra test against some blob (in the form of  ` var  expectedStructData = []byte{0x11,0x22,...}` (unless that is too long in which case we could use an external file).

bootloader/lk.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	return e
+}
+
+func (l *lk) Name() string {
+	return "lk"
+}
+
+func (l *lk) Dir() string {
```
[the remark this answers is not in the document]
mvo5: I pushed https://github.com/snapcore/snapd/pull/7122 and with that "Dir()" can just be removed entirely. I look at "ConfigFile()" next.

bootloader/lk.go
```diff
@@ -0,0 +1,202 @@
...
+
+func (l *lk) ConfigFile() string {
+	return l.envFile()
+}
+
+func (l *lk) envFile() string {
+	// as for dir, we have two scenarios, image building and runtime
+	if dirs.GlobalRootDir == "/" {
+		// TO-DO: this should be eventually fetched from gadget.yaml
+		return filepath.Join(l.Dir(), "snapbootsel")
```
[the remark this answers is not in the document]
mvo5: I think something like http://paste.ubuntu.com/p/46FCtnfFX2/ is easier to follow. Basicly this makes ConfigFile always look at the on-disk config file. And the detection of LK is done in newLk without the use of ConfigFile. For most boot loaders it make sense to just use "if FileExists(ConfigFile()))" - however because LK will put this into a partition I think we need to separate detection of LK and config file handling. I hope I do not miss something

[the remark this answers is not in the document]
mvo5: More like this: http://paste.ubuntu.com/p/qJjCzjYSjW/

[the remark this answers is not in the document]
mvo5: Btw, I wonder how the /boot/lk/snapbootsel.bin becomes the snapbootsel partition - also - do we need ConfigFile() at image build time at all?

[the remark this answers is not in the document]
mvo5: I.e. could we simply *not* install /boot/lk/snapbootsel.bin (because its not really used, just the partition is used)?

bootloader/lk.go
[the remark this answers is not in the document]
kubiko: @mvo5 I checked PR that looks good, indeed we can remove `Dir()`, I also commented in the PR

bootloader/lk.go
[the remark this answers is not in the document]
kubiko: It actually operates in quite same way as other bootloaders here. There are two files 

> Btw, I wonder how the /boot/lk/snapbootsel.bin becomes the snapbootsel partition - also - do we need ConfigFile() at image build time at all?

We do need `snapbootsel.bin` as this is initial boot configuration for device to boot at all. It works same way as for u-boot. We generate initial uboot.env at image build time ( which is based on default env file from gadget with added `snap_kernel` `snap_core` values.
Same way operates `snapbootsel.bin` only difference is that image builder instead of putting file to `boot` partition, it uses file as image for boot select partition.
This is also paving way for `uboot.env` being separate partition. You need to generate correct `uboot.env` file at image build time, which is used by image builder as partition image, and at run time partition is used instead.

bootloader/lkenv/lkenv_test.go
kubiko: @mvo5 I added now gzipped sample data test case.
Let me know if more tests are needed :)

bootloader/lk.go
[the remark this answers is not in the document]
kubiko: Code updated with `Dir()` now un-exported as well
I also introduced new common function where run mode is determine (image building/runtime)

zyga:
@chipaca this will need to be adjusted with the new boot refactoring.

bootloader/lk.go
```diff
@@ -0,0 +1,211 @@
...
+	return nil
+}
+
+// for lk we need to flash boot image to free bootimg partition
+// first make sure there is free boot part to use
+// if this is image creation, we just extract file
+func (l *lk) ExtractKernelAssets(s *snap.Info, snapf snap.Container) error {
+	blobName := filepath.Base(s.MountFile())
+
+	logger.Debugf("ExtractKernelAssets (%s)\n", blobName)
```
mvo5: (nitpick) The \n is not needed here.

bootloader/lk.go
```diff
@@ -0,0 +1,211 @@
...
+	if inRuntimeMode() {
+		logger.Debugf("ExtractKernelAssets handling run time usecase\n")
+		// this is live system, extracted bootimg needs to be flashed to
+		// free bootimg partition and env has be updated boot slop mapping
+		tmpdir, err := ioutil.TempDir("", "bootimg")
+		if err != nil {
+			return fmt.Errorf("Failed to create tmp directory %v", err)
+		}
+		defer os.RemoveAll(tmpdir)
+		if err := snapf.Unpack(env.GetBootImageName(), tmpdir); err != nil {
```
mvo5: (nitpick) you could assign bootImage here and avoid the calls, something like:
```
diff --git a/bootloader/lk.go b/bootloader/lk.go
index 5b554c702f..4341dd8239 100644
--- a/bootloader/lk.go
+++ b/bootloader/lk.go
@@ -149,14 +149,15 @@ func (l *lk) ExtractKernelAssets(s snap.PlaceInfo, snapf snap.Container) error {
                        return fmt.Errorf("Failed to create tmp directory %v", err)
                }
                defer os.RemoveAll(tmpdir)
-               if err := snapf.Unpack(env.GetBootImageName(), tmpdir); err != nil {
-                       return fmt.Errorf("Failed to unpack %s %v", env.GetBootImageName(), err)
+               bootImage := env.GetBootImageName()
+               if err := snapf.Unpack(bootImage, tmpdir); err != nil {
+                       return fmt.Errorf("Failed to unpack %s %v", bootImage, err)
                }
                // write boot.img to free boot partition
-               bootimgName := filepath.Join(tmpdir, env.GetBootImageName())
+               bootimgName := filepath.Join(tmpdir, bootImage)
                bif, err := os.Open(bootimgName)
                if err != nil {
-                       return fmt.Errorf("Failed to open unpacked %s %v", env.GetBootImageName(), err)
+                       return fmt.Errorf("Failed to open unpacked %s %v", bootImage, err)
                }
                defer bif.Close()
                bpart := filepath.Join(l.dir(), bootPartition)
```

bootloader/lk.go
```diff
@@ -0,0 +1,211 @@
...
+		defer bif.Close()
+		bpart := filepath.Join(l.dir(), bootPartition)
+
+		bpf, err := os.OpenFile(bpart, os.O_WRONLY, 0660)
+		if err != nil {
+			return fmt.Errorf("Failed to open boot partition [%s] %v", bpart, err)
+		}
+		defer bpf.Close()
+
+		buf := make([]byte, 1024)
```
mvo5: Silly question - could we use "io.Copy" here? Something like:
```
diff --git a/bootloader/lk.go b/bootloader/lk.go
index 5b554c702f..cdaef2477b 100644
--- a/bootloader/lk.go
+++ b/bootloader/lk.go
@@ -167,20 +168,8 @@ func (l *lk) ExtractKernelAssets(s snap.PlaceInfo, snapf snap.Container) error {
                }
                defer bpf.Close()
 
-               buf := make([]byte, 1024)
-               for {
-                       // read by chunks
-                       n, err := bif.Read(buf)
-                       if err != nil && err != io.EOF {
-                               return fmt.Errorf("Failed to read buffer chunk of %s %v", env.GetBootImageName(), err)
-                       }
-                       if n == 0 {
-                               break
-                       }
-                       // write a chunk
-                       if _, err := bpf.Write(buf[:n]); err != nil {
-                               return fmt.Errorf("Failed to write buffer chunk of %s %v", env.GetBootImageName(), err)
-                       }
+               if _, err := io.Copy(bpf, bif); err != nil {
+                       return err
                }
        } else {
                // we are preparing image, just extract boot image to bootloader directory

```
Doing this seems to not break anything - so either we don't have tests for this or it works :)

pedronis:
does this work out-of-the-box with image ATM ?

bootloader/lk.go
kubiko: I'm trying to remember reason I did it by chunks, but indeed if `io.Copy` works, then we should use it instead

kubiko:
> does this work out-of-the-box with image (aka prepare-image) ATM ?

Yes, this works out of the box

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,355 @@
...
+func copyString(b []byte, s string) {
+	copy(b[:], s)
+	b[len(s)] = 0
+}
+
+func NewEnv(path string) *Env {
+	// osutil.FileExists(path + "bak")
+	return &Env{
+		path:    path,
+		pathbak: path + "bak",
```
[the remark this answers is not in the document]
kubiko: yes, as this relates to partition name, so you will be having partition name `bootsel.bak`

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,355 @@
...
+)
+
+type SnapBootSelect struct {
+	/* Contains value BOOTSELECT_SIGNATURE defined above */
+	Signature uint32
+	/* snappy boot select version */
+	Version uint32
+
+	/* snap_mode, one of: 'empty', "try", "trying" */
+	Snap_mode [SNAP_MODE_LENGTH]byte
```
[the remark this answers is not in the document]
kubiko: Names are with underscore because this structure also exists in c form, to keep names in sync between go and c code version
otherwise no other reason

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,355 @@
...
+func cToGoString(c []byte) string {
+	if end := bytes.IndexByte(c, 0); end >= 0 {
+		return string(c[:end])
+	}
+	// XXX: no trailing \0 - should we return "" here?
+	return string(c)
+}
+
+// helper function to copy string into array and making sure it's terminated
+func copyString(b []byte, s string) {
```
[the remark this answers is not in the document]
kubiko: which case do you mean if string is empty, copy will not copy anything, if string is longer than byte array, it will only copy what fits in. Only case I can think of trailing zero
```

File: bootloader/lk.go [hunk 20]
```diff
@@ -0,0 +1,203 @@
...
+
+	bootPartition, err := env.FindFreeBootPartition(blobName)
+	if err != nil {
+		return err
+	}
+
+	if l.inRuntimeMode {
+		logger.Debugf("ExtractKernelAssets handling run time usecase")
+		// this is live system, extracted bootimg needs to be flashed to
+		// free bootimg partition and env has be updated boot slop mapping
```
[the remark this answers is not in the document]
kubiko: something like
```
		// this is live system, extracted bootimg needs to be flashed to
		// free bootimg partition and env has to be updated with
		// new kernel snap to bootimg partition mapping
``` ?

File: gadget/gadget.go [hunk 21]
```diff
@@ -328,7 +330,7 @@ func ReadInfo(gadgetSnapRootDir string, classic bool) (*Info, error) {
 		switch v.Bootloader {
 		case "":
 			// pass
-		case "grub", "u-boot", "android-boot":
+		case "grub", "u-boot", "android-boot", "lk":
 			bootloadersFound += 1
 		default:
 			return nil, errors.New("bootloader must be one of grub, u-boot or android-boot")
```
pedronis: this error message needs fixing

gadget/gadget.go
```diff
@@ -328,7 +330,7 @@ func ReadInfo(gadgetSnapRootDir string, classic bool) (*Info, error) {
 		switch v.Bootloader {
 		case "":
 			// pass
-		case "grub", "u-boot", "android-boot":
+		case "grub", "u-boot", "android-boot", "lk":
```
pedronis: do we need a test for this change, we don't have one for android-boot but not sure it's a reason

gadget/gadget.go
```diff
@@ -591,7 +593,7 @@ func validateRole(vs *VolumeStructure, vol *Volume) error {
 		if vs.Filesystem != "" && vs.Filesystem != "none" {
 			return errors.New("mbr structures must not specify a file system")
 		}
-	case SystemBoot, "":
+	case SystemBoot, BootImage, BootSelect, "":
```
pedronis: do we need a test here?

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,355 @@
...
+const BOOTIMG_DEFAULT_NAME = "boot.img"
+
+// for accessing the 	Bootimg_matrix
+const (
+	MATRIX_ROW_PARTITION = 0
+	MATRIX_ROW_KERNEL    = 1
+)
+
+type SnapBootSelect struct {
+	/* Contains value BOOTSELECT_SIGNATURE defined above */
```
pedronis: this is not defined above, except in a commented out form ?

bootloader/lkenv/lkenv.go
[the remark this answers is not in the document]
pedronis: can we have a comment with a pointer to the repo and the c form of this and explaing it's following that?

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,355 @@
...
+			   or in the future by snap prepare image time
+			 - ExtractKernelAssets fills mapping for initial boot image
+	 snapd:
+			 - when new kernel snap is installed, ExtractKernelAssets updates mapping
+			   in matrix for bootloader to pick correct kernel snap to use for boot
+			 - snap_mode, snap_try_kernel, snap_try_core behaves same way as with u-boot
+			 - boot partition labels are never modified by snapd at run time
+	 bootloader:
+			 - Finds boot partition to use based on info in matrix and snap_kernel / snap_try_kernel
+			 - bootloaer does not alter matrix, only alters snap_mode
```
pedronis: typo bootloader.

bootloader/lkenv/lkenv.go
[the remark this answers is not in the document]
pedronis: yessish, missing test?

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,355 @@
...
+
+	 [ <bootimg 1 part label> ] [ <currently installed kernel snap revison> ]
+	 [ <bootimg 2 part label> ] [ <currently installed kernel snap revision> ]
+	*/
```
pedronis: I find this diagram slightly confusing,  we don't set the same current kernel in both row?  one has the current and one the previous? or am I misunderstanding this completely ?

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,355 @@
...
+	case "snap_gadget":
+		copyString(l.env.Snap_gadget[:], value)
+	case "snap_try_gadget":
+		copyString(l.env.Snap_try_gadget[:], value)
+	case "reboot_reason":
+		copyString(l.env.Reboot_reason[:], value)
+	}
+}
+
+// Configure partition labels for used boot partitions
```
pedronis: the doc comments don't follow go conventions  // Func  does ... .

bootloader/lk.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	// XXX: in the long run we want this to go away, we probably add
+	//      something like "boot.PrepareImage()" and add an (optional)
+	//      method "PrepareImage" to the bootloader interface that is
+	//      used to setup a bootloader from prepare-image if things
+	//      are very different from runtime vs image-building mode.
+	//
+	// determine mode we are in, runtime or image build
+	e.inRuntimeMode = (dirs.GlobalRootDir == "/")
```
pedronis: can't we use something different to detect runtime vs not?  if we keep this way we need some kind of image test because if we change how we GlobalRootDir there we won't notice

pedronis: it's also a bit confusing that we use envFile just above but it actually depends on this flag that we set here

bootloader/lk.go
```diff
@@ -0,0 +1,203 @@
...
+		return err
+	}
+
+	if l.inRuntimeMode {
+		logger.Debugf("ExtractKernelAssets handling run time usecase")
+		// this is live system, extracted bootimg needs to be flashed to
+		// free bootimg partition and env has be updated boot slop mapping
+		tmpdir, err := ioutil.TempDir("", "bootimg")
+		if err != nil {
+			return fmt.Errorf("Failed to create tmp directory %v", err)
```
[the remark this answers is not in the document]
pedronis: yes, all errors should start with lower case, and be formulated as "cannot ..." whenever that is possible

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,355 @@
...
+// this should not be used at run time, it should be used
+// once at image build time, if part labels are not filled by gadget build
+func (l *Env) ConfigureBootPartitions(boot_1, boot_2 string) {
```
pedronis: this is called only in tests atm?

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,355 @@
...
+	copyString(l.env.Bootimg_matrix[0][MATRIX_ROW_PARTITION][:], boot_1)
+	copyString(l.env.Bootimg_matrix[1][MATRIX_ROW_PARTITION][:], boot_2)
+}
+
+// Configure boot image file name to be used at kernel extraction time
+// this should not be used at run time, it should be used
+// once at image build time if default boot.img is not used
+func (l *Env) ConfigureBootimgName(bootimgName string) {
```
pedronis: this is called only in tests atm?

bootloader/lkenv/lkenv.go
pedronis: not pushed yet?

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,355 @@
...
+	if err := binary.Read(f, binary.LittleEndian, &l.env); err != nil {
+		return fmt.Errorf("error reading lk environment from file %v", err)
+	}
+
+	// calculate crc32 to validate structure
+	w := bytes.NewBuffer(nil)
+	ss := binary.Size(l.env)
+	w.Grow(ss)
+	if err := binary.Write(w, binary.LittleEndian, &l.env); err != nil {
+		return fmt.Errorf("error writing lk environment to buffer for validation %v", err)
```
pedronis: "cannot write ..."

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,355 @@
...
+	return nil
+}
+
+func (l *Env) Save() error {
+	logger.Debugf("Save")
+	w := bytes.NewBuffer(nil)
+	ss := binary.Size(l.env)
+	w.Grow(ss)
+	if err := binary.Write(w, binary.LittleEndian, &l.env); err != nil {
+		return fmt.Errorf("error writing lk environment to buffer for saving %v", err)
```
pedronis: "cannot write..."

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,355 @@
...
+
+func (l *Env) SaveEnv(path string, buf *bytes.Buffer) error {
+	f, err := os.OpenFile(path, os.O_WRONLY, 0660)
+	if err != nil {
+		return fmt.Errorf("error opening lk env file for env storing %v", err)
+	}
+	defer f.Close()
+
+	if _, err := f.Write(buf.Bytes()); err != nil {
+		return fmt.Errorf("error writing lk env buf to lk env file %v", err)
```
pedronis: cannot write...

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,355 @@
...
+	}
+	if err := f.Sync(); err != nil {
+		return fmt.Errorf("error syncying lk environment file %v", err)
```
pedronis: cannot sync...

bootloader/lkenv/lkenv.go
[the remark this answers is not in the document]
kubiko: @pedronis ideally I'd add that header here, as best place of reference
alternatively we can have reference lk implementation for Dragonboard410 repo

bootloader/lk.go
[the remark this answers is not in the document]
kubiko: updated

bootloader/lk.go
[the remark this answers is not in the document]
kubiko: fixed

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,355 @@
...
+	"hash/crc32"
+	"os"
+
+	"github.com/snapcore/snapd/logger"
+	"github.com/snapcore/snapd/osutil"
+)
+
+const SNAP_BOOTSELECT_VERSION = 0x00010001
+
+// const SNAP_BOOTSELECT_SIGNATURE ('S' | ('B' << 8) | ('s' << 16) | ('e' << 24))
```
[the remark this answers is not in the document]
kubiko: version and signature are now checked, and tested with sample blob

gadget/gadget.go
kubiko: what happened to error starting with 'cannot ....' here :-P
Error updated :)

bootloader/lkenv/lkenv.go
kubiko: This is now used

bootloader/lkenv/lkenv.go
kubiko: fixed

bootloader/lk.go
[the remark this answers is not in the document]
kubiko: All errors are now fixed to start with "cannot "

gadget/gadget.go
kubiko: and also added test for BootImage and BootSelect roles

gadget/gadget.go
kubiko: Test added for this

bootloader/lkenv/lkenv.go
kubiko: fixed

bootloader/lkenv/lkenv.go
kubiko: fixed

bootloader/lkenv/lkenv.go
kubiko: fixed

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,369 @@
...
+	if err != nil {
+		return l.LoadEnv(l.pathbak)
+	}
+	return nil
+}
+
+func (l *Env) LoadEnv(path string) error {
+	f, err := os.Open(path)
+	if err != nil {
+		return fmt.Errorf("cannot open LK environment file: %v", err)
```
pedronis: s/LK/lk ?

bootloader/lkenv/lkenv.go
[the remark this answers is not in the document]
kubiko: there was test, but test was testing for panic :)
I updated now implementation that it concatenate string to fit into array
and added test to validate there is no panic and more test of extreme cases

bootloader/lkenv/lkenv.go
kubiko: fixed

bootloader/lk.go
pedronis: @mvo5  I'll do something intermediate in a separate PR, I think I wanted anyway to turn:

booloader.Find() ->  bootloader.Find(rootdir)  where rootdir = "" uses GlobalRootDir

I can easily change that a well into  booloader.Find(string, *Options), the we can a flag to here.

pedronis: I proposed https://github.com/snapcore/snapd/pull/7418, once that is merged, it can be merged back here and at the cost of some churn avoid the need for this kind of code

mvo5: Thanks @pedronis - I merged 7418 and updated the code here to use it.

mvo5: My commit does the minimal change - we could expand it further and use your new "Options" argument in Find() for LK. We probably also need to tweak InstallBootConfig so that ConfigFiles takes an argument as well or is just replaced with something else.

pedronis: yes, that was the point of the new argument. We don't need to tweak InstallBootConfig a lot, we know the different bootloaders there

include/snappy_boot_v1.h
```diff
@@ -0,0 +1,117 @@
+/**
```
pedronis: this should be in include/lk  , no?

not completely a fan of growing an include dir just for this.  @mvo5 thoughts?

kubiko: @pedronis happy to move it where you like. This was suggestion from @zyga 
I'm myself wondering what's best place to put this. But for reference I do want to have it somewhere in snapd repo

kubiko: moved to `include/lk`

include/lk/snappy_boot_v1.h
```diff
@@ -0,0 +1,117 @@
...
+               bootloader can pick correct kernel snap to use for boot
+             - snap_mode, snap_try_kernel, snap_try_core behaves same way as with u-boot
+             - boot partition labels are never modified by snapd at run time
+        - bootloader:
+             - Finds boot partition to use based on info in matrix and snap_kernel / snap_try_kernel
+             - bootloaer does not alter matrix, only alters snap_mode
+
+        [ <bootimg 1 part label> ] [ <currently installed kernel snap revison> ]
+        [ <bootimg 2 part label> ] [ <currently installed kernel snap revision> ]
+    */
```
pedronis: the comment here doesn't match the go code one anymore, they need syncing

include/lk/snappy_boot_v1.h
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     GADGET assets: Matrix for mapping of gadget asset partions
+     Optional boot asset tracking, based on bootloader support
+     Some boot chains support A/B boot assets for increased robustness
+     example being A/B TrustExecutionEnvironment
+     This matrix can be used to track current and try boot assets for
+     robust updates
+
+     [ <boot assets 1 part label> ] [ <currently installed assets revison> ]
```
pedronis: typo "revison", also this has the same problem of the preceding diagram

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,406 @@
...
+	 *      kernel snap revision name.
+	 *    - snap_mode, snap_try_kernel, snap_try_core behaves same way as with u-boot
+	 *  - bootloader:
+	 *    - bootloader reads snap_mode to determine if snap_kernel or snap_kernel is used
+	 *      to get kernel snap revision name
+	 *      kernel snap revision is then used to search matrix to determine
+	 *      partition label to be used for current boot
+	 *    - bootloader NEVER alters this matrix values
+	 *
+	 * [ <bootimg 1 part label> ] [ <kernel snap revison installed in this boot partition> ]
```
pedronis: typo "revison"

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,406 @@
...
+	 * [ <bootimg 2 part label> ] [ <kernel snap revison installed in this boot partition> ]
```
pedronis: same

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,406 @@
...
+
+	/**
+	 * Gadget assets: Matrix for mapping of gadget asset partions
+	 * Optional boot asset tracking, based on bootloader support
+	 * Some boot chains support A/B boot assets for increased robustness
+	 * example being A/B TrustExecutionEnvironment
+	 * This matrix can be used to track current and try boot assets for
+	 * robust updates
+	 *
+	 * [ <boot assets 1 part label> ] [ <currently installed assets revison> ]
```
pedronis: typo "revison" , also these should say "in this partition" as well ?

bootloader/lkenv/lkenv.go
```diff
@@ -0,0 +1,406 @@
...
+	/* current core snap revision */
+	Snap_core [SNAP_NAME_MAX_LEN]byte
+	/* try core snap revision */
+	Snap_try_core [SNAP_NAME_MAX_LEN]byte
+	/* current kernel snap revision */
+	Snap_kernel [SNAP_NAME_MAX_LEN]byte
+	/* current kernel snap revision */
+	Snap_try_kernel [SNAP_NAME_MAX_LEN]byte
+
+	/* GADGET assets: current gadget assets revision */
```
pedronis: s/GADGET/gadget/ ?

kubiko: yep, I have already fixed this with next update

include/lk/snappy_boot_v1.h
kubiko: yep, I already synched those with next update

bootloader/lkenv/lkenv.go
kubiko: typo fixed and "in this partition" added

bootloader/lkenv/lkenv.go
kubiko: fixed

bootloader/lkenv/lkenv.go
kubiko: fixed

include/lk/snappy_boot_v1.h
kubiko: typo fixed
diagram was updated as part of the sync between c and go files

PR #6711, opened 2019-04-12
tests/main/selinux-lxd: make sure LXD from snaps works cleanly with enforcing SELinux

bboozzoo:
Add a test that checks whether LXD from snaps works with SELinux in enforcing mode and raises no denials related to snapd policy.

Since we do not have a sysctl backend, add a workaround for a known problem with LXD on CentOS.

This should be the final PR in the SELinux series.

(cc @Conan-Kudo )

### DIFF
--- a/data/sysctl/rhel7-snap.conf
+++ b/data/sysctl/rhel7-snap.conf
@@ -2,3 +2,10 @@
 # Unexpected "Device or resource busy" error when removing a directory
 # see https://access.redhat.com/articles/3128691 for details
 fs.may_detach_mounts=1
+
+# RHEL 7.4+ specific:
+# A configuration change was made in the 7.4 release, the default number of user
+# namespaces is set to zero, see https://access.redhat.com/solutions/3188102 for
+# details
+# LXD specific fix
+user.max_user_namespaces=15000
--- a/data/sysctl/rhel7-snap.conf
+++ b/data/sysctl/rhel7-snap.conf
@@ -3,9 +3,9 @@
 # see https://access.redhat.com/articles/3128691 for details
 fs.may_detach_mounts=1
 
-# RHEL 7.4+ specific:
+# RHEL 7.4+ (and LXD) specific:
 # A configuration change was made in the 7.4 release, the default number of user
 # namespaces is set to zero, see https://access.redhat.com/solutions/3188102 for
-# details
-# LXD specific fix
+# details. For details about LXD side, see
+# https://discuss.linuxcontainers.org/t/lxd-on-centos-7/1250
 user.max_user_namespaces=15000
--- a/tests/main/selinux-lxd/task.yaml
+++ b/tests/main/selinux-lxd/task.yaml
@@ -1,0 +1,52 @@
+summary: Check that LXD snap works on a system using SELinux in enforcing mode
+
+description: |
+    Make sure that LXD snap can be installed and used on a SELinux enable
+    system, with enforcing mode on.
+
+# Systems with SELinux enabled out of the box.
+systems: [fedora-*, centos-*]
+
+# lxd downloads can be quite slow
+kill-timeout: 25m
+
+# Start before anything else as it can take a really long time.
+priority: 1000
+
+prepare: |
+    # Enable enforcing mode, our policy is already marked as permissive, so we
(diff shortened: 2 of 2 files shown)


data/sysctl/rhel7-snap.conf
```diff
@@ -2,3 +2,10 @@
 # Unexpected "Device or resource busy" error when removing a directory
 # see https://access.redhat.com/articles/3128691 for details
 fs.may_detach_mounts=1
+
+# RHEL 7.4+ (and LXD) specific:
+# A configuration change was made in the 7.4 release, the default number of user
+# namespaces is set to zero, see https://access.redhat.com/solutions/3188102 for
+# details. For details about LXD side, see
+# https://discuss.linuxcontainers.org/t/lxd-on-centos-7/1250
+user.max_user_namespaces=15000
```
stolowski: It's a bit unfortunate we have to touch a global limit, but well... 
The forum topic you linked seems to be suggesting much lower value, any reason to go higher?

tests/main/selinux-lxd/task.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+description: |
+    Make sure that LXD snap can be installed and used on a SELinux enable
+    system, with enforcing mode on.
+
+# Systems with SELinux enabled out of the box.
+systems: [fedora-*, centos-*]
+
+# lxd downloads can be quite slow
+kill-timeout: 25m
```
stolowski: Do we really need this (to be that large)? FWTW tests/regression/lp-1815869 installs lxd snap with no kill-timeout defined.

tests/main/selinux-lxd/task.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+# Start before anything else as it can take a really long time.
+priority: 1000
```
stolowski: Why do we want to start it with higher prio?

data/sysctl/rhel7-snap.conf
bboozzoo: That's the value suggested by RHEL knowledge base solution.

tests/main/selinux-lxd/task.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+    lxd.lxc launch "ubuntu:18.04" my-ubuntu
+
+    echo "Ensure we can run things inside"
+    lxd.lxc exec my-ubuntu echo hello | MATCH hello
+
+    echo "Stop the container"
+    lxd.lxc stop --force my-ubuntu
+
+    snap remove lxd
```
stolowski: Should we do this in restore, and also remove the container? (see the existing lxd test I mentioned earlier).

tests/main/selinux-lxd/task.yaml
bboozzoo: Took the timeout and priority values from here: https://github.com/snapcore/snapd/blob/master/tests/main/lxd/task.yaml#L12-L16

tests/main/selinux-lxd/task.yaml
bboozzoo: Interstingly, the `tests/main/lxd` test does not remove the containers it creates. Perhaps what the regression test does is unnecessary?

stolowski: IMHO we should remove the containers not to have leftovers that could affect other tests. At the moment they are probably wiped from the disk on lxd snap removal, but with the automatic snapshots we will be needlessly creating a snapshot here.

tests/main/selinux-lxd/task.yaml
stolowski: Hmmm, ok.

tests/main/selinux-lxd/task.yaml
bboozzoo: Sam motivation as here https://github.com/snapcore/snapd/pull/6711#discussion_r274840547

tests/main/selinux-lxd/task.yaml
bboozzoo: Pushed an update

stolowski: Thanks!

PR #6713, opened 2019-04-12
tests/upgrade/basic, packaging/fedoar: restore SELinux context of /var/cache/fontconfig, patch pre-2.39 mount units

bboozzoo:
The test installs an older version of snapd that may be missing proper context
transitions for /var/cache/fontconfig when created in the context of snapd
daemon (snappy_t). Avoid breaking tests that run after this one and restore the
context to whatever is the default.

### DIFF
--- a/packaging/fedora/snapd.spec
+++ b/packaging/fedora/snapd.spec
@@ -832,7 +832,7 @@ fi
 
 # TODO: the trigger relies on a very specific snapd version that introduced SELinux
 # mount context, figure out how to update the trigger condition to run when needed
-%triggerun -- snapd < 2.38
+%triggerun -- snapd < 2.39
 # Trigger on uninstall, with one version of the package being pre 2.38 see
 # https://rpm-packaging-guide.github.io/#triggers-and-scriptlets for details
 # when triggers are run
--- a/packaging/fedora/snapd.spec
+++ b/packaging/fedora/snapd.spec
@@ -840,6 +840,13 @@ fi
 if [ "$1" -eq 2 -a "$2" -eq 1 ]; then
    # Upgrade from pre 2.38 version
    %{_libexecdir}/snapd/snap-mgmt-selinux --patch-selinux-mount-context=system_u:object_r:snappy_snap_t:s0 || :
+
+   # snapd might have created fontconfig cache directory earlier, but with
+   # incorrect context due to bugs in the policy, make sure it gets the right one
+   # on upgrade when the new policy was introduced
+   if [ -d "%{_localstatedir}/cache/fontconfig" ]; then
+      restorecon -R %{_localstatedir}/cache/fontconfig || :
+   fi
 elif [ "$1" -eq 1 -a "$2" -eq 2 ]; then
    # Downgrade to a pre 2.38 version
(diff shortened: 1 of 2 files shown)


tests/upgrade/basic/task.yaml
```diff
@@ -12,6 +12,17 @@ restore: |
     rm -f /var/tmp/myevil.txt
     rm -f ./do-classic ./prevsnapdver.out
 
+    # an older version of snapd might have been missing proper context
+    # transitions for fontconfig cache locations, attempt to restore the context
+    # to avoid breaking tests that run after this one
```
stolowski: Isn't this a concern also for normal update in the distro?

bboozzoo: It is, but I would avoid fixing it programmatically in snapd as this may be overstepping distro boundaries a bit. Probably best if @Conan-Kudo could chime in. I fear, the situation regarding fontconfig cache in Fedora is not yet fully addressed.

Conan-Kudo: Yes, it's a concern sometimes, but if it is our fault, we might want a trigger in the package to restore the context during the package upgrade.

bboozzoo: I'll add a bit to the upgrade trigger.

mvo5:
Do we need a review from Neal for this or can this go in? It looks fine to me FWIW.

packaging/fedora/snapd.spec
```diff
@@ -840,6 +840,13 @@ fi
 if [ "$1" -eq 2 -a "$2" -eq 1 ]; then
    # Upgrade from pre 2.38 version
    %{_libexecdir}/snapd/snap-mgmt-selinux --patch-selinux-mount-context=system_u:object_r:snappy_snap_t:s0 || :
+
+   # snapd might have created fontconfig cache directory earlier, but with
+   # incorrect context due to bugs in the policy, make sure it gets the right one
+   # on upgrade when the new policy was introduced
+   if [ -d "%{_localstatedir}/cache/fontconfig" ]; then
+      restorecon -R %{_localstatedir}/cache/fontconfig || :
```
zyga: Curious about the `|| :` pattern used in the scriptlet, is that typical?

tests/upgrade/basic/task.yaml
```diff
@@ -12,6 +12,17 @@ restore: |
     rm -f /var/tmp/myevil.txt
     rm -f ./do-classic ./prevsnapdver.out
 
+    # an older version of snapd might have been missing proper context
```
zyga: Please capitalise sentences.

```suggestion
    # An older version of snapd might have been missing proper context
```

packaging/fedora/snapd.spec
mvo5: Heh, I noticed the same and assumed its a RPM pattern but its good to double check.

bboozzoo: Yes, it's a pattern that repeats in the macros and snippets: https://fedoraproject.org/wiki/PackagingDrafts/SELinux#File_contexts

PR #6715, opened 2019-04-12
interfaces/builtin/desktop: fonconfig v6/v7 cache handling on Fedora

bboozzoo:
Snap mount namespace includes host's /etc. An old fontconfig version may fail to
parse /etc/fonts/fonts.conf and fall back to using default locations, among them
/var/cache/fontconfig as cache directory. Since fc-cache-v6 would have created
the directory on the host side, make sure to make it available for the snap.


Conan-Kudo:
What is this still blocked on?

PR #6718, opened 2019-04-15
spread, tests: do not leave mislabeled files in restorecon test, attempt to catch similar files

bboozzoo:
The test created a ~/snap with entries that were intentionally mislabeled, but failed to remove the directory afterwards.

As part of the solution, try to catch when tests are leaving such entries behind so that we can enforce proper cleanup.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -409,6 +409,11 @@ debug-each: |
                 else
                     ausearch -m AVC || true
                 fi
+                (
+                    find /root/snap -printf '%Z\t%H/%P\n' || true
+                    find /home/test/snap -printf '%Z\t%H/%P\n' || true
+                ) | grep -v snappy_home_t || true
+                find /var/snap -printf '%Z\t%H/%P' | grep -v snappy_var_t || true
                 ;;
             *)
                echo '# apparmor denials '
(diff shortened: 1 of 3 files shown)


spread.yaml
```diff
@@ -409,6 +409,11 @@ debug-each: |
                 else
                     ausearch -m AVC || true
                 fi
+                (
+                    find /root/snap -printf '%Z\t%H/%P\n' || true
+                    find /home/test/snap -printf '%Z\t%H/%P\n' || true
```
zyga: Nitpick: can we expand that to:
```suggestion
                    find /home/ -path "/home/*/snap/" -printf '%Z\t%H/%P\n' || true
```

spread.yaml
```diff
@@ -577,6 +582,18 @@ restore: |
 restore-each: |
     "$TESTSLIB"/prepare-restore.sh --restore-project-each
     if journalctl -u snapd.service | grep -F "signal: terminated"; then exit 1; fi
+    case "$SPREAD_SYSTEM" in
+        fedora-*|centos-*)
+            # Make sure that we are not leaving behind incorrectly labeled snap
+            # files on systems supporting SELinux
+            (
+                find /root/snap -printf '%Z\t%H/%P\n' || true
+                find /home/test/snap -printf '%Z\t%H/%P\n' || true
```
zyga: Same suggestion as above.

PR #6721, opened 2019-04-16
tests: new profiler snap used to track cpu and memory for snapd and snap commands

sergiocazzolato:
This snap can be used to get profile information of snapd and snap coomands.

This is the initial step, then the profiler should evolve being more smart and capturing other information that could help to understand issues.


tests/lib/snaps/test-snapd-profiler-core18/profiler.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
+
+import datetime
```
bboozzoo: I would suggest to tweak the tool like this:
```
#!/usr/bin/env python3

import logging
import time
import json
import psutil

PROCATTRS = set(['pid', 'name', 'cmdline', 'cpu_times',
                 'create_time', 'cpu_percent', 'memory_info',
                 'memory_percent'])
PROCS = set(["snap", "snapd"])
RATE = {"snap": 3, "snapd": 1}
INTERVAL = 1


def main():
    logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(message)s')

    count = 0
    while True:
        procs = [proc for proc in psutil.process_iter() if proc.name() in PROCS]
        for proc in procs:
            if count % RATE.get(proc.name(), 1) == 0:
                logging.info(json.dumps(proc.as_dict(attrs=PROCATTRS)))
        time.sleep(INTERVAL)


if __name__ == '__main__':
    main()
```

Plus the changes to log into a file. Instead of logging a dictionary of attributes (some of which may be converted to a string in an unexpected way), the output is JSON. This would enable us to apply further processing with ease.

tests/lib/snaps/test-snapd-profiler/profiler.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
```
bboozzoo: It'd be nice if we could find a way to share the file in the source tree.

zyga: What did you mean by share?

bboozzoo: Having only one copy of the file in snapd source tree. I recall we have the same problem with shell wrappers for snap, there's a bunch of files under `tests/lib/snaps/*/bin` that are copies.

sergiocazzolato: Totally agree on this. I think I could create a new PR with that change fixing all the repeated files. Are you ok with this?

bboozzoo: Yes, a follow up sounds ok.

tests/lib/snaps/test-snapd-profiler-core18/profiler.py
```diff
@@ -0,0 +1,85 @@
...
+        return None
+
+
+def main():
+    logging.basicConfig(filename=LOG_PATH, filemode='w', level=logging.INFO, format='%(asctime)s - %(message)s')
+    config = get_config()
+
+    count = 0
+    while True:
+        if count % config.get(RATE_CONFIG, 1) == 0:
```
bboozzoo: Why the need to reload the configuration periodically?

tests/lib/snaps/test-snapd-profiler-core18/profiler.py
sergiocazzolato: done

tests/lib/snaps/test-snapd-profiler-core18/profiler.py
sergiocazzolato: now it is just reloading the config when it starts or when the flag has been removed.
Thanks for the suggestion.

bboozzoo:
Shellcheck does not approve:
```
In tests/lib/prepare-restore.sh line 549:
        logs_id=$(ls "$logs_dir"/*.journal.log 2>/dev/null | wc -l)
                  ^-- SC2012: Use find instead of ls to better handle non-alphanumeric filenames.
```

PR #5623, opened 2018-08-10
advise-snap: add --dump-db which dumps the command database

shawnl:
This is a hook needed for my new command-not-found implementation[1], which
was written to be fast, and remove the dependency on Python.

Currently Ubuntu is running two programs (one which requires an
interpreter) and opening two huge databases
for every command_not_found_handler invocation.

I also will need a hook somewhere so snap can call "update-command-not-found"
whenever it updates its command database.

[1] https://github.com/shawnl/command-not-found

Would you like me to export this data in a more neutral format?, such as JSON?

PR #6734, opened 2019-04-16
advise-snap: add --dump-db which dumps the command database

shawnl:
This is a hook needed for my new command-not-found implementation[1], which
was written to be fast, and remove the dependency on Python.

Currently Ubuntu is running two programs (one which requires an
interpreter) and opening two huge databases
for every command_not_found_handler invocation.

I also will need a hook somewhere so snap can call "update-command-not-found"
whenever it updates its command database.

[1] https://github.com/shawnl/command-not-found

see  #5623


mvo5:
Thanks for this PR! I ran `go fmt` over it and pushed to this PR. This should make CI happy :)

shawnl:
Whoops, I thought that I didn't submit the whole patch, and forced over that....I just ran it myself and then pushed.

mvo5:
Hey @shawnl - the PR looks good but it seems that `shawn@git.icu                       has no Launchpad account` is unkown to launchpad, could you please add it as an address to your launchpad account so that the CLA check will work? Thank you!

shawnl:
bump

shawnl:
What's next?

mvo5:
Hey @shawnl ! Sorry that this was dragging on for so long. This looks fine now, we will do a followup that adds tests but that should not concern you, we will take care of it :) Thanks for your contribution and I look forward to the new fast! command-not-found :)

PR #6735, opened 2019-04-17
tests: wait for man db cache is updated before after install snapd on Fedora

sergiocazzolato:
The service run-<id>.service which runs systemctl start
man-db-cache-update fails when we do daemon-reexec after install snpad,
leaving the system in degraded state. This is happening with the new
images for fedora.

To test this error and the fix use:
- fedora-29-64:
    image: test-fedora-1

The idea of the fix is to wait until this service finishes and then make
daemon-reexec.

### DIFF
--- a/tests/lib/pkgdb.sh
+++ b/tests/lib/pkgdb.sh
@@ -475,6 +475,16 @@ distro_install_build_snapd(){
 
         case "$SPREAD_SYSTEM" in
             fedora-*|centos-*)
+                # We need to wait until the man db cache is updated before do daemon-reexec
+                # Otherwise the service fails and the system will be degraded during tests executions
+                for i in $(seq 20); do
+                    man_db_cache_update_service=$(systemctl list-units | grep -E 'run-.*.service' | awk '{print $1;}')
+                    if [ -z "$man_db_cache_update_service" ]; then
+                        break
+                    fi
+                    sleep .5
+                done
+
                 # systemd caches SELinux policy data and subsequently attempts
                 # to create sockets with incorrect context, this installation of
                 # socket activated snaps fails, see:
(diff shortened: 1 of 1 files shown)


tests/lib/pkgdb.sh
```diff
@@ -475,6 +475,16 @@ distro_install_build_snapd(){
 
         case "$SPREAD_SYSTEM" in
             fedora-*|centos-*)
+                # We need to wait until the man db cache is updated before do daemon-reexec
+                # Otherwise the service fails and the system will be degraded during tests executions
+                for i in $(seq 20); do
+                    man_db_cache_update_service=$(systemctl list-units | grep -E 'run-.*.service' | awk '{print $1;}')
```
zyga: Perhaps this can be simplified to use `systemctl is-active`?

sergiocazzolato: Thanks for reviewing this. I didn't know is-active could take a pattern, I already made the change. 
Thanks for the suggestion.

bboozzoo:
If the problem is with `man-db-cache-update.service`, can't we just disable/mask it instead? 

Also:
```
[guest@localhost ~]$ cat /etc/sysconfig/man-db 
# Set this to "no" to disable man-db update triggered by installation
# of any package containing manual pages
SERVICE="yes"
```
Perhaps using SERVICE="no" would be enough?

zyga:
@bboozzoo the problem is not with that service, it is that reloading system breaks ongoing jobs (apparently). I think this is more of a systemd bug than anything else.

sergiocazzolato:
@bboozzoo I already tried disabling the service but didn't work. Thanks for reviewing.

PR #6736, opened 2019-04-17
cmd/snap-update-ns: rename variable "up" to "ctx"

zyga:
The up variable used to stand for "mount profile update" but during the
last review pass it was renamed to "mount update context", hence the
"ctx".

A few leftovers remained though, this patch renames them too.


zyga:
NOTE: I have a late follow up patch that renames `ctx` to `upCtx` but for the purpose of not wasting time rebasing the set again I chose to use `ctx` here and in the subsequent patch-set and only fix it all in the final patch.

PR #6741, opened 2019-04-17
osutil,cmdutil: move CommandFromCore and make it use the snapd snap (if available)

mvo5:
The CommandFromCore helper is currently always looking at the
core snap. With the snapd snap we need to refactor it so that
it uses the snapd snap if available and just falls back to the
core snap. 

This PR also moves the code from osutil (it is a bit misplaced there)
to cmd/cmdutil - we cannot have it in cmd/ because there is also
cmd/snapinfo.go which will trigger an import cycle if we move it
there. We should also move cmd.InternalToolPath to cmdutil but that
should be done in a separate PR.

This should unblock #6404. 

For easier review I could split this and just do a pure rename in the
first PR and the tweak the code in the second (and third?) PR. But
its *not* that much change so maybe its ok to review in one go.
Feedback welcome.

### DIFF
--- a/cmd/cmd_linux.go
+++ b/cmd/cmd_linux.go
@@ -114,7 +114,7 @@ func coreSupportsReExec(corePath string) bool {
 	return true
 }
 
-// TODO: move to cmd/cmdutil.go
+// TODO: move to cmd/cmdutil/
 //
 // InternalToolPath returns the path of an internal snapd tool. The tool
 // *must* be located inside the same tree as the current binary.
--- a/cmd/cmd_linux.go
+++ b/cmd/cmd_linux.go
@@ -114,6 +114,8 @@ func coreSupportsReExec(corePath string) bool {
 	return true
 }
 
+// TODO: move to cmd/cmdutil.go
+//
 // InternalToolPath returns the path of an internal snapd tool. The tool
 // *must* be located inside the same tree as the current binary.
 //
--- a/cmd/cmdutil/cmdutil.go
+++ b/cmd/cmdutil/cmdutil.go
@@ -112,7 +112,7 @@ func CommandFromSystemSnap(name string, cmdArgs ...string) (*exec.Cmd, error) {
 	}
 	coreLdSo := filepath.Join(root, interp)
 	// we cannot use EvalSymlink here because we need to resolve
-	// relative and absolute symlinks differently. A absolute
+	// relative and an absolute symlinks differently. A absolute
 	// symlink is relative to root of the snapd/core snap.
 	seen := map[string]bool{}
 	for osutil.IsSymlink(coreLdSo) {
--- a/cmd/cmdutil/cmdutil.go
+++ b/cmd/cmdutil/cmdutil.go
@@ -1,0 +1,139 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+package cmdutil
+
+import (
+	"bufio"
+	"bytes"
+	"debug/elf"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/snapcore/snapd/dirs"
+	"github.com/snapcore/snapd/osutil"
+)
+
+func elfInterp(cmd string) (string, error) {
+	el, err := elf.Open(cmd)
+	if err != nil {
+		return "", err
+	}
+	defer el.Close()
+
+	for _, prog := range el.Progs {
+		if prog.Type == elf.PT_INTERP {
+			r := prog.Open()
+			interp, err := ioutil.ReadAll(r)
+			if err != nil {
+				return "", nil
+			}
+
+			return string(bytes.Trim(interp, "\x00")), nil
+		}
+	}
+
+	return "", fmt.Errorf("cannot find PT_INTERP header")
+}
+
+func parseLdSoConf(root string, confPath string) []string {
+	f, err := os.Open(filepath.Join(root, confPath))
+	if err != nil {
+		return nil
+	}
+	defer f.Close()
+
+	var out []string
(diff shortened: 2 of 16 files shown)


snapcraft.yaml
```diff
@@ -29,6 +29,9 @@ parts:
     # FIXME: this should probably go upstream
     plugin: x-builddeb
     source: .
+    stage-packages:
+      - xdelta3
```
stolowski: Silly question, why do we need them in the snapd snap?

tests/main/snapd-without-core/task.yaml
```diff
@@ -0,0 +1,46 @@
...
+        exit 0
+    fi
+
+    echo "Create modified snapd snap"
+    UNPACK_DIR="./snapd-unpack"
+    snap download --edge snapd
+    unsquashfs -no-progress -d "$UNPACK_DIR" snapd_*.snap
+    dpkg-deb -x "$SPREAD_PATH"/../snapd_*.deb "$UNPACK_DIR"
+    cp /usr/lib/snapd/info "$UNPACK_DIR"/usr/lib/
+    snap pack "$UNPACK_DIR" /tmp
```
stolowski: The same logic exists in prepare.sh in setup_reflash_magic, could we have it in a common helper instead? I can imagine more tests requiring snapd snap where this will be needed.

snapcraft.yaml
mvo5: We use this in case the host has no delta installed when doing delta updates.

cmd/cmdutil/cmdutil.go
```diff
@@ -0,0 +1,139 @@
...
+
+// CommandFromSystemSnap runs a command from the snapd/core snap
+// using the proper interpreter and library paths.
+//
+// At the moment it can only run ELF files, expects a standard ld.so
+// interpreter, and can't handle RPATH.
+func CommandFromSystemSnap(name string, cmdArgs ...string) (*exec.Cmd, error) {
+	from := "snapd"
+	root := filepath.Join(dirs.SnapMountDir, "/snapd/current")
+	if !osutil.FileExists(root) {
```
bboozzoo: I think we're missing a unit test for the `snapd` snap case.

snapcraft.yaml
bboozzoo: For the record, the docs/README coming from the packages will be gone in the snapd snap?

cmd/cmd_linux.go
```diff
@@ -114,6 +114,8 @@ func coreSupportsReExec(corePath string) bool {
 	return true
 }
 
+// TODO: move to cmd/cmdutil.go
```
chipaca: minor nit: `cmd/cmdutil/`

cmd/cmdutil/cmdutil.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	cmdPath := filepath.Join(root, name)
+	interp, err := elfInterp(cmdPath)
+	if err != nil {
+		return nil, err
+	}
+	coreLdSo := filepath.Join(root, interp)
+	// we cannot use EvalSymlink here because we need to resolve
+	// relative and absolute symlinks differently. A absolute
```
chipaca: mother of all nitpicks: *An* absolute

snapcraft.yaml
chipaca: this change stages the whole package instead of picking out the binaries it needs, which leaves `/usr/share/doc/` of a lot of things, as well as manpages. Compared to the whole contents of the `snap` directory (from snapd's tree) getting copied, this is minor though. Could easily be part of a separate PR that tidied things up.

tests/lib/prepare.sh
```diff
@@ -307,6 +307,16 @@ prepare_classic() {
     fi
 }
 
+repack_snapd_snap_with_deb_content() {
+    TARGET="$1"
```
chipaca: use `local` a bit more?

snapcraft.yaml
mvo5: Good point, thanks, I updated this to use more parts and more targeted includes.

tests/lib/prepare.sh
```diff
@@ -307,6 +307,16 @@ prepare_classic() {
     fi
 }
 
+repack_snapd_snap_with_deb_content() {
```
stolowski: Neat, thank you!

bboozzoo:
More trouble ahead:
```
2019-04-29 16:06:24 Error preparing google:ubuntu-core-18-64:tests/core18/ : 
-----
+ /home/gopath/src/github.com/snapcore/snapd/tests/lib/prepare-restore.sh --prepare-suite
+ set -e
+ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/quiet.sh
+ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/boot.sh
++ GRUB_EDITENV=grub-editenv
++ case "$SPREAD_SYSTEM" in
+ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/dirs.sh
++ export SNAP_MOUNT_DIR=/snap
++ SNAP_MOUNT_DIR=/snap
++ export LIBEXECDIR=/usr/lib
++ LIBEXECDIR=/usr/lib
++ export MEDIA_DIR=/media
++ MEDIA_DIR=/media
++ case "$SPREAD_SYSTEM" in
+ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/pkgdb.sh
++ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/quiet.sh
+ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/random.sh
+ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/spread-funcs.sh
+ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/journalctl.sh
++ . /home/gopath/src/github.com/snapcore/snapd/tests/lib/state.sh
+++ SNAPD_STATE_PATH=/home/gopath/src/github.com/snapcore/snapd/tests/snapd-state
...
--extra-snaps is deprecated; use --snap instead
-o/--output is deprecated; use -O/--output-dir instead
+ rm -f './pc-kernel_*.snap' './pc-kernel_*.assert' './pc_*.snap' './pc_*.assert'
+ kpartx -avs /home/image/core18-amd64.img
add map loop3p1 (253:0): 0 2048 linear 7:3 2048
add map loop3p2 (253:1): 0 102400 linear 7:3 4096
add map loop3p3 (253:2): 0 793912 linear 7:3 106496
+ is_core18_system
+ [[ ubuntu-core-18-64 == ubuntu-core-18-* ]]
+ return 0
+ mount /dev/mapper/loop3p3 /mnt
+ mkdir -p /mnt/user-data/
+ rsync -a -C --exclude '*.a' --exclude '*.deb' --exclude /gopath/.cache/ --exclude /gopath/bin/govendor --exclude /gopath/pkg/ /home/gopath /mnt/user-data/
rsync: write failed on "/mnt/user-data/gopath/src/github.com/snapcore/snapd/tests/core18/kernel/snapd-unpack/usr/lib/snapd/snap-seccomp": No space left on device (28)
rsync error: error in file IO (code 11) at receiver.c(393) [receiver=3.1.1]
```

PR #6743, opened 2019-04-17
cmd/snap, client, daemon, store: create-cohort

chipaca:
This adds support for `snap create-cohort`, all the way from `store` through to `cmd/snap`.


chipaca:
The test failure raises a very good question: where should `create-cohort` be categorised?
One alternative would be to move `refresh` into `basics` ("basic snap management"), and add `create-cohort` to `...more` ("slightly more advanced snap management"),
```
Commands can be classified as follows:

         Basics: find, info, install, list, remove, refresh
        ...more: revert, switch, disable, enable, create-cohort
        History: changes, tasks, abort, watch
        Daemons: services, start, stop, restart, logs
       Commands: alias, aliases, unalias, prefer
  Configuration: get, set, wait
        Account: login, logout, whoami
    Permissions: connections, interface, connect, disconnect
      Snapshots: saved, save, check-snapshot, restore, forget
          Other: version, warnings, okay, ack, known
    Development: run, pack, try, download, prepare-image
```

but it doesn't *feel* like it's only "slightly" more advanced. It feels like it's a honkin' full-on advanced feature.

Another alternative would be to tuck it into `other` ("miscellanea"),
```
Commands can be classified as follows:

         Basics: find, info, install, list, remove
        ...more: refresh, revert, switch, disable, enable
        History: changes, tasks, abort, watch
        Daemons: services, start, stop, restart, logs
       Commands: alias, aliases, unalias, prefer
  Configuration: get, set, wait
        Account: login, logout, whoami
    Permissions: connections, interface, connect, disconnect
      Snapshots: saved, save, check-snapshot, restore, forget
          Other: version, warnings, okay, ack, known, create-cohort
    Development: run, pack, try, download, prepare-image
```
but 'other' does feel like a cop-out.

A third option would be to not categorise it, and have it be discoverable via the help for `--cohort` in `install` and `refresh`. It would be the only non-hidden non-debug command (other than help itself) that is not categorised in this way. I don't like this one.

zyga:
@chipaca about the help categorisation: cohort feels like an enterprise / cloud feature. Perhaps we need a section for that: `Enterprise: create-cohort`

daemon/api_cohort.go
```diff
@@ -0,0 +1,59 @@
...
+	if dec.More() {
+		return BadRequest("spurious content after cohort instruction")
+	}
+
+	if inst.Action != "create" {
+		return BadRequest("unknown cohort action %q", inst.Action)
+	}
+
+	if len(inst.Snaps) == 0 {
+		return BadRequest("snaps cannot be empty")
```
zyga:
```suggestion
		return BadRequest("cohort must relate to at least one snap")
```

image/helpers.go
```diff
@@ -206,38 +208,81 @@ func NewToolingStore() (*ToolingStore, error) {
...
+func elliptLeft(str string) string {
+	if len(str) < 10 {
+		// shouldn't happen outside of tests
+		return str
+	}
+	return "…" + str[len(str)-8:]
+}
+
+func (opts *DownloadOptions) String() string {
+	var spec []string
```
zyga: Nitpick:

```suggestion
	spec := make([]string, 0, 4)
```

store/store.go
```diff
@@ -2453,3 +2442,37 @@ func (s *Store) ConnectivityCheck() (status map[string]bool, err error) {
 
 	return status, nil
 }
+
+func (s *Store) CreateCohorts(ctx context.Context, snaps []string) (map[string]string, error) {
+
```
zyga: Stray newline?
```suggestion
```

tests/main/snap-download/task.yaml
```diff
@@ -33,3 +34,8 @@ execute: |
     su -l -c "SNAPPY_USE_STAGING_STORE=$SNAPPY_USE_STAGING_STORE HTTPS_PROXY=$HTTPS_PROXY snap download test-snapd-tools" test
     ls /home/test/test-snapd-tools_*.snap
     verify_asserts /home/test/test-snapd-tools_*.assert
+
+    echo "Snap download can download snaps from a cohort"
+    snap download --cohort="MSBlRmU4QlRSNUw1VjlGN3lIZU1BUHhrRXIyTmRVWE10dyAxNTU1NTE0MzA5IDg3ZmUwMjhkZDFjMTQ1MDY5N2QyZjdiMGZkMzgzODk0NjMzMmFhOTZmZmFjZmFlNmU2MGQyOTNjYzE1OTE3NWY=" test-snapd-tools
```
zyga: Perhaps add a note about the relevance of the value (that is, it should be not relevant?)

cmd/snap/cmd_create_cohort.go
```diff
@@ -0,0 +1,84 @@
...
+package main
+
+import (
+	"github.com/jessevdk/go-flags"
+	"gopkg.in/yaml.v2"
+
+	"github.com/snapcore/snapd/i18n"
+)
+
+var shortCreateCohortHelp = i18n.G("Create cohort keys for a series of snaps")
```
zyga: This and the description below read well.

cmd/snap/cmd_create_cohort.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func init() {
+	addCommand("create-cohort", shortCreateCohortHelp, longCreateCohortHelp, func() flags.Commander { return &cmdCreateCohort{} }, nil, nil)
+}
+
+type cohortInner struct {
+	CohortKey string `yaml:"cohort-key"`
+}
+type cohortOut struct {
```
zyga: Perhaps name the two structures with the word YAML somewhere to make it clear that it is purely about the yaml representation of the response.

daemon/api_cohort.go
chipaca: yes i agree the error message is poor.
but, there is no 'relation': you're asking for a cohort key for each of those snaps, not one cohort that covers them all.
`"cannot request cohorts for 0 snaps"`?

chipaca: or, just detect it, accept it, and return a `{}`

image/helpers.go
chipaca: fair

tests/main/snap-download/task.yaml
chipaca: I'll add a note in the other pr

cmd/snap/cmd_create_cohort.go
chipaca: I might have parapharsed the description from the [forum](https://forum.snapcraft.io/t/managing-cohorts/8995).

cmd/snap/cmd_create_cohort.go
chipaca: k

pedronis:
@chipaca I would put it under Other for the time being, Development is not the right category, don't think we want to introduce a category with just one command either

daemon/api_cohort.go
chipaca: In the end I'm just returning an empty map in this case

cmd/snap/cmd_create_cohort.go
chipaca: done

cmd/snap/cmd_create_cohort.go
```diff
@@ -0,0 +1,85 @@
...
+	"gopkg.in/yaml.v2"
+
+	"github.com/snapcore/snapd/i18n"
+)
+
+var shortCreateCohortHelp = i18n.G("Create cohort keys for a series of snaps")
+var longCreateCohortHelp = i18n.G(`
+The create-cohort command creates a set of cohort keys for a given set of snaps.
+
+A cohort is a view or snapshot of a snap's "channel map" at a given point in
```
stolowski: Should we avoid the word "snapshot" here, not to create confusion wrt actual "snapshots"?

chipaca: I think it saying "view or snapshot" avoids the confusion.

If it doesn't, we also need to change [the docs](https://forum.snapcraft.io/t/managing-cohorts/8995).

store/store.go
```diff
@@ -2480,3 +2482,36 @@ func (s *Store) ConnectivityCheck() (status map[string]bool, err error) {
...
+	}
+	resp, err := s.retryRequestDecodeJSON(ctx, reqOptions, nil, &remote, nil)
+	if err != nil {
+		return nil, err
+	}
+	switch resp.StatusCode {
+	case 200:
+		// OK
+	case 404:
+		return nil, ErrSnapNotFound
```
stolowski: Just wondering, would be nice to say which particular snap was not found when you pass multiple.. but that I guess would need changes in the store.

chipaca: we'd have to parse the error response; compare:
```
~$ http https://api.staging.snapcraft.io/v2/cohorts snaps:='["test-snapd-tools-foo", "xyzzy-moopmoop"]'
HTTP/1.1 404 NOT FOUND
Content-Length: 189
Content-Type: application/json
Date: Mon, 29 Apr 2019 09:20:02 GMT
Server: gunicorn/19.7.1
Snap-Store-Version: 13
X-Request-Id: D9A9015EDB1C0A324E7501BB5CC6C1C211766
X-VCS-Revision: 165f1b8
X-View-Name: snapdevicegw.webapi_cohort.create_cohort

{
    "error-list": [
        {
            "code": "resource-not-found",
            "message": "No snap with name 'xyzzy-moopmoop' found"
        },
        {
            "code": "resource-not-found",
            "message": "No snap with name 'test-snapd-tools-foo' found"
        }
    ]
}

~$ http https://api.staging.snapcraft.io/v2/cohorts snaps:='["test-snapd-tools", "xyzzy-moopmoop"]'
HTTP/1.1 404 NOT FOUND
Content-Length: 100
Content-Type: application/json
Date: Mon, 29 Apr 2019 09:20:39 GMT
Server: gunicorn/19.7.1
Snap-Store-Version: 13
X-Request-Id: D9A9015EDB200A324E7501BB5CC6C1E61176D
X-VCS-Revision: 165f1b8
X-View-Name: snapdevicegw.webapi_cohort.create_cohort

{
    "error-list": [
        {
            "code": "resource-not-found",
            "message": "No snap with name 'xyzzy-moopmoop' found"
        }
    ]
}
```

cmd/snap/cmd_create_cohort_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	var v map[string]map[string]map[string]string
+	c.Assert(yaml.Unmarshal(s.stdout.Bytes(), &v), check.IsNil)
+	c.Check(v, check.DeepEquals, map[string]map[string]map[string]string{
+		"cohorts": {
+			"foo": {"cohort-key": "what"},
+			"bar": {"cohort-key": "this"},
+		},
+	})
+}
```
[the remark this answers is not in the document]
chipaca: Done.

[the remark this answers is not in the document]
stolowski: Thank you!

store/store.go
stolowski: I see, thanks for the example, it would be nice to do at some point IMHO.

cmd/snap/cmd_create_cohort.go
stolowski: I see, although IMHO it's still an "OR", so a bit confusing. But it's a tiny detail which we can quickly fix afterwards if we decide so.

PR #6744, opened 2019-04-17
tests: make test parallel-install-interfaces work for boards with pre-installed snaps

sergiocazzolato:
This change is to make this test work on boards wich already have other
snaps connected to home interface.

This is the current error: https://paste.ubuntu.com/p/QwwTDp9c7H/


zyga:
As a style nitpick, the pastebin URL will eventually stop working. It would be much better to incorporate the *essential* part of the log into the commit message for everyone to see straight away.

zyga:
@sergiocazzolato please amend the commit message and force push here. Adjust the PR commit message as well with the same text.

sergiocazzolato:
@zyga info added to the commit msg

PR #6745, opened 2019-04-17
tests: make snap-connections test work on boards with snaps pre-installed

sergiocazzolato:
This snap is failing in boards because some of them have snaps
pre-installed.

Log:
https://paste.ubuntu.com/p/9SXT5pKtQG/


zyga:
As a style nitpick, the pastebin URL will eventually stop working. It would be much better to incorporate the essential part of the log into the commit message for everyone to see straight away.

PR #6749, opened 2019-04-18
cmd/libsnap: remove fringe error function

zyga:
The codebase before snap-confine had a family of functions such as
debug, error and die. The error function simply printed a message to
stderr. We have exactly one use of that function:

        sc_error *err SC_CLEANUP(sc_cleanup_error) = NULL;
        snap_context =
            sc_cookie_get_from_snapd(invocation.snap_instance, &err);
        if (err != NULL) {
            error("%s\n", sc_error_msg(err));
        }

When we cannot get a cookie from snapd, we print the error and carry on.
This is not really useful, the possible errors, apart from memory
allocation and I/O, are simply the absence of the cookie.

This code was originally written this way because ancient snapd was not
writing the cookies and, when upgrading, the previously fatal error
would prevent apps from running.

The error message is confusing and doesn't help: users cannot do
anything about it. As such, don't print the error at all.

### DIFF
--- a/cmd/snap-confine/snap-confine.c
+++ b/cmd/snap-confine/snap-confine.c
@@ -358,7 +358,13 @@ int main(int argc, char **argv)
 		sc_error *err SC_CLEANUP(sc_cleanup_error) = NULL;
 		snap_context =
 		    sc_cookie_get_from_snapd(invocation.snap_instance, &err);
-		/* The error is explicitly ignored because the cookies are optional. */
+		/* While the cookie is normally present due to various protection
+		 * mechanisms ensuring its creation from snapd, we are not considering
+		 * it a critical error for snap-confine in the case it is absent. When
+		 * absent snaps attempting to utilize snapctl to interact with snapd
+		 * will fail but it is more important to run a little than break
+		 * entirely in case snapd-side code is incorrect. Therefore error
+		 * information is collected but discarded. */
 	}
 
 	struct sc_apparmor apparmor;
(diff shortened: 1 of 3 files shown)


cmd/snap-confine/snap-confine.c
```diff
@@ -358,9 +358,7 @@ int main(int argc, char **argv)
 		sc_error *err SC_CLEANUP(sc_cleanup_error) = NULL;
 		snap_context =
 		    sc_cookie_get_from_snapd(invocation.snap_instance, &err);
-		if (err != NULL) {
-			error("%s\n", sc_error_msg(err));
-		}
+		/* The error is explicitly ignored because the cookies are optional. */
```
stolowski: I think the change is fine, what that means in practice is we will only see an error if the snap actually attempts to run snapctl from its app (and the cookie is missing). In such case snapctl should report "invalid snap cookie requested" if I read the code correctly (the critical bit is in `func (m *HookManager) Context(cookieID string) (*Context, error) {...` and `func (m *HookManager) ephemeralContext(cookieID string) (context *Context, err error) {....` in hookmgr.go, called from `runSnapctl` in api.go. But this is theoretical, because `SyncCookies()` run when overlord starts makes sure all cookies are created if missing after an upgrade from old snapd.

The only remark I've is the comment here, I think it's slightly misleading. Could you please expand it a little bit by mentioning SyncCookies, snapctl error and that cookies are only needed if snapctl is run by an app?

cmd/snap-confine/snap-confine.c
```diff
@@ -358,9 +358,13 @@ int main(int argc, char **argv)
...
-		if (err != NULL) {
-			error("%s\n", sc_error_msg(err));
-		}
+		/* While the cookie is normally present due to various protection
+		 * mechanisms ensuring its creation from snapd, we are not considering
+		 * it a critical error for snap-confine in the case it is absent. When
+		 * absent snaps attempting to utilize snapctl to interact with snapd
+		 * will fail but it is more important to run a little than break
+		 * entirely in case snapd-side code is incorrect. Therefore error
+		 * information is collected but discarded. */
```
stolowski: Perfect, thank you!

PR #6753, opened 2019-04-18
tests: fix spaces issue in the base snaps names to remove during reset phase

sergiocazzolato:
This will prevent this error in reset.sh like this one:

```
+ snap remove ' core16'
snap " core16" is not installed
```

To reproduce the error execute:
`spread -debug -repeat 5
google:ubuntu-core-16-64:tests/main/core16-provided-by-core`


sergiocazzolato:
@zyga @mvo5 The reset.sh was not failing because of this:

sergio@cachiomachine:~/workspace/snapcore/snapd$ sudo snap remove eiwtepw && echo No failed
snap "eiwtepw" is not installed
No failed

Is it working as designed or it is a bug?

PR #6756, opened 2019-04-19
snap: filter-out nil seed snaps

zyga:
We're seeing errors from fresh Ubuntu 19.04 installs where
snapd panics on a nil pointer:

    panic: runtime error: invalid memory address or nil pointer dereference
    [signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0x55675388c716]
    goroutine 38 [running]:
    github.com/snapcore/snapd/snap.ReadSeedYaml(0xc4207640c0, 0x1d, 0x1, 0xc4207640c0, 0x1d)
            github.com/snapcore/snapd/snap/seed_yaml.go:73 +0x236
    github.com/snapcore/snapd/overlord/devicestate.populateStateFromSeedImpl(0xc4200f3f80, 0x556753afe9ee, 0x4, 0x556753d5b700, 0xc4203bbb7f, 0x556753e9f3c0)
            github.com/snapcore/snapd/overlord/devicestate/firstboot.go:110 +0x24f
    github.com/snapcore/snapd/overlord/devicestate.(*DeviceManager).ensureSeedYaml(0xc4203685f0, 0x0, 0x0)
            github.com/snapcore/snapd/overlord/devicestate/devicemgr.go:354 +0x190
    github.com/snapcore/snapd/overlord/devicestate.(*DeviceManager).Ensure(0xc4203685f0, 0x0, 0x0)
            github.com/snapcore/snapd/overlord/devicestate/devicemgr.go:465 +0x42
    github.com/snapcore/snapd/overlord.(*StateEngine).Ensure(0xc4202fee10, 0x0, 0x0)
            github.com/snapcore/snapd/overlord/stateengine.go:100 +0x10b
    github.com/snapcore/snapd/overlord.(*Overlord).Loop.func1(0x0, 0x0)
            github.com/snapcore/snapd/overlord/overlord.go:290 +0x5e
    github.com/snapcore/snapd/vendor/gopkg.in/tomb%2ev2.(*Tomb).run(0xc4202dc320, 0xc4203d40f0)
            github.com/snapcore/snapd/vendor/gopkg.in/tomb.v2/tomb.go:163 +0x2d
    created by github.com/snapcore/snapd/vendor/gopkg.in/tomb%2ev2.(*Tomb).Go
            github.com/snapcore/snapd/vendor/gopkg.in/tomb.v2/tomb.go:159 +0xbb

The only possible way to panic there if the SeedSnap pointer is nil,
that is, if reading /var/lib/snapd/seed/seed.yaml gives us an element
that is nil (a list containing a nil pointer).

This patch filters out such entries until we know more about the
problem.

Fixes: https://bugs.launchpad.net/snapd/+bug/1825437

### DIFF
--- a/snap/seed_yaml.go
+++ b/snap/seed_yaml.go
@@ -68,6 +68,18 @@ func ReadSeedYaml(fn string) (*Seed, error) {
 		return nil, fmt.Errorf("cannot unmarshal %q: %s", yamlData, err)
 	}
 
+	// XXX: filter-out nil snaps
+	// For context see: https://bugs.launchpad.net/snapd/+bug/1825437
+	filtered := make([]*SeedSnap, 0, len(seed.Snaps))
+	for _, sn := range seed.Snaps {
+		if sn != nil {
+			filtered = append(filtered, sn)
+		}
+	}
+	if len(filtered) != len(seed.Snaps) {
+		seed.Snaps = filtered
+	}
+
 	// validate
 	for _, sn := range seed.Snaps {
 		if strings.Contains(sn.File, "/") {
--- a/snap/seed_yaml.go
+++ b/snap/seed_yaml.go
@@ -68,8 +68,8 @@ func ReadSeedYaml(fn string) (*Seed, error) {
 		return nil, fmt.Errorf("cannot unmarshal %q: %s", yamlData, err)
 	}
 
-	// XXX: filter-out nil snaps
-	// For context see: https://bugs.launchpad.net/snapd/+bug/1825437
+	// Filter out nil entries. The Ubuntu 19.04 installer can sometimes create them
+	// See: https://bugs.launchpad.net/snapd/+bug/1825437
 	filtered := make([]*SeedSnap, 0, len(seed.Snaps))
 	for _, sn := range seed.Snaps {
 		if sn != nil {
--- a/snap/seed_yaml_test.go
+++ b/snap/seed_yaml_test.go
@@ -83,7 +83,10 @@ func (s *seedYamlTestSuite) TestNoPathAllowed(c *C) {
 }
 
 var nilSeedItemYaml = []byte(`
-snaps: [null]
+snaps:
+ - name: foo
+ -
+ - name: bar
 `)
 
 func (s *seedYamlTestSuite) TestNilItems(c *C) {
@@ -93,5 +96,5 @@ func (s *seedYamlTestSuite) TestNilItems(c *C) {
 
 	seed, err := snap.ReadSeedYaml(fn)
 	c.Assert(err, IsNil)
-	c.Assert(seed.Snaps, HasLen, 0)
+	c.Assert(seed.Snaps, HasLen, 2)
 }
--- a/snap/seed_yaml_test.go
+++ b/snap/seed_yaml_test.go
@@ -81,3 +81,17 @@ func (s *seedYamlTestSuite) TestNoPathAllowed(c *C) {
 	_, err = snap.ReadSeedYaml(fn)
 	c.Assert(err, ErrorMatches, `"foo/bar.snap" must be a filename, not a path`)
 }
+
+var nilSeedItemYaml = []byte(`
+snaps: [null]
+`)
+
+func (s *seedYamlTestSuite) TestNilItems(c *C) {
+	fn := filepath.Join(c.MkDir(), "seed.yaml")
+	err := ioutil.WriteFile(fn, nilSeedItemYaml, 0644)
+	c.Assert(err, IsNil)
+
+	seed, err := snap.ReadSeedYaml(fn)
+	c.Assert(err, IsNil)
+	c.Assert(seed.Snaps, HasLen, 0)
+}


pedronis:
We should probably error, not panic on this, I wonder what happened to get seeds to contain that, and also why testing didn't catch it ??

pedronis:
Is this from a static seed.yaml or a seed yaml created by the installer?

zyga:
@pedronis it is a seed created by the installer but we have not seen the actual contents yet. I've asked the reporter to attach this to the launchd bug report.

pedronis:
Still a bit unclear, anyway I also commented a bit in the bug, if it it's really dynamically created by the installer there's a bug there too. I have vague memories that now it can auto-update so it is worth fixing it there too.  Do talk with foundations.

zyga:
I have a hunch that some of the issues that people report with regards to Ubuntu 19.04 and app startup failure is related to broken seeds on some systems. Perhaps there is a race somewhere in the installer or something else gets broken and the seed is messed up.

I've asked the OP on https://forum.snapcraft.io/t/snap-apps-not-working/11000/4 to provide their seed file, perhaps this will shed some light.

As for panicking, I will rework this to return an error if we encounter `nil` seed snaps.

zyga:
Heh, a moment after posting I noticed that the reported or the original bug provided the seed. It is indeed broken and if this issue is widespread it will cause massive failures for people using Ubuntu 19.04

```
snaps:
  -
    name: core18
    channel: stable/ubuntu-19.04
    file: core18_941.snap
  -
  -
    name: gnome-3-28-1804
    channel: stable/ubuntu-19.04
    file: gnome-3-28-1804_23.snap
  -
    name: gnome-calculator
    channel: stable/ubuntu-19.04
    file: gnome-calculator_406.snap
  -
    name: gnome-characters
    channel: stable/ubuntu-19.04
    file: gnome-characters_254.snap
  -
    name: gnome-logs
    channel: stable/ubuntu-19.04
    file: gnome-logs_61.snap
  -
    name: gnome-system-monitor
    channel: stable/ubuntu-19.04
    file: gnome-system-monitor_77.snap
  -
    name: core
    channel: stable
    file: core_6673.snap
  -
    name: gtk-common-themes
    channel: stable/ubuntu-19.04
    file: gtk-common-themes_1198.snap
```

pedronis:
Whatever is writing this seed with empty entries is broken and needs
fixing, we need to talk with Foundations/Desktop and find out what it
is/what can be done

zyga:
I asked for help https://twitter.com/zygoon/status/1119567700832018434

zyga:
Paging @kenvandine - ^ can you look at the circumstances in which the installer would create a seed like pasted in the comment above? I'd like to understand if this is a widespread issue.

kenvandine:
We need to talk to foundations, I don't know much about how that file is written.

zyga:
@kenvandine who would be the best person to talk to?

snap/seed_yaml.go
```diff
@@ -68,6 +68,18 @@ func ReadSeedYaml(fn string) (*Seed, error) {
...
+	filtered := make([]*SeedSnap, 0, len(seed.Snaps))
+	for _, sn := range seed.Snaps {
+		if sn != nil {
+			filtered = append(filtered, sn)
+		}
+	}
+	if len(filtered) != len(seed.Snaps) {
+		seed.Snaps = filtered
+	}
+
```
chipaca: I think this is the wrong approach. Yes we should not assume it's going to be valid, but if it's not valid (and an entry being nil is just one way for it not to be valid), we should ignore the whole thing (and log the fact, loudly).

zyga: @chipaca I agree in principle but not in this case. There is no loud logging that users will see. Users cannot fix that. There is only the yet-undiscovered issue that is affecting some ISOs that causes an otherwise valid YAML to have a single NULL entry.

The real bummer is that this should never have happened. The cleanup now should be focused on ensuring that broken installations can recover but IMO the only way to do that is with a snapd.deb update.

snap/seed_yaml_test.go
```diff
@@ -81,3 +81,17 @@ func (s *seedYamlTestSuite) TestNoPathAllowed(c *C) {
 	_, err = snap.ReadSeedYaml(fn)
 	c.Assert(err, ErrorMatches, `"foo/bar.snap" must be a filename, not a path`)
 }
+
+var nilSeedItemYaml = []byte(`
```
bboozzoo: Can you add a test case for the following seed.yaml:
```
snaps:
 -
   name: foo
   file: bar.snap

 -
```
This is close to the format in the ISO and the lone `-` at the end is sufficient to trigger the panic.

snap/seed_yaml_test.go
```diff
@@ -81,3 +81,20 @@ func (s *seedYamlTestSuite) TestNoPathAllowed(c *C) {
...
+snaps:
+ - name: foo
```
bboozzoo: I was actually thinking about replicating the odd format of seed.yaml in the images, `-` followed by a newline and actual properties in the subsequent lines.

mvo5:
This was attacked as part of https://github.com/snapcore/snapd/pull/6771 - it now errors for empty elements.

PR #6757, opened 2019-04-19
cmd/snap-update-ns: move Assumption to {System,User}ProfileUpdate

zyga:
The common implementation has been removed, replacing it with the
two existing implementations, now graduated to proper methods, with
tests and basic documentation.

The new logic is immediately used.


zyga:
Force pushed to fix commit typo 💃

PR #6758, opened 2019-04-19
cmd/snap-update-ns: move locking to the common layer

zyga:
The locking was, so far, only performed for system-wide updates.
This property is retained since this is simply a refactoring at this
stage. Locking is now fully tested and documented.

The common layer is now aware if snap-update-ns is invoked from
snap-confine or from snapd. When invoked from snapd locking is handled
entirely by snap-update-ns. When invoked from snap-confine (during
the namespace bootstrap process) locking is verified. In both cases
the snap processes are frozen.

### DIFF
--- a/cmd/snap-update-ns/user.go
+++ b/cmd/snap-update-ns/user.go
@@ -37,14 +37,22 @@ type UserProfileUpdateContext struct {
 }
 
 // NewUserProfileUpdateContext returns encapsulated information for performing a per-user mount namespace update.
-func NewUserProfileUpdateContext(instanceName string, uid int) *UserProfileUpdateContext {
+func NewUserProfileUpdateContext(instanceName string, fromSnapConfine bool, uid int) *UserProfileUpdateContext {
 	return &UserProfileUpdateContext{
 		CommonProfileUpdateContext: CommonProfileUpdateContext{
 			instanceName:       instanceName,
+			fromSnapConfine:    fromSnapConfine,
 			desiredProfilePath: desiredUserProfilePath(instanceName),
 		},
 		uid: uid,
 	}
+}
+
+// Lock acquires locks / freezes needed to synchronize mount namespace changes.
+func (ctx *UserProfileUpdateContext) Lock() (unlock func(), err error) {
+	// TODO: when persistent user mount namespaces are enabled, grab a lock
+	// protecting the snap and freeze snap processes here.
+	return func() {}, nil
 }
 
 // Assumptions returns information about file system mutability rules.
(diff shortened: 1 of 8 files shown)


cmd/snap-update-ns/user.go
```diff
@@ -72,7 +80,8 @@ func (ctx *UserProfileUpdateContext) LoadDesiredProfile() (*osutil.MountProfile,
 }
 
 func applyUserFstab(snapName string) error {
-	ctx := NewUserProfileUpdateContext(snapName, os.Getuid())
+	fromSnapConfine := true // currently always true, no persistence yet.
+	ctx := NewUserProfileUpdateContext(snapName, fromSnapConfine, os.Getuid())
```
stolowski: I think we should do `Lock()` and `defer unlock()` here even if they are stubs/no-ops to, so that high-level semantics are visible already.

zyga: Don't worry, we will, just not yet in this patch. This is coming up in https://github.com/zyga/snapd/commit/b3e0944ffb3d2d90aeeadca39c3941f90ff305ca

cmd/snap-update-ns/user.go
```diff
@@ -37,16 +37,24 @@ type UserProfileUpdateContext struct {
 }
 
 // NewUserProfileUpdateContext returns encapsulated information for performing a per-user mount namespace update.
-func NewUserProfileUpdateContext(instanceName string, uid int) *UserProfileUpdateContext {
+func NewUserProfileUpdateContext(instanceName string, fromSnapConfine bool, uid int) *UserProfileUpdateContext {
```
chipaca: I'd rather we had two initialisers, than one with a boolean flag. OTOH if there are going to be more ways, an options struct maybe?

cmd/snap-update-ns/user.go
```diff
@@ -37,16 +37,24 @@ type UserProfileUpdateContext struct {
...
 		},
 		uid: uid,
 	}
 }
 
+// Lock acquires locks / freezes needed to synchronize mount namespace changes.
+func (ctx *UserProfileUpdateContext) Lock() (unlock func(), err error) {
+	// TODO: when persistent user mount namespaces are enabled, grab a lock
+	// protecting the snap and freeze snap processes here.
+	return func() {}, nil
```
chipaca: why like this instead of the usual Lock/Unlock pair? this seems weird to me

zyga: Because this way the unlock function holds the reference to a flock-based lock. Otherwise the update context would have to hold more state. I don't really mind either way.

cmd/snap-update-ns/user.go
zyga: This is all that we have and we only invoke it in one place (never with a literal boolean) outside of test code. Once snap-boostrap-ns comes around this will be easier as the option will go away entirely. In that world snap-update-ns is always invoked by snapd and snap-boostrap-ns is always invoked by snap-confine. Then the options are gone :)

PR #6761, opened 2019-04-22
spread.yaml: add qemu:centos-7-64

jdstrand:
By going to https://cloud.centos.org/centos/7/images/ and using
https://blog.strandboge.com/2019/04/16/cloud-images-qemu-cloud-init-and-snapd-spread-tests/
it is possible to create a qemu:centos-7-64 VM for local testing (note,
the VM needs additional configuration, but this is all that is needed
for spread to connect, send the project data and start preparing).

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -201,6 +201,9 @@ backends:
             - fedora-28-64:
                 username: fedora
                 password: fedora
+            - centos-7-64:
+                username: centos
+                password: centos
     autopkgtest:
         type: adhoc
         allocate: |


jdstrand:
FYI, I updated my blog with the following:
"UPDATE 2019-04-22: the above is confirmed to work with CentOS 7 (though, if using the resulting image to test snapd, then be sure to 'yum update ; yum install epel-release ; yum install golang nc strace' in the image)."

With that I was able to run a single spread test. More may need to be added for all spread tests, but at least it is useful in some capacity. :)

jdstrand:
Thanks!

PR #6769, opened 2019-04-24
gadget: layout, smaller fixes

bboozzoo:
The PR introduces support for laying out volumes. This is needed both for updates and generating images.

In case of updates we need to know exactly if and how the volume has changed and can error out accordingly or proceed.

@cmatsuoka can you take a look as well?

### DIFF
--- a/gadget/gadget.go
+++ b/gadget/gadget.go
@@ -42,7 +42,7 @@ const validSnapIDLength = 32
 var (
 	validVolumeName = regexp.MustCompile("^[a-zA-Z0-9][a-zA-Z0-9-]+$")
 	validTypeID     = regexp.MustCompile("^[0-9A-F]{2}$")
-	validGUUID      = regexp.MustCompile("^([0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}|[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})$")
+	validGUUID      = regexp.MustCompile("^(?i)[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}$")
 )
 
 type Info struct {
--- a/gadget/gadget.go
+++ b/gadget/gadget.go
@@ -39,6 +39,13 @@ import (
 // The fixed length of valid snap IDs.
 const validSnapIDLength = 32
 
+const (
+	// MBR identifies a Master Boot Record partitioning schema, or an MBR like role
+	MBR = "mbr"
+	// GPT identifies a GUID Partition Table partitioning schema
+	GPT = "gpt"
+)
+
 var (
 	validVolumeName = regexp.MustCompile("^[a-zA-Z0-9][a-zA-Z0-9-]+$")
 	validTypeID     = regexp.MustCompile("^[0-9A-F]{2}$")
@@ -115,8 +122,8 @@ func (vs *VolumeStructure) EffectiveRole() string {
 	if vs.Role != "" {
 		return vs.Role
 	}
-	if vs.Role == "" && vs.Type == "mbr" {
-		return "mbr"
+	if vs.Role == "" && vs.Type == MBR {
+		return MBR
 	}
 	return ""
 }
@@ -315,7 +322,7 @@ func validateVolume(name string, vol *Volume) error {
 	if !validVolumeName.MatchString(name) {
 		return errors.New("invalid name")
 	}
-	if vol.Schema != "" && vol.Schema != "gpt" && vol.Schema != "mbr" {
+	if vol.Schema != "" && vol.Schema != GPT && vol.Schema != MBR {
 		return fmt.Errorf("invalid schema %q", vol.Schema)
 	}
 
@@ -366,7 +373,7 @@ func validateCrossVolumeStructure(structures []PositionedStructure, knownStructu
 	// - positioned structure overlap
 	// use structures positioned within the volume
 	for pidx, ps := range structures {
-		if ps.Role == "mbr" || ps.Type == "mbr" {
+		if ps.Role == MBR || ps.Type == MBR {
 			if ps.StartOffset != 0 {
 				return fmt.Errorf(`structure %v has "mbr" role and must start at offset 0`, ps)
 			}
@@ -471,7 +478,7 @@ func validateStructureType(s string, vol *Volume) error {
 
 	schema := vol.Schema
 	if schema == "" {
-		schema = "gpt"
+		schema = GPT
 	}
 
 	if s == "" {
@@ -483,7 +490,7 @@ func validateStructureType(s string, vol *Volume) error {
 		return nil
 	}
 
-	if s == "mbr" {
+	if s == MBR {
 		// backward compatibility for type: mbr
 		return nil
 	}
@@ -510,11 +517,11 @@ func validateStructureType(s string, vol *Volume) error {
 		}
 	}
 
-	if schema != "gpt" && isGPT {
+	if schema != GPT && isGPT {
 		// type: <uuid> is only valid for GPT volumes
 		return fmt.Errorf("GUID structure type with non-GPT schema %q", vol.Schema)
 	}
-	if schema != "mbr" && isMBR {
+	if schema != MBR && isMBR {
 		return fmt.Errorf("MBR structure type with non-MBR schema %q", vol.Schema)
 	}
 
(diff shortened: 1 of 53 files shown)


gadget/gadget.go
```diff
@@ -42,7 +42,7 @@ const validSnapIDLength = 32
 var (
 	validVolumeName = regexp.MustCompile("^[a-zA-Z0-9][a-zA-Z0-9-]+$")
 	validTypeID     = regexp.MustCompile("^[0-9A-F]{2}$")
-	validGUUID      = regexp.MustCompile("^[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}$")
+	validGUUID      = regexp.MustCompile("^([0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}|[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})$")
```
chipaca: silly question: why not simply `(?i)` it? Do we really care that it be all lowercase or all uppercase?

bboozzoo: We could, `ubuntu-image` kind of does that, but only because Python's UUID library does not care and converts the thing to a binary representation anyway. On the other hand, `00000000-0000-0000-abcd-dd00DeaDBeeF` looks a bit awkward.

gadget/gadget.go
```diff
@@ -625,8 +615,9 @@ func (e *editionNumber) UnmarshalYAML(unmarshal func(interface{}) error) error {
 type Size uint64
 
 const (
-	SizeMiB = Size(2 << 20)
-	SizeGiB = Size(2 << 30)
+	SizeKiB = Size(1 << 10)
```
stolowski: Ouch... ❤️

gadget/gadget.go
```diff
@@ -110,6 +110,17 @@ func (vs *VolumeStructure) IsBare() bool {
 	return vs.Filesystem == "none" || vs.Filesystem == ""
 }
 
+// EffectiveRole returns the role of given structure
+func (vs *VolumeStructure) EffectiveRole() string {
+	if vs.Role != "" {
+		return vs.Role
+	}
+	if vs.Role == "" && vs.Type == "mbr" {
+		return "mbr"
```
stolowski: A high-level wondering: would it make sense to have roles defined via consts in a central place, or it is not worth it? Is it just "gpt" and "mbr"?

gadget/layout.go
```diff
@@ -0,0 +1,214 @@
...
+
+	// StartOffset defines the start offset of this content image
+	StartOffset Size
+	// Size is the maximum size occupied by this image
+	Size Size
+}
+
+// LayOutVolume attempts to lay out the volume using constraints and returns a
+// fully positioned description of the resulting volume
+func LayOutVolume(gadgetRootDir string, volume *Volume, constraints LayoutConstraints) (*PositionedVolume, error) {
```
stolowski: Does it have to be exported?

gadget/layout.go
```diff
@@ -0,0 +1,214 @@
...
+		}
+
+		end := start + s.Size
+		ps := PositionedStructure{
+			VolumeStructure: &volume.Structure[idx],
+			StartOffset:     start,
+			Index:           idx,
+		}
+
+		if ps.EffectiveRole() != "mbr" {
```
stolowski: Back to my earlier comment, perhaps mbr (gpt..) could indeed be a constant instead?

gadget/layout.go
```diff
@@ -0,0 +1,214 @@
...
+package gadget
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+)
+
+type LayoutConstraints struct {
```
zyga: Naming nitpick: to avoid overloading the term "layout" could we instead use the word "positioning" here? PositioningConstraints, PositionVolume, etc.

stolowski: Oh yes, that's a good point, +1 for a different name

gadget/gadget.go
cmatsuoka: Mixed case hex could be unusual/ugly/awkward, but would it be **wrong**?

chipaca: https://youtu.be/k6SSPUcCln8?t=15

bboozzoo: Haha :smile: I'm ok with letting mixed casing, we can pacify it it elsewhere.

gadget/layout.go
cmatsuoka: +1 for that, it's always good to avoid ambiguities in terminology.

gadget/gadget.go
```diff
@@ -625,6 +615,7 @@ func (e *editionNumber) UnmarshalYAML(unmarshal func(interface{}) error) error {
 type Size uint64
 
 const (
+	SizeKiB = Size(1 << 10)
 	SizeMiB = Size(1 << 20)
```
cmatsuoka: I wonder if these constants are not interesting enough to be declared project-wide. There are other size-related variables in the codebase multiplying values by 1024 or 1024 * 1024. (And in this case maybe plain MiB, GiB, etc are more idiomatic in Go than SizeMiB and SizeGiB for project-wide units.)

bboozzoo: It'll probably need some consideration and a follow up though. I briefly looked at the places where 1024 is used in multiplication, it's not necessarily obvious there would be benefit from using common values.

gadget/layout.go
bboozzoo: Updated type/func names and some error messages.

gadget/gadget.go
bboozzoo: Consts added

gadget/layout.go
bboozzoo: Yes, ideally it will be used in a tool to generate images.

bboozzoo:
Pushed a bunch of updates. Please take another look.

gadget/gadget.go
```diff
@@ -39,10 +39,17 @@ import (
 // The fixed length of valid snap IDs.
 const validSnapIDLength = 32
 
+const (
+	// MBR identifies a Master Boot Record partitioning schema, or an MBR like role
+	MBR = "mbr"
```
stolowski: 👍

gadget/layout.go
stolowski: There is still a lot of "layout" wording (layout.go, layoutTestSuite, tests).

bboozzoo: Heh, I can do that too. Will break a bit my work-in-progress branch, but I can work around that probably.

gadget/gadget.go
```diff
@@ -39,10 +39,17 @@ import (
...
+	// GPT identifies a GUID Partition Table partitioning schema
+	GPT = "gpt"
+)
+
 var (
 	validVolumeName = regexp.MustCompile("^[a-zA-Z0-9][a-zA-Z0-9-]+$")
 	validTypeID     = regexp.MustCompile("^[0-9A-F]{2}$")
-	validGUUID      = regexp.MustCompile("^[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}$")
+	validGUUID      = regexp.MustCompile("^(?i)[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}$")
```
zyga: What does `(?i)`do?

gadget/gadget.go
```diff
@@ -110,6 +117,17 @@ func (vs *VolumeStructure) IsBare() bool {
 	return vs.Filesystem == "none" || vs.Filesystem == ""
 }
 
+// EffectiveRole returns the role of given structure
+func (vs *VolumeStructure) EffectiveRole() string {
```
zyga: Oh, I love this. Thank you!

gadget/gadget.go
```diff
@@ -376,7 +373,7 @@ func validateCrossVolumeStructure(structures []PositionedStructure, knownStructu
 	// - positioned structure overlap
 	// use structures positioned within the volume
 	for pidx, ps := range structures {
-		if ps.Role == "mbr" || ps.Type == "mbr" {
+		if ps.Role == MBR || ps.Type == MBR {
```
zyga: Should this be the effective role function call?

gadget/layout.go
```diff
@@ -0,0 +1,219 @@
...
+// within a volume
+type PositioningConstraints struct {
+	// NonMBRStartOffset is the default start offset of non-MBR structure in
+	// the volume.
+	NonMBRStartOffset Size
+	// SectorSize is the size of the sector to be used for calculations
+	SectorSize Size
+}
+
+type PositionedVolume struct {
```
zyga: Perhaps add something like this:
```
PositionedVolume defines the size of a volume and all the structures within it.
```

gadget/gadget.go
bboozzoo: ignore case

gadget/layout.go
bboozzoo: Applied thanks

bboozzoo:
@cmatsuoka somehow I cannot re-request a review from you using github's UI. Please take another look though.

PR #6770, opened 2019-04-24
overlord,tests: update static attrs during profile setup

zyga:
The persistent state of a snapd connection contains a copy of the static
interface attributes that were collected from plug and slot
participating in the connection. We were notified that snapd copies
those attributes only during the initial connection. As the snaps are
refreshed the attributes can change according to the change of snap.yaml
data in both snaps.

In the case of the content interface the attributes are essential for
correct operation of the system. The stale copy is part of a larger
problem that is described in detail in the bug referenced below. The
larger problem involves dynamic attributes that are equally stored in
the connect. A full solution would have to involve running interface
hooks to the extent necessary to inform both participating snaps about
the new peer and, if necessary, re-negotiate connection attributes. This
will likely take a longer moment to prepare.

Meanwhile, while interface hooks are not used as broadly as the content
interface is. In an effort to resolve the static part of the problem
snapd will now refresh the cached copy of the static connection
attributes whenever a snap has its security setup performed.

A regression test is attached as well. During the development of the
regression test I encountered another problem where the mount system
misbehaves when a snap is refreshed. This is similar to reports that
were earlier mentioned on the forum and in the launchpad bug tracker.
To avoid the bug, the content interface is mounted in a sub-directory of
the main snap, in a place that is otherwise irrelevant to the operation
of snapd.

Fixes: https://bugs.launchpad.net/snapd/+bug/1825883


overlord/ifacestate/helpers.go
```diff
@@ -331,6 +331,37 @@ func (m *InterfaceManager) reloadConnections(snapName string) ([]string, error)
 	return result, nil
 }
 
+func (m *InterfaceManager) refreshStaticConnectionAttrs(snapInfo *snap.Info) error {
```
mvo5: Can we unit test this somehow? Maybe something basic/trivial?

zyga: Yes, Pawel wrote unit tests yesterday. I will incorporate them into this PR now.

zyga:
This is insufficient. I am reworking this now.

zyga:
I will close this and reopen soon.

PR #6771, opened 2019-04-24
cmd: add `snap debug validate-seed <path>` cmd

mvo5:
We had a bug where one of the recent livecd-rootfs build images
had an incorrect seed.yaml. This caused a panic of snapd on when
seeding. To prevent this here is a small helper that we can run
inside livecd-rootfs to ensure seed.yaml is good enough to not
crash snapd.

This is for https://bugs.launchpad.net/ubuntu/+source/livecd-rootfs/+bug/1825437

### DIFF
--- a/cmd/snap/cmd_debug_validate_seed.go
+++ b/cmd/snap/cmd_debug_validate_seed.go
@@ -1,0 +1,55 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package main
+
+import (
+	"fmt"
+
+	"github.com/jessevdk/go-flags"
+
+	"github.com/snapcore/snapd/snap"
+)
+
+type cmdValidateSeed struct {
+	Positionals struct {
+		SeedYamlPath string `positional-arg-name:"<seed-yaml-path>"`
+	} `positional-args:"true"`
+}
+
+func init() {
+	cmd := addDebugCommand("validate-seed",
+		"(internal) validate seed.yaml",
+		"(internal) validate seed.yaml",
+		func() flags.Commander {
+			return &cmdValidateSeed{}
+		}, nil, nil)
+	cmd.hidden = true
+}
+
+func (x *cmdValidateSeed) Execute(args []string) error {
+	if len(args) > 0 {
+		return ErrExtraArgs
+	}
+
+	if _, err := snap.ReadSeedYaml(x.Positionals.SeedYamlPath); err != nil {
+		return fmt.Errorf("cannot read %q: %v", x.Positionals.SeedYamlPath, err)
+	}
+	return nil
+}
--- a/cmd/snap/cmd_debug_validate_seed.go
+++ b/cmd/snap/cmd_debug_validate_seed.go
@@ -20,8 +20,6 @@
 package main
 
 import (
-	"fmt"
-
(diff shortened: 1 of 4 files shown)


cmd/snap/cmd_debug_validate_seed.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+type cmdValidateSeed struct {
+	Positionals struct {
+		SeedYamlPath string `positional-arg-name:"<seed-yaml-path>"`
+	} `positional-args:"true"`
+}
+
+func init() {
+	cmd := addDebugCommand("validate-seed",
```
bboozzoo: or `validate-seed-file` ?

cmd/snap/cmd_debug_validate_seed.go
```diff
@@ -0,0 +1,55 @@
...
+		}, nil, nil)
+	cmd.hidden = true
+}
+
+func (x *cmdValidateSeed) Execute(args []string) error {
+	if len(args) > 0 {
+		return ErrExtraArgs
+	}
+
+	if _, err := snap.ReadSeedYaml(x.Positionals.SeedYamlPath); err != nil {
```
bboozzoo: When seeing, a bunch of checks is done after we load the seed.yaml file. Should we validate at least the name and channel here as well?

cmd/snap/cmd_debug_validate_seed_test.go
```diff
@@ -0,0 +1,68 @@
...
+	err := ioutil.WriteFile(tmpf, []byte(`
+snaps:
+ -
+   name: core
+   channel: foo
+   file: bar
+ -
+   name: gnome-foo
+   channel: foo
+   name: xxx
```
bboozzoo: duplicate `name`

cmd/snap/cmd_debug_validate_seed_test.go
```diff
@@ -0,0 +1,68 @@
...
+snaps:
+ -
+   name: core
+   channel: foo
+   file: bar
+ -
+ -
+   name: gnome-foo
+   channel: foo
+   name: xxx
```
bboozzoo: same here

cmd/snap/cmd_debug_validate_seed.go
stolowski: Not super important, but would it make sense to make seed yaml path argument optional, and use the default /var/lib/snapd/.../seed.yaml path if not provided, for convienience?

mvo5: Good point, thanks! I updated to the PR add basic checks.

cmd/snap/cmd_debug_validate_seed_test.go
mvo5: Thanks, fix this (and the other one) with more realistic inputs.

cmd/snap/cmd_debug_validate_seed.go
mvo5: Not sure, feel a bit long but I don't really mind strongly either way. Its a debug command after all :)

cmd/snap/cmd_debug_validate_seed.go
mvo5: I like the idea, but I'm also a bit lazy :) We could do it in a followup.

stolowski: Sure, no problem.

snap/seed_yaml.go
```diff
@@ -58,20 +58,36 @@ type Seed struct {
 }
 
 func ReadSeedYaml(fn string) (*Seed, error) {
+	errPrefix := "cannot read seed yaml: "
+
 	yamlData, err := ioutil.ReadFile(fn)
 	if err != nil {
-		return nil, fmt.Errorf("cannot read seed yaml: %s", fn)
+		return nil, fmt.Errorf(errPrefix+"%v", err)
```
stolowski: Nitpick: looks a bit weird to concatenate errorPrefix that way when using fmt formatting anyway, why not simply `fmt.Errorf("%s%v", errorPrefix, err)` (also below)?

snap/seed_yaml.go
```diff
@@ -58,20 +58,36 @@ type Seed struct {
...
 
 	// validate
 	for _, sn := range seed.Snaps {
+		if sn == nil {
+			return nil, fmt.Errorf(errPrefix + "empty element in seed")
+		}
+		if err := ValidateInstanceName(sn.Name); err != nil {
+			return nil, fmt.Errorf(errPrefix+"%v", err)
+		}
+		if sn.Channel != "" {
```
stolowski: Nice

bboozzoo:
Shall we land it?

PR #6773, opened 2019-04-24
osutil: make IsValidUsername public and fix regex

jdstrand:
The old regex, `^[a-z0-9][-a-z0-9+.-_]*$`, had a bug in it where it
would accept any character within the range of '.-_' where 'a:b'
would match since `.` &lt; `:` &lt; `_` in the ASCII table.

Make IsValidUsername public since it makes adding tests easier and
subsequent commits will use this.


jdstrand:
@mvo5 - I asked you to be a reviewer because I'm not sure if you want to consider this for 2.38 (.-_ is a big range in the ascii table: http://www.asciitable.com/)

chipaca:
I wrote that bad regexp 😧😧

PR #6780, opened 2019-04-25
seccomp: refactor ActLog check and expose via sandbox-features

jdstrand:
seccomp: add 'actlog-supported' command and expose via sandbox-features
    
snap-seccomp currently queries for ActLog support to decide if it should
be enabled or not. We want to expose this in sandbox-features and also
to other parts of snapd. Because snapd is intended to be compilable with
CGO_ENABLED=0, the simplest solution of refactoring the support check
does not work because it pulls in libseccomp. Therefore, add a new
snap-seccomp command, 'actlog-supported' and have the seccomp backend
call it on initialization so we only call this once on startup instead
of every time the query is performed.

### DIFF
--- a/cmd/snap-seccomp/main.go
+++ b/cmd/snap-seccomp/main.go
@@ -634,12 +634,11 @@ func preprocess(content []byte) (unrestricted, complain bool) {
 	return unrestricted, complain
 }
 
-func complainAction() seccomp.ScmpAction {
-	// XXX: Work around some distributions not having a new enough
-	// libseccomp-golang that declares ActLog. Instead, we'll guess at its
-	// value by adding one to ActAllow and then verify that the string
-	// representation is what we expect for ActLog. The value and string is
-	// defined in https://github.com/seccomp/libseccomp-golang/pull/29.
+func actLogSupported() bool {
+	// Guess at the ActLog value by adding one to ActAllow and then verify
+	// that the string representation is what we expect for ActLog. The
+	// value and string is defined in
+	// https://github.com/seccomp/libseccomp-golang/pull/29.
 	//
 	// Ultimately, the fix for this workaround is to be able to use the
 	// GetApi() function created in the PR above. It'll tell us if the
@@ -647,7 +646,16 @@ func complainAction() seccomp.ScmpAction {
 	var actLog seccomp.ScmpAction = seccomp.ActAllow + 1
 
 	if actLog.String() == "Action: Log system call" {
-		return actLog
+		return true
+	}
+	return false
+}
+
+func complainAction() seccomp.ScmpAction {
+	// XXX: Work around some distributions not having a new enough
+	// libseccomp-golang that declares ActLog.
+	if actLogSupported() {
+		return seccomp.ActLog
 	}
 
 	// Because ActLog is functionally ActAllow with logging, if we don't
@@ -771,6 +779,13 @@ func main() {
 		err = showSeccompLibraryVersion()
 	case "version-info":
 		err = showVersionInfo()
+	case "actlog-supported":
+		if actLogSupported() {
+			fmt.Printf("SCMP_ACT_LOG supported\n")
+			os.Exit(0)
+		}
+		fmt.Printf("SCMP_ACT_LOG not supported\n")
+		os.Exit(1)
 	default:
 		err = fmt.Errorf("unsupported argument %q", cmd)
 	}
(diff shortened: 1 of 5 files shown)


jdstrand:
Unfortunately, I hit the CGO_ENABLED=0 barrier:
```
2019-04-25 18:47:09 Error executing google:ubuntu-14.04-64:tests/main/static : 
-----
+ CGO_ENABLED=0
+ go build -o snapd.static github.com/snapcore/snapd/cmd/snapd
../../../sandbox/seccomp/compiler_linux.go:22:2: build constraints exclude all Go files in /home/gopath/src/github.com/snapcore/snapd/vendor/github.com/mvo5/libseccomp-golang
```
I guess I'll redo this as a snap-seccomp command and call out to it.

cmd/snap-seccomp/main.go
```diff
@@ -634,20 +634,28 @@ func preprocess(content []byte) (unrestricted, complain bool) {
...
+	// that the string representation is what we expect for ActLog. The
+	// value and string is defined in
+	// https://github.com/seccomp/libseccomp-golang/pull/29.
 	//
 	// Ultimately, the fix for this workaround is to be able to use the
 	// GetApi() function created in the PR above. It'll tell us if the
 	// kernel, libseccomp, and libseccomp-golang all support ActLog.
 	var actLog seccomp.ScmpAction = seccomp.ActAllow + 1
 
 	if actLog.String() == "Action: Log system call" {
```
zyga: Nitpick, you can just return ```actLog.String() == "Action: Log system call"```.

sandbox/seccomp/compiler_test.go
```diff
@@ -142,3 +142,48 @@ func (s *compilerSuite) TestCompilerNewUnhappy(c *C) {
...
+	c.Assert(err, IsNil)
+
+	res, err := compiler.ActLogSupported()
+	c.Assert(err, IsNil)
+	c.Assert(res, Equals, false)
+}
+
+func (s *compilerSuite) TestSupportsActLogError(c *C) {
+	cmd := testutil.MockCommand(c, "snap-seccomp", `
+echo "i will not"
```
zyga:
```suggestion
echo "Thou Shall Not Pass"
```

sandbox/seccomp/compiler_test.go
```diff
@@ -142,3 +142,48 @@ func (s *compilerSuite) TestCompilerNewUnhappy(c *C) {
...
+exit 1
+`)
+	defer cmd.Restore()
+
+	compiler, err := seccomp.New(fromCmd(c, cmd))
+	c.Assert(err, IsNil)
+
+	res, err := compiler.ActLogSupported()
+	c.Assert(err, ErrorMatches, "i will not")
```
zyga:
```suggestion
	c.Assert(err, ErrorMatches, "Thou Shall Not Pass")
```

:-)

interfaces/seccomp/backend.go
```diff
@@ -321,6 +332,10 @@ func (b *Backend) SandboxFeatures() []string {
 		tags = append(tags, "kernel:"+feature)
 	}
 	tags = append(tags, "bpf-argument-filtering")
+
+	if b.actLogSupported {
+		tags = append(tags, "bpf-actlog")
```
zyga: Nice touch, thank you!

cmd/snap-seccomp/main.go
```diff
@@ -771,6 +779,13 @@ func main() {
 		err = showSeccompLibraryVersion()
 	case "version-info":
 		err = showVersionInfo()
+	case "actlog-supported":
```
bboozzoo: Wonder if that should rather be part of version-info output. It conveys the information about capabilities of libseccomp-golang, and that in my view could count as version information the same way a libseccomp version and a hash of supported syscalls does.
Perhaps we could have a string like this:
```
<build-id> <libseccomp-version> <hash-of-syscall-names> <capability-tags>
```

jdstrand: @bboozzoo - it would mean I could make one less callout to snap-seccomp in my subsequent PR, but I was a bit concerned I'd get bogged down in system-key. The 'daemon user' PR has spun so far out and I'm desperately trying to reign it in. Perhaps I can have the command return both, but have system-key only care about what it cares about now.

jdstrand:
I'm closing this. I'm going to add the 4th field to system-key which will clean things up in the subsequent PR.

PR #6781, opened 2019-04-25
overlord: update static attrs when reloading connections

zyga:
The persistent state of a snapd connection contains a copy of the static
interface attributes that were collected from plug and slot
participating in the connection. We were notified that snapd copies
those attributes only during the initial connection. As the snaps are
refreshed the attributes can change according to the change of snap.yaml
data in both snaps.

In the case of the content interface the attributes are essential for
correct operation of the system. The stale copy is part of a larger
problem that is described in detail in the bug referenced below. The
larger problem involves dynamic attributes that are equally stored in
the connect. A full solution would have to involve running interface
hooks to the extent necessary to inform both participating snaps about
the new peer and, if necessary, re-negotiate connection attributes. This
will likely take a longer moment to prepare.

Meanwhile, while interface hooks are not used as broadly as the content
interface is. In an effort to resolve the static part of the problem
snapd will now refresh the cached copy of the static connection
attributes whenever a it is reloading connections. This is interestingly
just prior to re-generating security profiles on snapd startup and
during a snap refresh and associated security setup, just when we want
it.

A regression test is attached as well. During the development of the
regression test I encountered another problem where the mount system
misbehaves when a snap is refreshed. This is similar to reports that
were earlier mentioned on the forum and in the launchpad bug tracker.
To avoid the bug, the content interface is mounted in a sub-directory of
the main snap, in a place that is otherwise irrelevant to the operation
of snapd.

Several existing unit tests were affected by this behaviour.

Fixes: https://bugs.launchpad.net/snapd/+bug/1825883


overlord/ifacestate/helpers.go
```diff
@@ -296,34 +296,64 @@ func (m *InterfaceManager) reloadConnections(snapName string) ([]string, error)
...
 		} else {
+			// If the connection succeeded update the connection state and keep
+			// track of the snaps that were affected.
 			affected[connRef.PlugRef.Snap] = true
 			affected[connRef.SlotRef.Snap] = true
+			// XXX: ideally we'd know the revision associated with the
+			// attributes _or_ did a deep comparison but for the sake of
+			// simplicity this is not done and attributes are always refreshed.
+			connState.StaticPlugAttrs = updatedStaticPlugAttrs
+			connState.StaticSlotAttrs = updatedStaticSlotAttrs
```
stolowski: This is fine, we're updating connState reference, no need to insert back to conns map.

overlord/ifacestate/ifacestate_test.go
```diff
@@ -2373,6 +2377,221 @@ func (s *interfaceManagerSuite) TestDoSetupSnapSecurityKeepsExistingConnectionSt
...
+	// Put a connection in the state. The connection binds the two snaps we are
+	// adding below. The connection contains a copy of the static attributes
+	// but refers to the "old" values, in contrast to what the snaps define.
+	s.state.Lock()
+	s.state.Set("conns", map[string]interface{}{
+		"consumer:plug producer:slot": map[string]interface{}{
+			"interface":   "test",
+			"plug-static": map[string]interface{}{"attr": "old-plug-attr"},
+			"slot-static": map[string]interface{}{"attr": "old-slot-attr"},
+		},
```
stolowski: Could you please add a "canary" connection here (for other snap)? I had one in the original test, the idea is to make sure it remains untouched after the connection we're iterested in gets updated.

overlord/ifacestate/ifacestate_test.go
```diff
@@ -2373,6 +2377,221 @@ func (s *interfaceManagerSuite) TestDoSetupSnapSecurityKeepsExistingConnectionSt
...
+	repo := mgr.Repository()
+	conn, err := repo.Connection(connRef)
+	c.Assert(err, IsNil)
+	c.Check(conn.Plug.StaticAttrs(), DeepEquals, map[string]interface{}{"attr": "new-plug-attr"})
+	c.Check(conn.Slot.StaticAttrs(), DeepEquals, map[string]interface{}{"attr": "new-slot-attr"})
+
+	// Because of the fact that during testing the system key always
+	// mismatches, the security setup is performed.
+	c.Check(secBackend.SetupCalls, HasLen, 2)
+}
```
stolowski: I know backend setup callback check sort of covers it, but for completeness I'd check get "conns" from state here and check the expected map. That's also probably the easiest way to ensure the "canary" connection I mentioned above is in place.

overlord/ifacestate/ifacestate_test.go
```diff
@@ -2373,6 +2377,221 @@ func (s *interfaceManagerSuite) TestDoSetupSnapSecurityKeepsExistingConnectionSt
...
+	// producer snaps we introduce below.
+	secBackend := &ifacetest.TestSecurityBackend{
+		BackendName: "test",
+		SetupCallback: func(snapInfo *snap.Info, opts interfaces.ConfinementOptions, repo *interfaces.Repository) error {
+			// Whenever this function is invoked to setup security for a snap
+			// we check the connection attributes that it would act upon.
+			// Those attributes should always match those of the snap version.
+			conn, err := repo.Connection(connRef)
+			c.Assert(err, IsNil)
+			switch snapInfo.Version {
```
stolowski: Nice test!

overlord/ifacestate/helpers.go
```diff
@@ -296,34 +296,64 @@ func (m *InterfaceManager) reloadConnections(snapName string) ([]string, error)
...
+			connStateChanged = true
```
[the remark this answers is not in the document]
zyga: It is tempting but it is not quite the same. It is possible to have `len(affected) > 0` and `connStateChanged = false` at the same time. One example is where revisions didn't change, another example is when the attributes did not change across revisions. We really need to either use deep equals or revision check coupled with deep equals.

overlord/ifacestate/ifacestate_test.go
zyga: Sure, will do.

overlord/ifacestate/ifacestate_test.go
zyga: +1

overlord/ifacestate/helpers.go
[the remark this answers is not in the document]
stolowski: Ah, I see, good point.

overlord/ifacestate/ifacestate_test.go
```diff
@@ -2511,6 +2532,10 @@ slots:
 	s.mockSnap(c, consumerV1Yaml)
 	snaptest.MockSnapInstance(c, "", consumerV2Yaml, &snap.SideInfo{Revision: snap.R(2)})
 	snaptest.MockSnapInstance(c, "", producerV2Yaml, &snap.SideInfo{Revision: snap.R(2)})
+	// Mock two unrelated snaps, those will show that the state of unrelated
+	// snaps is not clobbered by the refresh process.
+	s.mockSnap(c, unrelatedAYaml)
+	s.mockSnap(c, unrelatedBYaml)
```
stolowski: 👍

PR #6782, opened 2019-04-26
osutil: use go build-id when no gnu build-id is available

mvo5:
The go compiler will only create a GNU build-id when either
build with -buildmode=pie or when we have a `import "C"`
somewhere. We do not use -buildmode=pie on 32bit systems
and with PR#6759 we will stop using cgo (`import "C"`) soon.

This means that we have no GNU build-id in our binaries
anymore on some sytems. The build-id is criticial for the
system-key generation. Fortunately go also provides a
build-id that we can use. It is available since at least
go-1.6 so should cause no problems. Its longer than the
GNU build id (~50 bytes) but not unwieldy.

We could also consider dropping the support for the GNU
build-id entirely the downside of that is that we will not be
able to take the build-id of any non-go program (which
does not matter right now).

### DIFF
--- a/osutil/buildid.go
+++ b/osutil/buildid.go
@@ -36,18 +36,16 @@ type elfNoteHeader struct {
 	Type   uint32
 }
 
-// ReadBuildID returns the GNU build ID note of the provided ELF executable.
-// The ErrNoBuildID error is returned when one is not found.
-//
-// Observed Go binaries presented one when built with:
-//
-//      go build -buildmode=pie
-//
-// See details at http://fedoraproject.org/wiki/Releases/FeatureBuildId
+// ReadBuildID will return the GNU build-id if available and else the
+// GO build-id (the go-buildid is only available
 func ReadBuildID(fname string) (string, error) {
-	const ELF_NOTE_GNU = "GNU\x00"
-	const NT_GNU_BUILD_ID uint32 = 3
+	if buildId, err := readGenericBuildID(fname, "GNU\x00", 3); err == nil {
+		return buildId, nil
+	}
+	return readGenericBuildID(fname, "Go\x00\x00", 4)
+}
 
+func readGenericBuildID(fname, elfNote string, hdrType uint32) (string, error) {
 	// Open the designated ELF file
 	f, err := elf.Open(fname)
 	if err != nil {
@@ -73,7 +71,7 @@ func ReadBuildID(fname string) (string, error) {
 		}
 
 		// We are looking for a specific type of note
-		if nHdr.Type != NT_GNU_BUILD_ID {
+		if nHdr.Type != hdrType {
 			continue
 		}
 
@@ -84,7 +82,7 @@ func ReadBuildID(fname string) (string, error) {
 		}
 
 		// We are only interested in GNU build IDs
-		if string(noteName) != ELF_NOTE_GNU {
+		if string(noteName) != elfNote {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


osutil/buildid.go
```diff
@@ -36,18 +36,16 @@ type elfNoteHeader struct {
...
-// See details at http://fedoraproject.org/wiki/Releases/FeatureBuildId
+// ReadBuildID will return the GNU build-id if available and else the
+// GO build-id (the go-buildid is only available
 func ReadBuildID(fname string) (string, error) {
-	const ELF_NOTE_GNU = "GNU\x00"
-	const NT_GNU_BUILD_ID uint32 = 3
+	if buildId, err := readGenericBuildID(fname, "GNU\x00", 3); err == nil {
+		return buildId, nil
+	}
+	return readGenericBuildID(fname, "Go\x00\x00", 4)
```
zyga: Can you provide two `const` definitions for the 3 and 4 magic values here?

As a separate question, why does go note name use two NUL bytes? To have the same length as the GNU note?

zyga:
@mvo5 the computer says no:
```
FAIL: buildid_test.go:103: buildIDSuite.TestReadBuildGo
buildid_test.go:119:
    c.Assert(err, IsNil)
... value *exec.ExitError = &exec.ExitError{ProcessState:(*os.ProcessState)(0xc4202a57c0), Stderr:[]uint8(nil)} ("exit status 2")
```

osutil/buildid.go
```diff
@@ -36,18 +36,23 @@ type elfNoteHeader struct {
...
-//
-// See details at http://fedoraproject.org/wiki/Releases/FeatureBuildId
+const (
+	gnuElfNote = "GNU\x00"
+	gnuHdrType = 3
+	goElfNote  = "Go\x00\x00"
+	goHdrType  = 4
+)
+
+// ReadBuildID will return the GNU build-id if available and else the
```
bboozzoo: I'd tweak the comment like this:
```
ReadBuildID returns the build ID of a given binary. GNU BuildID is is preferred over Go BuildID. 
Returns an error when neither is found
```

bboozzoo:
Pushed a little commit to figure out why it's failing on Travis. Turns out the `file` command, doesn't know about Go Build IDs:
```
file output: "/tmp/check-8406224889089203558/0/true: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, not stripped\n"
buildid_test.go:124:
    c.Assert(string(decoded), Equals, buildID(c, goTruth))
buildid_test.go:55:
    c.Assert(matches, HasLen, 3)
... obtained []string = []string(nil)
... n int = 3
```

osutil/buildid_test.go
```diff
@@ -78,7 +88,7 @@ func (s *buildIDSuite) TestReadBuildIDmd5(c *C) {
 	md5Truth := filepath.Join(c.MkDir(), "true")
 	err := ioutil.WriteFile(md5Truth+".c", []byte(`int main(){return 0;}`), 0644)
 	c.Assert(err, IsNil)
-	output, err := exec.Command(gccPath, "-Wl,-build-id=md5", "-xc", md5Truth+".c", "-o", md5Truth).CombinedOutput()
+	output, err := exec.Command(gccPath, "-Wl,--build-id=md5", "-xc", md5Truth+".c", "-o", md5Truth).CombinedOutput()
```
zyga: Ouch, did we always have this broken and not notice?

chipaca:
can we include this in 2.39.1?

PR #6786, opened 2019-04-26
daemon: verify snap instructions for multi-snap requests

chipaca:
Without this change, the checks done in `verifySnapInstructions` were
not done for requests done by POSTing to `/v2/snaps`. This meant that
you could install `ubuntu-core`, by doing `snap install ubuntu-core
hello`.

This change makes the multi-snap operation also check the instruction,
so now that multi-snap operation properly errors out.


zyga:
@mvo5 feels like another 2.39 back port.

PR #6787, opened 2019-04-26
snap-seccomp: add 4th field to version-info for golang-seccomp features

jdstrand:
After a small refactor to have actLogSupported(), have versionInfo()
call out to it and append the result to the version-info output,
defaulting to '-' with no detected features, then adjust tests.
    
In sandbox/seccomp/compiler.go, adjust the regex to have field lengths
and drop the fuzzy length check. The lengths could be more specific with
exact lengths for build-id and hash, but retain the previous behavior
and accept a length of one for these. Add a bunch of tests for the
regexp.

Finally, adjust sandbox-features to display bpf-actlog when available


cmd/snap-seccomp/versioninfo.go
```diff
@@ -65,3 +66,12 @@ func showVersionInfo() error {
 	fmt.Fprintln(os.Stdout, vi)
 	return nil
 }
+
+func GoSeccompFeatures() string {
+	features := "-"
+	// as more features are added, make this colon-separated
```
zyga: Colon is a bit unusual, is there a specific reason for it?

cmd/snapd/main_test.go
```diff
@@ -63,7 +63,7 @@ func (s *snapdSuite) TestSanityFailGoesIntoDegradedMode(c *C) {
 	restore = apparmor.MockIsHomeUsingNFS(func() (bool, error) { return false, nil })
 	defer restore()
 	restore = seccomp.MockSnapSeccompVersionInfo(func(s seccomp.Compiler) (string, error) {
-		return "abcdef 1.2.3 1234abcd", nil
+		return "abcdef 1.2.3 1234abcd -", nil
```
zyga: https://twitter.com/zygoon/status/1115944564441997312 (half serious ;-)

interfaces/seccomp/backend.go
```diff
@@ -311,16 +311,22 @@ func (b *Backend) NewSpecification() interfaces.Specification {
...
 	for _, feature := range features {
-		// Prepend "kernel:" to apparmor kernel features to namespace them and
-		// allow us to introduce our own tags later.
+		// Prepend "kernel:" to apparmor kernel features to namespace
+		// them.
 		tags = append(tags, "kernel:"+feature)
 	}
 	tags = append(tags, "bpf-argument-filtering")
+
+	if strings.Contains(b.versionInfo, "bpf-actlog") {
```
zyga: This is a bit crude. Not a priority but we could get a`userspaceFeatures` helper that does the heavy lifting to tell us what the features are in a more robust way.

sandbox/seccomp/compiler.go
```diff
@@ -28,11 +28,14 @@ import (
...
-	// The format of version-info: <build-id> <libseccomp-version> <hash>
+	// version-info format: <build-id> <libseccomp-version> <hash> <features>
 	// Where, the hash is calculated over all syscall names supported by the
-	// libseccomp library.
-	// Ex: 7ac348ac9c934269214b00d1692dfa50d5d4a157 2.3.3 03e996919907bc7163bc83b95bca0ecab31300f20dfa365ea14047c698340e7c
-	validVersionInfo = regexp.MustCompile(`^[0-9a-f]+ [0-9]+\.[0-9]+\.[0-9]+ [0-9a-f]+$`)
+	// libseccomp library. The build-id is a 160-bit SHA-1 (40 char) string
+	// and the hash is a 256-bit SHA-256 (64 char) string. Allow libseccomp
+	// version to be 1-5 chars per field (eg, 1.2.3 or 12345.23456.34567)
+	// and 1-30 chars of features.
```
zyga: It would be good to mention here that features are colon separated.

sandbox/seccomp/compiler_test.go
```diff
@@ -50,13 +50,24 @@ func (s *compilerSuite) TestVersionInfoValidate(c *C) {
 		err string
 	}{
 		// valid
-		{"7ac348ac9c934269214b00d1692dfa50d5d4a157 2.3.3 03e996919907bc7163bc83b95bca0ecab31300f20dfa365ea14047c698340e7c", "7ac348ac9c934269214b00d1692dfa50d5d4a157 2.3.3 03e996919907bc7163bc83b95bca0ecab31300f20dfa365ea14047c698340e7c", ""},
-		{"abcdef 0.0.0 abcd", "abcdef 0.0.0 abcd", ""},
+		{"7ac348ac9c934269214b00d1692dfa50d5d4a157 2.3.3 03e996919907bc7163bc83b95bca0ecab31300f20dfa365ea14047c698340e7c bpf-actlog", "7ac348ac9c934269214b00d1692dfa50d5d4a157 2.3.3 03e996919907bc7163bc83b95bca0ecab31300f20dfa365ea14047c698340e7c bpf-actlog", ""},
+		{"7ac348ac9c934269214b00d1692dfa50d5d4a157 2.3.3 03e996919907bc7163bc83b95bca0ecab31300f20dfa365ea14047c698340e7c -", "7ac348ac9c934269214b00d1692dfa50d5d4a157 2.3.3 03e996919907bc7163bc83b95bca0ecab31300f20dfa365ea14047c698340e7c -", ""},
+		{"abcdef 0.0.0 abcd bpf-actlog", "abcdef 0.0.0 abcd bpf-actlog", ""},
```
zyga: Could you add one more case that has two features please?

cmd/snap-seccomp/versioninfo.go
jdstrand: Well, passwd, group, shadow, gshadow beg to differ. The point is it shouldn't be a space. It doesn't matter otherwise.

cmd/snapd/main_test.go
jdstrand: The format is already decided: space separated fields. Changing that isn't a big deal to me, but should be in a different PR imho.

sandbox/seccomp/compiler.go
jdstrand: Done

sandbox/seccomp/compiler_test.go
jdstrand: Done

interfaces/seccomp/backend.go
jdstrand: Done

jdstrand:
FYI, this is a prerequisite for https://github.com/snapcore/snapd/pull/6681

cmd/snapd/main_test.go
zyga: I just found the - funny, it is a special marker to indicate that a variable list component is absent in mountinfo. I have no problem with this.

cmd/snap-seccomp/versioninfo.go
zyga: Indeed, I haven't thought about those. In either case I think this is fine, it's not a blocker from my side.

sandbox/seccomp/compiler.go
```diff
@@ -66,18 +70,47 @@ func (c *Compiler) VersionInfo() (string, error) {
...
+	return strings.Split(versionInfo, " ")[1], nil
+}
+
+// GetGoSeccompFeatures parses the output of VersionInfo and provides the
+// golang seccomp features
+func GetGoSeccompFeatures(versionInfo string) (string, error) {
+	if match := validVersionInfo.Match([]byte(versionInfo)); !match {
+		return "", fmt.Errorf("invalid format of version-info: %q", versionInfo)
+	}
+	return strings.Split(versionInfo, " ")[3], nil
```
bboozzoo: Maybe we should return `""` when there are no runtime detected featrures (i.e. features part is `-`).

cmd/snap-seccomp/versioninfo.go
```diff
@@ -65,3 +66,12 @@ func showVersionInfo() error {
 	fmt.Fprintln(os.Stdout, vi)
 	return nil
 }
+
+func GoSeccompFeatures() string {
```
mvo5: (nitpick) AFAICT we don't need to export this as it is only needed in tests, something like:
```
--- a/cmd/snap-seccomp/export_test.go
+++ b/cmd/snap-seccomp/export_test.go
@@ -20,9 +20,10 @@
 package main
 
 var (
-       Compile         = compile
-       SeccompResolver = seccompResolver
-       VersionInfo     = versionInfo
+       Compile           = compile
+       SeccompResolver   = seccompResolver
+       VersionInfo       = versionInfo
+       GoSeccompFeatures = goSeccompFeatures
 )
 
 func MockArchUbuntuArchitecture(f func() string) (restore func()) {
diff --git a/cmd/snap-seccomp/versioninfo.go b/cmd/snap-seccomp/versioninfo.go
index 99378bcbba..600bb80937 100644
--- a/cmd/snap-seccomp/versioninfo.go
+++ b/cmd/snap-seccomp/versioninfo.go
@@ -53,7 +53,7 @@ func versionInfo() (string, error) {
        }
 
        major, minor, micro := seccomp.GetLibraryVersion()
-       features := GoSeccompFeatures()
+       features := goSeccompFeatures()
 
        return fmt.Sprintf("%s %d.%d.%d %x %s", myBuildID, major, minor, micro, sh.Sum(nil), features), nil
 }
@@ -67,7 +67,7 @@ func showVersionInfo() error {
        return nil
 }
 
-func GoSeccompFeatures() string {
+func goSeccompFeatures() string {
        features := "-"
        // as more features are added, make this colon-separated
        if actLogSupported() {
```
should work, yes?

cmd/snap-seccomp/main.go
```diff
@@ -634,19 +634,25 @@ func preprocess(content []byte) (unrestricted, complain bool) {
...
+// defined in https://github.com/seccomp/libseccomp-golang/pull/29.
+//
+// Ultimately, the fix for this workaround is to be able to use the GetApi()
+// function created in the PR above. It'll tell us if the kernel, libseccomp,
+// and libseccomp-golang all support ActLog, but GetApi() is also not available
+// in golang-seccomp <= 0.9.0.
+const actLog seccomp.ScmpAction = seccomp.ActAllow + 1
+
+func actLogSupported() bool {
+	return actLog.String() == "Action: Log system call"
```
mvo5: Very nice and clever!

sandbox/seccomp/compiler.go
```diff
@@ -66,18 +70,47 @@ func (c *Compiler) VersionInfo() (string, error) {
...
 	if match := validVersionInfo.Match(raw); !match {
 		return "", fmt.Errorf("invalid format of version-info: %q", raw)
 	}
 
 	return string(raw), nil
 }
 
+// GetLibseccompVersion parses the output of VersionInfo and provides the
+// libseccomp version
+func GetLibseccompVersion(versionInfo string) (string, error) {
```
mvo5: (nitpick) I wonder if something like:
```
type VersionInfo string
func (vi VersionInfo) LibseccompVersion() (string, error)
func (vi VersionInfo) Features() (string, error)
func (vi VersionInfo) HasFeature(feature string) (bool, error)
```
would be more "go-ish". I played a bit with this and here is a full diff with my idea:  http://paste.ubuntu.com/p/WWg6Pm8jRG/

cmd/snap-seccomp/versioninfo.go
```diff
@@ -65,3 +66,12 @@ func showVersionInfo() error {
...
+	features := "-"
+	// as more features are added, make this colon-separated
+	if actLogSupported() {
+		features = "bpf-actlog"
+	}
+	return features
```
mvo5: (nitpick^2) I know this is a bit of "YAGNI" but I would prepare for the ":" a bit more, something like (mostly because the other side of the code in versioninfo.go is already prepared for this):
```
--- a/cmd/snap-seccomp/versioninfo.go
+++ b/cmd/snap-seccomp/versioninfo.go
@@ -23,6 +23,7 @@ import (
        "crypto/sha256"
        "fmt"
        "os"
+       "strings"
 
        "github.com/mvo5/libseccomp-golang"
 
@@ -68,10 +69,14 @@ func showVersionInfo() error {
 }
 
 func GoSeccompFeatures() string {
-       features := "-"
+       var features []string
        // as more features are added, make this colon-separated
        if actLogSupported() {
-               features = "bpf-actlog"
+               features = append(features, "bpf-actlog")
        }
-       return features
+
+       if len(features) == 0 {
+               return "-"
+       }
+       return strings.Join(features, ":")
 }
```

mvo5:
Given the number of :+1: and that it is important for #6681 I will proposed the diffs I added inline as a separate PR.

jdstrand:
Thanks!

PR #6789, opened 2019-04-26
tests: give more time until nc returns on appstream test

sergiocazzolato:
During executions the requests done on the appstream could take more
than 2 seconds. With 3 second of timeout the error cannot be reproduced
anymore.

Error:
https://paste.ubuntu.com/p/gbfkv796F4/

Test executions:
https://paste.ubuntu.com/p/37qryMK8zg/

### DIFF
--- a/tests/main/appstream-id/task.yaml
+++ b/tests/main/appstream-id/task.yaml
@@ -9,7 +9,7 @@ prepare: |
 execute: |
     echo "Verify that search results contain common-ids"
     printf 'GET /v2/find?name=test-snapd-appstreamid HTTP/1.0\r\n\r\n' | \
-        nc -U -q 1 /run/snapd.socket| grep '{'| \
+        nc -U -q 5 /run/snapd.socket| grep '{'| \
         jq -r ' .result[0]["common-ids"] | sort | join (",")' | \
         MATCH 'io.snapcraft.test-snapd-appstreamid.bar,io.snapcraft.test-snapd-appstreamid.foo'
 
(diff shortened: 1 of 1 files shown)


zyga:
@sergiocazzolato Thank you for the patch!

It would be more useful to provide the relevant portion of the two pastebins than the URLs as that part of the commit message will eventually stop working. In addition it requires reviewers to chase through a log file where normally a few lines would be sufficient.

tests/main/appstream-id/task.yaml
```diff
@@ -9,20 +9,20 @@ prepare: |
 execute: |
     echo "Verify that search results contain common-ids"
     printf 'GET /v2/find?name=test-snapd-appstreamid HTTP/1.0\r\n\r\n' | \
-        nc -U -q 1 /run/snapd.socket| grep '{'| \
+        nc -U -q 3 /run/snapd.socket| grep '{'| \
```
bboozzoo: Wouldn't mind if we increased the timeout even a bit more.

stolowski: +1

sergiocazzolato: @bboozzoo thanks for the review, I moved that to 5 seconds

sergiocazzolato:
@zyga @mvo5  thanks for the review, I already updated the commit message.

chipaca:
Why do we even have a timeout? and why are we using `nc` for this :-)
why not
    http snapd:///v2/find name==test-snapd-appstreamid | \
? (or `curl --unix-socket` if that works on trusty)

chipaca:
(in any case, thank you for the fix! )

PR #6790, opened 2019-04-29
interfaces: add login-session-control interface

AlanGriffiths:
Add a "login-session-control" interface that enables a "desktop session" to negotiate with logind from a confined snap.

As discussed here: https://forum.snapcraft.io/t/negotiating-a-vt-session-with-logind-needs-a-new-interface/10969

There's a proof-of-concept branch of Mir's "mir-test-tools" using this here: https://github.com/MirServer/mir-test-tools/tree/login-session-control (the observable difference being it can be used from a VT without needing sudo).

### DIFF
--- a/interfaces/builtin/login_session_control.go
+++ b/interfaces/builtin/login_session_control.go
@@ -38,7 +38,7 @@ dbus (send,receive)
     bus=system
     path=/org/freedesktop/login1/{seat,session}/*
     interface=org.freedesktop.DBus.Properties
-    member={GetAll,PropertiesChanged}
+    member={GetAll,PropertiesChanged,Get}
     peer=(label=unconfined),
 
 dbus (send,receive)
--- a/interfaces/builtin/login_session_control.go
+++ b/interfaces/builtin/login_session_control.go
@@ -54,8 +54,6 @@ dbus (send,receive)
     interface=org.freedesktop.login1.Session
     member={TakeControl,TakeDevice,PauseDevice,PauseDeviceComplete,ResumeDevice,ReleaseDevice,Active,State,Lock,Unlock}
     peer=(label=unconfined),
-
-/{,usr/}{,s}bin/loginctl ixr,
 `
 
 func init() {
--- a/interfaces/builtin/login_session_control.go
+++ b/interfaces/builtin/login_session_control.go
@@ -19,10 +19,10 @@
 
 package builtin
 
-const sessionSummary = `allows setup of login session & seat`
+const loginSessionControlSummary = `allows setup of login session & seat`
 
-const sessionBaseDeclarationSlots = `
-  session:
+const loginSessionControlBaseDeclarationSlots = `
+  login-session-control:
     allow-installation:
       slot-snap-type:
         - core
@@ -31,7 +31,7 @@ const sessionBaseDeclarationSlots = `
       on-classic: false
 `
 
-const sessionConnectedPlugAppArmor = `
+const loginSessionControlConnectedPlugAppArmor = `
 # Description: Can setup login session & seat.
 
 #include <abstractions/dbus-strict>
@@ -59,12 +59,12 @@ dbus (send)
 
 func init() {
 	registerIface(&commonInterface{
-		name:                  "session",
-		summary:               sessionSummary,
+		name:                  "login-session-control",
+		summary:               loginSessionControlSummary,
 		implicitOnCore:        true,
 		implicitOnClassic:     true,
-		baseDeclarationSlots:  sessionBaseDeclarationSlots,
-		connectedPlugAppArmor: sessionConnectedPlugAppArmor,
+		baseDeclarationSlots:  loginSessionControlBaseDeclarationSlots,
+		connectedPlugAppArmor: loginSessionControlConnectedPlugAppArmor,
 		reservedForOS:         true,
 	})
 }
--- a/interfaces/builtin/login_session_control.go
+++ b/interfaces/builtin/login_session_control.go
@@ -27,8 +27,6 @@ const loginSessionControlBaseDeclarationSlots = `
       slot-snap-type:
         - core
     deny-auto-connection: true
-    deny-connection:
-      on-classic: false
 `
 
 const loginSessionControlConnectedPlugAppArmor = `
--- a/interfaces/builtin/login_session_control.go
+++ b/interfaces/builtin/login_session_control.go
@@ -36,20 +36,21 @@ const loginSessionControlConnectedPlugAppArmor = `
 
 #include <abstractions/dbus-strict>
 
-dbus (send)
+dbus (send,receive)
     bus=system
     path=/org/freedesktop/login1/{seat,session}/*
     interface=org.freedesktop.DBus.Properties
-    member=GetAll,
+    member={GetAll,PropertiesChanged}
+    peer=(label=unconfined),
 
-dbus (send)
+dbus (send,receive)
     bus=system
     path=/org/freedesktop/login1/seat/*
     interface=org.freedesktop.login1.Seat
-    member=ActiveSession
+    member={ActiveSession,SwitchTo}
     peer=(label=unconfined),
 
-dbus (send)
+dbus (send,receive)
     bus=system
     path=/org/freedesktop/login1/session/*
     interface=org.freedesktop.login1.Session
--- a/interfaces/builtin/login_session_control.go
+++ b/interfaces/builtin/login_session_control.go
@@ -30,30 +30,32 @@ const loginSessionControlBaseDeclarationSlots = `
 `
 
 const loginSessionControlConnectedPlugAppArmor = `
-# Description: Can setup login session & seat.
+# Description: Can setup login session & seat. This grants privileged access to user sessions.
 
 #include <abstractions/dbus-strict>
 
 dbus (send,receive)
     bus=system
-    path=/org/freedesktop/login1/{seat,session}/*
+    path=/org/freedesktop/login1/{seat,session}/**
     interface=org.freedesktop.DBus.Properties
     member={GetAll,PropertiesChanged,Get}
     peer=(label=unconfined),
 
 dbus (send,receive)
     bus=system
-    path=/org/freedesktop/login1/seat/*
+    path=/org/freedesktop/login1/seat/**
     interface=org.freedesktop.login1.Seat
     member={ActiveSession,SwitchTo}
     peer=(label=unconfined),
 
 dbus (send,receive)
     bus=system
-    path=/org/freedesktop/login1/session/*
+    path=/org/freedesktop/login1/session/**
     interface=org.freedesktop.login1.Session
     member={TakeControl,TakeDevice,PauseDevice,PauseDeviceComplete,ResumeDevice,ReleaseDevice,Active,State,Lock,Unlock}
     peer=(label=unconfined),
+
+/{,usr/}{,s}bin/loginctl ixr,
 `
 
 func init() {
--- a/interfaces/builtin/login_session_control_test.go
+++ b/interfaces/builtin/login_session_control_test.go
@@ -85,7 +85,7 @@ func (s *loginSessionControlSuite) TestConnectedPlugSnippet(c *C) {
 	err := apparmorSpec.AddConnectedPlug(s.iface, s.plug, s.slot)
 	c.Assert(err, IsNil)
 	c.Assert(apparmorSpec.SecurityTags(), DeepEquals, []string{"snap.other.app"})
-	c.Assert(apparmorSpec.SnippetForTag("snap.other.app"), testutil.Contains, `org.freedesktop.login1`)
+	c.Assert(apparmorSpec.SnippetForTag("snap.other.app"), testutil.Contains, `Can setup login session & seat.`)
 }
 
 func (s *loginSessionControlSuite) TestInterfaces(c *C) {
--- a/interfaces/builtin/login_session_control_test.go
+++ b/interfaces/builtin/login_session_control_test.go
@@ -30,7 +30,7 @@ import (
 	"github.com/snapcore/snapd/testutil"
 )
 
-type SessionInterfaceSuite struct {
+type LoginSessionControlConnectedInterfaceSuite struct {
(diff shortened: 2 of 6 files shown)


interfaces/builtin/login_session_control_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"github.com/snapcore/snapd/interfaces"
+	"github.com/snapcore/snapd/interfaces/apparmor"
+	"github.com/snapcore/snapd/interfaces/builtin"
+	"github.com/snapcore/snapd/snap"
+	"github.com/snapcore/snapd/snap/snaptest"
+	"github.com/snapcore/snapd/testutil"
+)
+
+type LoginSessionControlConnectedInterfaceSuite struct {
```
zyga: Tip: the test suite could be private, so `loginSessionControlConnectedInterfaceSuite` could be just `loginSessionControlSuite` (which is also shorter).

interfaces/builtin/login_session_control.go
```diff
@@ -0,0 +1,71 @@
...
+    interface=org.freedesktop.login1.Session
+    member={TakeControl,TakeDevice,PauseDevice,PauseDeviceComplete,ResumeDevice,ReleaseDevice,Active,State,Lock,Unlock}
+    peer=(label=unconfined),
+`
+
+func init() {
+	registerIface(&commonInterface{
+		name:                  "login-session-control",
+		summary:               loginSessionControlSummary,
+		implicitOnCore:        true,
```
zyga: I honestly don't remember: is there a `logind` on core systems?

interfaces/builtin/login_session_control.go
```diff
@@ -0,0 +1,71 @@
...
+const loginSessionControlConnectedPlugAppArmor = `
+# Description: Can setup login session & seat.
+
+#include <abstractions/dbus-strict>
+
+dbus (send,receive)
+    bus=system
+    path=/org/freedesktop/login1/{seat,session}/*
+    interface=org.freedesktop.DBus.Properties
+    member={GetAll,PropertiesChanged}
```
zyga: I would also add `Get` method for completeness. I realise that this is probably coming from the PropertiesChanged signal subscription + GetAll to populate the initial set but there may be other users that also need the Get method.

interfaces/builtin/login_session_control.go
```diff
@@ -0,0 +1,71 @@
...
+const loginSessionControlSummary = `allows setup of login session & seat`
+
+const loginSessionControlBaseDeclarationSlots = `
+  login-session-control:
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
+    deny-connection:
+      on-classic: false
```
zyga: What is the need to allow connection on classic systems? Or am I reading this wrong?

interfaces/builtin/login_session_control.go
zyga: Are there any properties that need to be set as well?

interfaces/builtin/login_session_control.go
AlanGriffiths: Yes

interfaces/builtin/login_session_control.go
AlanGriffiths: Probably I read it wrong while experimenting (this is all unfamiliar syntax to me).

It works without this clause, so I'm just removing it.

interfaces/builtin/login_session_control.go
AlanGriffiths: Adding "Get".

I don't see Mir setting anything here, so AFAIK setting isn't needed.

interfaces/builtin/login_session_control.go
zyga: The question is: what should the default policy be? Should it allow or deny connecting to the interface?

AlanGriffiths: I want to enable creating a confined session that can be selected and started from the greeter. That currently works with an unconfined snap as it can negotiate with logind, from a confined snap the negotiation requires something like this interface.

interfaces/builtin/login_session_control.go
```diff
@@ -0,0 +1,69 @@
...
+`
+
+const loginSessionControlConnectedPlugAppArmor = `
+# Description: Can setup login session & seat.
+
+#include <abstractions/dbus-strict>
+
+dbus (send,receive)
+    bus=system
+    path=/org/freedesktop/login1/{seat,session}/*
```
jdstrand: I suggest '**' here if you expect paths in the form of /org/freedesktop/login1/{seat,session}/foo/bar ('path' in dbus rules operates like file rules)

interfaces/builtin/login_session_control.go
```diff
@@ -0,0 +1,69 @@
...
+    interface=org.freedesktop.DBus.Properties
+    member={GetAll,PropertiesChanged,Get}
+    peer=(label=unconfined),
+
+dbus (send,receive)
+    bus=system
+    path=/org/freedesktop/login1/seat/*
```
jdstrand: Same here

interfaces/builtin/login_session_control.go
```diff
@@ -0,0 +1,69 @@
...
+    interface=org.freedesktop.login1.Seat
+    member={ActiveSession,SwitchTo}
+    peer=(label=unconfined),
+
+dbus (send,receive)
+    bus=system
+    path=/org/freedesktop/login1/session/*
```
jdstrand: and here

interfaces/builtin/login_session_control.go
```diff
@@ -0,0 +1,69 @@
...
+    interface=org.freedesktop.login1.Session
+    member={TakeControl,TakeDevice,PauseDevice,PauseDeviceComplete,ResumeDevice,ReleaseDevice,Active,State,Lock,Unlock}
+    peer=(label=unconfined),
+`
+
+func init() {
+	registerIface(&commonInterface{
+		name:                  "login-session-control",
+		summary:               loginSessionControlSummary,
+		implicitOnCore:        true,
```
jdstrand: Curious, will this make it so that, say mir, could create a login seat for 'root' so you don't strictly need to create /run/user/0 yourselves?

interfaces/builtin/login_session_control.go
```diff
@@ -0,0 +1,69 @@
...
+const loginSessionControlBaseDeclarationSlots = `
+  login-session-control:
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
+`
+
+const loginSessionControlConnectedPlugAppArmor = `
+# Description: Can setup login session & seat.
```
jdstrand: Can you adjust to be:
```
# Description: Can setup login session & seat. This grants privileged access to user sessions.
```

jdstrand: I think you are going to be ok, but if you have no other interfaces other than this one, are there any seccomp denials?

interfaces/builtin/login_session_control.go
```diff
@@ -0,0 +1,69 @@
...
+    member={ActiveSession,SwitchTo}
+    peer=(label=unconfined),
+
+dbus (send,receive)
+    bus=system
+    path=/org/freedesktop/login1/session/*
+    interface=org.freedesktop.login1.Session
+    member={TakeControl,TakeDevice,PauseDevice,PauseDeviceComplete,ResumeDevice,ReleaseDevice,Active,State,Lock,Unlock}
+    peer=(label=unconfined),
+`
```
jdstrand: Should this rule also be present:
```
/{,usr/}{,s}bin/loginctl ixr,
```

Importantly, I do not want to add this rule unless it is known to work with only this interface connected since systemd commands have a tendency to spin out wrt to system access. Eg, systemctl ends up requiring the equivalent of device ownership, but hostnamectl works very cleanly with confinement. This could be demonstrated via spread tests.

interfaces/builtin/login_session_control_test.go
```diff
@@ -0,0 +1,93 @@
...
+func (s *loginSessionControlSuite) TestSanitizePlug(c *C) {
+	c.Assert(interfaces.BeforePreparePlug(s.iface, s.plugInfo), IsNil)
+}
+
+func (s *loginSessionControlSuite) TestConnectedPlugSnippet(c *C) {
+	apparmorSpec := &apparmor.Specification{}
+	err := apparmorSpec.AddConnectedPlug(s.iface, s.plug, s.slot)
+	c.Assert(err, IsNil)
+	c.Assert(apparmorSpec.SecurityTags(), DeepEquals, []string{"snap.other.app"})
+	c.Assert(apparmorSpec.SnippetForTag("snap.other.app"), testutil.Contains, `org.freedesktop.login1`)
```
jdstrand: This isn't specific enough to this interface. Can you perhaps pick a full rule or comment string? Eg:
```
c.Assert(apparmorSpec.SnippetForTag("snap.other.app"), testutil.Contains, `# Description: Can setup login session & seat.`
```

interfaces/builtin/login_session_control.go
AlanGriffiths: Interesting question. I'll have to experiment.

(The scenario I've used for testing is first signing into a VT as a normal user and then starting the snap without sudo as that uses the same Mir codepath as starting a user shell.)

interfaces/builtin/login_session_control.go
AlanGriffiths: I'm not sure how best to answer this. Obviously, Mir needs other interfaces (mir, wayland and opengl) and replicating its full use of org.freedesktop.login1 would be non-trivial.

interfaces/builtin/login_session_control.go
AlanGriffiths: Empirically, adding this rule doesn't break anything, but I don't know enough to know whether this should be present.

interfaces/builtin/login_session_control.go
AlanGriffiths: ...and the answer is: yes.

interfaces/builtin/login_session_control.go
AlanGriffiths: Let me try rephrasing:

On classic systems without this interface, a confined snap does not have the access needed to provide a graphical non-root shell.

That means the interface is useful on classic.

An example would be the "egmde-confined-desktop" snap, which currently runs as a daemon (i.e. root) as a way to gain the access it needs. That has the unfortunate side-effect "taking over" whenever the snap starts.

With this interface in place it will be possible to run an updated  "egmde-confined-desktop" snap as a strictly confined user session selected from the greeter.

interfaces/builtin/login_session_control.go
AlanGriffiths: We don't need this, we're not using `loginctl` - just talking to logind via dbus.

interfaces/builtin/login_session_control.go
AlanGriffiths: @jdstrand Having thought further I don't understand the question:  How could anything dbus (or transitively logind) does as a result of using this interface cause a seccomp denial?

interfaces/builtin/login_session_control.go
AlanGriffiths: Done

interfaces/builtin/login_session_control.go
```diff
@@ -0,0 +1,69 @@
...
+package builtin
+
+const loginSessionControlSummary = `allows setup of login session & seat`
+
+const loginSessionControlBaseDeclarationSlots = `
+  login-session-control:
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
```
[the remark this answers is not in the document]
AlanGriffiths: That's my intention. (I'm tempted by the convenience of auto-connect, but not sure I can justify it.)

AlanGriffiths:
The CI failure looks unrelated to my changes:
```plain
2019-04-30 12:58:35 Error preparing google:amazon-linux-2-64 : 
-----
...
quiet: exit status 2. Output follows:
# cd .; git clone https://go.googlesource.com/crypto /home/gopath/.cache/govendor/golang.org/x/crypto
Cloning into '/home/gopath/.cache/govendor/golang.org/x/crypto'...
fatal: unable to access 'https://go.googlesource.com/crypto/': error:04091068:rsa routines:INT_RSA_VERIFY:bad signature
# cd .; git clone https://gopkg.in/yaml.v2 /home/gopath/.cache/govendor/gopkg.in/yaml.v2
Cloning into '/home/gopath/.cache/govendor/gopkg.in/yaml.v2'...
fatal: unable to access 'https://gopkg.in/yaml.v2/': error:04091068:rsa routines:INT_RSA_VERIFY:bad signature
Error: Remotes failed for:
	Failed for "golang.org/x/crypto/cast5" (failed to clone repo): exit status 128
	Failed for "gopkg.in/yaml.v2" (failed to clone repo): exit status 128
```

interfaces/builtin/login_session_control_test.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func (s *loginSessionControlSuite) TestSanitizeSlot(c *C) {
+	c.Assert(interfaces.BeforePrepareSlot(s.iface, s.slotInfo), IsNil)
+	slot := &snap.SlotInfo{
+		Snap:      &snap.Info{SuggestedName: "some-snap"},
+		Name:      "login-session-control",
+		Interface: "login-session-control",
+	}
+	c.Assert(interfaces.BeforePrepareSlot(s.iface, slot), ErrorMatches,
```
mvo5: (nitpick²) The usual convention in go is not to wrap lines, so this line and the following should be in a single line (but its fine to ignore for this PR, mostly a FYI).

AlanGriffiths: OK. FWIW I was just following existing practice:
```plain
interfaces/builtin/account_control_test.go:87
interfaces/builtin/alsa_test.go:75
interfaces/builtin/autopilot_test.go:77
...
```

mvo5: Heh, sorry for that then - looks like we are less consistent about it then I thought :)Thanks for clarifying this.

PR #6792, opened 2019-04-29
gadget: embed volume in positioned volume, rename fields

bboozzoo:
Fairly trivial PR with no functional changes that would introduce yet more noise to #6769.

Embed Volume information in PositionedVolume. Rename Structures to PositionedStructure.

(cc @cmatsuoka)

### DIFF
--- a/gadget/position_test.go
+++ b/gadget/position_test.go
@@ -93,6 +93,8 @@ volumes:
           size: 100M
 `
 	vol := mustParseVolume(c, gadgetYaml, "first-image")
+	c.Assert(vol.Structure, HasLen, 2)
+
 	v, err := gadget.PositionVolume(p.dir, vol, defaultConstraints)
 	c.Assert(err, IsNil)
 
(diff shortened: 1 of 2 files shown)


gadget/position_test.go
```diff
@@ -137,40 +136,28 @@ volumes:
...
+				Index:           2,
 			},
 			{
-				VolumeStructure: &gadget.VolumeStructure{
-					Size: 100 * gadget.SizeMiB,
-					Type: "00000000-0000-0000-0000-aa00deadbeef",
-				},
-				StartOffset: 1001 * gadget.SizeMiB,
-				Index:       3,
+				VolumeStructure: &vol.Structure[3],
```
stolowski: Would it make sense to assert on len(vol.Structure) == 4 after `vol := mustParseVolume(...` (above and in other tests)?

bboozzoo: Pushed

PR #6795, opened 2019-04-30
cmd,sandbox: tweak seccomp version info handling

mvo5:
Some ideas from the review of https://github.com/snapcore/snapd/pull/6787 - all cosmetic but (IMO) nice to have.

1. Unexport GoSeccompFeatures in cmd/snap-seccomp (its only used internally)
2. Make VersionInfo a type with methods
3. Tweak goSeccompFeatures helper to include the fact that we need to use ":" as separator (this is what the other side already implements).

### DIFF
--- a/packaging/debian-sid/patches/0001-cmd-snap-seccomp-use-upstream-seccomp-package.patch
+++ b/packaging/debian-sid/patches/0001-cmd-snap-seccomp-use-upstream-seccomp-package.patch
@@ -21,11 +21,11 @@ Signed-off-by: Maciej Borzecki <maciej.zenon.borzecki@canonical.com>
  cmd/snap-seccomp/versioninfo_test.go | 2 +-
  4 files changed, 4 insertions(+), 7 deletions(-)
 
-diff --git a/cmd/snap-seccomp/main.go b/cmd/snap-seccomp/main.go
-index 7056c0309dbcc1b6bf9f3b69e6adf454bed4fcbd..dce0f086bde290839a108d90945c880d5f387e5d 100644
---- a/cmd/snap-seccomp/main.go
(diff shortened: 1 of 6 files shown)


packaging/debian-sid/patches/0001-cmd-snap-seccomp-use-upstream-seccomp-package.patch
```diff
@@ -21,11 +21,11 @@ Signed-off-by: Maciej Borzecki <maciej.zenon.borzecki@canonical.com>
  cmd/snap-seccomp/versioninfo_test.go | 2 +-
  4 files changed, 4 insertions(+), 7 deletions(-)
 
-diff --git a/cmd/snap-seccomp/main.go b/cmd/snap-seccomp/main.go
```
stolowski: Why all the .patch changes?

mvo5: Our debian-quilt patch touches cmd/snap-seccomp/versioninfo.go which is also modified by this branch. This leads to a "fuzzy" apply of the patch. Unfortunately dpkg-buildpackage complains when there is fuzziness in the diff so it needed to be refreshed.

sandbox/seccomp/compiler.go
```diff
@@ -78,28 +78,31 @@ func (c *Compiler) VersionInfo() (string, error) {
 	return string(raw), nil
 }
 
-// GetLibseccompVersion parses the output of VersionInfo and provides the
+// VersionInfo represents information about the seccomp compilter
+type VersionInfo string
```
bboozzoo: That's nice!

PR #6796, opened 2019-04-30
cmd/snap-update-ns: add no-op load/save current user profile logic

zyga:
Currently user mount profiles are not persisted. To allow using
common logic across system and user update apply functions we must
ensure that loading current per-user mount profile returns the empty
profile and that saving said profile is a no-op.

### DIFF
--- a/cmd/snap-update-ns/user.go
+++ b/cmd/snap-update-ns/user.go
@@ -41,10 +41,16 @@ func NewUserProfileUpdateContext(instanceName string, fromSnapConfine bool, uid 
 		CommonProfileUpdateContext: CommonProfileUpdateContext{
 			instanceName:       instanceName,
 			fromSnapConfine:    fromSnapConfine,
+			currentProfilePath: currentUserProfilePath(instanceName, uid),
 			desiredProfilePath: desiredUserProfilePath(instanceName),
 		},
 		uid: uid,
 	}
+}
+
+// UID returns the user ID of the mount namespace being updated.
+func (ctx *UserProfileUpdateContext) UID() int {
+	return ctx.uid
 }
 
 // Lock acquires locks / freezes needed to synchronize mount namespace changes.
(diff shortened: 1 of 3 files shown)


cmd/snap-update-ns/user.go
```diff
@@ -78,18 +84,44 @@ func (ctx *UserProfileUpdateContext) LoadDesiredProfile() (*osutil.MountProfile,
...
 	return err
 }
 
 // desiredUserProfilePath returns the path of the fstab-like file with the desired, user-specific mount profile for a snap.
 func desiredUserProfilePath(snapName string) string {
 	return fmt.Sprintf("%s/snap.%s.user-fstab", dirs.SnapMountPolicyDir, snapName)
 }
+
+// currentUserProfilePath returns the path of the fstab-like file with the applied, user-specific mount profile for a snap.
+func currentUserProfilePath(snapName string, uid int) string {
```
mvo5: How/where will this be used? And why is it different from desiredUserProfilePath which does not take a uid ? Mostly wondering if this should be part of a large function or not - its so trivial it seems we might not need it as a separate helper(?). Or maybe it should be a method on ctx *UserProfileUpdateContext ? But only more context can tell :)

zyga: This is used in the subsequent patches where we actually start saving the user profile. It is different from desired because the desired profile is a template where various variables derived from the UID are used (.e.g. SNAP_USER_DATA or XDG_RUNTIME_DIR). The current profile has an UID because those variables are already derived for that user there.

The desired mount profile may look like this:

```
tmpfs $SNAP_USER_DATA/foo ...
```

A current mount profile for user 1000 might look like:

```
tmpfs /home/zyga/snap/my-fancy-snap/42/foo ...
```

We need to have those separately because per-user mount namespaces are updated one-by-one, for each one that exists.

PR #6798, opened 2019-04-30
gadget: introduce checkers for sanitizing structure updates

bboozzoo:
Introduce code for resolving which is the matching voulme description between updated gadget YAML infos, and checkers for sanitizing structure updates.

The rule is that we allow changes in structure content, be it bare, or a filesystem one. However, none of upper level changes in structure are allowed. That is, structure's position (declared and resolved) must not change, it's type or role, the filesystem/bare transition is not allowed, nor is the type of filesystem or the label.

cc @cmatsuoka

### DIFF
--- a/gadget/update.go
+++ b/gadget/update.go
@@ -1,0 +1,27 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+package gadget
+
+import (
+	"errors"
+)
+
+var (
+	ErrNoUpdate = errors.New("no update needed")
+)
--- a/gadget/update.go
+++ b/gadget/update.go
@@ -28,7 +28,7 @@ var (
 )
 
 func resolveVolume(old *Info, new *Info) (oldVol, newVol *Volume, err error) {
-	// TODO: support only one volume
+	// support only one volume
 	if len(new.Volumes) != 1 || len(old.Volumes) != 1 {
 		return nil, nil, errors.New("cannot update with more than one volume")
 	}
--- a/gadget/update.go
+++ b/gadget/update.go
@@ -20,8 +20,102 @@ package gadget
 
 import (
 	"errors"
+	"fmt"
 )
 
 var (
 	ErrNoUpdate = errors.New("no update needed")
 )
+
+func resolveVolume(old *Info, new *Info) (oldVol, newVol *Volume, err error) {
+	// TODO: support only one volume
+	if len(new.Volumes) != 1 || len(old.Volumes) != 1 {
+		return nil, nil, errors.New("cannot update with more than one volume")
+	}
+
+	var name string
+	for n := range old.Volumes {
+		name = n
+		break
+	}
+	oldV := old.Volumes[name]
+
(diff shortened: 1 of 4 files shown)


gadget/update.go
```diff
@@ -20,8 +20,102 @@ package gadget
...
+	if one == nil && two == nil {
+		return true
+	}
+	if one != nil && two != nil {
+		return *one == *two
+	}
+	return false
+}
+
+func isSameRelativeOffset(one *RelativeOffset, two *RelativeOffset) bool {
```
cmatsuoka: It seems that theoretically we could have two RelativeOffsets with different contents but with the same meaning (i.e. different offsets to different structures but both resolving to the same absolute offset) but I'm assuming the idea here is to not allow this situation.

gadget/update.go
```diff
@@ -20,8 +20,102 @@ package gadget
...
+	if one != nil && two != nil {
+		return *one == *two
+	}
+	return false
+}
+
+func isLegacyMBRTransition(from *PositionedStructure, to *PositionedStructure) bool {
+	// legacy MBR could have been specified by setting type: mbr, with no
+	// role
+	return from.Type == MBR && to.EffectiveRole() == MBR
```
cmatsuoka: What if from.Type is MBR but from.Role is set to something else? Or this will never happen?

gadget/update.go
```diff
@@ -20,8 +20,102 @@ package gadget
...
 import (
 	"errors"
+	"fmt"
 )
 
 var (
 	ErrNoUpdate = errors.New("no update needed")
 )
+
+func resolveVolume(old *Info, new *Info) (oldVol, newVol *Volume, err error) {
```
cmatsuoka: This is defined and test but not used in this patch. I'm assuming it's preparation for a future change.

gadget/update.go
```diff
@@ -20,8 +20,102 @@ package gadget
...
+	return false
+}
+
+func isLegacyMBRTransition(from *PositionedStructure, to *PositionedStructure) bool {
+	// legacy MBR could have been specified by setting type: mbr, with no
+	// role
+	return from.Type == MBR && to.EffectiveRole() == MBR
+}
+
+func canUpdateStructure(from *PositionedStructure, to *PositionedStructure) error {
```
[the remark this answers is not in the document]
bboozzoo: Nope, we look at StartOffset (i.e. after the structure has been positioned in the volume) which is only in PositionedStructure

gadget/update.go
bboozzoo: Yeah, you could come up with such relative offset values that resolve to the same location. Though, as you noticed, it's probably too much hassle to support this, so for simplicity, we expect the unresolved offsets to be literally, the same.

gadget/update.go
bboozzoo: Yup, the next batch is already waiting for a PR. I have to control the dosage otherwise nobody would do the reviews :smile:

gadget/update.go
bboozzoo: Hm this should be caught by validation. Let me double check if it is. If not, I'll open a separate PR with the fix.

gadget/update.go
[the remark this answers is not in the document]
cmatsuoka: I mean, we test for many `VolumeStructure` fields such as `Size`, `Label`, `Offset`, `OffsetWrite`, `Type`, `Role` (effective), `ID` and `Filesystem` (if not bare), which are all fields except `Name`. I just wanted to make sure this is deliberate and not a missing test :)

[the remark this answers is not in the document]
bboozzoo: Ah, I totally misunderstood your question. The name is somewhat special and sometimes ends up as  filesystem label (if it's a filesystem structure). I need to double check the exact behavior of ubuntu-image here and look at generated images and will update the code later on. It's already quite confusing as it is, didn't to make it even worse.

gadget/update.go
bboozzoo: Added in a separate PR: #6832

gadget/update.go
```diff
@@ -0,0 +1,121 @@
...
+	"errors"
+	"fmt"
+)
+
+var (
+	ErrNoUpdate = errors.New("no update needed")
+)
+
+func resolveVolume(old *Info, new *Info) (oldVol, newVol *Volume, err error) {
+	// TODO: support only one volume
```
stolowski: Is this TODO relevant, we seem to be supporting exactly one volume below?

PR #6805, opened 2019-04-30
interfaces/docker-support: support overlayfs on ubuntu core

anonymouse64:
Aufs is being deprecated by upstream docker, new recommended storage-driver is overlayfs, but there are some bugs with overlayfs + apparmor, so add some additional accesses on ubuntu core to allow overlayfs to work there.

It would be great to get this into 2.39 if it's not too late yet...

### DIFF
--- a/interfaces/builtin/docker_support.go
+++ b/interfaces/builtin/docker_support.go
@@ -51,8 +51,8 @@ const dockerSupportConnectedPlugAppArmorCore = `
 # overlayfs mount correctly
 # the accesses show up as runc trying to read from
 # /system-data/var/snap/docker/common/var-lib-docker/overlay2/$SHA/diff/
-/system-data/var/snap/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}/common/{,**} rwl,
-/system-data/var/snap/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}/@{SNAP_REVISION}/{,**} rwl,
+/system-data/var/snap/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}/common/{,**/} rwl,
+/system-data/var/snap/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}/@{SNAP_REVISION}/{,**/} rwl,
 `
 
 const dockerSupportConnectedPlugAppArmor = `
--- a/interfaces/builtin/docker_support.go
+++ b/interfaces/builtin/docker_support.go
@@ -45,6 +45,16 @@ const dockerSupportBaseDeclarationSlots = `
     deny-auto-connection: true
 `
 
+const dockerSupportConnectedPlugAppArmorCore = `
+# these accesses are necessary for Ubuntu Core 16 and 18, likely due to the version 
+# of apparmor or the kernel which doesn't resolve the upper layer of an 
+# overlayfs mount correctly
+# the accesses show up as runc trying to read from
+# /system-data/var/snap/docker/common/var-lib-docker/overlay2/$SHA/diff/
+/system-data/var/snap/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}/common/{,**} rwl,
+/system-data/var/snap/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}/@{SNAP_REVISION}/{,**} rwl,
+`
+
 const dockerSupportConnectedPlugAppArmor = `
 # Description: allow operating as the Docker daemon. This policy is
 # intentionally not restrictive and is here to help guard against programming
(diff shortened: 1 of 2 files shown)


interfaces/builtin/docker_support.go
```diff
@@ -45,6 +45,16 @@ const dockerSupportBaseDeclarationSlots = `
...
 `
 
+const dockerSupportConnectedPlugAppArmorCore = `
+# these accesses are necessary for Ubuntu Core 16 and 18, likely due to the version 
+# of apparmor or the kernel which doesn't resolve the upper layer of an 
+# overlayfs mount correctly
+# the accesses show up as runc trying to read from
+# /system-data/var/snap/docker/common/var-lib-docker/overlay2/$SHA/diff/
+/system-data/var/snap/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}/common/{,**} rwl,
+/system-data/var/snap/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}/@{SNAP_REVISION}/{,**} rwl,
```
jdstrand: This is fine since it is in the docker snap's area and we have similar accesses for aufs (without the /system-data bit), but I'm wondering if `.../{,**}` is needed or if `.../{,**/}` is sufficient. I recall when exploring overlays that under some circumstances only the directories were needed, but I don't recall the details OTOH. Again, it is fine as is since these accesses are already allowed via the non-/system-data path.

interfaces/builtin/docker_support.go
```diff
@@ -593,7 +603,16 @@ func (iface *dockerSupportInterface) AppArmorConnectedPlug(spec *apparmor.Specif
 	spec.SetSuppressHomeIx()
 	spec.AddSnippet(dockerSupportConnectedPlugAppArmor)
 	if privileged {
+		// note we don't consider classic vs core here because if we're privileged
+		// then we are unconfined and don't need to worry about apparmor overlayfs
+		// bugs
```
jdstrand: 'privileged' here != unconfined and corresponds to the the `privileged-containers` interface attribute and privileged only gives `/** wl,`, not 'r' but in `dockerSupportConnectedPlugAppArmorCore` you added 'r'. I think the comment needs updating...

anonymouse64: Ah yes you're correct I had confused `docker-support` with `lxd-support`. The `'r'` is definitely needed for the `dockerSupportConnectedPlugAppArmorCore` snippet to work with overlayfs, but I had not tested running a privileged container with the overlayfs storage-driver. 
I'm not really sure if it is necessary to add `'r'` to the `dockerSupportPrivilegedAppArmor` snippet, running a privileged container seems to work without the `'r'` in `dockerSupportPrivilegedAppArmor`. What do you think?

interfaces/builtin/docker_support.go
anonymouse64: I checked and `.../{,**/}` is sufficient for overlayfs to work. Do you want me to change it to `.../{,**/}`?

interfaces/builtin/docker_support.go
anonymouse64: I adjusted the logic and removed the erroneous comment here.

anonymouse64:
Hi folks, any way we can get another review for this in? We'd like to update the snap to use overlayfs ASAP but don't want to enable it until it will work everywhere

interfaces/builtin/docker_support.go
anonymouse64: Done

jdstrand:
Note these test failures:
```
2019/05/23 02:08:52.780533 cmd_run.go:779: cannot extract runtime data: cannot parse end of exec profile: EOF

FAIL: cmd_snapshot_test.go:67: SnapSuite.TestSnapSnaphotsTest

Test: restore x
Test: saved --id=x
Test: saved --id=3
cmd_snapshot_test.go:85:
    c.Check(s.Stdout(), Matches, test.stdout)
... value string = "" +
...     "Set  Snap  Age     Version  Rev   Size    Notes\n" +
...     "3    htop  65d10h  2        1168      1B  auto\n"
... regex string = "" +
...     "Set  Snap  Age    Version  Rev   Size    Notes\n" +
...     "3    htop  .*  2        1168      1B  auto\n"

Test: saved
cmd_snapshot_test.go:85:
    c.Check(s.Stdout(), Matches, test.stdout)
... value string = "" +
...     "Set  Snap  Age     Version  Rev   Size    Notes\n" +
...     "1    htop  65d10h  2        1168      1B  -\n"
... regex string = "" +
...     "Set  Snap  Age    Version  Rev   Size    Notes\n" +
...     "1    htop  .*  2        1168      1B  -\n"

Test: forget x
Test: check-snapshot x
Test: restore 1
Test: forget 2
Test: forget 2 snap1 snap2
Test: check-snapshot 4
Test: check-snapshot 4 snap1 snap2
OOPS: 388 passed, 13 skipped, 1 FAILED
--- FAIL: Test (3.52s)
FAIL
FAIL	github.com/snapcore/snapd/cmd/snap	3.734s
```

These seem unrelated to your snap. Perhaps fetch/merge from master and try again?

anonymouse64:
@jdstrand yes agreed that the test failures seem unrelated. I was able to reproduce locally as well. I rebased to see if that resolves it

PR #6806, opened 2019-04-30
snapcraft.yaml: include libc6 in snapd

mvo5:
We added some tools like xdelta3 and squashfs-tools to the snapd
snap now so that we can use those even if they are not available
on the system. Unfortunately this means that we need to also
carry their libc6 to be truely independent of the host system.

This PR adds libc6 to the snapd snap

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -58,7 +58,7 @@ parts:
   # CommandFromSystemSnap
   libc6:
     plugin: nil
-    stage-packaes:
+    stage-packages:
       - libc6
     stage:
       - lib/*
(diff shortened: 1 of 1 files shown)


mvo5:
@chipaca I looked into this a bit last night, its doable but we need to also build fc-cache-* static which is only using the static libfontconfig right now but a dynamic libc. Doable but we need to look.

chipaca:
@mvo5 yeah as long as nothing touches users nor the network, fully static should be doable, right? Without having to look at musl or some other less-supported thing I mean.

PR #6820, opened 2019-05-03
snap-confine: improve error when running on a not /home homedir

mvo5:
We currently don't support running on a non /home homedirs. The
error in this case is very cryptic:
```
cannot create user data directory: /data1/home/username/snap/hello/20: Read-only file system
```

This PR detects the error in snap-confine and provides a clearer
error message to the user with a link to the relevant forum topic.

Ideally we would have a step-by-step guide in the forum docs how
to fix things there.

### DIFF
--- a/cmd/libsnap-confine-private/string-utils.c
+++ b/cmd/libsnap-confine-private/string-utils.c
@@ -56,6 +56,22 @@ bool sc_endswith(const char *str, const char *suffix)
 	return strncmp(str - xlen + slen, suffix, xlen) == 0;
 }
 
+bool sc_startswith(const char *str, const char *prefix)
+{
+	if (!str || !prefix) {
+		return false;
+	}
+
+	size_t xlen = strlen(prefix);
+	size_t slen = strlen(str);
+
+	if (slen < xlen) {
+		return false;
+	}
+
+	return strncmp(str, prefix, xlen) == 0;
+}
+
 char *sc_strdup(const char *str)
 {
 	size_t len;
--- a/cmd/snap-confine/user-support.c
+++ b/cmd/snap-confine/user-support.c
@@ -21,6 +21,7 @@
 #include <stdlib.h>
 #include <sys/stat.h>
 
+#include "../libsnap-confine-private/string-utils.h"
 #include "../libsnap-confine-private/utils.h"
 
 void setup_user_data(void)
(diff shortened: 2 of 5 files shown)


cmd/snap-confine/user-support.c
```diff
@@ -37,6 +38,11 @@ void setup_user_data(void)
 
 	debug("creating user data directory: %s", user_data);
 	if (sc_nonfatal_mkpath(user_data, 0755) < 0) {
+		if (errno == EROFS && !sc_startswith(user_data, "/home")) {
```
zyga: This would misbehave for /home2/bob. Please add a trailing slash:

```suggestion
		if (errno == EROFS && !sc_startswith(user_data, "/home/")) {
```

tests/main/non-home/task.yaml
```diff
@@ -0,0 +1,30 @@
+summary: Ensure running on none /home dirs gives a useful error
+
+# limit to ubuntu for easier user creation
+systems: [ubuntu-1*, ubuntu-2*]
+
+environment:
+    TUSER: jim
+    THOME: /var/home/jim
```
zyga: Could you add a variant with `/home2/bob` as well please?

cmd/libsnap-confine-private/string-utils.c
```diff
@@ -56,6 +56,22 @@ bool sc_endswith(const char *str, const char *suffix)
 	return strncmp(str - xlen + slen, suffix, xlen) == 0;
 }
 
+bool sc_startswith(const char *str, const char *prefix)
```
zyga: This is nice, thanks!

cmd/snap-confine/user-support.c
```diff
@@ -37,6 +38,11 @@ void setup_user_data(void)
 
 	debug("creating user data directory: %s", user_data);
 	if (sc_nonfatal_mkpath(user_data, 0755) < 0) {
+		if (errno == EROFS && !sc_startswith(user_data, "/home/")) {
+				// clear errno or it will be displayed in die()
+				errno = 0;
+				die("Sorry, cannot use home directories outside /home.\nSee https://forum.snapcraft.io/t/11209 for details.");
```
bboozzoo: Wonder if @degville has some ideas about tweaking the message.

degville: I'd probably slightly tweak:
`Sorry, home directories outside of /home are not currently supported. \nSee https://forum.snapcraft.io/t/11209 for details.`

I think it's worth updating the forum post too, including the bind mount workaround and adding it to the docs - I'm obviously happy to do this if it helps.

mvo5: Thanks, I updated the message. Agreed on the forum post - its a skeleton at this point mostly because I was too lazy^Wbusy to flesh it out. If you want to have a stab I would be very happy!

PR #6823, opened 2019-05-03
packaging: build empty package on powerpc

mvo5:
Snapd was never supported on powerpc because we never built a
core snap. However we did keep the package building mostly
because it would cause friction with the SRU process if we don't.

However since we moved to golang-1.10 we can not even build
the package on powerpc anymore. The previous release required
some manual override for the SRU and foundations team. To avoid
that they have to do this this PR will build an empty package
on powerpc.


mwhudson:
Obviously I'm coming here late and might be missing something, but this is still going to fail to install build dependencies on powerpc isn't it? (I'm stealing the idea for runc :slightly_smiling_face: )

PR #6824, opened 2019-05-03
release: 2.39

mvo5:
Merge the 2.39 changes back into master (changelog plus one extra check that made it into one of the backported branches).


overlord/snapstate/handlers_prereq_test.go
```diff
@@ -468,6 +468,7 @@ func (s *prereqSuite) testDoPrereqNoCorePullsInSnaps(c *C, base string) {
 		},
 	})
 
+	c.Check(t.Change().Err(), IsNil)
```
bboozzoo: Wonder how come this one did not end up in master before.

mvo5: Some of the changes had to be backported because master has changed. As part of this this line was added (because the backport itself was not passing tests and this gives a much better message than the one below).

PR #6844, opened 2019-05-09
interfaces: special-case "snapd" in sanitizeSlotReservedForOS* helpers

stolowski:
Special-case "snapd" in sanitizeSlotReservedForOSOrGadget and sanitizeSlotReservedForOS helpers used by most of interfaces in their sanitize* mathods, so that hotplug slots can be attached to snapd snap in core18+snapd system.

Note: no unit test for this, because these methods will be completely removed in a followup PR (because respective checks are now made by base declarations). I've verified this fixes hotplug on pi3 manually. Unfortunately removing these helpers affects ~12 interfaces directly, and ~80 tests of interfaces, so a quickfix here to make change minimal for 2.39. Followup will clean all the interfaces in master.

### DIFF
--- a/interfaces/builtin/utils.go
+++ b/interfaces/builtin/utils.go
@@ -75,7 +75,7 @@ func plugAppLabelExpr(plug *interfaces.ConnectedPlug) string {
 
 // sanitizeSlotReservedForOS checks if slot is of type os.
 func sanitizeSlotReservedForOS(iface interfaces.Interface, slot *snap.SlotInfo) error {
-	if slot.Snap.Type != snap.TypeOS {
+	if slot.Snap.Type != snap.TypeOS && slot.Snap.InstanceName() != "snapd" {
 		return fmt.Errorf("%s slots are reserved for the core snap", iface.Name())
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


interfaces/builtin/utils.go
```diff
@@ -75,15 +75,15 @@ func plugAppLabelExpr(plug *interfaces.ConnectedPlug) string {
 
 // sanitizeSlotReservedForOS checks if slot is of type os.
 func sanitizeSlotReservedForOS(iface interfaces.Interface, slot *snap.SlotInfo) error {
-	if slot.Snap.Type != snap.TypeOS {
+	if slot.Snap.Type != snap.TypeOS && slot.Snap.InstanceName() != "snapd" {
```
zyga: Should use use snap ID here instead? Though this will interfere with self-built snapd.
Is this in sync with how the policy checker does a similar verification?

stolowski: Good question. If I read it correctly, policy checker indeed uses snap ID (snapIDSnapd function in policy helpers.go). Note though, policy checker is in any case effective, and this temporary quickfix is to just stop custom sanitize from choking before it's removed completely. So If security is a concern, then policy checker takes care of it anyway. I hope that makes sense.

zyga: The policy checker is not considered when doing --dangerous installations so perhaps using snap name here is correct.

mvo5:
Looks great! Thanks! Is there an existing test we could extend to have some minimal test coverage? Or is that too difficult?

stolowski:
> Looks great! Thanks! Is there an existing test we could extend to have some minimal test coverage? Or is that too difficult?

@mvo Added a minimal test now. I didn't include it initially as this code will be go away almost instantly (famous last words ;))

PR #6845, opened 2019-05-09
cmd/snap-update-ns: move apply{Profile,{User,System}Fstab} to same file

zyga:
This sets the stage for the upcoming unification of all the three
functions into one.


cmd/snap-update-ns/update.go
```diff
@@ -39,3 +40,103 @@ type MountProfileUpdateContext interface {
...
+		return err
+	}
+	debugShowProfile(current, "current mount profile")
+
+	as := ctx.Assumptions()
+	_, err = applyProfile(ctx, current, desired, as)
+	return err
+}
+
+func applyProfile(ctx MountProfileUpdateContext, currentBefore, desired *osutil.MountProfile, as *Assumptions) (*osutil.MountProfile, error) {
```
mvo5: The first argument changed name from "up" to "ctx" - but as its not used in this function this has no consequences.

zyga: yeah, this prepares everything for eventual rename to `upCtx` in the last patch in this series.

PR #6847, opened 2019-05-09
cmd/snap-update-ns: make apply{User,System}Fstab identical

zyga:
This mechanical step is done to ensure that subsequent changes
did not break the system. Note that the per-user mount profile update
logic now performs locking (which is no-op), loads the current profile
(which is empty), synthesizes additional filesystem assumptions based
on that (empty) profile (so no new assumptions). Lastly the updated
current profile is saved (which is a no-op).

In short, this makes user and system profile application use identical
code without changing anything in practice.

### DIFF
--- a/cmd/snap-update-ns/update.go
+++ b/cmd/snap-update-ns/update.go
@@ -48,9 +48,6 @@ func applySystemFstab(ctx MountProfileUpdateContext) error {
 	}
 	defer unlock()
 
-	// Read the desired and current mount profiles. Note that missing files
-	// count as empty profiles so that we can gracefully handle a mount
-	// interface connection/disconnection.
 	desired, err := ctx.LoadDesiredProfile()
 	if err != nil {
 		return err
@@ -62,6 +59,7 @@ func applySystemFstab(ctx MountProfileUpdateContext) error {
 		return err
 	}
 	debugShowProfile(currentBefore, "current mount profile (before applying changes)")
+
 	// Synthesize mount changes that were applied before for the purpose of the tmpfs detector.
 	as := ctx.Assumptions()
 	for _, entry := range currentBefore.Entries {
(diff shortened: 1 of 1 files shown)


cmd/snap-update-ns/update.go
```diff
@@ -77,21 +75,36 @@ func applySystemFstab(ctx MountProfileUpdateContext) error {
...
 	}
 	debugShowProfile(desired, "desired mount profile")
 
-	current, err := ctx.LoadCurrentProfile()
+	currentBefore, err := ctx.LoadCurrentProfile()
 	if err != nil {
 		return err
 	}
-	debugShowProfile(current, "current mount profile")
+	debugShowProfile(currentBefore, "current mount profile (before applying changes)")
```
mvo5: Should this say "current user mount profile..."?

zyga: No, because the whole point is to unify this into one function.

zyga:
@mvo5 I understand the frustration. This is branch series contains patches from https://github.com/zyga/snapd/commits/feature/user-mount-ns-20.9-split-rebased where the exact idea is spelled out. That branch is in turn a split of this original pull request with the whole changes and justifications https://github.com/snapcore/snapd/pull/6360.

The next patch immediately after this one is https://github.com/zyga/snapd/commit/a5ecb6bd2d277a7abd6f1f625b86cc7669bad79a which removes one of the duplicates. I was assuming you knew this context so I didn't provide it separately in the pull request.

On the upside this is almost over, we only have a pair of patches left, on the downside I think this is also justified as a way of proposing changes as there's no sensible way, in GitHub, to provide a path series and allow reviewers to look at individual patches. It doesn't work because those changes together are too big and discouraging (this was written many months ago) and also because GitHub UI is really tailored into "this is a diff of the entire pull reuqest" but not everything works well in that model, IMO.

PR #6848, opened 2019-05-09
many: introduce healthstate, run check-health post-(install/refresh/try/revert)

chipaca:
This change introduces `overlord/healthstate` and sets things up so that after installing / refreshing / trying / reverting a snap, the `check-health` hook is run if it exists.

This is the first step towards implementing [health checks](https://forum.snapcraft.io/t/rfc-health-checks/10605).


overlord/healthstate/healthstate.go
```diff
@@ -0,0 +1,163 @@
...
+		Message:   "hook failed",
+	})
+}
+
+func (h *healthHandler) appendHealth(health *HealthState) error {
+	st := h.context.State()
+	st.Lock()
+	defer st.Unlock()
+
+	var hs map[string]*HealthState
```
mvo5: (idle wonder) Should this be a list of HealthState? So that we can keep e.g. the last 2/3/5 or whatever of them?

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,143 @@
...
+
+Note the health is of the snap, not of the apps it contains; it’s up to the
+snap developer to determine how the health of the individual apps add up to
+the health of the snap as a whole.
+
+status can be one of
+
+- okay: the snap is healthy. This status takes no message and no code.
+
+- waiting: some resource the snap needs isn’t ready yet; the user just needs
```
mvo5: Maybe add something like: `some resource the snap needs isn’t ready yet;` -> `some resource the snap needs (e.g. an internet connection) isn’t ready yet;`

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,143 @@
...
+
+		// n is the number of runes
+		// j is the index of the 71st rune
+		var j, n int
+		for i, r := range c.Message {
+			if strutil.IsCtrl(r) {
+				return fmt.Errorf(`character %+q at position %d is not allowed`, r, i)
+			}
+			if n == 70 {
+				j = i
```
mvo5: Can we break here or am I misreading it?

strutil/strutil.go
```diff
@@ -209,3 +210,19 @@ func ElliptRight(str string, n int) string {
 	// this is expensive; look into a cheaper way maybe sometime
 	return string([]rune(str)[:n-1]) + "…"
 }
+
+// IsCtrl reports whether the rune is a control character.
+func IsCtrl(r rune) bool {
```
mvo5: Nice! I was initially wondering if we should be even more conservative and only allow [a-Z0-9+*-] but that is probably really overdoing it. Plus it won't allow the important: `¯\_(ツ)_/¯`.

overlord/healthstate/healthstate.go
chipaca: after discussing it with pedronis we agreed we can always add a `Previous *HealthState` if and when needed

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,143 @@
...
+status can be one of
+
+- okay: the snap is healthy. This status takes no message and no code.
+
+- waiting: some resource the snap needs isn’t ready yet; the user just needs
+  to wait.  The message (and optionally the code) must explain what it’s
+  waiting for.
+
+- blocked: the user needs to do something for the snap to do something; the
+  message (and optionally the) must say what it is needs doing.
```
stolowski: and optionally the code?

overlord/healthstate/healthstate.go
```diff
@@ -0,0 +1,163 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2017 Canonical Ltd
```
stolowski: year+=2

overlord/hookstate/hookmgr.go
```diff
@@ -302,6 +302,8 @@ func (m *HookManager) runHook(task *state.Task, tomb *tomb.Tomb, snapst *snapsta
 		}
 	}
 
+	// XXX: why not return nil at this point if !(hookExists || mustHijack) ?
```
stolowski: This is a very good question, I'll dig a bit.

overlord/hookstate/ctlcmd/health.go
chipaca: we could (although then we'd "accept" a message that contains bad chars as long as they are beyond a certain point in the message, which would make the issue harder to spot when developing). In any case I'm dropping this bit of validation in the next commit, and will add it back later -- keeps this pr minimal

overlord/hookstate/hookmgr.go
stolowski: Ok, I looked a bit on it and I *think* I *vaguely* (very vaguely) remember the idea behind this was to still allow our handler to do something even if there was no hook script. But I think that not being able to tell if a script was actually executed inside Done() handler is an oversight/limitation, so perhaps we should improve it.

mvo5:
Woah, lots of new stuff \o/ - should this go out of draft stage now?

chipaca:
This one is code-complete, I'm just waiting for its predecessor to land so
I can shrink the diff. I'm working on the next phase already.

overlord/healthstate/healthstate_test.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+func (s *healthSuite) TearDownTest(c *check.C) {
+	s.hookMgr.StopHooks()
+	s.se.Stop()
+	s.BaseTest.TearDownTest(c)
+}
+
+func (s *healthSuite) TestHealthNoHook(c *check.C) {
+	s.testHealth(c, false, false)
```
mvo5: I'm not a fan of these booleans as I need to look at how testHealth is defined to see what they do. Granted its super close by and has good names so I'm probably overly pedantic. I looked at this a bit and the only other suggestion I have is something like this:
```
--- a/overlord/healthstate/healthstate_test.go
+++ b/overlord/healthstate/healthstate_test.go
@@ -90,26 +90,37 @@ func (s *healthSuite) TearDownTest(c *check.C) {
 	s.BaseTest.TearDownTest(c)
 }
 
+type healthHookTestCondition int
+
+const (
+	noHook = iota
+	badHook
+	goodHook
+	captainHook
+)
+
 func (s *healthSuite) TestHealthNoHook(c *check.C) {
-	s.testHealth(c, false, false)
+	s.testHealth(c, noHook)
 }
 
 func (s *healthSuite) TestHealthFailingHook(c *check.C) {
-	s.testHealth(c, true, true)
+	s.testHealth(c, badHook)
 }
 
 func (s *healthSuite) TestHealth(c *check.C) {
-	s.testHealth(c, true, false)
+	s.testHealth(c, goodHook)
 }
 
-func (s *healthSuite) testHealth(c *check.C, withScript, failScript bool) {
+func (s *healthSuite) testHealth(c *check.C, cond healthHookTestCondition) {
 	var cmd *testutil.MockCmd
-	if failScript {
+	switch cond {
+	case badHook:
 		cmd = testutil.MockCommand(c, "snap", "exit 1")
-	} else {
+	default:
 		cmd = testutil.MockCommand(c, "snap", "exit 0")
 	}
-	if withScript {
+
+	if cond != noHook {
 		hookFn := filepath.Join(s.info.MountDir(), "meta", "hooks", "check-health")
 		c.Assert(os.MkdirAll(filepath.Dir(hookFn), 0755), check.IsNil)
 		// the hook won't actually be called, but needs to exist
@@ -150,19 +161,20 @@ func (s *healthSuite) testHealth(c *check.C, withScript, failScript bool) {
 	err = s.state.Get("health", &healths)
 	s.state.Unlock()
 
-	if failScript {
+	switch cond {
+	case badHook:
 		c.Assert(status, check.Equals, state.ErrorStatus)
-	} else {
+	default:
 		c.Assert(status, check.Equals, state.DoneStatus)
 	}
-	if withScript {
+	if cond != noHook {
 		c.Assert(err, check.IsNil)
 		c.Assert(healths, check.HasLen, 1)
 		c.Assert(healths["test-snap"], check.NotNil)
 		health := healths["test-snap"]
 		c.Check(health.Revision, check.Equals, snap.R(42))
 		c.Check(health.Status, check.Equals, healthstate.UnknownStatus)
-		if failScript {
+		if cond == badHook {
 			c.Check(health.Message, check.Equals, "hook failed")
 			c.Check(health.Code, check.Equals, "snapd-hook-failed")
 		} else {
```
But feel free to ignore.

overlord/healthstate/healthstate_test.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+func (*healthSuite) TestStatus(c *check.C) {
+	for i, str := range healthstate.KnownStatuses {
+		status, err := healthstate.StatusLookup(str)
+		c.Check(err, check.IsNil, check.Commentf("%v", str))
+		c.Check(status, check.Equals, healthstate.HealthStatus(i), check.Commentf("%v", str))
+		c.Check(healthstate.HealthStatus(i).String(), check.Equals, str, check.Commentf("%v", str))
+	}
+	status, err := healthstate.StatusLookup("rabbits")
```
mvo5: (nitpick^2) Might be worthwhile to split here into TestStatusUnhappy

overlord/hookstate/ctlcmd/health_test.go
```diff
@@ -0,0 +1,144 @@
...
+	c.Assert(err, check.IsNil)
+	s.mockContext = ctx
+}
+
+func (s *healthSuite) TestBadArgs(c *check.C) {
+	type tableT struct {
+		args []string
+		err  string
+	}
+	table := []tableT{
```
mvo5: Nice test!

overlord/hookstate/ctlcmd/health_test.go
```diff
@@ -0,0 +1,144 @@
...
+			[]string{"set-health", "blocked"},
+			`when status is not "okay", message is required`,
+		}, {
+			[]string{"set-health", "blocked", "message", "--code=xx"},
+			`code should have between 3 and 30 bytes, got 2`,
+		}, {
+			[]string{"set-health", "blocked", "message", "--code=abcdefghijklmnopqrstuvwxyz12345"},
+			`code should have between 3 and 30 bytes, got 31`,
+		}, {
+			[]string{"set-health", "blocked", "message", "--code=☠☢☣💣💢🐍✴👿‼"},
```
mvo5: hahahah

overlord/hookstate/ctlcmd/services_test.go
```diff
@@ -309,8 +350,8 @@ func (s *servicectlSuite) TestQueuedCommands(c *C) {
 	c.Assert(err, IsNil)
 	c.Check(installed, DeepEquals, []string{"one", "two"})
 	c.Assert(tts, HasLen, 2)
-	c.Assert(tts[0].Tasks(), HasLen, 13)
-	c.Assert(tts[1].Tasks(), HasLen, 13)
+	c.Assert(taskKinds(tts[0].Tasks()), DeepEquals, installTaskKinds)
```
mvo5: Nice!

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,130 @@
...
+
+	"github.com/snapcore/snapd/i18n"
+	"github.com/snapcore/snapd/overlord/healthstate"
+)
+
+var (
+	shortHealthHelp = i18n.G("Report on snap's health")
+	longHealthHelp  = i18n.G(`
+The set-health command can signal to the system and the user that something is
+not well with the snap.
```
pedronis: Given that if there's a hook one must use this, the emphasis only on negative situations seems unwarranted. We need a more neutral explanation I think.

Also this should probably mention the hook as a common usage place.

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,130 @@
...
+
+status can be one of
+
+- okay: the snap is healthy. This status takes no message and no code.
+
+- waiting: some resource (e.g. a device, network, or service) the snap needs
+  isn’t ready yet; the user just needs to wait.  The message (and optionally
+  the code) must explain what it’s waiting for.
+
+- blocked: the user needs to do something for the snap to do something; the
```
pedronis: "...the snap to proceed; ..."?

maybe @degville can look at this overall

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,130 @@
...
+  message (and optionally the code) must say what it is needs doing.
```
pedronis: "what it is that needs doing"?  "what needs doing"? "what is required"?

overlord/healthstate/healthstate.go
```diff
@@ -0,0 +1,164 @@
...
+
+	if err != nil {
+		if err != state.ErrNoState {
+			return err
+		}
+		health = HealthState{
+			Revision:  h.context.SnapRevision(),
+			Timestamp: time.Now(),
+			Status:    UnknownStatus,
+			Message:   "hook did not call set-health",
```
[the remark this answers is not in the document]
chipaca: I don't mind, but note this behaviour is what we [documented](https://forum.snapcraft.io/t/health-checks/10605),

> There is another status that can’t be set by the snap directly:
> 
> * `unknown` : the default, no hook provided or the hook did not call `snapctl set-health` (no `code` ), [...]

I think distinguishing 'no hook provided' from 'hook did not do the only thing it's expected to do' is sane, and will gladly change the documentation to reflect this.

How does `snapd-no-health-set` sound?

overlord/hookstate/ctlcmd/health.go
chipaca: How about
```
The set-health command is called from a hook (primarily the "check-health" hook)
to inform the system of the overall health of the snap.
```
?

overlord/hookstate/ctlcmd/health.go
chipaca: How about
```
- blocked: something needs doing to unblock the snap; the message must be
  sufficient to point the user in the right direction.
```
?

overlord/healthstate/healthstate.go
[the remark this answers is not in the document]
pedronis: snapd-hook-no-health-set,  it's clunkier but makes clearer that is the hook that didn't set it ?

overlord/hookstate/ctlcmd/health.go
pedronis: It doesn't need to be called from hook, though the main usage is like that, no?  can be called from the snap at any point, or is that not true?

overlord/hookstate/ctlcmd/health.go
pedronis: sounds good

overlord/hookstate/ctlcmd/health.go
chipaca: You're right, a snap could call `snapctl set-health` from an app directly and maybe that's something we want to encourage, even. How about
```
The set-health command is called from a snap to inform the system of the snap's
overall health.
```
?

overlord/healthstate/healthstate.go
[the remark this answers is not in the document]
chipaca: SGTM. OTOH if that sounds too much like it's talking about a `snapd-hook` (?) it could be `snapd-no-health-set-in-hook`.

overlord/hookstate/ctlcmd/health.go
pedronis: actually that doesn't work, the only hook  that works is check-health, because how setting health in state is split out, but we want "snapctl set-health" to work at any point, so we need to think about that

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return err
+	}
+	if status == healthstate.UnknownStatus {
+		return fmt.Errorf(`status cannot be manually set to "unknown"`)
+	}
+
+	if len(c.Code) > 0 {
+		if len(c.Code) < 3 || len(c.Code) > 30 {
+			return fmt.Errorf("code should have between 3 and 30 bytes, got %d", len(c.Code))
```
stolowski: s/bytes/characters/? "bytes" sounds a little bit too "technical".

overlord/healthstate/healthstate.go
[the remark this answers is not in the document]
chipaca: this is done now (`snapd-hook-no-health-set`), including the docs

overlord/hookstate/ctlcmd/health.go
chipaca: `snapctl set-health` now works from everywhere (hooks or apps). I'll be updating the help to the last one I suggested, in view of this. Shout if you want it different :-)

overlord/hookstate/ctlcmd/health.go
chipaca: Yes, I changed it to 'characters'. In a sense it's now wrong, because if you try to set code to `á` it will tell you you're trying to set it to something 2 characters long (instead of 1 character, or 2 bytes), but I think that's acceptable.

overlord/healthstate/healthstate.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+		hs = map[string]*HealthState{}
+	}
+	hs[ctx.InstanceName()] = health
+	st.Set("health", hs)
+
+	return nil
+}
+
+func SetHealth(ctx *hookstate.Context) error {
```
pedronis: this doesn't seem to be unit tested

pedronis: I would also call it SetHealthFromHookContext,  I would expect a different signature for something just called SetHealth, (it would take the HealthState directly I think)

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,142 @@
...
+	ctx.Lock()
+	defer ctx.Unlock()
+
+	var v struct{}
+
+	// if 'health' is there we've either already added an OnDone (and the
+	// following Set("health"), or we're in the set-health hook itself
+	// (which sets it to a dummy entry for this purpose).
+	if err := ctx.Get("health", &v); err == state.ErrNoState {
+		ctx.OnDone(func() error {
```
pedronis: +1

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,142 @@
...
+
+var (
+	shortHealthHelp = i18n.G("Report on snap's health")
+	longHealthHelp  = i18n.G(`
+The set-health command is called from a snap to inform the system of the snap's
+overall health.
+
+Note the health is of the snap, not of the apps it contains; it’s up to the
+snap developer to determine how the health of the individual apps add up to
+the health of the snap as a whole.
```
pedronis: didn't we discuss mentioning check-health explicitly here though?

overlord/healthstate/healthstate.go
chipaca: okay

overlord/hookstate/ctlcmd/health.go
chipaca: I don't remember that, but maybe we did. Can you point me at it?

pedronis: this comment mentions it:  https://github.com/snapcore/snapd/pull/6848#discussion_r288614070

> Also this should probably mention the hook as a common usage place.

chipaca: How about this, just before the Note:
> It can be called from any hook, and even from the apps themselves,
although a snap can provide a 'check-health' hook that is solely
concerned with this; if it does, that hook is called periodically
(with increased frequency while the snap is "unhealthy"). Any health
regression issues a warning to the user.

the last line would be untrue until we implemented that (I could add it then)

chipaca:
@pedronis addressed your points, pr is green and ready for re-review.

overlord/healthstate/healthstate.go
```diff
@@ -0,0 +1,195 @@
...
+			checkTimeout = to
+		} else {
+			logger.Debugf("cannot override check-health timeout: %v", err)
+		}
+	}
+
+	snapstate.HealthCheckHook = CheckHook
+}
+
+func CheckHook(st *state.State, snapName string, snapRev snap.Revision) *state.Task {
```
pedronis: should this just be Hook ?

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,147 @@
...
+	Code             string `long:"code" value-name:"<code>" description:"optional tool-friendly value representing the problem that makes the snap unhealthy.  Not a number, but a word with 3-30 bytes matching [a-z](-?[a-z0-9])+"`
+}
+
+var (
+	validCode = regexp.MustCompile(`^[a-z](?:-?[a-z0-9])+$`).MatchString
+)
+
+func (c *healthCommand) Execute([]string) error {
+	if c.Status == "okay" && (len(c.Message) > 0 || len(c.Code) > 0) {
+		return fmt.Errorf(`when status is "okay", message and code should be empty`)
```
pedronis: s/should/must/ ?

overlord/snapstate/snapstate.go
```diff
@@ -380,6 +384,10 @@ var SetupRemoveHook = func(st *state.State, snapName string) *state.Task {
 	panic("internal error: snapstate.SetupRemoveHook is unset")
 }
 
+var HealthCheckHook = func(st *state.State, snapName string, rev snap.Revision) *state.Task {
```
pedronis: should this be `CheckHealthHook` to reflect the actual hook name?

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,147 @@
...
+
+	"github.com/snapcore/snapd/i18n"
+	"github.com/snapcore/snapd/overlord/healthstate"
+	"github.com/snapcore/snapd/overlord/state"
+)
+
+var (
+	shortHealthHelp = i18n.G("Report on snap's health")
+	longHealthHelp  = i18n.G(`
+The set-health command is called from a snap to inform the system of the snap's
```
degville: maybe `The set-health command is called from *within* a snap to inform the system...` for greater accuracy?

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,147 @@
...
+overall health.
+
+It can be called from any hook, and even from the apps themselves, although a
+snap can provide a 'check-health' hook that is solely concerned with this; if it
+does, that hook is called periodically (with increased frequency while the snap
+is "unhealthy"). Any health regression will issue a warning to the user.
+
```
degville: I'd split the sentence:

It can be called from any hook, and even from the apps themselves. A snap can optionally provide a `check-health` hook to better manage these calls, which is then called periodically and with increased frequency when when a snap is "unhealthy".  Any health regression will issue a warning to the user.

(if you don't change this, there's a typo in the quotes around `check-health`)

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,147 @@
...
+Note the health is of the snap, not of the apps it contains; it’s up to the
```
degville: *Note the health...* sounds a little like an instruction. Maybe use a colon? ie:

Note: the reported health is of the snap only, and not of the apps it contains . The snap developer must determine how the health of individual apps is reflected in the overall health of the snap.

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,147 @@
...
+snap developer to determine how the health of the individual apps add up to
+the health of the snap as a whole.
+
+status can be one of
```
degville: maybe add a colon to stop this looking like an orphaned sentence?

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,147 @@
...
+
+- okay: the snap is healthy. This status takes no message and no code.
+
+- waiting: some resource (e.g. a device, network, or service) the snap needs
+  isn’t ready yet; the user just needs to wait.  The message must explain what
+  it’s waiting for.
```
degville: waiting: a resource needed by the snap (e.g. a device, network, or service)  is not ready and the user will need to wait. The message must explain which resource is being waited for.

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,147 @@
...
+
+- blocked: something needs doing to unblock the snap; the message must be
```
degville: could we add a small example for this (such as device, network and service above)? I'd find it useful.

overlord/hookstate/ctlcmd/health.go
```diff
@@ -0,0 +1,147 @@
...
+`)
+)
+
+func init() {
+	addCommand("set-health", shortHealthHelp, longHealthHelp, func() command { return &healthCommand{} })
+}
+
+type healthPositional struct {
+	Status  string `positional-arg-name:"<status>" required:"yes" description:"a valid health status; required."`
+	Message string `positional-arg-name:"<message>" description:"a short human-readable explanation of the status (when not okay). Must be longer than 7 characters, and will be truncated if over 70. Message cannot be provided if status is okay, and is required otherwise."`
```
degville: I'd swap out *and* at the end to make *but is required otherwise* .

chipaca:
@pedronis @degville thanks! I'll get right on those.

PR #6849, opened 2019-05-09
many: introduce a gadget helper for locating device matching given structure

bboozzoo:
Add a helper for locating a device that matches given structure, by inspecting the partition and filesystem labels.

Reshuffle some systemd specific code to strutil package.

cc @cmatsuoka

### DIFF
--- a/gadget/device.go
+++ b/gadget/device.go
@@ -87,8 +87,6 @@ func encodeLabel(in string) string {
 		switch {
 		case utf8.RuneLen(r) > 1:
 			buf.WriteRune(r)
-		case r == '\\':
-			fallthrough
 		case strings.IndexRune(allowed, r) == -1:
 			fmt.Fprintf(buf, `\x%x`, r)
 		default:
--- a/gadget/device.go
+++ b/gadget/device.go
@@ -1,0 +1,73 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+package gadget
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/snapcore/snapd/dirs"
+	"github.com/snapcore/snapd/osutil"
+	"github.com/snapcore/snapd/strutil"
+)
+
+var ErrDeviceNotFound = errors.New("device not found")
+
+// FindDeviceForStructure attempts to find an existing device matching given
+// volume structure, by inspecting it's name and, optionally, the filesystem
+// label. Assumes that the host's udev has set up device symlinks correctly.
+func FindDeviceForStructure(ps *PositionedStructure) (string, error) {
+	var candidates []string
+
+	if ps.Name != "" {
+		byPartlabel := filepath.Join(dirs.GlobalRootDir, "/dev/disk/by-partlabel/", strutil.HexEscapePath(ps.Name))
+		candidates = append(candidates, byPartlabel)
+	}
+
+	if ps.Label != "" {
+		byFsLabel := filepath.Join(dirs.GlobalRootDir, "/dev/disk/by-label/", strutil.HexEscapePath(ps.Label))
+		candidates = append(candidates, byFsLabel)
+	}
+
+	var found string
+	for _, candidate := range candidates {
+		if !osutil.FileExists(candidate) {
+			continue
+		}
+		target, err := os.Readlink(candidate)
+		if err != nil {
+			return "", fmt.Errorf("cannot read device link: %v", err)
+		}
+		if found != "" && target != found {
+			// partition label and filesystem label links point to
+			// different devices
+			return "", fmt.Errorf("conflicting device match, %q points to %q, while other match points to %q",
+				candidate, target, found)
+		}
+		found = target
+	}
+
+	if found == "" {
+		return "", ErrDeviceNotFound
+	}
+
+	return found, nil
+}
--- a/gadget/device.go
+++ b/gadget/device.go
@@ -32,7 +32,7 @@ import (
 var ErrDeviceNotFound = errors.New("device not found")
 
 // FindDeviceForStructure attempts to find an existing device matching given
-// volume structure, by inspecting it's name and, optionally, the filesystem
+// volume structure, by inspecting its name and, optionally, the filesystem
 // label. Assumes that the host's udev has set up device symlinks correctly.
 func FindDeviceForStructure(ps *PositionedStructure) (string, error) {
 	var candidates []string
--- a/gadget/device.go
+++ b/gadget/device.go
@@ -48,6 +48,7 @@ func FindDeviceForStructure(ps *PositionedStructure) (string, error) {
 	}
 
 	var found string
+	var match string
 	for _, candidate := range candidates {
 		if !osutil.FileExists(candidate) {
 			continue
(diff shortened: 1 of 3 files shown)


strutil/escape.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"bytes"
+	"fmt"
+	"path/filepath"
+	"strings"
+)
+
+const allowed = `:_.abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789`
+
+// HexEscapePath works like systemd-escape --path
```
cmatsuoka: There are a few corner cases where `HexEscapePath()` and `systemd-escape --path` behave differently, but unsure if this difference is relevant in this case. Examples: `.`, `a/..`, `a/../b`, `a/../../b` (generally `systemd-escape --path` returns an error for anything that goes back to `.`).

strutil/escape.go
```diff
@@ -0,0 +1,66 @@
...
+
+	// replace all special chars
+	for i := 0; i < len(in); i++ {
+		c := in[i]
+		if c == '/' {
+			buf.WriteByte('-')
+		} else if strings.IndexByte(allowed, c) >= 0 {
+			buf.WriteByte(c)
+		} else {
+			fmt.Fprintf(buf, `\x%x`, in[i])
```
cmatsuoka: We could use `c` here for consistency?

bboozzoo: Pushed

strutil/escape.go
bboozzoo: `systemd-escape` is really picky about dots. I looked at the use cases in current code, and the only one is mount unit names for snap mount points. Those paths should be safe and dot-less.

gadget/device.go
```diff
@@ -0,0 +1,73 @@
...
+			continue
+		}
+		target, err := os.Readlink(candidate)
+		if err != nil {
+			return "", fmt.Errorf("cannot read device link: %v", err)
+		}
+		if found != "" && target != found {
+			// partition label and filesystem label links point to
+			// different devices
+			return "", fmt.Errorf("conflicting device match, %q points to %q, while other match points to %q",
```
cmatsuoka: Perhaps we could say what the matches come from in the error message? (Like partlabel points to this, label points to that?)

gadget/device.go
```diff
@@ -0,0 +1,73 @@
...
+		candidates = append(candidates, byPartlabel)
+	}
+
+	if ps.Label != "" {
+		byFsLabel := filepath.Join(dirs.GlobalRootDir, "/dev/disk/by-label/", strutil.HexEscapePath(ps.Label))
+		candidates = append(candidates, byFsLabel)
+	}
+
+	var found string
+	for _, candidate := range candidates {
```
cmatsuoka: Initially I thought that building and iterating over a two-element slice is strange, but it seems that doing it in a single pass would involve a lot of ugly conditionals and the result wouldn't be more readable anyway.

gadget/device.go
bboozzoo: I've tweaked the error message to include the previous match. The path we tried already includes `by-label` or `by-partlabel` so it's clear what type of the property matched.

cmatsuoka: Right, what it is is encoded in the path, I missed that.

gadget/device_test.go
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package gadget_test
+
+import (
+	// "fmt"
```
stolowski: "fmt" can be removed.

gadget/device_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+type deviceSuite struct {
+	dir string
+}
+
+var _ = Suite(&deviceSuite{})
+
+func (d *deviceSuite) SetUpTest(c *C) {
+	d.dir = c.MkDir()
+	dirs.SetRootDir(d.dir)
```
stolowski: nitpick, can you add restore for this?

strutil/escape.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+const allowed = `:_.abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789`
+
+// HexEscapePath works like systemd-escape --path
+// FIXME: we could use github.com/coreos/go-systemd/unit/escape.go
+//        and EscapePath from it.
+//
+//        But thats not in the archive and it won't work with go1.3
+func HexEscapePath(in string) string {
```
stolowski: I'm not totally sure HexEscapePath belongs to strutil given that it implements systemd-specific mangling, but not going to block on that.

strutil/escape_test.go
```diff
@@ -0,0 +1,41 @@
...
+	. "gopkg.in/check.v1"
+
+	. "github.com/snapcore/snapd/strutil"
+)
+
+type escapeSuite struct{}
+
+var _ = Suite(&escapeSuite{})
+
+func (e *escapeSuite) TestEscape(c *C) {
```
stolowski: Thanks for adding this test!

strutil/escape.go
bboozzoo: Actually, I'm looking more into this and maybe I'll have to revert this change and write something myself. The escaping done by util-linux and systemd varies slightly for some edge cases.

bboozzoo: I'll force push and drop the patches to systemd/strutil.

gadget/device.go
```diff
@@ -73,3 +75,25 @@ func FindDeviceForStructure(ps *PositionedStructure) (string, error) {
...
+// encodeLabel encodes a name for use a partition or filesystem label symlink by
+// udev. The result matches the output of blkid_encode_string().
+func encodeLabel(in string) string {
+	const allowed = `#+-.:=@_abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789`
+
+	buf := &bytes.Buffer{}
+
+	for _, r := range in {
+		switch {
+		case utf8.RuneLen(r) > 1:
```
cmatsuoka: What happens if the label or partlabel uses multibyte characters with an encoding that's not utf-8 (or any invalid utf-8)?

bboozzoo: I think we do not have to care for other encodings. We're passing a string as input, which already is a valid utf-8 sequence. The source is a gadget YAML, which again, uses utf-8. If I understand the plumbing in systemd and udev, but the time any links are created by udev, everything is utf-8 only.

gadget/device.go
```diff
@@ -73,3 +75,25 @@ func FindDeviceForStructure(ps *PositionedStructure) (string, error) {
...
+			buf.WriteRune(r)
+		case r == '\\':
```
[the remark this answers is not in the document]
bboozzoo: Updated

gadget/device.go
cmatsuoka: A scenario that would potentially give us some problem is when you set your ext2/3/4 filesystem label to something containing utf-8 multibyte characters with a total byte size that's larger than the available space in such a way that it would be truncated to  16 bytes in the middle of a character code. When recovered, it could potentially lead to invalid utf-8. I didn't actually try this situation to see if something wrong happens, but as you mentioned before it would happen at recovery time when building the input string and woudn't even reach `encodeLabel`.

gadget/device.go
```diff
@@ -0,0 +1,97 @@
...
+		case utf8.RuneLen(r) > 1:
+			buf.WriteRune(r)
+		case strings.IndexRune(allowed, r) == -1:
+			fmt.Fprintf(buf, `\x%x`, r)
+		default:
+			buf.WriteRune(r)
+		}
+	}
+	return buf.String()
+}
```
stolowski: This looks fine.

gadget/device_test.go
```diff
@@ -0,0 +1,205 @@
...
+	err := os.MkdirAll(filepath.Join(d.dir, "/dev/disk/by-label"), 0755)
+	c.Assert(err, IsNil)
+	err = os.MkdirAll(filepath.Join(d.dir, "/dev/disk/by-partlabel"), 0755)
+	c.Assert(err, IsNil)
+	err = ioutil.WriteFile(filepath.Join(d.dir, "/dev/fakedevice"), []byte(""), 0644)
+	c.Assert(err, IsNil)
+}
+
+func (d *deviceSuite) TearDownTest(c *C) {
+	dirs.SetRootDir("/")
```
stolowski: Thanks.

gadget/device_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (d *deviceSuite) TestDeviceFindByStructureName(c *C) {
+	names := []struct {
+		escaped   string
+		structure string
+	}{
+		{"foo", "foo"},
+		{"123", "123"},
+		{"foo\\x20bar", "foo bar"},
+		{"foo#bar", "foo#bar"},
```
stolowski: Nitpick, I think this test (and the one below) could also check a multi-byte utf label case, for completeness.

PR #6850, opened 2019-05-09
gadget: add volume level update checks

bboozzoo:
Add volume level update sanity checks. We effectively assume there is no room for changes in the volume across updates. The rule may be lifted in the future, but currently, the number of structures, schema and ID must remain unchanged.

cc @cmatsuoka

### DIFF
--- a/gadget/gadget.go
+++ b/gadget/gadget.go
@@ -82,7 +82,7 @@ type Volume struct {
 
 func (v *Volume) EffectiveSchema() string {
 	if v.Schema == "" {
-		return "gpt"
+		return GPT
 	}
 	return v.Schema
 }
--- a/gadget/gadget.go
+++ b/gadget/gadget.go
@@ -80,6 +80,13 @@ type Volume struct {
 	Structure []VolumeStructure `yaml:"structure"`
 }
 
+func (v *Volume) EffectiveSchema() string {
+	if v.Schema == "" {
+		return "gpt"
+	}
+	return v.Schema
+}
+
 // VolumeStructure describes a single structure inside a volume. A structure can
 // represent a partition, Master Boot Record, or any other contiguous range
 // within the volume.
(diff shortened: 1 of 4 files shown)


gadget/gadget.go
```diff
@@ -80,6 +80,13 @@ type Volume struct {
 	Structure []VolumeStructure `yaml:"structure"`
 }
 
+func (v *Volume) EffectiveSchema() string {
```
zyga: Nice :)

gadget/gadget.go
```diff
@@ -80,6 +80,13 @@ type Volume struct {
...
+	if v.Schema == "" {
+		return "gpt"
```
zyga: Wasn't there a constant for that?

gadget/update.go
```diff
@@ -119,3 +119,16 @@ func canUpdateStructure(from *PositionedStructure, to *PositionedStructure) erro
...
 }
+
+func canUpdateVolume(from *PositionedVolume, to *PositionedVolume) error {
+	if from.ID != to.ID {
+		return fmt.Errorf("cannot change volume ID from %q to %q", from.ID, to.ID)
+	}
+	if from.EffectiveSchema() != to.EffectiveSchema() {
+		return fmt.Errorf("cannot change volume schema from %q to %q", from.EffectiveSchema(), to.EffectiveSchema())
+	}
+	if len(from.PositionedStructure) != len(to.PositionedStructure) {
```
zyga: Curious, why cannot we change the number of structures?

gadget/gadget.go
bboozzoo: Ah, right. The commit was added almost a month ago, way before the consts were added. I'll update the PR accordingly.

gadget/update.go
bboozzoo: To fully support changing the number of structure would mean we'd need to allow adding or removing partition on the disk during update. We may allow this in future, but not in the initial version.

PR #6860, opened 2019-05-10
tests: running tests on fedora 30

sergiocazzolato:
This branch is using the new fedora 30 image. 

Fedora 29 is set as manual as part of this PR as well


bboozzoo:
There's this bit in the logs, but it doesn't make any sense:
```
May 13 07:34:31 may130724-660556 snapd[24501]: main.go:123: system does not fully support snapd: cannot mount squashfs image using "squashfs": mount: /tmp/sanity-mountpoint-630627329: unknown filesystem type 'squashfs'.
```

bboozzoo:
@sergiocazzolato can you update the image? The latest kernel is kernel-core-5.0.13-300.fc30.x86_64, while the image is running kernel-core-5.0.9-301.fc30.x86_64. The kernel-modules package got installed, but only for the latest kernel.

sergiocazzolato:
@zyga @bboozzoo 
The error related to fedora 30 out of space I think it is related to these issue:

https://bugzilla.redhat.com/show_bug.cgi?id=1584461
https://bugzilla.redhat.com/show_bug.cgi?id=1612430

It is based on the mount output after executing some tests: 
https://paste.ubuntu.com/p/zc8jR9DGJT

bboozzoo:
#6874 will fix the problem with nsswitch.conf. 

The problem has not come up sooner, because /etc/nsswitch.conf is modified in our Fedora 29 images (this was probably true for Fedora 28 image too). Only by stroke of luck, Fedora 30 image has not been adjusted in the same way.

@sergiocazzolato any particular reason we edit /etc/nsswitch.conf ? I did not find anything in spread and our tests, so it's probably the image preparation scripts that introduce the change.

sergiocazzolato:
@bboozzoo we are not editing this file (/etc/nsswitch.conf) in the image creation scripts (since fedora 27 to 30), at least not directly. 

Perhaps installing dependencies or doing update/upgrade is editing that file. What I can do is go step by step of the script until I detect what is happening with that file. 

Thanks for the feedback

bboozzoo:
Investigated the problem a bit more. I started with `Fedora-Cloud-Base-29-1.2.x86_64.qcow2` image. `/etc/nsswitch.conf` is a symlink. Ran `dnf upgrade -y`, and this is what I got:
```
[guest@localhost ~]$ ls -l /etc/nsswitch.conf*
-rw-r--r--. 1 root root 2371 May 21 05:41 /etc/nsswitch.conf
lrwxrwxrwx. 1 root root   29 Oct 24  2018 /etc/nsswitch.conf.bak -> /etc/authselect/nsswitch.conf
```
Perhaps it's  prroblem with the image or some package upgrade is breaking things. Found a some reports of issues like this happening: https://bugzilla.redhat.com/show_bug.cgi?id=1622272 and https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/PNKKVG3K6WAU42CCPVIEV6LZY7PWUG4P/

bboozzoo:
The remaining SELinux denial is addressed in #6881

PR #6873, opened 2019-05-16
gadget: improve device lookup, add helper for mount point lookup

bboozzoo:
Add a helper for a filesystem structure mount point lookup. Take a live core18 system, for a structure `EFI System` defined like this:
```
volumes:
  pc:
    bootloader: grub
   ...
      - name: EFI System
        type: EF,C12A7328-F81F-11D2-BA4B-00A0C93EC93B
        filesystem: vfat
        filesystem-label: system-boot
        size: 50M
        content:
          - source: grubx64.efi
            target: EFI/boot/grubx64.efi
          - source: shim.efi.signed
            target: EFI/boot/bootx64.efi
          - source: grub.cfg
            target: EFI/ubuntu/grub.cfg
```
we should be able to idenfity `/boot/efi` as its mount point.

Improve the lookup of matching devices. Although `os.FileExists()` already does `os.Stat()` and that will catch symlinks pointing nowhere, try the belt and suspenders approach to be on the safe side should somebody tweak the implementation.

cc @cmatsuoka


gadget/device_linux.go
```diff
@@ -95,3 +100,42 @@ func encodeLabel(in string) string {
...
+		return "", err
+	}
+
+	var mountPoint string
+	mountInfo, err := osutil.LoadMountInfo(filepath.Join(dirs.GlobalRootDir, osutil.ProcSelfMountInfo))
+	if err != nil {
+		return "", fmt.Errorf("cannot read mount info: %v", err)
+	}
+	for _, entry := range mountInfo {
+		if entry.MountSource != devpath || entry.FsType != ps.Filesystem {
```
zyga: I would add a small snippet like this here:
```
if mountPoint != "" && (mountPoint == entry.MountDir || strings.HasPrefix(mountPoint, entry.MountDir + "/")) {
  // mount entry clobbers visibility to previously discovered match
  mountPoint = "" // clobbered
}
```

bboozzoo: There's a `break` couple of lines down, so the loop will finish once a match is found. This makes me think, I should restructure the loop bit to make the lookup easier to read.

bboozzoo: Pushed an update, should be clearer now.

PR #6874, opened 2019-05-16
cmd/snap-confine: do not mount over non files/directories

bboozzoo:
When bringing entries in /etc back from the desired rootfs, make sure that those
are either regular files or directories, and do not attempt to mount over
anything that is not a file or directory.

Otherwise we can break the host mount namespace.

This happens on Fedora, where /etc/nsswitch.conf is a symlink to
/etc/authselect/nsswitch.conf. Mounting over it breaks the host and the
following entries appear:
```
/var/lib/snapd/snaps/core_6818.snap on /etc/authselect/nsswitch.conf type squashfs (ro,nodev,relatime,context=system_u:object_r:snappy_snap_t:s0)
/var/lib/snapd/snaps/core_6818.snap on /etc/authselect/nsswitch.conf type squashfs (ro,nodev,relatime,context=system_u:object_r:snappy_snap_t:s0)
/var/lib/snapd/snaps/core_6818.snap on /etc/authselect/nsswitch.conf type squashfs (ro,nodev,relatime,context=system_u:object_r:snappy_snap_t:s0)
```
At this point, the file in the host mount namespace will also have the SELinux
label from the snap's mount context.

This seems to go back to Fedora 28. See: https://bugzilla.redhat.com/show_bug.cgi?id=1584461

### DIFF
--- a/cmd/snap-confine/mount-support.c
+++ b/cmd/snap-confine/mount-support.c
@@ -324,30 +324,29 @@ static void sc_bootstrap_mount_namespace(const struct sc_mount_config *config)
 					continue;
 				}
 				die("cannot stat %s from desired rootfs", src);
-			} else {
-				if (!S_ISREG(src_stat.st_mode)
-				    && !S_ISDIR(src_stat.st_mode)) {
-					debug
-					    ("entry %s from the desired rootfs is not a file or directory, skipping mount",
-					     src);
-					continue;
-				}
 			}
+			if (!S_ISREG(src_stat.st_mode)
+			    && !S_ISDIR(src_stat.st_mode)) {
+				debug
+				    ("entry %s from the desired rootfs is not a file or directory, skipping mount",
+				     src);
+				continue;
+			}
+
 			if (lstat(dst, &dst_stat) != 0) {
 				if (errno == ENOENT) {
 					continue;
 				}
 				die("cannot stat %s from host", src);
-			} else {
-				if (!S_ISREG(dst_stat.st_mode)
-				    && !S_ISDIR(dst_stat.st_mode)) {
-					debug
-					    ("entry %s from the host is not a file or directory, skipping mount",
-					     src);
-					continue;
-				}
-
 			}
+			if (!S_ISREG(dst_stat.st_mode)
+			    && !S_ISDIR(dst_stat.st_mode)) {
+				debug
+				    ("entry %s from the host is not a file or directory, skipping mount",
+				     src);
+				continue;
+			}
+
 			if ((dst_stat.st_mode & S_IFMT) !=
 			    (src_stat.st_mode & S_IFMT)) {
 				debug
--- a/cmd/snap-confine/mount-support.c
+++ b/cmd/snap-confine/mount-support.c
@@ -310,14 +310,29 @@ static void sc_bootstrap_mount_namespace(const struct sc_mount_config *config)
 		};
 		for (const char **dirs = dirs_from_core; *dirs != NULL; dirs++) {
 			const char *dir = *dirs;
-			struct stat buf;
 			if (access(dir, F_OK) == 0) {
+				struct stat dst_stat;
+				struct stat src_stat;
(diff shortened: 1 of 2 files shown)


cmd/snap-confine/mount-support.c
```diff
@@ -310,20 +310,55 @@ static void sc_bootstrap_mount_namespace(const struct sc_mount_config *config)
...
+			sc_must_snprintf(src, sizeof src, "%s%s",
+					 config->rootfs_dir, dir);
+			sc_must_snprintf(dst, sizeof dst, "%s%s",
+					 scratch_dir, dir);
+			if (lstat(src, &src_stat) != 0) {
+				if (errno == ENOENT) {
+					continue;
+				}
+				die("cannot stat %s from desired rootfs", src);
+			} else {
```
zyga: You don't need the else part, do you? It just increases indent without any semantic change since you either continue in the loop or die in the if-part above.

cmd/snap-confine/mount-support.c
```diff
@@ -310,20 +310,55 @@ static void sc_bootstrap_mount_namespace(const struct sc_mount_config *config)
...
+					     src);
+					continue;
+				}
+			}
+			if (lstat(dst, &dst_stat) != 0) {
+				if (errno == ENOENT) {
+					continue;
 				}
+				die("cannot stat %s from host", src);
+			} else {
```
zyga: Same remark about else as above.

bboozzoo:
Added a spread test

bboozzoo:
This landed a bit early. @jdstrand can you take a look at the changes in case there is a follow up needed?

cmd/snap-confine/mount-support.c
```diff
@@ -310,20 +310,54 @@ static void sc_bootstrap_mount_namespace(const struct sc_mount_config *config)
...
+			struct stat src_stat;
+			sc_must_snprintf(src, sizeof src, "%s%s",
+					 config->rootfs_dir, dir);
+			sc_must_snprintf(dst, sizeof dst, "%s%s",
+					 scratch_dir, dir);
+			if (lstat(src, &src_stat) != 0) {
+				if (errno == ENOENT) {
+					continue;
+				}
+				die("cannot stat %s from desired rootfs", src);
```
jdstrand: I'm not saying this is a problem, but it is a change in behavior. Before we would have silently ignored this error and moved on. This seems more correct, but I wonder if we'll see any regressions here....

cmd/snap-confine/mount-support.c
```diff
@@ -310,20 +310,54 @@ static void sc_bootstrap_mount_namespace(const struct sc_mount_config *config)
...
+
+			if ((dst_stat.st_mode & S_IFMT) !=
+			    (src_stat.st_mode & S_IFMT)) {
+				debug
+				    ("entries %s and %s are of different types, skipping mount",
+				     dst, src);
+				continue;
 			}
+			// both source and destination exist and are either files or
+			// directories
```
jdstrand: This comment probably should've read:
```
// both source and destination exist where both are either files
// or both are directories
```

Not a blocker.

tests/regression/rhbz-1584461/task.yaml
```diff
@@ -0,0 +1,46 @@
...
+        rm /etc/ssl
+        mv /etc/ssl-backup /etc/ssl
+    fi
+
+    rm -f ./in-snap ./on-host
+
+execute: |
+    # on Fedora, /etc/nsswitch.conf is a symlink to /etc/authselect/nsswitch.conf
+    # TODO: uncomment once we fix this in spread images
+    # test -s /etc/nsswitch.conf
```
jdstrand: Did you mean 'test -h /etc/nsswitch.conf'?

bboozzoo: Heh, yes! Thanks for spotting this.

bboozzoo:
@jdstrand thanks for the review. I've opened #6916 with the fixes.

PR #6876, opened 2019-05-20
interfaces/optical-drive: add scsi-generic type 4 and 5 support

diddledan:
ref: https://forum.snapcraft.io/t/openat-returning-eperm-for-dev-sg0/10069/

This PR is to add scsi_generic block devices to the optical-media interface. The udev rule specifically includes ONLY optical media devices. I have confirmed that no other scsi_generic devices on my system are exposed via this modification. The pre-existing `write` configuration also applies to the scsi_generic `/dev/sg*` devices exposed to the snap. By default the scsi_generic device(s) are readonly via apparmor.

@jdstrand could you take a look, please?


interfaces/builtin/optical_drive.go
```diff
@@ -46,13 +46,15 @@ const opticalDriveConnectedPlugAppArmor = `
...
 /dev/scd[0-9]* r,
+/dev/sg[0-9]* r,
 @{PROC}/sys/dev/cdrom/info r,
 /run/udev/data/b11:[0-9]* r,
 `
 
 var opticalDriveConnectedPlugUDev = []string{
 	`KERNEL=="sr[0-9]*"`,
 	`KERNEL=="scd[0-9]*"`,
+	`SUBSYSTEM=="scsi_generic", SUBSYSTEMS=="scsi", ATTRS{type}=="4|5"`,
```
mvo5: Might be nice to have a comment somewhere what the values of "4|5" means in this context.

diddledan: I plucked those out of the default rules file in `/lib/udev/rules.d/50-udev-default.rules` from the rule that assigns the group of `cdrom` to appropriate devices. According to https://en.wikipedia.org/wiki/SCSI_Peripheral_Device_Type the types are:

4. Write-once device
5. CD/DVD-ROM device

I can add those definitions into the source as a comment if it would help.

diddledan: We might get away with dropping type `4`?

mvo5: Thanks for adding this comment !

interfaces/builtin/optical_drive.go
```diff
@@ -46,13 +46,18 @@ const opticalDriveConnectedPlugAppArmor = `
 # Allow read access to optical drives
 /dev/sr[0-9]* r,
 /dev/scd[0-9]* r,
+/dev/sg[0-9]* r,
```
jdstrand: This isn't necessarily just optical drives, so adding a clarifying comment would be good:
```
# allow all generic scsi devices here and use the device cgroup to
# differentiate optical drives
/dev/sg[0-9]* r,
```

interfaces/builtin/optical_drive.go
```diff
@@ -46,13 +46,18 @@ const opticalDriveConnectedPlugAppArmor = `
...
 /run/udev/data/b11:[0-9]* r,
 `
 
 var opticalDriveConnectedPlugUDev = []string{
 	`KERNEL=="sr[0-9]*"`,
 	`KERNEL=="scd[0-9]*"`,
+	// ATTRS{type} below takes scsi peripheral device types.
+	// Type 4 is 'Write-once device'; type 5 is 'CD/DVD-ROM device'
+	// ref: https://en.wikipedia.org/wiki/SCSI_Peripheral_Device_Type
+	`SUBSYSTEM=="scsi_generic", SUBSYSTEMS=="scsi", ATTRS{type}=="4|5"`,
```
jdstrand: This is quite nice. As for getting rid of type 4, I think it is ok to leave. Type 4 is defined as 'write-once, read many' (aka WORM) and there are historical cases of optical WORM devices (eg, https://www.computerhistory.org/collections/catalog/102671161), though the original WORM devices were punch cards. :)

diddledan:
The travis-ci failure seems to be "The job exceeded the maximum log length, and has been terminated."

mvo5:
@diddledan Yeah, looks unrelated, we will take care of this, sometimes there are network issues in the cloud or similar that lead to this kind of unexpected timeouts.

anonymouse64:
The most recent test failure is now on OS X:
```
Installing Homebrew Packages
7.14s$ rvm $brew_ruby do brew bundle --verbose --global
/usr/local/bin/brew tap homebrew/bundle
==> Tapping homebrew/bundle
Cloning into '/usr/local/Homebrew/Library/Taps/homebrew/homebrew-bundle'...
remote: Enumerating objects: 81, done.
remote: Counting objects: 100% (81/81), done.
remote: Compressing objects: 100% (74/74), done.
remote: Total 81 (delta 7), reused 26 (delta 4), pack-reused 0
Unpacking objects: 100% (81/81), done.
Tapped 0 formulae (172 files, 244.5KB)
Error: Your Homebrew is outdated. Please run `brew update`.
Error: Kernel.exit
```

Doesn't seem related to anything setup in the tests so I'm hoping it's a transient thing with Travis and will go away upon restarting again

mvo5:
Indeed, sorry for letting this hanging for so long - I merged master which will fix the osx issue and hopefully other spurious issues.

PR #6879, opened 2019-05-20
[RFC] overlord: add context.Context to snapstate.Install()

mvo5:
This is a very mechanical PR - it just adds a context to the snapstate.Install(). If this looks sane I can do the same for InstallMany()  and possible more. Once we have a context there we can use this to add the `Snap-Client-User-Agent` to install operations.

(the changes for InstallMany are similar in https://github.com/snapcore/snapd/commit/edefff5ec423ef14bd6dfb857b85b63e94edc2ad)


bboozzoo:
Conflicts.

mvo5:
Given that there is no feedback on this approach I will not resolve the conflicts but instead simply close the PR and redo it once we have a clear path forward.

PR #6880, opened 2019-05-20
daemon: refactor user ops to api_users

chipaca:
This is a pure refactor: it moves login/logout/create user/list users
out to api_users.go.  It also moves the test for create/list to
api_users_test.go; the moving of the login/logout tests is left for
later (it's fiddly and intertwined), as is making the tests sit in the
daemon_test package.


chipaca:
there's another PR where we make the users api conditional, which is also
needed before we go back to Claudio's user work

pedronis:
as I mentioned instead of waiting for conflicts either side, is probably best to reapply as a new commit the changes in #6834 here

PR #6882, opened 2019-05-21
cmd: add snap-verify stub binary (UC20)

mvo5:
We will need a snap-verify binary in snapd that is able to validate
snaps from e.g. the recovery partition using a set of assertions
from a given directory.

This PR adds a stub implementation of the tool and makes the
packaging aware of it. For now it will only be available in
the Ubuntu packaging as it is a rather specialized tool. This
may change in the future.

Unless people feel this is ok on master this can go into the new
"20" branch of snapd and we can start building a "20" track for
snapd (or a branch).


packaging/opensuse/snapd.spec
```diff
@@ -393,6 +393,8 @@ fi
 %{_libexecdir}/snapd/snap-device-helper
 %{_libexecdir}/snapd/snap-discard-ns
 %{_libexecdir}/snapd/snap-exec
+# Not making snap-verify available as its specialized for Ubuntu Core systems
```
zyga: This is incorrect because it is illegal to comment-away %{...} expressions. Just leave this out since both Fedora and openSUSE will fail to build when there are installed files that are not motioned by %files.

packaging/snapd.mk
```diff
@@ -86,6 +86,7 @@ install:: snap | $(DESTDIR)$(bindir)
 	install -m 755 $^ $|
 
 # Install snapctl snapd, snap-{exec,update-ns,seccomp} into /usr/lib/snapd/
+# (but not snap-verify)
```
zyga: There's a fragment for core systems (though we're not using this in Debian yet) that you could extend as well.

packaging/ubuntu-14.04/snapd.install
```diff
@@ -7,6 +7,7 @@ usr/bin/snap-failure /usr/lib/snapd/
 usr/bin/snap-update-ns /usr/lib/snapd/
 usr/bin/snapd /usr/lib/snapd/
 usr/bin/snap-seccomp /usr/lib/snapd/
+usr/bin/snap-verify /usr/lib/snapd/
```
zyga: I was not expecting to see it on 14.04

PR #6883, opened 2019-05-21
tests: fix how strings are matched on auto-refresh-retry test

sergiocazzolato:
The problem was that sometimes the first MATCH was matching when the
catalog refresh failed to refresh, so then the second match failed to
match the whole string.

This was matched initially using MATCH "state ensure error", but it was
not related to the auto-refresh:

May 21 12:48:57 may201007-857826 snapd[30902]: catalogrefresh.go:94:
DEBUG: Catalog refresh failed: decode new commands catalog: net/http:
request canceled (Client.Timeout exceeded while reading body).
May 21 12:48:57 may201007-857826 snapd[30902]: stateengine.go:102: state
ensure error: decode new commands catalog: net/http: request canceled
(Client.Timeout exceeded while reading body)

Also was fixed the restore and the last MATCH is done using
get_journalctl_log because if we use the journalctl command it brings
the whole journal and it was creating a huge log.

Full test (error) log: https://paste.ubuntu.com/p/MHByKPsPJH/


tests/lib/pkgdb.sh
```diff
@@ -441,6 +441,12 @@ distro_install_build_snapd(){
         apt install -y --only-upgrade snapd
         mv sources.list.back /etc/apt/sources.list
         apt update
+
+        # In case the PPA does not have an updated snapd the apt install snapd call will not
+        # error and we only test the distro snapd. So we make apt show to double check the
+        # snapd package comes from a ppa
+        apt show snapd | grep "APT-Sources: http.*ppa.launchpad.net"
```
mvo5: This looks good - but also unrelated to this PR? Did this accidentally slip into this PR?

tests/main/auto-refresh-retry/task.yaml
```diff
@@ -7,6 +7,7 @@ prepare: |
 
 restore: |
     rm -f /etc/systemd/system/snapd.service.d/override.conf
+    ip netns delete testns || true
```
stolowski: Thanks!

tests/main/auto-refresh-retry/task.yaml
```diff
@@ -41,15 +42,15 @@ execute: |
...
+        if get_journalctl_log | MATCH "state ensure error: persistent network error"; then
             break
         fi
         echo "Ensure refresh"
         snap debug ensure-state-soon
         sleep 5
     done
 
-    journalctl | MATCH "state ensure error: persistent network error"
+    get_journalctl_log | MATCH "state ensure error: persistent network error"
```
stolowski: I remember having issues with this test and get_journal_log (would fail sometimes on some systems even though there was a matching line in the logs - perhaps something with log cursors).

tests/lib/pkgdb.sh
sergiocazzolato: yes it is accidental, It is already on master, perhaps it was introduced on the initial marge.
But I think is shouldn't affect due to this is already merged on master https://github.com/snapcore/snapd/blob/master/tests/lib/pkgdb.sh#L448

tests/main/auto-refresh-retry/task.yaml
sergiocazzolato: @stolowski, thanks for the review, in this case we shouldnt make journalctl becuase it brings the whole history, and that string could match with something that happend on another test which ran before.

tests/lib/pkgdb.sh
mvo5: When I expand the diff to see more context I see that this ppa is added under ` if [ "$SRU_VALIDATION" = "1" ]; then`. When expanding below I see ` elif [ -n "$PPA_VALIDATION_NAME" ]; then` and the PPA check is there too. So it seems like this will add it again in the wrong place (in the SRU_VALIDATION path)? Or am I misreading this?

sergiocazzolato: Well what you found is that the check was duplicated and in the incorrect place. I just deleted that and if you look few lines below you will see the same check but in the ppa section (which is correct).
Thanks for finding this.

PR #6888, opened 2019-05-22
tests: add mountinfo-tool

zyga:
During testing, when working with mount namespaces, we often need to ask
questions about the property of a specific mount point. Traditionally
the findmnt tool can be used for very nice, graphical exploration of the
mount namespaces. This tool has two problematic properties:

1) It doesn't show mount propagation settings with detail required for
testing. Specifically it uses custom, though arguably more readable,
vocabulary when describing propagation settings. It would be nicer to be
able to show kernel language rather than the custom language, especially
when cross-referencing kernel documentation. In addition it doesn't show
the essential property of peer group number, which is invaluable when
writing precise tests.

2) It doesn't offer unique features useful when testing across wide
range of machines. Specifically details like /dev/sda vs /dev/vda, which
are different but irrelevant to the test at hand need to be unified in
post-processing. Similarly, when tests want to measure expected
properties of a mount namespace, they will see a lot of variability
resulting from dynamic allocation of peer group numbers, block device
numbers, loopback device numbers or snap revisions. The new tool offers
a convenient option to deterministically re-number and re-name
appropriate things so that test output can be entirely deterministic if
desired.

This small python tool addresses the two shortcomings. It has convenient
syntax useful for scripting. Basic documentation and is short enough for
anyone to dive in and adjust if necessary.

It has two shortcomings that I would like to address separately:
 - doesn't have unit tests
 - it doesn't support Python 3.5 or earlier

At this time the tool is mypy and flake8 clean, though no CI is
established around that. I believe it can be addressed when unit tests
are added to the tool.


tests/lib/mountinfo-tool
```diff
@@ -0,0 +1,379 @@
...
+
+    @property
+    def major(self) -> int:
+        """major is the higher 16 bits of the device number."""
+        return self >> 16
+
+    @property
+    def minor(self) -> int:
+        """minor is the lower 16 bits of the device number."""
+        return self & ((1 << 16) - 1)
```
bboozzoo: or simply `self & 0xffff` without being too fancy

tests/lib/mountinfo-tool
```diff
@@ -0,0 +1,379 @@
...
+    """
+    Device is a device number with major and minor components.
+
+    Note that this class does not attempt to mimic peculiar
+    encoding used by the Linux kernel.
+    """
+
+    @classmethod
+    def pack(cls, major: int, minor: int) -> 'Device':
+        return cls((major << 16) | (minor & (1 << 16) - 1))
```
bboozzoo: Why going back to int if you're handled a `(maj, min)` tuple already? Would `Device(Tuple[int,int])` work?

tests/lib/mountinfo-tool
```diff
@@ -0,0 +1,379 @@
...
+        renumber_opt_fields(entry, seen_opt_fields)
+        renumber_loop_devices(entry, seen_loops)
+
+
+def rewrite_rename(entries: List[MountInfoEntry]) -> None:
+    """rewrite_renameapplies all re-naming helpers."""
+    # TODO: allocate devices like everything else above.
+    for entry in entries:
+        entry.mount_source = re.sub(
+            '/dev/[sv]d([a-z])', '/dev/sd\\1', entry.mount_source)
```
bboozzoo: Should be good enough for spread tests and VMs

tests/lib/mountinfo-tool
zyga: well... yeah though I don't think this is a blocker :)

tests/lib/mountinfo-tool
zyga: Because `Device` inherits from `int`. It is not an object with a dict but a real int with custom methods.

tests/lib/mountinfo-tool
zyga: Yeah, it's written based on experience :)

tests/lib/bin/mountinfo-tool
```diff
@@ -0,0 +1,379 @@
...
+    mount_opts:     options applying to the mount point only
+    sb_opts:        options applying to the mounted filesystem
+    opt_fields:     optional fields, used for propagation information
+    mount_id:       mount point identifier
+    parent_id:      identifier of parent mount point
+    dev:            major:minor numbers of the mounted device
+    root_dir:       subtree of the mounted filesystem exposed at mount_point
+    """, formatter_class=RawTextHelpFormatter)
+    parser.add_argument('-v', '--version', action="version", version="1.0")
+    parser.add_argument(
```
sergiocazzolato: I tested that with just -f and it is not using the default value:
> ./mountinfo-tool -f
usage: mountinfo-tool [-h] [-v] [-f MOUNTINFO] [--one] [--renumber] [--rename]
                      [EXPRESSION [EXPRESSION ...]]
mountinfo-tool: error: argument -f: expected one argument

zyga: The `-f` option accepts a file name to use instead of `/proc/self/mountinfo`, normally you can just skip it. Alternatively you can use `-f-` to parse stdin.

PR #6889, opened 2019-05-22
tests/main/auto-refresh-retry: fix loop/break condition

chipaca:
auto-refresh-retry checked for a condition in a limited loop, and on
exiting the loop checked the condition again. Unfortunately those two
conditions were different, so in some cases could break the test.

While I was there I limited it to only check snapd's journal instead
of the whole thing, and to only use MATCH on exit of the loop, so the
test output wasn't so needlessly big.


chipaca:
#6883 fixes the same issue and a bit more

PR #6892, opened 2019-05-22
tests: reuse the image created initially for nested tests execution

sergiocazzolato:
The idea of this change is to speedup the execution of the nested test
suite by reusing the base image for all the tests and creating the
instances with the snapshot feature of qemu.

Also it is included a change to move common parts of the prepare and
restore to the spread.yaml

Test results:
https://paste.ubuntu.com/p/kBqdSFWm9v/
https://paste.ubuntu.com/p/Q25bKnYDDN/


tests/lib/nested.sh
```diff
@@ -127,72 +128,73 @@ refresh_to_new_core(){
...
         -machine accel=kvm"
+
     if ! wait_for_ssh; then
         systemctl restart nested-vm
     fi
+
+    if wait_for_ssh; then
+        prepare_ssh
+    else
+        echo "ssh not stablished, exiting..."
```
chipaca: *established*, or *stabilised*?

sergiocazzolato: established :)

tests/lib/nested.sh
```diff
@@ -127,72 +128,73 @@ refresh_to_new_core(){
...
     fi
 }
 
 start_nested_core_vm(){
-    local QEMU
+    local IMAGE QEMU
     QEMU=$(get_qemu_for_nested_vm)
-    systemd_create_and_start_unit nested-vm "${QEMU} -m 2048 -nographic \
+    # As core18 systems use to fail to start the assetion disk when using the
+    # snapshot feature, we copy the original image and use that copy to start
```
stolowski: This sounds a bit weird, snapshots should be completely transparent to the guest system, what was the error you experienced?

bboozzoo:
Conflicts. Please merge master.

tests/lib/nested.sh
sergiocazzolato: Thanks for reviewing!! 
I thought the same but using snapshot I was happening that the vm was not working well with the user assertion, so the image vm was starting but the user assertion was discarded and console-conf was triggered instead. 
I was researching that isssue but couldn't find nothing useful, if you have any idea about which could be the reason please tell me.

bboozzoo: That's weird, I'm using snapshots locally all the time, although I'm normally passing an additional image with cloud-init config, rather than a user assertion. No issues observed with core18 images. Let me try the figure out how do generate an assertion locally and run that.

bboozzoo: So I built the image, and passed an assertions disk to qemu as well. I could see that snapd.autoimport.service complains about unknown account key:
```
-- Logs begin at Wed 2019-06-12 08:10:55 UTC, end at Wed 2019-06-12 08:11:38 UTC. --
Jun 12 08:10:57 localhost systemd[1]: Starting Auto import assertions from block devices...
Jun 12 08:10:57 localhost snap[781]: cmd_auto_import.go:142: error: cannot import /run/snapd/auto-import/7k87yxVV-aCE6l1RYK6z11i1HPy0v57GH4arAJcD_50fTfjmKcnbOVcJlWGdG9eo.assert: assert failed: cannot find account-key (0u-IKZyVlTxwiBVOv7oT1FNKdfkKJyOSwbEWJPurlvtaGZxX6VrMHSESnIPeaNMC)
Jun 12 08:10:57 localhost systemd[1]: Started Auto import assertions from block devices.
```
I've build snapd with test keys enabled, but maybe I should have used a different model?

sergiocazzolato: @the account-key has to be the same you used to sign the model. So you will need to create the new model with your account-key and sign the model doing: 
snap sign  -k <key-name>  < <name>.model.json > <name>.model

sergiocazzolato:
@bboozzoo could you please take a look to the latest changes?

PR #6894, opened 2019-05-23
tests: add "not" command as replacement for "!" in tests

zyga:
We've recently learned that shell's behaviour regarding "set -e" and "!
command" is rather unexpected. To preserve the ability to fail execution when a
command succeeds, when we were expecting it to fail, we can introduce a new
executable, "not" that acts like the "!" operator, but is exempt from shell's
logic as to when "set -e" is effective.

All spread tests are adjusted with the following shell command:

    find -name 'task.yaml' -exec sed -i -e 's/^\( \+\)!/\1not/g' {} \;

The remaining hits on the use of leading "!" in other scripts were adjusted
manually.


zyga:
This needs some more work:

- `not` needs to be a shell function so that it can be used to negate shell functions
- `not` needs to be placed in the correct spot in pipe chains, e.g. `echo foo bar | not grep foo` vs `not echo foo bar | grep foo`

PR #6902, opened 2019-05-23
spread.yaml: make HOST: usage shellcheck-clean

zyga:
spread-shellcheck is about to perform rudimentary checks of environment
variables. With this is has complained about spread.yaml's own
environment section, full of HOST: shell calls that lacked some serious
number of double quotes.

Signed-off-by: Zygmunt Krynicki <me@zygoon.pl>

PR #6903, opened 2019-05-23
spread-shellcheck: add support for variants and environment

zyga:
Spread support test variants where the environment section can be used
to convey specific variant information. In terms of YAML it looks like
this:

    environment:
        key/A: variant-a
        key/B: variant-b
    execute: |
        echo "$key"

Here key/A and key/B define the value of the environment variable "key"
and depending on the execution variant, the values will be "variant-a"
and "variant-b" respectively.

Currently spread-shellcheck would complain that the variable "$key" is
unset. With this patch it will know the definition of some of the
variant values. This is suboptimal but allows spread-shellcheck to not
complain about something that is clearly not broken.

As a future extension the checker could re-check the script with each
variant, carrying the exact values each time, in the hope that constant
propagation in shellcheck can make use of that information.

### DIFF
--- a/spread-shellcheck
+++ b/spread-shellcheck
@@ -118,6 +118,10 @@ def checksection(data, env: Dict[str, str]):
         # quotes so that one value can refer to another value.
         if '"' in value:
             value = value.replace('"', '\"')
+        # converts
+        # FOO: "$(HOST: echo $foo)"     -> FOO="$(echo $foo)"
+        # FOO: "$(HOST: echo \"$foo\")" -> FOO="$(echo \"$foo\")"
+        # FOO: "foo"                    -> FOO="foo"
         script_data.append("{}=\"{}\"".format(key, value))
         script_data.append("export {}".format(key))
     script_data.append(data)
--- a/spread-shellcheck
+++ b/spread-shellcheck
@@ -20,6 +20,7 @@ import subprocess
 import argparse
 from concurrent.futures import ThreadPoolExecutor
 from multiprocessing import cpu_count
+from typing import Dict
 
 import yaml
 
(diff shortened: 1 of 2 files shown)


zyga:
This is stacked on top of https://github.com/snapcore/snapd/pull/6902

spread-shellcheck
```diff
@@ -101,15 +102,30 @@ class ShellcheckFailures(Exception):
...
+        # Unpack the special "$(HOST: ...) syntax and tell shellcheck not to
+        # worry about the use of echo to print variable value.
+        if value.startswith("$(HOST:") and value.endswith(")"):
+            script_data.append("# shellcheck disable=SC2116")
+            value = "$({})".format(value[len("$(HOST:"):-1])
+        # XXX: poor man's shell key=value assignment with values in double
+        # quotes so that one value can refer to another value.
+        if '"' in value:
+            value = value.replace('"', '\"')
+        script_data.append("{}=\"{}\"".format(key, value))
```
bboozzoo: Could use an example in the comments, eg:
```
# converts
# FOO: "$(HOST: echo $foo)"     -> FOO="$(echo $foo)"
# FOO: "$(HOST: echo \"$foo\")" -> FOO="$(echo \"$foo\")"
# FOO: "foo"                    -> FOO="foo"
```

spread.yaml
```diff
@@ -3,7 +3,7 @@ project: snapd
 environment:
     GOHOME: /home/gopath
     GOPATH: $GOHOME
-    REUSE_PROJECT: "$(HOST: echo $REUSE_PROJECT)"
+    REUSE_PROJECT: "$(HOST: echo \"$REUSE_PROJECT\")"
```
bboozzoo: `\"` looks kinda ugly, I think this would work with:
```
REUSE_PROJECT: '$(HOST: echo "$REUSE_PROJECT")'
```

zyga: I wasn't aware that YAML had both type of strings. If that works +1 for changing it.

zyga: +1, I'll open a separate PR for that though.

zyga: Actually, it's here now.

spread-shellcheck
```diff
@@ -120,14 +140,21 @@ def checkfile(path):
...
     errors = ShellcheckError(path)
-
+    # TODO: handle stacking of environment from other places that influence it:
+    # spread.yaml -> global env + backend env + suite env -> task.yaml (task
+    # env + variant env).
+    env = {}
+    for key, value in data.get("environment", {}).items():
+        if "/" in key:
+            # TODO: re-check with each variant's value set.
+            key = key.split('/', 1)[0]
```
stolowski: Looks ok.

spread-shellcheck
```diff
@@ -101,15 +102,34 @@ class ShellcheckFailures(Exception):
...
     # spread shell snippets are executed under 'set -e' shell, make sure
     # shellcheck knows about that
-    data = 'set -e\n' + data
+    script_data = []
+    script_data.append('set -e')
+    for key, value in env.items():
+        value = str(value)
+        # Unpack the special "$(HOST: ...) syntax and tell shellcheck not to
+        # worry about the use of echo to print variable value.
+        if value.startswith("$(HOST:") and value.endswith(")"):
```
stolowski: +0.85 on this matching and magling, seems a bit delicate. Can we just disable SC2116 for entire environment block unconditionally? This shellcheck warning seems mildly useful in environment block anyway, no?

zyga: About delicate: yeah, if we had spread support for emitting the effective script for a given test, we'd not have to use this hackery.

About SC2116, I don't know of a way to disable warnings for the entire block and I also don't know what you mean by block here. It's all one long shell script, with environment emitted up front.

The problem about this warning is that `spread.yaml` is coded to use `echo "$FOO"` because those have to be executable commands. When they are placed inside `$(...)` and used just for assignment then spread-shellcheck effectively sees, in many cases, `foo="$(echo "$bar")"` and says this can be simplified to just `foo="$bar"`, rightfully so.

This however is just an artefact of how we use `$(HOST: ...)` syntax and perform spellchecking **and** is not universally true because some parts of our code use actual logic to decide what to print.

stolowski: Ok, fair enough, thanks for explanation. What I meant by block was all the environment we iterate and output within this loop. Fair point with $(HOST:..) vs everything else.

PR #6904, opened 2019-05-23
timings: always store ensure timings as long as they have an associated change

stolowski:
Always store root "Timings" object for ensure timings, even if it has no nested timings (either because they were not created, or because they were too fast), as long as it has "change-id" set, meaning a related change was created. Requiring "change-id" tag to be present means we will not store unnecessary ensure timings that had no change, e.g. because of erroring out.

This is needed to find related changes when using `snap debug timings --ensure=xyz` syntax (the "ensure" timing keeps related "change-id" in its tags, and it's the only way to correlate things) - it may currently fail to find an ensure if it was too fast, and in case of "become-operational" it won't find it at all because that ensure loop never creates nested timings.

### DIFF
--- a/timings/state.go
+++ b/timings/state.go
@@ -65,19 +65,29 @@ var timeDuration = func(start, end time.Time) time.Duration {
 // flatten flattens nested measurements into a single list within rootTimingJson.NestedTimings
 // and calculates total duration.
 func (t *Timings) flatten() interface{} {
-	if len(t.timings) == 0 {
+	var hasChangeID, hasTaskID bool
+	if t.tags != nil {
+		_, hasChangeID = t.tags["change-id"]
+		_, hasTaskID = t.tags["task-id"]
+	}
+	// store the root timing object even if it has no nested timings as long as it has a "change-id" tag
+	// but not a "task-id" tag - this signifies an "ensure" timing which we always want to record as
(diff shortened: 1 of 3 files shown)


timings/state.go
```diff
@@ -65,19 +65,29 @@ var timeDuration = func(start, end time.Time) time.Duration {
...
-	if len(t.timings) == 0 {
+	var hasChangeID, hasTaskID bool
+	if t.tags != nil {
+		_, hasChangeID = t.tags["change-id"]
+		_, hasTaskID = t.tags["task-id"]
+	}
+	// store the root timing object even if it has no nested timings as long as it has a "change-id" tag
+	// but not a "task-id" tag - this signifies an "ensure" timing which we always want to record as
+	// long as it created a change.
+	if len(t.timings) == 0 && (hasTaskID || !hasChangeID) {
```
bboozzoo: Maybe it'd be a bit easier to follow the comment if the code read:
```
// ensure timings which created a change, have the corresponding
// change-id tag, but no task-id
isEnsureWithChange := hasChangeID && !hasTaskID

// store the root timing object ...
if len(t.timings) == 0 && !isEnsureWithChange {
    return nil
}
```

stolowski: Addressed in #6920

PR #6910, opened 2019-05-24
spread.yaml: use "snap connections" in debug

zyga:
The debug-each section calls "snap interfaces" which has rather quirky
output. We now have a much better way to display that via "snap
connections", let's use it.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -437,8 +437,8 @@ debug-each: |
         esac
         echo '# seccomp denials (kills) '
         dmesg --ctime | grep type=1326 || true
-        echo '# snap interfaces'
-        snap interfaces || true
+        echo '# snap connections'
+        snap connections || true
         echo '# tasks executed on system'
         cat "$RUNTIME_STATE_PATH/runs" || true
         echo '# free space'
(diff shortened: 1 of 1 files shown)


spread.yaml
```diff
@@ -437,8 +437,8 @@ debug-each: |
         esac
         echo '# seccomp denials (kills) '
         dmesg --ctime | grep type=1326 || true
-        echo '# snap interfaces'
-        snap interfaces || true
+        echo '# snap connections'
+        snap connections || true
```
bboozzoo: `snap connections --all`

zyga:
```suggestion
        snap connections --all || true
```

spread.yaml
```diff
@@ -437,8 +437,8 @@ debug-each: |
         esac
         echo '# seccomp denials (kills) '
         dmesg --ctime | grep type=1326 || true
-        echo '# snap interfaces'
-        snap interfaces || true
+        echo '# snap connections'
```
zyga:
```suggestion
        echo '# snap connections --all'
```

PR #6913, opened 2019-05-24
tests: force snap-reexec test to fail to reproduce the error on ubuntu 14.04

sergiocazzolato:


sergiocazzolato:
This is an exemple of the output when snap-confine-from-core fails after running snap-set test
https://paste.ubuntu.com/p/t7qMJ86PDq/

sergiocazzolato:
Error not being reproduced last week, this PR will be reopen if needed

PR #6916, opened 2019-05-27
cmd/snap-confine, tests: tweak comments, reenable symlink check in RHBZ 1584461 regression

bboozzoo:
Based on comments from @jdstrand in #6874

### DIFF
--- a/tests/regression/rhbz-1584461/task.yaml
+++ b/tests/regression/rhbz-1584461/task.yaml
@@ -20,7 +20,7 @@ restore: |
 
 execute: |
     # on Fedora, /etc/nsswitch.conf is a symlink to /etc/authselect/nsswitch.conf
-    test -h /etc/nsswitch.conf
+    test -L /etc/nsswitch.conf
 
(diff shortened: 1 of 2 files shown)


tests/regression/rhbz-1584461/task.yaml
```diff
@@ -20,8 +20,7 @@ restore: |
 
 execute: |
     # on Fedora, /etc/nsswitch.conf is a symlink to /etc/authselect/nsswitch.conf
-    # TODO: uncomment once we fix this in spread images
-    # test -s /etc/nsswitch.conf
+    test -h /etc/nsswitch.conf
```
mvo5: Silly question - why `test -h` instead of `test -L` ? Mostly asking because 'L' -> 'link' in my brain but 'h' -> 'manpage'. Anyway, not a real issue, mostly curious.

bboozzoo: Hahah, I can update it to -L

PR #6918, opened 2019-05-27
snaptest: add helper for mocking snap with contents

bboozzoo:
Add a helper for mocking an installed snap with given contents.

### DIFF
--- a/snap/snaptest/snaptest.go
+++ b/snap/snaptest/snaptest.go
@@ -154,6 +154,18 @@ func MockInvalidInfo(c *check.C, yamlText string, sideInfo *snap.SideInfo) *snap
 	return snapInfo
 }
 
+// MockSnapWithFiles does the same as MockSnap, but also populates the snap
+// directory with given content
+//
+// The caller is responsible for mocking root directory with dirs.SetRootDir()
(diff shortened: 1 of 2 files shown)


bboozzoo:
It's for #6750

snap/snaptest/snaptest.go
```diff
@@ -154,6 +154,18 @@ func MockInvalidInfo(c *check.C, yamlText string, sideInfo *snap.SideInfo) *snap
 	return snapInfo
 }
 
+// MockSnapWithFiles does the same as MockSnap, but also populates the snap
+// directory with given content
+//
+// The caller is responsible for mocking root directory with dirs.SetRootDir()
+//and for altering the overlord state if required.
+func MockSnapWithFiles(c *check.C, yamlText string, si *snap.SideInfo, files [][]string) *snap.Info {
```
stolowski: Note this will create a snap that will likely fail snap checks (e.g. due to missing exec bits on apps). That's probably fine for many tests that avoid snap checks though.

bboozzoo: Yes, that's intentional. Once, there's a need we can switch the internal implementation to use osutil.EnsureDirState()

PR #6919, opened 2019-05-27
cmd/okay: Remove err message when warning file not exist

ardaguclu:
cmd/okay: Remove err message when warning file not exist

if previously there is no warning message and hence there is no warning file;

`snap okay` returns 

> error: no client-side warning timestamp found: cannot open timestamp file: open /home/user/.snap/warnings.json: no such file or directory

Checking if file exists and if not, returning nothing can be more appropriate.

### DIFF
--- a/cmd/snap/cmd_warnings.go
+++ b/cmd/snap/cmd_warnings.go
@@ -138,7 +138,7 @@ func (cmd *cmdOkay) Execute(args []string) error {
 
 	last, err := lastWarningTimestamp()
 	if err != nil {
-		return fmt.Errorf("no client-side warning timestamp found: %v", err)
+		return fmt.Errorf("%v", err)
 	}
 
 	return cmd.client.Okay(last)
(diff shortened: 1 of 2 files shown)


chipaca:
Also it might be an error to do a `snap okay` more than a certain time after `snap warnings`. I'm not sure how long that time should be though: a day is too long between one and t'other, but an hour might be too short (read the warnings, go to have lunch, come back and carry on). Maybe @pedronis has input on this.

ardaguclu:
If there is no other concern about the change, I could change the warning message if the warning file does not exist like you said.

chipaca:
On second reading, another issue is that there's a race in the code, and you could just check for IsNotExist on the os.Open and avoid it

ardaguclu:
Thank you for you feedback, I will make the necessary changes. I thought that testing failed because I sent pr without executing formatting tool. @chipaca

ardaguclu:
I added changes in this pull request referring the issue below;

[(https://bugs.launchpad.net/snapd/+bug/1794006)](url)

If there is no network connection. HTTP requests may return no such host or service misbehaving. These errors also should return DNSError message.

chipaca:
Please don't glom separate issues into a single pull request...

ardaguclu:
I will revert second issue from this pull request.

cmd/snap/cmd_warnings.go
```diff
@@ -138,7 +138,7 @@ func (cmd *cmdOkay) Execute(args []string) error {
 
 	last, err := lastWarningTimestamp()
 	if err != nil {
-		return fmt.Errorf("no client-side warning timestamp found: %v", err)
+		return fmt.Errorf("%v", err)
```
chipaca: a nit (for the followup): this should probably be just `return err`

PR #6924, opened 2019-05-28
packaging/fedora: force external linker to ensure static linking and -extldflags use

bboozzoo:
Go toolchain will use external linking only under specfiic conditions, such as
enabling PIE buildmode, using code that depends on CGO or an GOARCH/GOOS
unsupported by the internal linker.

Make sure that Go binaries are always using the external linker. This ensures
that external flags defined by distro policy, or package (such as -static) are
actually used.

Fixes the problem with non-statically built snapctl in #6759 on CentOS. For that particular combination, the PIE buildmode is not enabled as libc was built without -fPIC. In this case, go tool link falls back to internal linker and -extldflags (including -static) are ignored. The resulting binary is dynamically linked.

Add checks to make sure that binaries we need statically built are such.

cc @Conan-Kudo

### DIFF
--- a/packaging/fedora/snapd.spec
+++ b/packaging/fedora/snapd.spec
@@ -63,15 +63,17 @@
 
 # Until we have a way to add more extldflags to gobuild macro...
 %if 0%{?fedora}
+# buildmode PIE triggers external linker consumes -extldflags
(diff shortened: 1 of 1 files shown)


packaging/fedora/snapd.spec
```diff
@@ -675,6 +677,10 @@ sort -u -o devel.file-list devel.file-list
 %endif
 
 %check
+for binary in snap-exec snap-update-ns snapctl; do
+    ldd bin/$binary | grep 'not a dynamic executable'
+done
```
Conan-Kudo: Is this supposed to _not_ fail it?

bboozzoo: Yes, ldd output is like:
```
$ ldd bin/snapct
        not a dynamic executable
```

PR #6926, opened 2019-05-29
tests: stop using ! for naive negation in shell scripts

zyga:
This branch contains several patches that together eliminate all use of 
`! foo...` in shell commands. There are several distinct commits, each with their
respective commit message, that explain how the replacements were performed.
The branch also reverts the patch that uses stable shellcheck, so that we get
the extra test analysis. The branch also contains executable documentation
that show how `!` and `not` behave in practice.

### DIFF
--- a/tests/main/install-remove-multi/task.yaml
+++ b/tests/main/install-remove-multi/task.yaml
@@ -18,4 +18,4 @@ execute: |
     test -e /var/lib/snapd/desktop/applications/basic-desktop_io.snapcraft.echoecho.desktop
     echo "Removing a snap with a desktop file removes the desktop file again"
     snap remove  basic-desktop
-    ! test -e /var/lib/snapd/desktop/applications/basic-desktop_io.snapcraft.echoecho.desktop
+    not test -e /var/lib/snapd/desktop/applications/basic-desktop_io.snapcraft.echoecho.desktop
--- a/tests/main/interfaces-process-control/task.yaml
+++ b/tests/main/interfaces-process-control/task.yaml
@@ -33,7 +33,7 @@ execute: |
     ps ax | grep -Pq "^ *$pid"
     process-control-consumer.signal "SIGTERM" $pid
     #shellcheck disable=SC2009
-    ! ps ax | grep -Pq "^ *$pid"
+    ps ax | not grep -Pq "^ *$pid"
 
     if [ "$(snap debug confinement)" = partial ] ; then
         exit 0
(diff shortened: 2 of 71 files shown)


tests/main/interfaces-process-control/task.yaml
```diff
@@ -33,7 +33,7 @@ execute: |
     ps ax | grep -Pq "^ *$pid"
     process-control-consumer.signal "SIGTERM" $pid
     #shellcheck disable=SC2009
-    ! ps ax | grep -Pq "^ *$pid"
+    ps ax | not grep -Pq "^ *$pid"
```
bboozzoo: Bit on a fence about which version is more readable. This or  `not ( ps ax | grep -Pq "^ *$pid" )`. Feels like `.. | not ..` is not as visible.

tests/regression/lp-1797556/task.yaml
```diff
@@ -19,9 +19,9 @@ execute: |
     test ! -e /home/test/bin/evil-1
     test ! -e /home/test/bin/evil-2
     if [ "$(snap debug confinement)" = "strict" ]; then
-        ! su -l -c 'test-snapd-sh.with-home-plug -c "touch /home/test/bin/evil-1"' test 2>&1 | MATCH '.* Permission denied'
+        su -l -c 'test-snapd-sh.with-home-plug -c "touch /home/test/bin/evil-1"' test 2>&1 | MATCH '.* Permission denied'
```
bboozzoo: Was this one an actual bug?

zyga: Yeah, this was failing but it didn’t break the test.

tests/main/interfaces-process-control/task.yaml
zyga: I think it requires some training to read it this way but I grew to like the fact that “not” is coupled with the command that actually fails.

zyga: I think it requires some training to read it this way but I grew to like the fact that “not” is coupled with the command that actually fails.

tests/main/install-remove-multi/task.yaml
```diff
@@ -18,4 +18,4 @@ execute: |
     test -e /var/lib/snapd/desktop/applications/basic-desktop_io.snapcraft.echoecho.desktop
     echo "Removing a snap with a desktop file removes the desktop file again"
     snap remove  basic-desktop
-    ! test -e /var/lib/snapd/desktop/applications/basic-desktop_io.snapcraft.echoecho.desktop
+    not test -e /var/lib/snapd/desktop/applications/basic-desktop_io.snapcraft.echoecho.desktop
```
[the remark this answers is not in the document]
zyga: Indeed, this could become “test ! ...”. I’m happy to do a pass to make everything consistent next.

PR #6928, opened 2019-05-29
packaging: fix build-depends on powerpc

mvo5:
We do not support powerpc anymore and just build an empty package
there. Powerpc never worked because we never had a core snap for
powerpc so this is not a regression.

However right now powerpc fails to build because it cannot find
the right golang build-depends. This PR fixes this by making sure
we:
a) only build using golang (not gccgo)
b) excluding this build-dep on powerpc

There is also a pointer to the work my Michael Hudson-Doyle on
runc which took this idea and polished it with a debconf prompt:

https://github.com/tianon/debian-runc/compare/70957b315f82170dc2ab7085d39c23835c0fa996...xenial

But we don't need this debconf prompt.


mvo5:
Hrm, it looks like this needs slightly more work - the PPA at least is rejecting completely empty deb packages. So we need to put some sort of README.Debian into the package on powerpc sadly to make this fully work.

PR #6930, opened 2019-05-29
tests: make more robust how we check the journal log is ready to start a test

sergiocazzolato:
This is done because some tests (random) are failing to detect when we do:
>  echo "Running test: $marker" | systemd-cat -t snapd-test

This is mostly happening with the opensuse systemd (242) version.

### DIFF
--- a/tests/lib/journalctl.sh
+++ b/tests/lib/journalctl.sh
@@ -24,10 +24,14 @@ start_new_journalctl_log(){
 
 check_journalctl_ready(){
     marker="test-${RANDOM}${RANDOM}"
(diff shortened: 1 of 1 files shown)


sergiocazzolato:
@zyga, thanks for taking a look to this one, this is happening with the last version of the systemd 242 just on opensuse. And what I see is that is only failing when we use systemd-cat.

The solution is not nice, I know :(, I was researching and there is not any error raised related to this. 

I try to dig a bit more on another solution, any suggestions?

sergiocazzolato:
This PR was replaced with the #6989 and will be closed.

PR #6933, opened 2019-05-29
[RFC] snapd: ensure GOMAXPROCS is at least 2

mvo5:
This is a request from our friends from Rigardo. We should
measure it with the nice new tooling we have before merging.
Maybe @stolowski can help with that? The key to measure
is the firstboot seeding on a single core device (e.g. by
forcing "maxcpus=1 " in the kernel commandline on a PI
or in qemu).

The Go scheduler by default has a single operating system
thread per processor core, and does it's own goroutine
scheduling inside of that thread.  For I/O operations that
the Go runtime knows about, it has mechanisms to reschedule
goroutines so the system thread isn't blocked waiting for
I/O.  If a goroutine performs a blocking system call which
the go runtime doesn't have special optimizations for, the
system thread can become blocked waiting for the syscall.
This can dramatically reduce runtime performance, and the
problem is much worse on single processor systems because
there is normally only a single system thread.

We workaround by increasing the armound of procs to a
minimum of two.

### DIFF
--- a/cmd/snapd/main.go
+++ b/cmd/snapd/main.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"runtime"
 	"syscall"
 	"time"
 
@@ -52,6 +53,24 @@ func init() {
 
 func main() {
 	cmd.ExecInSnapdOrCoreSnap()
+
+	// The Go scheduler by default has a single operating system
+	// thread per processor core, and does it's own goroutine
+	// scheduling inside of that thread.  For I/O operations that
+	// the Go runtime knows about, it has mechanisms to reschedule
+	// goroutines so the system thread isn't blocked waiting for
+	// I/O.  If a goroutine performs a blocking system call which
+	// the go runtime doesn't have special optimizations for, the
+	// system thread can become blocked waiting for the syscall.
+	// This can dramatically reduce runtime performance, and the
+	// problem is much worse on single processor systems because
+	// there is normally only a single system thread.
+	//
+	// We workaround by increasing the armound of procs to a
+	// minimum of two.
+	if runtime.GOMAXPROCS(-1) == 1 {
+		runtime.GOMAXPROCS(2)
+	}
 
 	ch := make(chan os.Signal, 2)
 	signal.Notify(ch, syscall.SIGINT, syscall.SIGTERM)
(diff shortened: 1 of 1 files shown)


cmd/snapd/main.go
```diff
@@ -53,6 +54,24 @@ func init() {
...
+	// the Go runtime knows about, it has mechanisms to reschedule
+	// goroutines so the system thread isn't blocked waiting for
+	// I/O.  If a goroutine performs a blocking system call which
+	// the go runtime doesn't have special optimizations for, the
+	// system thread can become blocked waiting for the syscall.
+	// This can dramatically reduce runtime performance, and the
+	// problem is much worse on single processor systems because
+	// there is normally only a single system thread.
+	//
+	// We workaround by increasing the armound of procs to a
```
bboozzoo: s/armound/number/

stolowski:
Happy to help, will mesaure firstboot seeding with current code on pi3 and maxcpus=1.

stolowski:
@mvo Here are the timings with maxcpus=1 & current master on pi3:

https://pastebin.ubuntu.com/p/K3hBfggFhn/

(many thanks to @bboozzoo for crash-course on u-boot ;))

chipaca:
@stolowski I think the intent was to measure both before and after =)

stolowski:
@mvo I got through a few hoops to get custom snapd from this branch installed with edge, it eventually failed on console-conf & add user. I think it will be easiest to simply land this PR, then I'll just re-test and collect new timings with new edge image when available. We can easily revert this PR if the result is not satisfying.

cmd/snapd/main.go
chipaca:
```suggestion
	// We workaround by increasing the number of procs to a
```

cmd/snapd/main.go
```diff
@@ -53,6 +54,24 @@ func init() {
 func main() {
 	cmd.ExecInSnapdOrCoreSnap()
 
+	// The Go scheduler by default has a single operating system
+	// thread per processor core, and does it's own goroutine
```
chipaca:
```suggestion
	// thread per processor core, and does its own goroutine
```

chipaca:
@stolowski measure away

stolowski:
Here are the results for the version with the patch (2.39.1+git703.g8999873) - afict there is no improvement, in fact the new version scored slightly lower than before-the-patch version, but probably just random fluctuations.

https://pastebin.ubuntu.com/p/b25v2JdYtp/

mvo5:
Thanks! I will let Rigardo know that this is now in edge and what our results are. They can do their own measurement with it too and if there is no improvements we simply revert it again.

PR #6935, opened 2019-05-30
cmd/snap: support snap debug timings --startup=.. and measure loadState time

stolowski:
Add --startup=.. option for snap debug timings and enable measurements of loadState execution.


cmd/snap/cmd_debug_timings.go
```diff
@@ -152,6 +165,9 @@ func (x *cmdChangeTimings) Execute(args []string) error {
 		return ErrExtraArgs
 	}
 
+	if x.StartupTag != "" && (x.Positional.ID != "" || x.EnsureTag != "") {
```
chipaca: I'd say, either have one single "these three options are mutually exclusive" error, or have one for each invalid combination. Otherwise it's a bit weird.

While I'm at it I'll point out that the error when you specify no options is a bit misleading / frustrating, as it comes exclusively from `GetChangeID`,

    ~$ snap debug timings 
    error: please provide change ID or type with --last=<type>

(no mention of `--statup`/`--ensure`)

cmd/snap/cmd_debug_timings_test.go
```diff
@@ -39,6 +39,9 @@ var timingsTests = []timingsCmdArgs{{
 }, {
 	args:  "debug timings --ensure=foo 9",
 	error: "cannot use 'ensure' and change id together",
+}, {
+	args:  "debug timings --ensure=foo --startup=bar",
+	error: "cannot use 'startup' with 'ensure' or change id",
```
chipaca: interestingly you don't have a test for startup + change id :-)

cmd/snap/cmd_debug_timings.go
```diff
@@ -177,7 +194,7 @@ func (x *cmdChangeTimings) Execute(args []string) error {
 	} else {
 		allEnsures = "false"
 	}
-	if err := x.client.DebugGet("change-timings", &timings, map[string]string{"change-id": chgid, "ensure": x.EnsureTag, "all": allEnsures}); err != nil {
+	if err := x.client.DebugGet("change-timings", &timings, map[string]string{"change-id": chgid, "ensure": x.EnsureTag, "all": allEnsures, "startup": x.StartupTag}); err != nil {
```
chipaca: could you use a struct? just so you can have `omitempty` all over the place i mean

cmd/snap/cmd_debug_timings.go
```diff
@@ -32,9 +32,10 @@ import (
 
 type cmdChangeTimings struct {
 	changeIDMixin
-	EnsureTag string `long:"ensure"`
-	All       bool   `long:"all"`
-	Verbose   bool   `long:"verbose"`
+	EnsureTag  string `long:"ensure"`
+	All        bool   `long:"all"`
+	StartupTag string `long:"startup"`
```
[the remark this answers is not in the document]
stolowski: Good idea, done.

cmd/snap/cmd_debug_timings.go
stolowski: Fair point, that's improved now.

cmd/snap/cmd_debug_timings_test.go
stolowski: Good catch, added.

cmd/snap/cmd_debug_timings.go
stolowski: That's a good observation, however `DebugGet` is used in other places and I wanted to avoid changing it with this PR.

PR #6937, opened 2019-05-30
cmd/snap-update-ns: detach unused mount points

zyga:
When snap-update-ns determines that a mount point is no longer needed
and should be unmounted it really tried to unmount things. This was
working because usually there were no nested mount structures present.

We were alerted that a particular layout would fail to refresh. The
layout in question is simply:

    layout:
      $SNAP/bar/mount:
	bind: $SNAP/foo
      $SNAP/baz:
	bind: $SNAP/bar

That is:
 - $SNAP/bar/mount is a view into $SNAP/foo
 - $SNAP/baz is a view into $SNAP/bar
    (which contains a view into SNAP/foo)

Since layout elements use $SNAP, which contains the revision number,
each time such layout is refreshed it will really be re-created as the
old paths will no longer be used. This will create a situation where
the mount point $SNAP/baz cannot be unmounted because there's
$SNAP/baz/mount is a mount point. This can be seen by looking at the
output of snap-update-ns debug log:

    desired mount profile:
    	/snap/layout-content-test/x2/foo /snap/layout-content-test/x2/bar/mount none rbind,rw,x-snapd.origin=layout 0 0
    	/snap/layout-content-test/x2/bar /snap/layout-content-test/x2/baz none rbind,rw,x-snapd.origin=layout 0 0
    current mount profile (before applying changes):
    	/snap/layout-content-test/x1/foo /snap/layout-content-test/x1/bar/mount none rbind,rw,x-snapd.origin=layout 0 0
    	/snap/layout-content-test/x1/bar /snap/layout-content-test/x1/baz none rbind,rw,x-snapd.origin=layout 0 0
    desiredIDs: map[/snap/layout-content-test/x2/bar/mount:true /snap/layout-content-test/x2/baz:true]
    reuse: map[]
    mount changes needed:
    	unmount (/snap/layout-content-test/x1/bar /snap/layout-content-test/x1/baz none rbind,rw,x-snapd.origin=layout 0 0)
    	unmount (/snap/layout-content-test/x1/foo /snap/layout-content-test/x1/bar/mount none rbind,rw,x-snapd.origin=layout 0 0)
    	mount (/snap/layout-content-test/x2/foo /snap/layout-content-test/x2/bar/mount none rbind,rw,x-snapd.origin=layout 0 0)
    	mount (/snap/layout-content-test/x2/bar /snap/layout-content-test/x2/baz none rbind,rw,x-snapd.origin=layout 0 0)
    performing mount changes:
    	 * unmount (/snap/layout-content-test/x1/bar /snap/layout-content-test/x1/baz none rbind,rw,x-snapd.origin=layout 0 0)
    umount "/snap/layout-content-test/x1/baz" (error: device or resource busy)

This is obviously something where we should have used MNT_DETACH, since
that will have the kernel handle the inner elements ahead of the outer
elements.

In fact, we should never really unmount anything as running applications
may hold open file descriptors and that would cause the mount namespace
update to fail miserably with partially-applied content changes. To
avoid this problem, whenever a mount entry, that doesn't represent a
symlink, is no longer needed it is simply detached, rather than
unmounted.

Fixes: https://bugs.launchpad.net/snapd/+bug/1831010

### DIFF
--- a/cmd/snap-update-ns/change.go
+++ b/cmd/snap-update-ns/change.go
@@ -500,7 +500,8 @@ func neededChangesImpl(currentProfile, desiredProfile *osutil.MountProfile) []*C
 		} else {
 			var entry osutil.MountEntry = current[i]
 			entry.Options = append([]string(nil), entry.Options...)
-			// Detach directory bind mounts and filesystem mounts.
+			// If the mount entry can potentially host nested mount points then detach
+			// rather than unmount, since detach will always succeed.
 			if (entry.Type == "tmpfs" || entry.OptBool("bind") || entry.OptBool("rbind")) && !entry.XSnapdDetach() {
 				entry.Options = append(entry.Options, osutil.XSnapdDetach())
 			}
--- a/cmd/snap-update-ns/change.go
+++ b/cmd/snap-update-ns/change.go
@@ -498,7 +498,13 @@ func neededChangesImpl(currentProfile, desiredProfile *osutil.MountProfile) []*C
 		if reuse[current[i].Dir] {
 			changes = append(changes, &Change{Action: Keep, Entry: current[i]})
 		} else {
-			changes = append(changes, &Change{Action: Unmount, Entry: current[i]})
+			var entry osutil.MountEntry = current[i]
+			entry.Options = append([]string(nil), entry.Options...)
+			// Detach directory bind mounts and filesystem mounts.
+			if (entry.Type == "tmpfs" || entry.OptBool("bind") || entry.OptBool("rbind")) && !entry.XSnapdDetach() {
+				entry.Options = append(entry.Options, osutil.XSnapdDetach())
+			}
+			changes = append(changes, &Change{Action: Unmount, Entry: entry})
 		}
 	}
 
--- a/cmd/snap-update-ns/change.go
+++ b/cmd/snap-update-ns/change.go
@@ -502,7 +502,8 @@ func neededChangesImpl(currentProfile, desiredProfile *osutil.MountProfile) []*C
 			entry.Options = append([]string(nil), entry.Options...)
 			// If the mount entry can potentially host nested mount points then detach
 			// rather than unmount, since detach will always succeed.
-			if (entry.Type == "tmpfs" || entry.OptBool("bind") || entry.OptBool("rbind")) && !entry.XSnapdDetach() {
+			shouldDetach := entry.Type == "tmpfs" || entry.OptBool("bind") || entry.OptBool("rbind")
+			if shouldDetach && !entry.XSnapdDetach() {
 				entry.Options = append(entry.Options, osutil.XSnapdDetach())
 			}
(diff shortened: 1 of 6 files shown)


cmd/snap-update-ns/change.go
```diff
@@ -498,7 +498,13 @@ func neededChangesImpl(currentProfile, desiredProfile *osutil.MountProfile) []*C
 		if reuse[current[i].Dir] {
 			changes = append(changes, &Change{Action: Keep, Entry: current[i]})
 		} else {
-			changes = append(changes, &Change{Action: Unmount, Entry: current[i]})
+			var entry osutil.MountEntry = current[i]
+			entry.Options = append([]string(nil), entry.Options...)
+			// Detach directory bind mounts and filesystem mounts.
```
bboozzoo: IIRC these could be files too, also tmpfs is the only fs we mount. The comment should probably say `// Detach bind mounts and tmpfs mounts`

cmd/snap-update-ns/change.go
```diff
@@ -498,7 +498,13 @@ func neededChangesImpl(currentProfile, desiredProfile *osutil.MountProfile) []*C
...
+			if (entry.Type == "tmpfs" || entry.OptBool("bind") || entry.OptBool("rbind")) && !entry.XSnapdDetach() {
```
bboozzoo: Nested condition makes it a bit hard to read, maybe organizing the code like this would make it clearer:
```
needsDetach := entry.Type == "tmpfs" || entry.OptBoot("bind") || entry.OptBool("rbind")
if needsDetach && !enrty.XSnapdDetach() {
   ...
}
```

cmd/snap-update-ns/change.go
zyga: Yes although files don't have this problem. We technically only mount `tmpfs` but the property holds for all filesystems, since they can by definition hold further mount entries underneath. I will expand the comment to explain why this is done.

cmd/snap-update-ns/change.go
```diff
@@ -498,7 +498,15 @@ func neededChangesImpl(currentProfile, desiredProfile *osutil.MountProfile) []*C
 		if reuse[current[i].Dir] {
 			changes = append(changes, &Change{Action: Keep, Entry: current[i]})
 		} else {
-			changes = append(changes, &Change{Action: Unmount, Entry: current[i]})
+			var entry osutil.MountEntry = current[i]
```
chipaca: typically we just write this as `entry := current[i]`.

Also, any reason we're copying the whole MountEntry struct around every time instead of using pointers?

zyga: Ah, that `var` is a leftover from earlier experiment.

As for copying, it's not a big deal in this program but since there are pointers inside the mount entry I wanted to avoid having two entries sharing options by accident. Since go doesn't have const / mutability as a part of the variable type declaration it is hard to say "you can use this object but don't change it", hence I'm copying it as the entry is now distinct from the original.

PR #6939, opened 2019-05-31
gadget: main update routine

bboozzoo:
Add the main updater routine. The update creates a contextual updater for each pair of changed structures. Before any updates are applied, all data that would be modified, is backed up. On errors, the modified structures are restored from backups.


zyga:
```
/home/travis/gopath/src/github.com/snapcore/snapd/strutil/shlex/shlex.go:375:8: ineffectual assignment to state
```

CC @bboozzoo

gadget/update.go
```diff
@@ -132,3 +207,82 @@ func canUpdateVolume(from *PositionedVolume, to *PositionedVolume) error {
...
+	}
+
+	for _, one := range updaters {
+		if err := one.Backup(); err != nil {
+			return fmt.Errorf("cannot backup structure: %v", err)
+		}
+	}
+
+	var updateErr error
+	var updateLastAttempted int
```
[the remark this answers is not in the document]
bboozzoo: That, or file stamps that are used for backup pass.

I'm thinking, we should also return ErrNoUpdate when there actually no update deployed after backup/update passes. That should cover the case when one declares a different `edition`, but the content is unchanged.

gadget/update.go
```diff
@@ -132,3 +207,82 @@ func canUpdateVolume(from *PositionedVolume, to *PositionedVolume) error {
...
+
+type updatePair struct {
+	from *PositionedStructure
+	to   *PositionedStructure
+}
+
+func resolveUpdate(oldVol *PositionedVolume, newVol *PositionedVolume) (updates []updatePair) {
+	for j, oldStruct := range oldVol.PositionedStructure {
+		newStruct := newVol.PositionedStructure[j]
+		if newStruct.Update.Edition != oldStruct.Update.Edition {
```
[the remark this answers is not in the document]
bboozzoo: In that case, the code It defaults to edition 0. I discussed this a bit with @mvo5 and the conclusion was we should enforce that the new edition higher or equal than the old one. Once added, it would not be removed.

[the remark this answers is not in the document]
bboozzoo: The check has been updated as we discussed. Updates are only picked up if `edition` is larger than the old one.

gadget/update.go
```diff
@@ -21,12 +21,78 @@ package gadget
 import (
 	"errors"
 	"fmt"
+
+	"github.com/snapcore/snapd/logger"
 )
 
 var (
-	ErrNoUpdate = errors.New("no update needed")
```
pedronis: nitpick: I would use different var groupings for Err* variable and the non-exported ones

gadget/update.go
```diff
@@ -21,12 +21,78 @@ package gadget
...
+// GadgetData holds references to a gadget revision metadata and its data directory.
+type GadgetData struct {
+	// Info is the gadget metadata
+	Info *Info
+	// RootDir is the root directory of gadget snap data
+	RootDir string
+}
+
+// Update applies the gadget update given the gadget information and data from
+// old and new revisions. It errors out when update is not possible or illegal,
```
pedronis: the update

gadget/update.go
```diff
@@ -21,12 +21,78 @@ package gadget
...
+// or a failure occurs at any of the steps. When there is no update, a special
+// error ErrNoUpdate is returned.
+//
+// Updates are opt-in, and are only applied to structures which have a different
```
pedronis: s/different/higher/ ?

gadget/update.go
```diff
@@ -132,3 +198,97 @@ func canUpdateVolume(from *PositionedVolume, to *PositionedVolume) error {
...
+// TODO: replace placeholder implementation with real one
+var updaterForStructure = func(_ *PositionedStructure, rootDir, rollbackDir string) (Updater, error) { return &nopUpdater{}, nil }
+
+func applyUpdates(new GadgetData, updates []updatePair, rollbackDir string) error {
+	updaters := make([]Updater, 0, len(updates))
+
+	for _, one := range updates {
+		up, err := updaterForStructure(one.to, new.RootDir, rollbackDir)
+		if err != nil {
+			return fmt.Errorf("cannot prepare updater: %v", err)
```
pedronis: s/updater/update to structure/ ?

gadget/update.go
```diff
@@ -21,12 +21,78 @@ package gadget
...
+// rollback directory. Should the apply step fail, the modified data is
+// recovered.
+func Update(old, new GadgetData, rollbackDirPath string) error {
+	oldVol, newVol, err := resolveVolume(old.Info, new.Info)
+	if err != nil {
+		return err
+	}
+
+	// layout old
+	pOld, err := PositionVolume(old.RootDir, oldVol, defaultConstraints)
```
pedronis: sorry for noticing this only now,  but "lay out" would probably be a more typical verb for this?  did position come from the ubuntu-image code?

gadget/update.go
```diff
@@ -21,12 +21,78 @@ package gadget
...
+	if err != nil {
+		return fmt.Errorf("cannot position old volume: %v", err)
```
pedronis: here then is unclear what "position" means, also because it could also mean find.  "cannot compute layout of the" or "cannot lay out" would be better

gadget/update.go
```diff
@@ -132,3 +198,97 @@ func canUpdateVolume(from *PositionedVolume, to *PositionedVolume) error {
...
+		up, err := updaterForStructure(one.to, new.RootDir, rollbackDir)
+		if err != nil {
+			return fmt.Errorf("cannot prepare updater: %v", err)
+		}
+		updaters = append(updaters, up)
+	}
+
+	for _, one := range updaters {
+		if err := one.Backup(); err != nil {
+			return fmt.Errorf("cannot backup structure: %v", err)
```
pedronis: I wonder if in external message we should always have "volume structure" and not just structure

gadget/update.go
bboozzoo: As discussed yesterday, we'll update the naming to reflect the fact that it's about volume/partition layout. However, so as to keep the conflicts limited, this will happen once more of the pending PRs land.

gadget/update.go
bboozzoo: Added info about the actual structure that failed too.

gadget/update.go
bboozzoo: Error message updated

gadget/update_test.go
```diff
@@ -573,3 +578,618 @@ func (u *updateTestSuite) TestCanUpdateVolume(c *C) {
...
+	oldInfo := &gadget.Info{
+		Volumes: map[string]gadget.Volume{
+			"foo": {
+				Bootloader: "grub",
+				Schema:     gadget.GPT,
+				Structure:  []gadget.VolumeStructure{bareStruct, fsStruct, lastStruct},
+			},
+		},
+	}
+	newInfo := &gadget.Info{
```
stolowski: Could use a small comment, e.g. "oldInfo and newInfo intentionally the same, they get updated by the caller".

gadget/update_test.go
```diff
@@ -573,3 +578,618 @@ func (u *updateTestSuite) TestCanUpdateVolume(c *C) {
...
+	err := gadget.Update(oldData, newData, rollbackDir)
+	c.Assert(err, ErrorMatches, `cannot lay out the old volume: cannot position structure #0 \("foo"\): content "first.img": .* no such file or directory`)
+
+	makeSizedFile(c, filepath.Join(oldRootDir, "first.img"), gadget.SizeMiB, nil)
+
+	// cannot position the new volume
+	err = gadget.Update(oldData, newData, rollbackDir)
+	c.Assert(err, ErrorMatches, `cannot lay out the new volume: cannot position structure #0 \("foo"\): content "first.img": .* no such file or directory`)
+
+	makeSizedFile(c, filepath.Join(newRootDir, "first.img"), 900*gadget.SizeKiB, nil)
```
stolowski: Is this achieving anything without `gadget.Update(...)`?

gadget/update.go
```diff
@@ -132,3 +200,97 @@ func canUpdateVolume(from *PositionedVolume, to *PositionedVolume) error {
...
 }
+
+type updatePair struct {
+	from *PositionedStructure
+	to   *PositionedStructure
+}
+
+func resolveUpdate(oldVol *PositionedVolume, newVol *PositionedVolume) (updates []updatePair) {
+	for j, oldStruct := range oldVol.PositionedStructure {
+		newStruct := newVol.PositionedStructure[j]
```
stolowski: The length of PositionedStructure for old and new are checked earlier by `canUpdateVolume` so this is fine, I wonder if we should have a comment about that here, or an early check on entering the function (error out with internal error if lengths are different).

PR #6940, opened 2019-06-01
tests: support mountinfo-tool on Python 2.7+

zyga:
Since mountinfo-tool is pretty useful and it is used by a significant
regression test, we want to support all the operating systems we are
testing against. As such, drop the nice Python 3.6+ annotations and
replace them with comments that are compatible with all the versions of
Python.

In addition, since abc.ABC class (which avoids the syntax
incompatibility in defining meta-classes across the major versions of
Python) is not available in Python 2.7, drop the use of abstract base
classes and replace it with an empty dummy class.

Attribute filters used annotations to know about attributes that can be
filtered, this is now replaced with an explicit map of attributes that
can be queried.

Small structural changes were done to renumber_snap_revision, so that it
would type check and be flake8 clean at the same time. Instead of using
a lambda we use helper functions instead.

All of the __str__ and __repr__ functions return correct values on both
Python 2 and Python 3 *while* being type-correct according to strict
mypy type checker. This required, in the Python 2 case, encoding the
Unicode return values to byte string equivalents.

The OptFields type, which was just syntactic sugar on top of a list of
strings, was dropped, because it was causing problems with proper type
checking in absence of the typing module. Specifically the class
declaration:
```
    class OptFields(List[Text]):
	pass
```
Was causing problems when defined as plain OptFields(list). I chose to
give up given the limited value of the output. To compensate we
special-case the list-of-strings value in the printer.

Lastly, to allow running mountinfo-tool directly, without worrying which
python interpreter is actually used, a new helper script "any-python" is
provided and used as the interpreter of mountinfo-tool.

### DIFF
--- a/tests/lib/bin/any-python
+++ b/tests/lib/bin/any-python
@@ -1,0 +1,17 @@
+#!/bin/sh
+# Use any-python as interpreter in portable Python scripts.
+# Instead of /usr/bin/env python use /usr/bin/env any-python
+# The preferred choices are: Python 3, Python 2, anything called "python"
+py3="$(command -v python3)"
+py2="$(command -v python2)"
+py_dunno="$(command -v python)"
+if [ -n "$py3" ]; then
+    exec "$py3" "$@"
+elif [ -n "$py2" ]; then
+	exec "$py2" "$@"
+elif [ -n "$py_dunno" ]; then
+	exec "$py_dunno" "$@"
+else
+	echo "cannot find any Python installation" >&2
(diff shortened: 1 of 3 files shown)


zyga:
This failed with:
```
Checking for ineffective assignments
/home/travis/gopath/src/github.com/snapcore/snapd/strutil/shlex/shlex.go:375:8: ineffectual assignment to state
```
Huh?

Is that a non-deterministic test? How did it not fail before?

PR #6941, opened 2019-06-03
strutil/shlex: fix ineffassign

bboozzoo:
The `state` variable is function local, assigning before return does not make
sense.

Fixes:
github.com/snapcore/snapd/strutil/shlex/shlex.go:375:8: ineffectual assignment to state


zyga:
CC @mvo5, @chipaca

PR #6944, opened 2019-06-03
tests: skip installing a new core for autopkgtests

sergiocazzolato:
On autopkgtests we dont need to install a new core when preparing the test suite for classic

### DIFF
--- a/tests/lib/prepare.sh
+++ b/tests/lib/prepare.sh
@@ -261,19 +261,20 @@ prepare_classic() {
         # shellcheck disable=SC2086
         cache_snaps ${PRE_CACHE_SNAPS}
 
-        # Just install core and configure reexec when the backend is not autopkgtest
-        if [ "$SPREAD_BACKEND" != autopkgtest ]; then
+        # When REUSE_SNAPD is set to 1, use snapd and the core snapd as they are installed
+        # Otherwise, install the new core from the defined channel
+        if [ "$REUSE_SNAPD" != 1 ]; then
             snap list | not grep core || exit 1
             # use parameterized core channel (defaults to edge) instead
             # of a fixed one and close to stable in order to detect defects
             # earlier
             snap install --"$CORE_CHANNEL" core
-            snap list | grep core
-
-            systemctl stop snapd.{service,socket}
-            update_core_snap_for_classic_reexec
-            systemctl start snapd.{service,socket}
         fi
+        snap list | grep core
+
+        systemctl stop snapd.{service,socket}
+        update_core_snap_for_classic_reexec
+        systemctl start snapd.{service,socket}
 
         disable_refreshes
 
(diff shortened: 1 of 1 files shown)


zyga:
There  are two selinux denials, CC @bboozzoo 

```
type=AVC msg=audit(06/03/19 16:11:26.125:1194) : avc:  denied  { map } for  pid=14159 comm=unsquashfs path=/etc/passwd dev="sda1" ino=137138 scontext=system_u:system_r:snappy_t:s0 tcontext=system_u:object_r:passwd_file_t:s0 tclass=file permissive=1 
----
type=AVC msg=audit(06/03/19 16:11:26.232:1197) : avc:  denied  { read } for  pid=14184 comm=selinux-autorel name=nsswitch.conf dev="loop0" ino=748 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:snappy_snap_t:s0 tclass=file permissive=0
```

mvo5:
Unfortunately it seems like this is not a sufficient fix, when I run this with autopkgtest, I get:
```
+ systemctl stop snapd.service snapd.socket
+ save_snapd_state
+ is_core_system
+ [[ ubuntu-19.04-amd64 == ubuntu-core-* ]]
+ return 1
+ systemctl daemon-reload
++ systemd-escape --path /snap
+ escaped_snap_mount_dir=snap
++ cut -f1 -d ' '
++ grep -e '^snap[-.].*\.mount' -e '^snap[-.].*\.service'
++ systemctl list-unit-files --full
+ units='snap-core-6964.mount
snap-go-3739.mount'
+ for unit in $units
+ systemctl stop snap-core-6964.mount
+ for unit in $units
+ systemctl stop snap-go-3739.mount
+ snapd_env='/etc/environment /etc/systemd/system/snapd.service.d /etc/systemd/system/snapd.socket.d'
++ ls '/etc/apparmor.d/snap.core.*'
ls: cannot access '/etc/apparmor.d/snap.core.*': No such file or directory
++ true
+ snap_confine_profiles=
+ tar cf /home/gopath/src/github.com/snapcore/snapd/tests/snapd-state/snapd-state.tar /var/lib/snapd /snap /etc/systemd/system/snap-core-6964.mount /etc/systemd/system/multi-user.target.wants/snap-core-6964.mount /etc/environment /etc/systemd/system/snapd.service.d /etc/systemd/system/snapd.socket.d
tar: /home/gopath/src/github.com/snapcore/snapd/tests/snapd-state/snapd-state.tar: Cannot open: No such file or directory
tar: Error is not recoverable: exiting now
...
```

mvo5:
I poked around a bit and I wonder if using the main spread.yaml is actually the right approach, when I tried to untangle this I got lost too quickly. Here is a bit of a different approach: https://github.com/snapcore/snapd/pull/6948 to untangle this.

sergiocazzolato:
@mvo5 , it makes sense to me, otherwise we will be allways messing up the autopkgtest run when we update the prepare/restore of the suite. I'll close this one.

PR #6945, opened 2019-06-04
overlord: mock store to avoid net requests

mvo5:
This is the uncontroversial bit of https://github.com/snapcore/snapd/pull/6932
We should merge it IMHO while discussing/untangling the retry logic.

The patch itself is from Samuele, this prevents the unit tests from
hitting the snapcraft web APIs. In certain environments unit tests run
without any network and using anything but local services is disallowed.


overlord/managers_test.go
```diff
@@ -3584,6 +3584,10 @@ func (s *mgrsSuite) TestRemodelStoreSwitch(c *C) {
 }
 
 func (s *mgrsSuite) TestHappyDeviceRegistrationWithPrepareDeviceHook(c *C) {
+	// just to 404 locally eager account-key requests
+	mockStoreServer := ms.mockStore(c)
```
zyga:
```suggestion
	mockStoreServer := s.mockStore(c)
```

PR #6946, opened 2019-06-04
data/selinux: permit init_t to remount snappy_snap_t

bboozzoo:
Systemd services that get a new mount namespace on startup fail due to SELinux
blocking system from remounting /var/lib/snapd/snap into the new mount ns.

Since we cannot declare fs_type() and generate contexts through genfscon outside
of core policy, add an explicit permission for init_t.

Corresponding denial:
```
type=AVC msg=audit(1559625610.991:67): avc:  denied  { remount } for  pid=578
         comm="(resolved)"
         scontext=system_u:system_r:init_t:s0
         tcontext=system_u:object_r:snappy_snap_t:s0
         tclass=filesystem permissive=0
```
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1708991

cc @Conan-Kudo

### DIFF
--- a/tests/regression/rhbz-1708991/task.yaml
+++ b/tests/regression/rhbz-1708991/task.yaml
@@ -22,7 +22,8 @@ restore: |
     rm -f stamp enforcing.mode
 
 execute: |
-    # prepare does systemctl daemon-reexec to workaround:
+    # global prepare calls systemctl daemon-reexec to make systemd run with
+    # additional init_t permissions from snapd package, workaround for:
     # https://bugzilla.redhat.com/show_bug.cgi?id=1660141
     # https://github.com/systemd/systemd/issues/9997
 
--- a/tests/regression/rhbz-1708991/task.yaml
+++ b/tests/regression/rhbz-1708991/task.yaml
@@ -1,0 +1,32 @@
+summary: Check that snapd SELinux policy does not break systemd services with private mount ns
+
+systems: [fedora-*, centos-*]
+
+prepare: |
+    #shellcheck source=tests/lib/pkgdb.sh
+    . "$TESTSLIB"/pkgdb.sh
+
+    if [[ "$SPREAD_SYSTEM" == centos-7-* ]]; then
+        distro_install_package systemd-resolved
+    fi
+
+    getenforce > enforcing.mode
+
+    # Enable enforcing mode, our policy is already marked as permissive, so we
+    # will get audit entries but the program will not be stopped by SELinux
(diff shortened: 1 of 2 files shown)


tests/regression/rhbz-1708991/task.yaml
```diff
@@ -0,0 +1,32 @@
...
+        distro_install_package systemd-resolved
+    fi
+
+    getenforce > enforcing.mode
+
+    # Enable enforcing mode, our policy is already marked as permissive, so we
+    # will get audit entries but the program will not be stopped by SELinux
+    setenforce 1
+    ausearch --checkpoint stamp -m AVC || true
+
```
stolowski: Would it make sense to have a debug: section and print some extra selinux-related diagnostics in case we have any failures in the future? (FWTW I looked at existing selinux spread tests and we don't do that atm.)

tests/regression/rhbz-1708991/task.yaml
```diff
@@ -0,0 +1,32 @@
...
+restore: |
+    setenforce "$(cat enforcing.mode)"
+    rm -f stamp enforcing.mode
+
+execute: |
+    # prepare does systemctl daemon-reexec to workaround:
```
mvo5: I don't see any `systemctl daemon-reexec` in this test. Maybe rephrase the comment to something like "the global prepare does ..." and also outline why this comment is relevant for this test?

tests/regression/rhbz-1708991/task.yaml
bboozzoo: We already show the audit log in spread.yaml.

tests/regression/rhbz-1708991/task.yaml
```diff
@@ -0,0 +1,32 @@
...
+    if [[ "$SPREAD_SYSTEM" == centos-7-* ]]; then
+        distro_install_package systemd-resolved
+    fi
+
+    getenforce > enforcing.mode
+
+    # Enable enforcing mode, our policy is already marked as permissive, so we
+    # will get audit entries but the program will not be stopped by SELinux
+    setenforce 1
+    ausearch --checkpoint stamp -m AVC || true
```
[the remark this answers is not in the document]
bboozzoo: Yes. The command fails when checkpoint does not exist.

Conan-Kudo:
@zyga @bboozzoo If you know how to fix it, propose a fix here: https://github.com/fedora-selinux/selinux-policy

PRs are welcome there. :)

PR #6948, opened 2019-06-04
debian: rework how we run autopkgtests

mvo5:
Our current autopkgtest setup reuses the spread.yaml that we use
for regular spread runs. The regular spread.yaml will modify the
host heavily which is not what we want for the autopkgtest
tests. Here the idea is to use the binaries that got build as
part of the distro build and run a smoke test against the
most pristine system possible.

After poking in {prepare-restore,prepare,reset}.sh I think we
need something radically simpler: a minimal spread.yaml that
is just there for autopkgtest. This PR goes this way.

The downside of this approach is of couse that thing may
get out of sync between the main spread.yaml and the autopkgtest
spread.yaml. To fix this we need a test that runs the autopkgtest
regularly against a clean environment (probably in the nested
suite).

### DIFF
--- a/packaging/ubuntu-16.04/tests/integrationtests
+++ b/packaging/ubuntu-16.04/tests/integrationtests
@@ -49,12 +49,35 @@ systemctl restart snapd
 snap install --classic go
 
 # and now run spread against localhost
-# shellcheck disable=SC1091
-. /etc/os-release
 export GOPATH=/tmp/go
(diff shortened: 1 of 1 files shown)


mvo5:
Tested this running a manual "autpkgtest" on ubuntu-16.04-32, ubuntu-18.04-64, ubuntu-16.04-64, ubuntu-18.04-32, ubuntu-19.04-64, ubuntu-18.10-64

PR #6949, opened 2019-06-04
httputil: handle "no such host" error explicitly and do not retry it

stolowski:
Handle "no such host" error explicitly in retry code and do not retry it - this is the error observed when querying for an unknown host as long as network and dns server work. Tested manually under a network namespace with only lo interface, and with `osc build` on opensuse.


zyga:
This patch doesn't fix the issue yet. Perhaps reopen when you have completed the suse build investigation?

stolowski:
Marking blocked until I manage to sort out suse build issues.

stolowski:
> Marking blocked until I manage to sort out suse build issues.

Passed in my tests on opensuse, unblocking.

httputil/retry.go
```diff
@@ -85,6 +85,10 @@ func ShouldRetryError(attempt *retry.Attempt, err error) bool {
 	// The CDN sometimes resets the connection (LP:#1617765), also
 	// retry in this case
 	if opErr, ok := err.(*net.OpError); ok {
+		// "no such host" is a permanent error and should not be retried.
+		if opErr.Op == "dial" && strings.Contains(opErr.Error(), "no such host") {
```
bboozzoo: Looking at Go source code, errNoSuchHost is used with either pure Go resolver or Cgo resolver. For the Cgo one, it's used when getaddrinfo() returns EAI_NONAME. We should be able to correctly catch the case regardless of how system environment is set up, or libc used.

DNSError also has IsNotFound field that basically conveys this information. In which case, perhaps the check would better suited here: https://github.com/snapcore/snapd/pull/6949/files#diff-5cf048fe4361da077de704b9876215fdR116
```
if dnsErr, ok := opErr.Err.(*net.DNSError); ok {
    if dnsErr.IsNotFound {
        // no such host
        return false
    }
    ...
}
```

stolowski: Unfortunately `IsNotfound` seems to be a new addition to DNSError, it's not present in go 1.10.4 😢

bboozzoo: We've tracked it down to code in Go which is not part of any release yet. Once it is released, we will still be a number of versions behind. Probably, it's best to stay with the unfortunate string check.

PR #6952, opened 2019-06-05
packaging: fix permissons powerpc docs dir

mvo5:
When building powerpc, use install -d instead of mkdir to install
the base directory for the README.powerpc.

This fixes a lintian warning about unusual permissions on the
powerpc build.


mvo5:
@zyga This was only very recently added when we did the hack to build an empty powerpc package of snapd.

PR #6958, opened 2019-06-06
Added api endpoint for downloading snaps

glower:


daemon/api_download_test.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	for _, s := range []scenario{
+		{
+			dataJSON: `{"action": "download", "snaps": ["doom"]}`,
+			status:   404,
+			err:      "snap not found",
+		},
+		{
+			dataJSON: `{"action": "download", "snaps": ["foo"]}`,
```
mvo5: I would rename "foo" to something like "download-error-trigger-snap" to make reading a bit easier. I stumbled a bit reading this as the dataJSON that triggers the error looks almost the same as the one that does not, of course the snap is different but "foo" looks innocent :)

daemon/export_api_download_test.go
```diff
@@ -0,0 +1,25 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2018 Canonical Ltd
```
mvo5: (c) 2019

daemon/api_download.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	downloadInfo := info.DownloadInfo
+	r, err := getStore(c).DownloadStream(context.TODO(), snapName, &downloadInfo, user)
+	if err != nil {
+		return InternalError(err.Error())
+	}
+
+	return FileStream{
+		FileName: snapName,
```
chipaca: the `snapName` is not the filename. That would be something like `filepath.Base(info.MountFile())`.

daemon/api_download.go
```diff
@@ -0,0 +1,93 @@
...
+	"encoding/json"
+	"net/http"
+
+	"github.com/snapcore/snapd/overlord/auth"
+	"github.com/snapcore/snapd/store"
+)
+
+var snapDownloadCmd = &Command{
+	Path:     "/v2/download",
+	UserOK:   true,
```
chipaca: you don't need `UserOK` if there is no `GET`.

OTOH I think this _should_ be a `GET`, and not a `POST` (otherwise `snap download` would ask you for credentials...). @pedronis WDYT?

daemon/response.go
```diff
@@ -246,6 +246,26 @@ func makeErrorResponder(status int) errorResponder {
...
+// A FileStream ServeHTTP method streams the snap
+type FileStream struct {
+	FileName string
+	Info     snap.DownloadInfo
+	stream   io.ReadCloser
+}
+
+// ServeHTTP from the Response interface
+func (s FileStream) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
+	w.Header().Set("Content-Type", "application/octet-stream")
```
chipaca: just call `w.Header()` once please

daemon/response.go
```diff
@@ -246,6 +246,26 @@ func makeErrorResponder(status int) errorResponder {
...
+	size := fmt.Sprintf("%d", s.Info.Size)
+	w.Header().Set("Content-Length", size)
+	defer s.stream.Close()
+
+	if _, err := io.Copy(w, s.stream); err != nil {
```
chipaca: you should be checking how many bytes are copied here I think (and failing with EOF/[502] if it's too short)

[502]: https://httpstatusdogs.com/502-bad-gateway

daemon/response.go
```diff
@@ -246,6 +246,26 @@ func makeErrorResponder(status int) errorResponder {
...
+	FileName string
+	Info     snap.DownloadInfo
+	stream   io.ReadCloser
+}
+
+// ServeHTTP from the Response interface
+func (s FileStream) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
+	w.Header().Set("Content-Type", "application/octet-stream")
+	size := fmt.Sprintf("%d", s.Info.Size)
+	w.Header().Set("Content-Length", size)
```
chipaca: can you also set content-disposition (see `FileResponse`)?

daemon/api_download_test.go
```diff
@@ -0,0 +1,192 @@
...
+			},
+		}, nil
+	}
+	return nil, store.ErrSnapNotFound
+}
+
+func (s *snapDownloadSuite) DownloadStream(ctx context.Context, name string, downloadInfo *snap.DownloadInfo, user *auth.UserState) (io.ReadCloser, error) {
+	if name == "bar" {
+		return ioutil.NopCloser(bytes.NewReader([]byte("SNAP"))), nil
+	} else {
```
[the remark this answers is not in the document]
glower: You are totally right here. this will trigger me in golang code review as well :), will fix it

daemon/response.go
glower: Do you mean this?
```
hdr := w.Header()
hdr.Set(...)
```

daemon/api_download.go
chipaca: talked about it in the standup, and yeah, we need it to be POST. So, drop the UserOK :-)

daemon/response.go
chipaca: yeap

daemon/response.go
glower: Added the check, but the problem with reporting errors, in this case, is this: after something is written to ResponseWriter, status is automatically set to 200. Setting w.WriteHeader after this make no effect. Don't know how to report an error in this case propperly

mvo5:
Thanks for the update! Some go vet issues reported by travis (should be trivial to fix):
```
...
Check for usages of http.Status*
Usages of http.Status*, we prefer the numeric values directly:
/home/travis/gopath/src/github.com/snapcore/snapd/daemon/response.go:270:		http.Error(w, err.Error(), http.StatusInternalServerError)
/home/travis/gopath/src/github.com/snapcore/snapd/daemon/response.go:274:		http.Error(w, io.EOF.Error(), http.StatusBadGateway)
```

daemon/response.go
```diff
@@ -248,21 +248,30 @@ func makeErrorResponder(status int) errorResponder {
...
 	Info     snap.DownloadInfo
 	stream   io.ReadCloser
 }
 
 // ServeHTTP from the Response interface
 func (s FileStream) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
-	w.Header().Set("Content-Type", "application/octet-stream")
+	hdr := w.Header()
+	hdr.Set("Content-Type", "application/octet-stream")
+	hdr.Set("Content-Disposition", s.SnapName)
```
chipaca: that's wrong: it should be `attachment; filename=<the filename>`, assuming the filename is just ascii.

daemon/response.go
chipaca: good point, i guess all we can do is shuffle things on and let the client figure things out from the content-length (and the hash).

glower: yes, this is the way to check it, I think

daemon/response.go
glower: Fixed, my bad

PR #6961, opened 2019-06-06
overlord/snapstate, & fallout: give Install a *RevisionOptions

chipaca:
This change is a pure refactor that drops `snapstate.InstallCohort`,
and changes the signature of `snapstate.Install` so it's more like
`snapstate.Update`, taking a `*snapstate.RevisionOptions`,

    func Install(st *state.State, name string, opts *RevisionOptions, userID int, flags Flags) (*state.TaskSet, error)

of course this also means changing InstallWithDeviceContext, in the
obvious way,

    func InstallWithDeviceContext(st *state.State, name string, opts *RevisionOptions, userID int, flags Flags, deviceCtx DeviceContext) (*state.TaskSet, error)

HTH, HAND!

### DIFF
--- a/overlord/snapstate/storehelpers.go
+++ b/overlord/snapstate/storehelpers.go
@@ -82,7 +82,7 @@ func refreshOptions(st *state.State, origOpts *store.RefreshOptions) (*store.Ref
 	return &opts, nil
 }
 
-func installInfo(st *state.State, name, channel, cohort string, revision snap.Revision, userID int, deviceCtx DeviceContext) (*snap.Info, error) {
+func installInfo(st *state.State, name string, revOpts *RevisionOptions, userID int, deviceCtx DeviceContext) (*snap.Info, error) {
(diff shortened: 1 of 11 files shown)


overlord/snapstate/storehelpers.go
```diff
@@ -108,11 +103,17 @@ func installInfo(st *state.State, name, channel, cohort string, revision snap.Re
...
 		// the desired channel
-		Channel: channel,
+		action.Channel = revOpts.Channel
+		// the desired cohort key
+		action.CohortKey = revOpts.CohortKey
+	} else {
 		// the desired revision
-		Revision:  revision,
-		CohortKey: cohort,
+		action.Revision = revOpts.Revision
```
bboozzoo: Wondered whether we should drop some internal error if cohort key is set, but there's a similar check before we reach this place already, so probably not worth it.

PR #6962, opened 2019-06-07
interfaces: allow locking of pppd files

alfonsosanchezbeato:
Allow file locking in files in [/var]/run/ppp* for pppd.


bboozzoo:
Does this need a quick look from @jdstrand ?

PR #6970, opened 2019-06-07
data/selinux: fix policy for snaps with bases and classic snaps

bboozzoo:
When a classic snap with service or hooks is installed on SELinux system, the
policy will prevent snapd from running the hooks or, should the snap have a
service, it will fail to be started by systemd -> snap-confine chain.

The reason is that for classic snaps, snap-confine executes snap-exec from the
host (labeled as snappy_exec_t) instead of the one from a mounted snap (labeled
as snappy_snap_t). The policy does not allow that.

Allow snappy_exec_t to be an entrypoint to unconfined_service_t.

cc @Conan-Kudo


bboozzoo:
Will look into adding a spread test.

Conan-Kudo:
I don't think I like this being unconditionally allowed by default. In general, classic snaps are "unsafe" snaps, since they violate most of the rules that snaps try to guarantee. I think this should be locked behind a tunable and people should manually turn this on with a boolean.

bboozzoo:
@Conan-Kudo as discussed on IRC, the tunable (which is off by default) would break services and hooks in snaps with bases. The reason is, when a snap with base is started, the $(libexecdir)/snapd from the host is bind mounted into the snap's namespace, carrying over the SELinux contexts from the host. The policy must allow snap-confine to transition to `unconfined_service_t` via execution of `snappy_exec_t`, which is identical to the classic snap.

Conan-Kudo:
@bboozzoo Please bump the version of the snappy SELinux module too. It should be bumped to `0.1.0` at this point.

PR #6972, opened 2019-06-07
many: install snap-confine-debug while testing

zyga:
For the longest time, snap-confine had a sibling, snap-confine-debug,
built with additional debugging messages. Those were almost never visble
outside of a particular contributor's computer. This patch creates a new
binary package, snap-confine-debug, which exists solely to ship the binary
so that it can be injected into the repackaged core or snapd snaps and
installed as a replacement for snap-confine during testing.

The patch is partial since it needs to be compensated in other packaging
systems, so that the new file is not installed anywhere.

### DIFF
--- a/packaging/ubuntu-16.04/control
+++ b/packaging/ubuntu-16.04/control
@@ -111,6 +111,14 @@ Depends: snapd (= ${binary:Version}), ${misc:Depends}
 Description: Transitional package for snapd
  This is a transitional dummy package. It can safely be removed.
 
+Package: snap-confine-debug
+Architecture: any
+Depends: snapd (= ${binary:Version}), ${misc:Depends}
+Description: snap-confine built with additional debugging.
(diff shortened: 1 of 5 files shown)


packaging/ubuntu-16.04/control
```diff
@@ -111,6 +111,14 @@ Depends: snapd (= ${binary:Version}), ${misc:Depends}
 Description: Transitional package for snapd
  This is a transitional dummy package. It can safely be removed.
 
+Package: snap-confine-debug
```
mvo5: Why do we install it into an extra pacakge instead of just shipping it as part of snapd?

zyga: Because nobody outside of the team would need to use it and it is somewhat larger than the regular build.

zyga:
This branch would be nice to have but we are dealing with larger issues so I'll close it for now.

PR #6974, opened 2019-06-10
spread-shellcheck: use safe loader

bboozzoo:
Fix the warning when loading PyYAML:
```
./spread-shellcheck:140: YAMLLoadWarning: calling yaml.load() without Loader=...
  is deprecated, as the default Loader is unsafe.
  Please read https://msg.pyyaml.org/load for full details.
```

### DIFF
--- a/spread-shellcheck
+++ b/spread-shellcheck
@@ -137,7 +137,7 @@ def checksection(data, env: Dict[str, str]):
 def checkfile(path):
     logging.debug("checking file %s", path)
     with open(path) as inf:
-        data = yaml.load(inf)
+        data = yaml.safe_load(inf)
 
     errors = ShellcheckError(path)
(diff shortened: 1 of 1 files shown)


spread-shellcheck
```diff
@@ -137,7 +137,7 @@ def checksection(data, env: Dict[str, str]):
 def checkfile(path):
     logging.debug("checking file %s", path)
     with open(path) as inf:
-        data = yaml.load(inf)
+        data = yaml.load(inf, Loader=yaml.CSafeLoader)
```
zyga: Any reason over just using `yaml.safe_load`?

bboozzoo: Started out with that, @chipaca suggested, CSafeLoader since it can be slightly faster. In my totally unscientific benchmark, there was a tiny difference, so went with that.

PR #6993, opened 2019-06-12
devicestate: make TestUpdateGadgetOnClassicErrorsOut less racy

mvo5:
The test sometimes fails with:
```
FAIL: devicestate_test.go:3237: deviceMgrSuite.TestUpdateGadgetOnClassicErrorsOut

devicestate_test.go:3261:
    c.Assert(chg.IsReady(), Equals, true)
... obtained bool = false
... expected bool = true
```

Easy to reproduce with
```
$ go test -count 100 -check.f TestUpdateGadgetOnClassicErrorsOut
```

### DIFF
--- a/overlord/devicestate/devicestate_test.go
+++ b/overlord/devicestate/devicestate_test.go
@@ -3252,9 +3252,18 @@ func (s *deviceMgrSuite) TestUpdateGadgetOnClassicErrorsOut(c *C) {
 
 	s.state.Unlock()
 
-	for i := 0; i < 6; i++ {
+	// we cannot use "s.o.Settle()" here because this change has an
+	// error which means that the settle will never converge
+	for i := 0; i < 50; i++ {
 		s.se.Ensure()
 		s.se.Wait()
+
+		s.state.Lock()
+		ready := chg.IsReady()
+		s.state.Unlock()
+		if ready {
+			break
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


overlord/devicestate/devicestate_test.go
```diff
@@ -3252,10 +3252,8 @@ func (s *deviceMgrSuite) TestUpdateGadgetOnClassicErrorsOut(c *C) {
 
 	s.state.Unlock()
 
-	for i := 0; i < 6; i++ {
-		s.se.Ensure()
-		s.se.Wait()
-	}
+	// we could use:
+	s.o.Settle(0.5 * time.Second)
```
bboozzoo: vet complains:
```
overlord/devicestate/devicestate_test.go:3256:17: constant 0.5 truncated to integer
```

stolowski: FWTW the timeout here is really relevant if settle is not converging (e.g. we have a real bug in code or test), there is no real benefit with fractions of seconds here, it's fine to have large timeout; it might actually be better to have a few seconds here (we seem to be using 5*Second in many places) to avoid failures on very slow boards.

mvo5:
Sorry, I just forced pushed a real fix, the settle() one does not work unfortunately - silly me. It does not work because the test creates a change in error state and settle will not converge for changs in error state.

PR #6995, opened 2019-06-12
gadget: fallback device lookup

bboozzoo:
The `/dev/disk/by-{partlabel,label}` symlinks made by udev are often relative. Make sure that code supports that.

Add a fallback path to device lookup for a given structure. Base on the assumption that there is a filesystem labeled 'writable'.


gadget/device_linux.go
```diff
@@ -54,11 +55,23 @@ func FindDeviceForStructure(ps *PositionedStructure) (string, error) {
 	var found string
 	var match string
 	for _, candidate := range candidates {
-		if !osutil.FileExists(candidate) {
-			continue
+		st, err := os.Lstat(candidate)
```
stolowski: Can we skip lstat here and simply do `target, err := filepath.EvalSymlinks(candidate)`, will EvalSymlink error out if candidate is not a symlink?

bboozzoo: From earlier testing, EvalSymlink would not error out. That's why I added an explicit check for a symlink.

gadget/device_linux.go
```diff
@@ -137,3 +200,45 @@ func FindMountPointForStructure(ps *PositionedStructure) (string, error) {
...
+
+	// do not bother with investigating major/minor devices (inconsistent
+	// across block device types) or mangling strings, but look at sys
+	// hierarchy for block devices instead:
+	// /sys/block/sda               - main SCSI device
+	// /sys/block/sda/sda1          - partition 1
+	// /sys/block/sda/sda<n>        - partition n
+	// /sys/block/nvme0n1           - main NVME device
+	// /sys/block/nvme0n1/nvme0n1p1 - partition 1
+	matches, err := filepath.Glob(filepath.Join(dirs.GlobalRootDir, "/sys/block/*/", devname))
```
stolowski: Looks fine. I was trying to see if there is a simpler and more direct way that doesn't require globbing, but couldn't find it.

gadget/device_linux.go
```diff
@@ -137,3 +200,45 @@ func FindMountPointForStructure(ps *PositionedStructure) (string, error) {
...
+
+func findParentDeviceWithWritableFallback() (string, error) {
+	byWritableLabel := filepath.Join(dirs.GlobalRootDir, "/dev/disk/by-label/", writableFsLabel)
+	target, err := filepath.EvalSymlinks(byWritableLabel)
+
+	if err != nil {
+		if os.IsNotExist(err) {
+			return "", ErrDeviceNotFound
+		}
+		return "", fmt.Errorf("cannot resolve device symlink for filesystem label %q: %v", writableFsLabel, err)
```
stolowski: Do we have a test for this and not-exists error condition above?

bboozzoo: IsNotExist is already tested, I'll add something for the else case.

gadget/device_test.go
```diff
@@ -194,8 +219,158 @@ func (d *deviceSuite) TestDeviceFindNotFoundNotASymlink(c *C) {
...
+		StartOffset: 123,
+	}
+	found, offs, err := gadget.FindDeviceForStructureWithFallback(psFs)
+	c.Check(err, ErrorMatches, "internal error: cannot use with filesystem structures")
+	c.Check(found, Equals, "")
+	c.Check(offs, Equals, gadget.Size(0))
+}
+
+func (d *deviceSuite) TestDeviceFindFallbackBadEvalSymlinks(c *C) {
+	if os.Geteuid() == 0 {
```
stolowski: Can we instead have `var evalSymlinks = filepath.EvalSymlinks` and mock it in the test to inject an error, so this restriction can be avoided?

bboozzoo: added

bboozzoo:
Temporarily marked as blocked while I tweak the fallback device lookup.

bboozzoo:
Split out relative symlink handling into #7009

gadget/device_linux.go
```diff
@@ -81,6 +96,57 @@ func FindDeviceForStructure(ps *PositionedStructure) (string, error) {
...
+//   by which we could find it)
+//
+// The fallback mechanism uses the fact that Core devices always have a mount at
+// /writable. The system is booted from the parent of the device mounted at
+// /writable.
+//
+// Returns the device name and an offset at which the stricture content starts
+// within the device or an error.
+func FindDeviceForStructureWithFallback(ps *PositionedStructure) (dev string, offs Size, err error) {
+	if !ps.IsBare() {
```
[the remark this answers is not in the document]
bboozzoo: Perhaps it would make sense to rename the helper to something like `HasFilesystem()`. On the same note `ps.Type == "bare"` could be `HasPartition()`.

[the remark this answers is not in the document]
bboozzoo: While at it, I'll do a follow up with renames.

gadget/device_test.go
```diff
@@ -194,8 +221,156 @@ func (d *deviceSuite) TestDeviceFindNotFoundNotASymlink(c *C) {
 			Label: "foo",
 		},
 	})
-	c.Check(err, ErrorMatches, `cannot read device link: .*`)
+	c.Check(err, ErrorMatches, `candidate .*/dev/disk/by-label/foo is not a symlink`)
+	c.Check(found, Equals, "")
+}
+
+const writableMountInfo = `
```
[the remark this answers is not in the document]
bboozzoo: Dropped the whole newline thing. The entry is short enough to be on the same line as variable declaration.

gadget/device_linux.go
```diff
@@ -82,6 +85,55 @@ func FindDeviceForStructure(ps *PositionedStructure) (string, error) {
...
+// - the structure has no filesystem
+// - and the structure is of type: bare (no partition table entry)
+// - or the structure has no name, but a partition table entry (hence no label
+//   by which we could find it)
+//
+// The fallback mechanism uses the fact that Core devices always have a mount at
+// /writable. The system is booted from the parent of the device mounted at
+// /writable.
+//
+// Returns the device name and an offset at which the stricture content starts
```
stolowski: s/stricture/structure/

gadget/device_linux.go
```diff
@@ -138,3 +190,61 @@ func FindMountPointForStructure(ps *PositionedStructure) (string, error) {
 
 	return mountPoint, nil
 }
+
+func isWritableMount(entry *osutil.MountInfoEntry) bool {
+	// example mountinfo entry:
+	// 26 27 8:3 / /writable rw,relatime shared:7 - ext4 /dev/sda3 rw,data=ordered
+	return entry.Root == "/" && entry.MountDir == "/writable" && entry.FsType == "ext4"
```
stolowski: Silly question, do we always assume ext4 for /writable?

gadget/device_linux.go
```diff
@@ -138,3 +190,61 @@ func FindMountPointForStructure(ps *PositionedStructure) (string, error) {
...
+}
+
+func findDeviceForWritable() (device string, err error) {
+	mountInfo, err := osutil.LoadMountInfo(filepath.Join(dirs.GlobalRootDir, osutil.ProcSelfMountInfo))
+	if err != nil {
+		return "", fmt.Errorf("cannot read mount info: %v", err)
+	}
+	for _, entry := range mountInfo {
+		if isWritableMount(entry) {
+			device = entry.MountSource
```
stolowski: Nitpick, this could return device immediately instead of breaking out from the loop.

gadget/device_linux.go
bboozzoo: That's how the fs is created in all images.

gadget/device_linux.go
bboozzoo: Good catch!

PR #6996, opened 2019-06-13
tests/main/appstream-id: collect debug info

bboozzoo:
The test fails often. Try to collect some information to find out what is going
on.

### DIFF
--- a/tests/main/appstream-id/task.yaml
+++ b/tests/main/appstream-id/task.yaml
@@ -1,7 +1,8 @@
 summary: Verify AppStream ID integration
 
-# fedora-*, amazon-*, centos-*: use nmap netcat by default (https://nmap.org/ncat/)
-systems: [-fedora-*, -amazon-*, -centos-*]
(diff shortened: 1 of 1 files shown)


tests/main/appstream-id/task.yaml
```diff
@@ -1,28 +1,34 @@
...
+systems: [-ubuntu-14.04-*, -ubuntu-core*]
 
 prepare: |
     snap install jq
 
+debug: |
+    if [ -e response ]; then
+        grep -n '' response
+    fi
+
```
stolowski: response file can be removed afterwards?

bboozzoo: Added the cleanup. Although, IIRC @sergiocazzolato added something to clean up the test tree automatically on each prepare/restore. Isn't that right?

mvo5: Yes, the test dir is cleaned everytime iirc

PR #6997, opened 2019-06-13
gadget: add wrapper for creating and populating filesystems

bboozzoo:
Add wrappers for creating and populating filesystems that can specified in
gadget YAML. Only ext4 and vfat can be used, thus the patch adds support for
`mkfs.ext4` and `mkfs.vfat`.
    
For VFAT we also use `mtools` to populate the filesystem with contents of given
directory.

### DIFF
--- a/gadget/mkfs.go
+++ b/gadget/mkfs.go
@@ -37,8 +37,9 @@ func MkfsExt4(img, label, contentsRootDir string) error {
 		"mkfs.ext4",
 		// default usage type
 		"-T", "default",
-		// disable metadata checksum
-		// XXX: why?
+		// disable metadata checksum, which were unsupported in Ubuntu
+		// 16.04 and Ubuntu Core 16 systems and would lead to a boot
+		// failure if enabled
 		"-O", "-metadata_csum",
 		// allow uninitialized block groups
 		"-O", "uninit_bg",
--- a/gadget/mkfs.go
+++ b/gadget/mkfs.go
@@ -33,7 +33,6 @@ import (
 // directory.
 func MkfsExt4(img, label, contentsRootDir string) error {
 	// taken from ubuntu-image
-	// TODO: support e2fsprogs 1.42 without -d in Ubuntu 16.04
 	mkfsArgs := []string{
 		"mkfs.ext4",
 		// default usage type
@@ -45,12 +44,14 @@ func MkfsExt4(img, label, contentsRootDir string) error {
 		"-O", "uninit_bg",
 		// mkfs.ext4 can populate the filesystem with contents of given
 		// root directory
+		// TODO: support e2fsprogs 1.42 without -d in Ubuntu 16.04
 		"-d", contentsRootDir,
 	}
 	if label != "" {
 		mkfsArgs = append(mkfsArgs, "-L", label)
 	}
 	mkfsArgs = append(mkfsArgs, img)
+	// run through fakeroot so that files are owned by root
 	cmd := exec.Command("fakeroot", mkfsArgs...)
 	out, err := cmd.CombinedOutput()
 	if err != nil {
--- a/gadget/mkfs.go
+++ b/gadget/mkfs.go
@@ -1,0 +1,121 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+package gadget
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+
+	"github.com/snapcore/snapd/osutil"
+)
+
+// MkfsExt4 creates an EXT4 filesystem in given image file, with an optional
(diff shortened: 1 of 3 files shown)


gadget/mkfs.go
```diff
@@ -0,0 +1,121 @@
...
+	"os/exec"
+	"path/filepath"
+
+	"github.com/snapcore/snapd/osutil"
+)
+
+// MkfsExt4 creates an EXT4 filesystem in given image file, with an optional
+// filesystem label, and populates it with the contents of provided root
+// directory.
+func MkfsExt4(img, label, contentsRootDir string) error {
```
mvo5: (Thinking aloud) This is probably overdoing it but I wonder if putting this into `gadget/mkfs/mkfs.go` would make sense. Then it would read `mkfs.Ext4(...)` and mkfs.VFat(...)` which has a nice ring to it (and is close to how its layed out on the filesystem mkfs.ext4, mkfs.vfat). But probably silly (also easy to tweak later :)

gadget/mkfs.go
```diff
@@ -0,0 +1,121 @@
...
+	// taken from ubuntu-image
+	// TODO: support e2fsprogs 1.42 without -d in Ubuntu 16.04
+	mkfsArgs := []string{
+		"mkfs.ext4",
+		// default usage type
+		"-T", "default",
+		// disable metadata checksum
+		// XXX: why?
+		"-O", "-metadata_csum",
```
mvo5: I think we should clarify with foundations (cc @vorlonofportland) why ubuntu-image is using these defaults. It seems beneficial to use metadata checksums so having at least a comment about the "why disabled" here would be good.

gadget/mkfs.go
```diff
@@ -0,0 +1,121 @@
...
+		"-O", "uninit_bg",
+		// mkfs.ext4 can populate the filesystem with contents of given
+		// root directory
+		"-d", contentsRootDir,
+	}
+	if label != "" {
+		mkfsArgs = append(mkfsArgs, "-L", label)
+	}
+	mkfsArgs = append(mkfsArgs, img)
+	cmd := exec.Command("fakeroot", mkfsArgs...)
```
mvo5: Maybe a comment why we need fakeroot here? Also we probably don't want/need it if we are already root. I'm mostly thinking of an early boot context (initramfs) in which we *may* need to create a filesystem but have only a very limited selection of tools available. This may not be a concern in practise because it looks like we will always have a base snap available in recover mode but being defensive won't hurt I guess.

gadget/mkfs.go
```diff
@@ -0,0 +1,121 @@
...
+		// default usage type
+		"-T", "default",
+		// disable metadata checksum
+		// XXX: why?
+		"-O", "-metadata_csum",
+		// allow uninitialized block groups
+		"-O", "uninit_bg",
+		// mkfs.ext4 can populate the filesystem with contents of given
+		// root directory
+		"-d", contentsRootDir,
```
mvo5: I have some vague memory that older version of mkfs.ext4 do not support `-d`. IIRC on xenial (core16). I guess that is not a concern but might be worth a comment?

gadget/mkfs_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+	cmdMkfsVfat := testutil.MockCommand(c, "mkfs.vfat", "echo 'override in test'; exit 1")
+	m.cleanup = append(m.cleanup, func() { cmdMkfsVfat.Restore() })
+
+	cmdMcopy := testutil.MockCommand(c, "mcopy", "echo 'override in test'; exit 1")
+	m.cleanup = append(m.cleanup, func() { cmdMcopy.Restore() })
+}
+
+func (m *mkfsSuite) TearDownTest(c *C) {
+	for _, r := range m.cleanup {
```
mvo5: (nitpick) might be easier to just use testutil.BaseTest and m.AddClean(cmdFakeroot.Restore) there

gadget/mkfs_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+	cmdMcopy := testutil.MockCommand(c, "mcopy", "")
+	defer cmdMcopy.Restore()
+
+	err := gadget.MkfsVfat("foo.img", "my-label", d)
+	c.Assert(err, IsNil)
+	c.Assert(cmdMkfs.Calls(), HasLen, 1)
+
+	c.Assert(cmdMcopy.Calls(), DeepEquals, [][]string{
+		{"mcopy", "-s", "-i", "foo.img", filepath.Join(d, "bar"), filepath.Join(d, "foo"), "::"},
```
mvo5: Having an integration test for this at some point would be cool. But probably not in this PR :)

gadget/mkfs.go
bboozzoo: There's a `TODO:..` just above.

gadget/mkfs.go
bboozzoo: Wonder if it would be beneficial to put all image building pieces under a separate package. We have the mkfs wrappers, parted/sfdisk wrapper and offset writer that are needed only in such scenarios.

gadget/mkfs_test.go
```diff
@@ -0,0 +1,188 @@
...
+)
+
+type mkfsSuite struct {
+	testutil.BaseTest
+}
+
+var _ = Suite(&mkfsSuite{})
+
+func (m *mkfsSuite) SetUpTest(c *C) {
+	// fakeroot, mkfs.ext4, mkfs.vfat and mcopy are commonly installed in
```
mvo5: This should call `m.BaseTest.SetUpTest(c)`

mvo5: (not doing it will cause no harm at least today :)

gadget/mkfs.go
vorlonofportland: @mvo5 metadata checksums were unsupported by the e2fsprogs in Ubuntu 16.04 (and therefore in Ubuntu Core 16).  Creating filesystems with metadata checksums (which is the default with the e2fsprogs in Ubuntu 18.04!) would directly lead to a boot failure.

bboozzoo: @vorlonofportland thanks for the input. I'll update the comment accordingly.

PR #6999, opened 2019-06-13
tests: retry installing the failing snap on failover test

sergiocazzolato:
This loop retries the installation until it finishes correctly. The
install process could fail and rollback the installation, see the error:

snap install --dangerous failing.snap
Automatically connect eligible plugs and slots of snap "core" (cannot
finish core installation, there was a rollback across reboot)

### DIFF
--- a/tests/main/failover/task.yaml
+++ b/tests/main/failover/task.yaml
@@ -95,8 +95,11 @@ execute: |
             exit 1
         fi
 
-        # install new target snap
-        snap install --dangerous failing.snap
+        # Retry installing new target snap until it is intalled correctly. The installation
+        # could fail with: cannot finish core installation, there was a rollback across reboot
+        while ! snap install --dangerous failing.snap; do
+            sleep 1
+        done
 
         # use journalctl wrapper to grep only the logs collected while the test is running
         while ! check_journalctl_log "Waiting for system reboot"; do


chipaca:
without your explanation in the standup i wouldn't've understood this, so maybe work on that a little next time :-)

tests/main/failover/task.yaml
```diff
@@ -95,8 +95,11 @@ execute: |
             exit 1
         fi
 
-        # install new target snap
-        snap install --dangerous failing.snap
+        # Retry installing new target snap until it is intalled correctly. The installation
+        # could fail with: cannot finish core installation, there was a rollback across reboot
+        while ! snap install --dangerous failing.snap; do
```
stolowski: Isn't this hiding the real problem we might have?

sergiocazzolato: @stolowski that could be true, initially I though there was a problem with the snap that we are modifying, so I created this workaround. 
Could you confirm that the solution you are implementing would fix this problem too?

stolowski: No, my fix is unrelated to this problem. However it seems this problem of rollback across reboot is now understood and might have a fix soon, @mvo5 has more details.

mvo5: We have fixed a bug about the "rollback accross reboots" just recently - its in edge now. It would be great to check (cc @sergiocazzolato) if this bug is still reproducible with the "edge" snap. There is one more PR in flight (#7020) that may also help/fix this issue.

sergiocazzolato: @mvo5 sure, will test that

PR #7003, opened 2019-06-14
gadget: add writer for offset-write

bboozzoo:
Add writer for deploying offset-write entries to the image


gadget/offset.go
```diff
@@ -0,0 +1,83 @@
...
+func (w *OffsetWriter) Write(out io.WriteSeeker) error {
+	// positioning guarantees that start offset is aligned to sector size
+
+	if w.ps.PositionedOffsetWrite != nil {
+		if err := deployOffsetWrite(out, *w.ps.PositionedOffsetWrite, asLBA(w.ps.StartOffset, w.sectorSize)); err != nil {
+			return err
+		}
+	}
+
+	if !w.ps.IsBare() {
```
stolowski: I think this case is tested by giving and non-giving Filesystem in the tests inputs, is that right?

gadget/offset.go
```diff
@@ -0,0 +1,83 @@
...
+		// only raw content uses offset-writes
+		return nil
+	}
+
+	for _, pc := range w.ps.PositionedContent {
+		if pc.PositionedOffsetWrite == nil {
```
stolowski: Is that case tested?

gadget/offset.go
bboozzoo: Yup

gadget/offset.go
bboozzoo: Yes, see https://github.com/snapcore/snapd/pull/7003/commits/ece80c25f1039e8c3d640d0664c3ceddcaa9be56#diff-341e2c070839d7b2c0b6c99cc764f8d7R45 structure declares offset-write but its content does not.

gadget/offset.go
```diff
@@ -0,0 +1,83 @@
...
+type OffsetWriter struct {
+	ps         *PositionedStructure
+	sectorSize Size
+}
+
+func asLBA(value, sectorSize Size) uint32 {
+	return uint32(value / sectorSize)
+}
+
+func deployOffsetWrite(out io.WriteSeeker, offset Size, value uint32) error {
```
zyga: Nitpick: in the kernel things like that are called `pwrite`, perhaps we could mimic the name?

PR #7008, opened 2019-06-17
cmd/snap-exec: fix snap completion for classic snaps with non /usr/lib/snapd libexecdir

bboozzoo:
Fix completion for classic snaps on systems where snapd libexecdir is not
/usr/lib/snapd. These are for instance Fedora and CentOS, where libexecdir is
set to /usr/libexec/snapd.

When run for completions, snap-exec would always attempt to run the completer
from the path which is correct given the snap gets a mount namespace (i.e. is
not classic).

This worked by accident on systems that use /usr/lib/snapd (eg. Ubuntu, Debian,
Arch), but failed on those using /usr/libexec/snapd.

For instance, on CentOS, one would see:
``` 
$ cmake -G Ni<tab>
/bin/bash: /usr/lib/snapd/etelpmoc.sh: No such file or directory
```

Fixes: https://bugs.launchpad.net/snapd/+bug/1833043

### DIFF
--- a/cmd/snap-exec/main.go
+++ b/cmd/snap-exec/main.go
@@ -201,8 +201,12 @@ func execApp(snapApp, revision, command string, args []string) error {
 		cmdArgs = nil
 	case "complete":
 		fullCmd[0] = defaultShell
+		completionHelper := dirs.CompletionHelperInCore
+		if info.NeedsClassic() {
+			completionHelper = filepath.Join(dirs.SnapMountDir, "core/current", dirs.CompletionHelperInCore)
+		}
 		cmdArgs = []string{
-			dirs.CompletionHelper,
+			completionHelper,
 			filepath.Join(app.Snap.MountDir(), app.Completer),
 		}
 	case "gdb":
--- a/cmd/snap-exec/main.go
+++ b/cmd/snap-exec/main.go
@@ -36,6 +36,7 @@ import (
 
 // for the tests
 var syscallExec = syscall.Exec
+var osReadlink = os.Readlink
 
 // commandline args
 var opts struct {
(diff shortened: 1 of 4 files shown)


cmd/snap-exec/main.go
```diff
@@ -201,8 +201,12 @@ func execApp(snapApp, revision, command string, args []string) error {
 		cmdArgs = nil
 	case "complete":
 		fullCmd[0] = defaultShell
+		completionHelper := dirs.CompletionHelperInCore
+		if info.NeedsClassic() {
+			completionHelper = filepath.Join(dirs.SnapMountDir, "core/current", dirs.CompletionHelperInCore)
```
mvo5: Hm, I think this also needs to take the "snapd" snap into account, i.e. check if there is snapd/current and if so use that. Or should we maybe just use the host system one instead of using the one coming from the snap?

zyga: The location of the completion helpers must be dictated by the re-execution policy. We cannot just use the ones from the core or from snapd snap if the distribution has opted out of the re-exec feature.

bboozzoo: Tweaked this to something similar to `cmd.InternalToolPath()` but without the distro specific shenanigans, to not break when `/etc/os-release` is a file rather than a symlink (eg. on CentOS/AMZN2)

cmd/snap-exec/main_test.go
```diff
@@ -65,6 +66,7 @@ apps:
    BASE_PATH: /some/path
    LD_LIBRARY_PATH: ${BASE_PATH}/lib
    MY_PATH: $PATH
+  completer: you/complete/me
```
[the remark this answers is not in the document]
bboozzoo: Moved the line up a bit.

bboozzoo:
@mvo5 would you like one last pass before landing this?

PR #7009, opened 2019-06-17
gadget: support relative symlinks in device lookup

bboozzoo:
The /dev/disk/by-{partlabel,label} symlink made by udev are often relative. Make
sure that code supports that.


gadget/device_linux.go
```diff
@@ -54,11 +54,23 @@ func FindDeviceForStructure(ps *PositionedStructure) (string, error) {
 	var found string
 	var match string
 	for _, candidate := range candidates {
-		if !osutil.FileExists(candidate) {
```
mvo5: I'm not sure I get the aim 100% but I think this might be written in a slightly simpler way by using "osutil.IsSymlink()". I find the following (or some variation of it) a bit easier to read because its more linear and less nested. but maybe its more subtle and I miss something, in this case we need more tests :) With the change all tests still pass:
```go
                if !osutil.FileExists(candidate) {
			continue
		}
		if !osutil.IsSymlink(candidate) {
			return "", fmt.Errorf("candidate %v is not a symlink", candidate)
		}
		target, err := filepath.EvalSymlinks(candidate)
		if err != nil {
			return "", fmt.Errorf("cannot read device link: %v", err)
		}
		if found != "" && target != found {
                ...
```
(full diff: http://paste.ubuntu.com/p/Jd98SW3Xqk/)

gadget/device_linux.go
```diff
@@ -54,11 +54,23 @@ func FindDeviceForStructure(ps *PositionedStructure) (string, error) {
...
+		st, err := os.Lstat(candidate)
+		if err != nil {
+			if os.IsNotExist(err) {
+				continue
+			}
+			return "", fmt.Errorf("cannot stat device: %v", err)
+		}
+		if (st.Mode() & os.ModeSymlink) == 0 {
+			// /dev/disk/by-label/* and /dev/disk/by-partlabel/* are
+			// expected to be symlink
```
zyga: Is this an implementation detail or an advertised feature of udev?

bboozzoo: It's part of persistent device naming. Though I don't recall seeing any official docs on this aide from bits of information scattered around different wikis, one is here https://wiki.archlinux.org/index.php/Persistent_block_device_naming

PR #7010, opened 2019-06-17
interfaces/docker-support: declare controls-device-cgroup

anonymouse64:
With this PR, the docker-support interface now will disable the device cgroup / udev tagging when it is connected. This is because when attempting to use docker with device specific interfaces such as gpio, or network-control, etc. that enable a device cgroup and udev tagging, container creation fails because dockerd isn't able to create it's own device cgroup for the container as it is already confined by one. A remedy for this is to disable the device cgroup when docker-support is connected.

The eventual solution of this would be for runC to learn how to nest it's device cgroup within the one setup by snapd, but this likely won't happen for some time, and by that point snapd will likely also have migrated to cgroups v2 which is entirely different.

Also update the tests:
* Add test for when docker-support is connected after an interface that triggers the device cgroup and udev rules, network-control, is already connected to ensure that the previously defined udev specifications now disappear
* Add test for when an interface that triggers the device cgroup and udev rules, network-control, is connected after docker-support to ensure that there are still no udev specifications generated
* Create the plug/plugInfos for the various forms of the docker-support interface tested here in SetUpTest with MockConnected{Plug,Slot} to simplify the test bodies to just show the business logic behind connecting various forms of the docker-support

Note @jdstrand, @pedronis and I discussed this in person during Malta and tentatively agreed this was okay to do, due to it's similarity with AWS IoT Greengrass because with a snap of Azure IoT Edge (which uses docker) a primary use case is accessing device hardware using device interfaces that trigger the device cgroup to be enabled.


zyga:
Note that currently snapd does not support the transition where a snap process (or set of processes) were using the device cgroup and due to interface operation, no longer do so. The process will continue to inhabit the cgroup they were placed in, along with any restrictions.

To perform this kind of transition we'd need per-command process tracking (behind experimental branch) and a helper that would move processes in and out of the control group on demand.

anonymouse64:
> Note that currently snapd does not support the transition where a snap process (or set of processes) were using the device cgroup and due to interface operation, no longer do so. The process will continue to inhabit the cgroup they were placed in, along with any restrictions.

Yes, this was also acknowledged for `greengrass-support`, since both AWS IoT Greengrass and Azure IoT Edge both operate as daemons, this was agreed to be okay for now. We also document for AWS IoT Greengrass specifically that the `greengrass-support` should always be connected first before any other interfaces, so I imagine documented the same for Azure IoT Edge should be sufficient in the meantime.

anonymouse64:
The tests seem to have failed due to some network issues, i.e.

```
+ snap install --classic --edge ubuntu-image
2019-06-17T21:39:30Z INFO Waiting for restart...
- Fetch and check assertions for snap "ubuntu-image" (143) (cannot fetch assertion: got unexpected HTTP status code 503 via GET to "https://api.snapcraft.io/api/v1/snaps/assertions/snap-revision/VLrD5YsCsoEb2jrCj7AnLSSABzxYs2UdE8uAWDgNTUZDhGcDl2cHyzCDH5nJAJJs?max-format=0")
```
and 
```
+ snap download --edge core
Fetching snap "core"
Fetching assertions for "core"
error: cannot fetch snap signatures/assertions: cannot fetch assertion: got unexpected HTTP status code 503 via GET to "https://api.snapcraft.io/api/v1/snaps/assertions/snap-declaration/16/99T7MUlRhtI3U0QFgl5mXXESAiSwt776?max-format=3"
```

interfaces/builtin/docker_support.go
```diff
@@ -594,6 +595,16 @@ var (
 	parserFeatures = release.AppArmorParserFeatures
 )
 
+func (iface *dockerSupportInterface) UDevConnectedPlug(spec *udev.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	var controlsDeviceCgroup bool
+	_ = plug.Attr("controls-device-cgroup", &controlsDeviceCgroup)
+	if controlsDeviceCgroup {
+		spec.SetControlsDeviceCgroup()
+	}
```
jdstrand: I do recall discussing this and saying I thought it was fine to use spec.SetControlsDeviceCgroup() in this interface, but I'm a little surprised you aren't just setting `controlsDeviceCgroup: true,` (like we did in greengrass) and instead using an interface attribute. Using an attribute suggests that you expect a mixture of cases for when snaps plugging this interface sometimes want the device cgroup and sometimes don't. Is there ever a time when the cgroup is in effect that snaps would want the device cgroup? It seems like the plugging snap would always die when other interfaces are used, so I'm confused why this needs to be conditional on the snap developer's choice. Perhaps I am missing something?

Put another way, if something plugs docker-support, should we assume that it will manage cgroups and therefore use `controlsDeviceCgroup: true,`. I think we can, but again, I maybe am missing something.

anonymouse64: Well we have a specific use case with Azure IoT Edge in that it uses docker/moby for container management on specifically IoT devices, and thus needs to manage cgroups for containers (called modules in Azure IoT Edge speak) that interact with devices such as gpio, etc.

However, I don't see why we need to also enable this for the other users of the docker-support interface such as the docker snap itself or microk8s, etc. From my notes I think it was @pedronis who recommended using an attribute for it in Malta. However, at least for the docker snap currently, no interfaces are plugged that trigger udev/device cgroup currently so it probably won't trigger any new issues with the docker snap.

jdstrand: > However, at least for the docker snap currently, no interfaces are plugged that trigger udev/device cgroup currently

Right. I think thus far we've been fortunate or the docker snap was written in such a way as to specifically avoid this. My understanding of docker is that use of cgroups is not optional (please correct me) but even it it was, they are always expected to work when configured. I maintain it should be unconditional since:
* this will avoid a potential future bug if the docker snap adds a new interface that triggers them (or more likely, an existing interface that docker plugs starts to trigger them)
* there is no regression in security (and the lack of the attribute is doing nothing for the real-world docker snap)
* docker-support policy is advisory only (it is trivial for something plugging docker-support to own the system, with or without the device cgroup in effect)

I appreciate the desire to limit this to where it is needed, but this interface is the *docker*-support interface and I can't imagine a docker install that would be fully functional without being able to modify cgroups.

anonymouse64: Okay, I will update this to be unconditional

interfaces/builtin/docker_support.go
```diff
@@ -629,6 +636,11 @@ func (iface *dockerSupportInterface) BeforePreparePlug(plug *snap.PlugInfo) erro
 			return fmt.Errorf("docker-support plug requires bool with 'privileged-containers'")
 		}
 	}
+	if v, ok := plug.Attrs["controls-device-cgroup"]; ok {
+		if _, ok = v.(bool); !ok {
+			return fmt.Errorf("docker-support plug requires bool with 'controls-device-cgroup'")
```
zyga: I think the message is a little confusing, perhaps more so to reviewers than to users. It seems to say that a plug must always have the attribute named `controls-device-cgroup`. I think the desired message is that if present, the attribute must be of type bool.

This is not a blocker but if I'm reading this right, please adjust this in a follow up.

anonymouse64: actually this is no longer an attribute as per @jdstrand, so this whole block shouldn't be here, my apologies

anonymouse64:
Merging master here in the hope that it fixes the spread tests...

interfaces/builtin/docker_support_test.go
```diff
@@ -26,33 +26,70 @@ import (
...
+  plugs: 
+   - docker-support
+   - network-control
+`
+
+const dockerSupportPrivilegedContainersFalseMockPlugSnapInfoYaml = `name: docker
+version: 1.0
+plugs:
+ privileged:
+  interface: docker-support
```
jdstrand: In the previous tests we used `privileged-containers: false`.

anonymouse64: Ah thanks for catching this. I added it back

jdstrand:
Still LGTM

PR #7019, opened 2019-06-20
interfaces: miscellaneous policy updates

jdstrand:
- interfaces/removable-media: allow 'l' on files in /media and /mnt
- interfaces/network-observe: allow read access to /etc/iproute2/**
- interfaces/network-control: allow read to /etc/iproute2/** and write to .d/*
- interfaces/network-setup-observe: allow read on /etc/systemd/network/{,**}
- interfaces/network-setup-control: allow rw on /etc/systemd/network/{,**}
- interfaces/system-observe: allow read on /proc/*/{cgroup,wchan}
    Allow read access to `/proc/<pid>/cgroup` for top/ps/system monitor
    utilities to also display cgroups.
    
    Also allow read access to `/proc/<pid>/wchan` for showing ksym in waiting.
    Note that we already allow /proc/*/stat and at column 35 the wchan field
    is available to processes running with PTRACE_MODE_READ_FSCREDS which on
    kernels lower than (at least) 4.4, shows the "wait channel" (ie, the
    location in the kernel where the process is sleeping) as an address.
    Depending on kaslr, kptr_restrict, access to /proc/kallsyms and/or
    access to System.map (or from well-known distro kernels), this address
    allows mapping the address to the kernel symbol (indeed, this is how ps
    used to operate). 4.4 (at least) and later kernels no longer expose an
    address in the wchan field since it could be used to calculate the kaslr
    offset. These kernels will instead populate the wchan field with 0 or 1
    depending on if the channel is blocked or not (ie, 0 if unblocked).
    
    The 2.5 kernel introduced the `/proc/<pid>/wchan` file which will show the
    kernel symbol in which the process is waiting without leaking addresses,
    having to process the System.map or /proc/kallsyms and works regardless
    of what is in `/proc/<pid>/stat`. Legitimate programs (eg, ps) have long
    since moved to `/proc/<pid>/wchan` instead. Access to this file also
    requires the process is running with PTRACE_MODE_READ_FSCREDS.
    
    The manually connected system-observe interface already grants 'ptrace
    (read)' for other ps accesses, so now adjust system-observe to also
    allow access to `/proc/<pid>/wchan`.
    
    Additionally, expand comments for 'ptrace (read)' and why system-observe
    doesn't allow 'capability syslog' ir access to /proc/kallsyms, etc.
- interfaces/kernel-module-{control,observe}: comment on kernel addresses
- interfaces/spi: add more test cases for expanded regex
- interfaces/raw-usb: allow NETLINK_KOBJECT_UEVENT for libusb_open()
- interfaces/cpu-control: allow read on /sys/devices/system/cpu/smt/*
- interfaces/browser-support: allow read to /run/udev/data/b230:* (zvol)
- interface: add 'deny capability ptrace' to ptraceTraceDenySnippet
    Some applications (eg, Discord) trigger a capability ptrace denial but
    otherwise work fine. Silence this noisy denial by default unless
    plugging and interfaces that spec.UsesPtraceTrace(). Note, no interfaces
    currently allow 'capability ptrace' so spec.UsesPtraceTrace() wasn't
    added to any existing interfaces.
- interfaces/network-setup-{control,observe}: update for netplan generate
    Add several rules to network-setup-control for 'netplan generate' as
    driven by cloud-init (which stages netplan):
    
        # cloud-init clean --logs
        # cloud-init init --local
        # cloud-init init
    
    The systemd-networkd accesses are because netplan transitions from
    NetworkManager to systemd-networkd. Also add corresponding read access
    to network-setup-observe.
    
    This intentionally does not implement rules for 'netplan apply' since
    netplan should grow support for working with snapd to run 'netplan
    apply'.
    
    References:
    - https://forum.snapcraft.io/t/network-control-network-setup-control-doesnt-enable-running-netplan-generate-on-classic/11126
    - https://github.com/snapcore/snapd/pull/5915#issuecomment-504413932
- interfaces/apparmor: allow reading session user state files by default
    These files are root:root and world readable and don't contain sensitive
    information (much of it is already exposed via getent passwd).
    
    References:
    https://forum.snapcraft.io/t/dbus-run-session-not-working-with-core18/9509
- interfaces/apparmor: allow read of /usr/share/dbus-1/{,system-}services
    dbus-run-session tries to read /usr/share/dbus-1/{,system-}services/*
    but continues without trouble. These files come from the base runtime of
    the snap and are therefore not sensitive, so allow the files by default.
- interfaces/hardware-observe: accesses for systemd-detect-virt
    References:
    - https://launchpad.net/bugs/1831473
- interfaces: allow reading /proc/sys/fs/pipe-max-size by default
- interfaces/system-observe: allow read on /proc/$pid/comm
    /proc/$pid/comm is a subset of /proc/$pid/cmdline and we allow cmdline
    in system-observe. We already allow /proc/$pid/comm with owner match by
    default, so adjust the comment in the default policy.
    
    References:
    - https://launchpad.net/bugs/1831457
- interface/camera: allow /dev/vchiq for rpi cameras (Broadcom VideoCore)
    /dev/vchiq is shared by VideoCore graphics (EQL) and VideoCore camera.
    While the opengl interface provides access to /dev/vchiq, there are
    applications that do not use GL but need access to the camera. Update
    the camera interface to include /dev/vchiq and update the comments in
    the opengl interface to mentioned the device has shared functionality.


jdstrand:
Note to reviewers, I'm done adding stuff to this PR and it is ready to review.

jdstrand:
The only failing test is google:ubuntu-18.04-64:tests/main/snap-seccomp-syscalls, which is unrelated to this PR. I created a topic here: https://forum.snapcraft.io/t/spread-test-failing-due-to-new-syscalls/12018

interfaces/builtin/system_observe.go
```diff
@@ -38,7 +38,15 @@ const systemObserveConnectedPlugAppArmor = `
...
-# since since ps doesn't actually need to trace other processes.
+# since since ps doesn't actually need to trace other processes. Note this
+# allows a number of accesses (assuming the associated /proc file is allowed),
+# such as various memory address locations and esp/eip via /proc/*/stat,
+# /proc/*/mem, /proc/*/personality, /proc/*/stack, /proc/*/syscall,
+# /proc/*/timerslack_ns and /proc/*/wchan (see man proc).
+#
+# Some files like /proc/kallsyms (but anything using %pK format specifier) need
+# 'capability syslog' when /proc/sys/kernel/kptr_restrict=1, but we
+# intentionally do not allow since it could be used to default KASLR.
```
zyga: I think you meant to say `defeat` here
```suggestion
# intentionally do not allow since it could be used to defeat KASLR.
```

jdstrand: Done

jdstrand:
Thanks for the reviews and merge!

PR #7022, opened 2019-06-21
interfaces/bluetooth-control: add mtk BT device node

kubiko:
Extending allowed vevice nodes for bluetooth control as
some mtk drivers (mt7668) expose /dev/stpbt, /dev/stpbtfwlog device node
instead of usual /dev/vhci

### DIFF
--- a/interfaces/builtin/bluetooth_control.go
+++ b/interfaces/builtin/bluetooth_control.go
@@ -50,6 +50,9 @@ const bluetoothControlConnectedPlugAppArmor = `
 
   # Requires CONFIG_BT_VHCI to be loaded
   /dev/vhci                       rw,
+
+  # Device nodes for MTK combo driver
+  /dev/stpbt                      rw,
 `
 
 const bluetoothControlConnectedPlugSecComp = `


kubiko:
@jdstrand any concerns from security?

interfaces/builtin/bluetooth_control.go
```diff
@@ -49,7 +49,10 @@ const bluetoothControlConnectedPlugAppArmor = `
   /sys/devices/**/bluetooth/**    rw,
 
   # Requires CONFIG_BT_VHCI to be loaded
+  # some mtk drivers (mt7668) expose stpbt, stpbtfwlog device node
```
bboozzoo: I'd put this in a separate section like so:
```
  # Requires CONFIG_BT_VHCI to be loaded
  /dev/vhci                       rw,

  # Device nodes for MTK combo driver
  /dev/stpbt                      rw,
  /dev/stpbtfwlog                 rw,
```

bboozzoo: I could only find this source code of the driver: https://github.com/MotorolaMobilityLLC/kernel-mtk/blob/f008e679440bbc0f77229df9d6520b2831bee8f8/drivers/misc/mediatek/connectivity/common/common_main/linux/stp_chrdev_bt.c

It seems like it's not really adding anything that would match with `SUBSYSTEM=="bluettooth"` rule. Have you checked that the access is not blocked by device cgroup?

zyga:
I'm somewhat worried about the potential explosion of random devices like that. Do we know why this driver doesn't follow established rules?

interfaces/builtin/bluetooth_control.go
```diff
@@ -50,6 +50,10 @@ const bluetoothControlConnectedPlugAppArmor = `
 
   # Requires CONFIG_BT_VHCI to be loaded
   /dev/vhci                       rw,
+
+  # Device nodes for MTK combo driver
+  /dev/stpbt                      rw,
+  /dev/stpbtfwlog                 rw,
```
jdstrand: /dev/stpbt seems fine. I can't find any reference to /dev/stpbtfwlog. What is it? If it is 'log', why does it need write? 'fw' seems like it could be 'firmware' or 'firewall'...

jdstrand: I looked at some code and, again, /dev/stpbt seems fine. I didn't spend a whole lot of time with the code so don't take my comments as authoritative, but it seems clear that /dev/stpbtfwlog is the device for working with the firmware of the hardware (the 'log' in 'stpbtfwlog' seems a misnomer) and as such, it shouldn't be included in bluetooth-control.

We only have one existing interface for working with any sort of firmware: fwupd. This interface "allows operating as the fwupd service" and is not appropriate for /dev/stpbtfwlog.

The normal pattern is that we let the gadget and kernel sort out firmware so that snaps themselves don't have to and I wonder if this would be sufficient for your use case. It does appear that this chip supports being inserted as a "card" that can be ejected which then causes module unload, so perhaps if you need to support inserting and removing a card, the gadget and kernel configuration wouldn't be the right place since on insert the device is brought up and patched with the firmware that must be loaded (that said, the kernel knows how to do this with other firmware, why not this device?). It appears that some chips might support different firmwares (eg, single purpose firmware, combo firmware (eg, bt and wifi)), so this might also complicate things.

If this can't be solved in the kernel/gadget, I see two potential paths forward:
1. create a firmware backend in snapd that works like the kmod backend, and have bluetooth-control call into this backend to load the firmware (bluetooth-control may need to also call into kmod)
2. create something along the lines of a 'firmware-control' that operates similarly to kernel-module-control that allows a snap to load firmware. This interface would require an installation constraint.

'1' might suffer from the same issues as kernel/gadget since snapd would have to learn about when to load firmware on card insert/module load and how to poweroff/disable the device on module unload. If the 'combo mode' is supported, snapd would need to be configurable on what to load. This all seems a bit messy and perhaps snapd is not at the right level (though, it could, in theory, add a udev rule (the code I looked at appears to be GPLv2 so it is eligible for integrating with udev) that calls a tool to do all this (ie, the same tool that prompted this PR)).

'2' is a workable solution but the interface would need design. Should the interface try to be all encompassing and allow loading all kinds of firmware? Should it operate like system-files where you specify the devices, sysfs, proc entries, etc via interface attributes? Should it be an interface with predefined attributes (eg, "btmtk", "intel...") that represent different snippets? Should it be different interfaces for each (eg, firmware-control-btmtk, firmware-control-intel, ...)?

I suppose it is possible to simply use system-files to grant access to /dev/stpfwlog. We've not yet granted use of system-files for things outside of /etc, but in theory it could be done. The interface would have to somehow grow udev backend support which I think would complicate the interface and ultimately, we know exactly what to grant and we know that the devices will be in more than one snap/device, so having a specific firmware interface seems best.

I've asked @pedronis to review this PR to weigh in on stpfwlog. If you remove stpfwlog, then we could move forward as the PR would become straightforward.

kubiko: thank for looking into this @jdstrand 
I dropped `stpfwlog` device node, as we have other options to support this. I have used system-files already before this PR as temporary measures, so I can can confirm it does resolve the problem. And since device connect to dedicated brand store, this should be straight forward permissions wise.
So I reduced PR to  `stpf` device node only

jdstrand: > I have used system-files already before this PR as temporary measures, so I can can confirm it does resolve the problem.

I'm surprised this works since system-files doesn't do anything to add the device to the device cgroup. Is your snap plugging greengrass-support by chance? (if so, that would explain it because it sets `controlsDeviceCgroup: true`

jdstrand: > > I have used system-files already before this PR as temporary measures, so I can can confirm it does resolve the problem.
> 
> I'm surprised this works since system-files doesn't do anything to add the device to the device cgroup. Is your snap plugging greengrass-support by chance? (if so, that would explain it because it sets `controlsDeviceCgroup: true`

Oh, stpfwlog is probably part of `SUBSYSTEM=="bluetooth"`, which explains it.

jdstrand:
With stpfwlog removed, no need for @pedronis' review. There are 3 approvals. Commiting.

PR #7023, opened 2019-06-21
interfaces: spi: update regex rules to accept spi nodes like spidev12…

kubiko:
…345.0

updating allowed spi device node names
There are platforms (Mediatek) where spi device node is not single digit
e.g. /dev/spidev34566.0

### DIFF
--- a/interfaces/builtin/spi.go
+++ b/interfaces/builtin/spi.go
@@ -55,7 +55,7 @@ func (iface *spiInterface) StaticInfo() interfaces.StaticInfo {
 	}
 }
 
-var spiDevPattern = regexp.MustCompile("^/dev/spidev[0-9].[0-9]+$")
+var spiDevPattern = regexp.MustCompile(`^/dev/spidev[0-9]+\.[0-9]+$`)
 
 func (iface *spiInterface) path(slotRef *interfaces.SlotRef, attrs interfaces.Attrer) (string, error) {
 	var path string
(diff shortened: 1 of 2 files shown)


interfaces/builtin/spi.go
```diff
@@ -55,7 +55,7 @@ func (iface *spiInterface) StaticInfo() interfaces.StaticInfo {
 	}
 }
 
-var spiDevPattern = regexp.MustCompile("^/dev/spidev[0-9].[0-9]+$")
+var spiDevPattern = regexp.MustCompile("^/dev/spidev[0-9]+.[0-9]+$")
```
chipaca: AIUI the period is a literal. As such, can you escape it in the regexp, please? I.e. something like

    `^/dev/spidev[0-9]+\.[0-9]+$`

(note the backticks to avoid having to double up the backslash).

kubiko: @chipaca thanks for spotting this!
PR is now updated

interfaces/builtin/spi.go
```diff
@@ -55,7 +55,7 @@ func (iface *spiInterface) StaticInfo() interfaces.StaticInfo {
 	}
 }
 
-var spiDevPattern = regexp.MustCompile("^/dev/spidev[0-9].[0-9]+$")
+var spiDevPattern = regexp.MustCompile(`^/dev/spidev[0-9]\+.[0-9]+$`)
```
bboozzoo: Did you mean `^/dev/spidev[0-9]+.[0-9]+$` ?

kubiko: @bboozzoo yeah mistake, should be 
`^/dev/spidev[0-9]+\.[0-9]+$`

PR #7024, opened 2019-06-24
many: add strutil.ElliptLeft, use it for shortening cohorts

chipaca:
When I first wrote some cohort code, for 'snap download', I very
carefully pointed out that

    // cohort keys are really long, and the rightmost bit being the
    // interesting bit, so ellipt the rest

and then wrote a private helper

    // TODO: maybe move this to strutil next to ElliptRight

When I then moved to write the rest of the cohort shortening code I
forgot about this and just used ElliptRight (which throws away the
interesting bit, and keeps the bit that changes less).

This makes moves `elliptLeft` to `strutil`, and uses that in all
cohort-shortening code.

HTH, HAND!


strutil/strutil.go
```diff
@@ -209,3 +209,19 @@ func ElliptRight(str string, n int) string {
...
+
+// ElliptLeft returns a string that is at most n runes long,
+// replacing the first rune with an ellipsis if necessary. If N is less
+// than 1 it's treated as a 1.
+func ElliptLeft(str string, n int) string {
+	if n < 1 {
+		n = 1
+	}
+	// this is expensive; look into a cheaper way maybe sometime
+	rstr := []rune(str)
```
bboozzoo: Maybe checking `utf8.RuneCountInstring(str) < n` before making a `[]rune` would be lest expensive. Or maybe not? The cost is probably negligible given where it's called.

chipaca: I need to measure (note `ElliptRight` does do that), but it's essentially doing the decode twice.

chipaca:
@mvo I used Right instead of Left because ... Left didn't exist :-)
The first three bytes are always MSB as they are a base64-encoded 1, space, unix timestamp in seconds:
```
snap create-cohort test-snapd-tools http hello-world
cohorts:
  hello-world:
    cohort-key: MSBidVBLVUQzVEtxQ09nTEVqakh4NWtTaUNwSXM1Y011USAxNTYxMzgxMDQ4IGMxY2FlYzlkZTY1NTA0NzUyMzg3ZTU0ZmRhNGZkMTRmNTA0MDc5NWYwZjk0ZDc5YmVmNTgxMGM3NGVjM2JmOTM=
  http:
    cohort-key: MSBUSkVmZ2dOaGdFSjRYS0o4bzdhaHN2UmtsejVrUks1dyAxNTYxMzgxMDQ4IDhjZTZlM2Y0M2ExMjliNmQ5MTQxOWIyZTQzMDE5NzYwYzQ3OTg3ZmFiM2ViOGMxZDVjNGFkZjk5Y2Y2ZWY4YTg=
  test-snapd-tools:
    cohort-key: MSBlRmU4QlRSNUw1VjlGN3lIZU1BUHhrRXIyTmRVWE10dyAxNTYxMzgxMDQ4IDkxNGMxNzRmMmQwNmUzMDQyNThhZjU2NzIyNzRmMDg1YTQ0Mzk0OTRhYTE0ZjRlNzliYjIzMTBkNTY5MGRiNGY=

the last character is often `=` and I almost wrote a thing that pruned that, but thought it was overkill.

PR #7028, opened 2019-06-25
systemd: add InstanceMode enumeration to control which systemd instance to control

jhenstridge:
This has been split out from PR #6954.  On a desktop system, there will usually be multiple instances of systemd running: one system instance as pid 1, and one for each logged in user on the system.  All can be configured in roughly the same way through the `systemctl` command, so it seems reasonable to control them through the same Go API.

This PR adds an `InstanceMode` argument to `systemd.New`, taking one of the following arguments:

* `SystemMode`: this was previously the only mode offered by the API.
* `UserMode`: this controls the user mode systemd instance for the current user.
* `GlobalUserMode`: this edits configuration used by all user mode instances on the system.  As this doesn't control a single instance, some operations do not make sense and are caught with `panic()`.

`SystemMode` and `GlobalUserMode` are both useful within snapd, while `UserMode` will be useful within the user session agent (which runs in the context of the user, so is allowed to communicate with that user's systemd instance).

### DIFF
--- a/systemd/systemd.go
+++ b/systemd/systemd.go
@@ -202,13 +202,36 @@ type reporter interface {
 }
 
 // New returns a Systemd that uses the given rootDir
-func New(rootDir string, rep reporter) Systemd {
-	return &systemd{rootDir: rootDir, reporter: rep}
+func New(rootDir string, mode InstanceMode, rep reporter) Systemd {
+	return &systemd{rootDir: rootDir, mode: mode, reporter: rep}
 }
+
+// InstanceMode determines which instance of systemd to talk to
+type InstanceMode int
+
+const (
+	SystemMode InstanceMode = iota
+	UserMode
+	GlobalUserMode
+)
 
 type systemd struct {
 	rootDir  string
 	reporter reporter
+	mode     InstanceMode
+}
+
+func (s *systemd) systemctl(args ...string) ([]byte, error) {
+	switch s.mode {
+	case SystemMode:
+	case UserMode:
+		args = append([]string{"--user"}, args...)
+	case GlobalUserMode:
+		args = append([]string{"--user", "--global"}, args...)
+	default:
+		panic("unknown InstanceMode")
+	}
+	return systemctlCmd(args...)
 }
 
 // DaemonReload reloads systemd's configuration.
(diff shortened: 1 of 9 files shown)


systemd/systemd.go
```diff
@@ -202,17 +202,52 @@ type reporter interface {
...
+	default:
+		panic("unknown InstanceMode")
+	}
+	return systemctlCmd(args...)
 }
 
 // DaemonReload reloads systemd's configuration.
 func (s *systemd) DaemonReload() error {
+	if s.mode == GlobalUserMode {
+		panic("cannot call daemon-reload with GlobalUserMode")
```
zyga: Should this be an error instead? I'm a bit uneasy about panics. I was thinking about this briefly and was wondering if we could have a collection of types related to systemd, e.g. `systemd.SystemInstance`, `systemd.UserInstance` and `systemd.SettingsForAllUsers` (names are obviously provisional), where each type would only implement the set of operations that make sense and the actual interactions would be through a set of interfaces, exposing them.

I'm not -1 on the PR in the current form, just wanted to voice the idea where a particular error path would not be necessary through the use of the type system.

systemd/systemd.go
```diff
@@ -222,43 +257,49 @@ func (s *systemd) DaemonReload() error {
...
+	_, err := s.systemctl(append([]string{"start"}, serviceNames...)...)
 	return err
 }
 
 // StartNoBlock starts the given service or services non-blocking
-func (*systemd) StartNoBlock(serviceNames ...string) error {
-	_, err := systemctlCmd(append([]string{"start", "--no-block"}, serviceNames...)...)
+func (s *systemd) StartNoBlock(serviceNames ...string) error {
+	if s.mode == GlobalUserMode {
+		panic("cannot call start with GlobalUserMode")
```
zyga:
```suggestion
		panic("cannot call start --no-block with GlobalUserMode")
```

systemd/systemd.go
chipaca: maybe we want different types, I like the idea, but that's a refactor we can do at our leisure.

chipaca:
Things are green, and what remains can be addressed in passing in a followup. Merging.

systemd/systemd.go
jhenstridge: None of the existing uses of this API really have a separation between instantiating the `Systemd` instance and calling its methods, and I'm not sure whether there are many cases where this wouldn't be true.  As this is more of a programming error rather than a runtime error, panics seemed appropriate.

As an alternative to three types, another option would be to split the methods into two types: those that manipulate configuration files vs. those that communicate with a systemd instance.

PR #7029, opened 2019-06-25
udevmon: allocate much larger buffer for event data based on reference implementation of systemd

stolowski:
Fix for LP:1833707. The fix for upstream go-udev package is waiting for the review, this PR fixes the bug in our fork go-dev. 

Fix for upstream: https://github.com/pilebones/go-udev/pull/17

Pre-allocate large buffer for event data based on reference implementation from systemd. Buffer creation has been moved to the Connect method of go-udev Conn, so it's taking place just once. Remove existing resizing logic as it was based on wrong assumption regarding MSG_PEEK semantics - this flag simply returns data from the beginning of the receive queue without removing that data from the queue, so a subsequent receive call will return the same data, but it is not for peeking for message size.

### DIFF
--- a/osutil/udev/netlink/conn.go
+++ b/osutil/udev/netlink/conn.go
@@ -20,6 +20,7 @@ const (
 type NetlinkConn struct {
 	Fd   int
 	Addr syscall.SockaddrNetlink
+	buf  []byte
 }
 
 type UEventConn struct {
@@ -47,6 +48,10 @@ func (c *UEventConn) Connect(mode Mode) (err error) {
 		syscall.Close(c.Fd)
 	}
 
+	// this must be rather large for some events, see the reference implementation from systemd:
+	// https://github.com/systemd/systemd/blob/0d92a3088a50212f16bf72672832b2b61dfca551/src/udev/udevadm-monitor.c#L73
+	c.buf = make([]byte, 128*1024*1024)
+
 	return
 }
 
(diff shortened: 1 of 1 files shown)


bboozzoo:
```
Invalid PR title: "Allocate much larger buffer for event data - 128kB, based on referenc…"
```

osutil/udev/netlink/conn.go
```diff
@@ -47,6 +48,10 @@ func (c *UEventConn) Connect(mode Mode) (err error) {
 		syscall.Close(c.Fd)
 	}
 
+	// this must be rather large for some events, see the reference implementation from systemd:
+	// https://github.com/systemd/systemd/blob/0d92a3088a50212f16bf72672832b2b61dfca551/src/udev/udevadm-monitor.c#L73
```
bboozzoo: AFAICT the systemd code changes socket receive buffer size at the kernel level via `setsockopt(SOL_SOCKET, SO_RECVBUF)`. This isn't exactly the same as we do. I would like to learn more about the problem here. 

Also what udevadm does is interesting. Looks like the default buffer size on my machine is 52 pages:
```
In [13]: s = socket.socket(socket.AF_NETLINK, socket.SOCK_RAW)                         

In [14]: s.getsockopt(socket.SOL_SOCKET, socket.SO_RCVBUF)                             
Out[14]: 212992
```
Wonder why they'd force it to 128k rather than have at least 128k.

stolowski: Interesting, FWTW what we had so far was Getpagesize(), which on my system is 4096 bytes.

stolowski: It's actually 128MB, not 128kB, my bad.

stolowski:
Marking blocked, per discussion with Maciej on IRC it needs more investigation.

stolowski:
Closing, problem caused by low value of /proc/sys/net/core/rmem_default, needs a different fix.

PR #7030, opened 2019-06-25
gadget: add a wrapper for generating partitioned images with sfdisk

bboozzoo:
Add a wrapper for generating partitioned images using sfdisk.


gadget/partition.go
```diff
@@ -0,0 +1,118 @@
...
+
+		fmt.Fprintf(script, "\n")
+	}
+	return runSfdisk(image, script.String())
+}
+
+func runSfdisk(image string, script string) error {
+	logger.Debugf("running sfdisk script:\n%v", script)
+	cmd := exec.Command("sfdisk", image)
+	cmd.Stdin = bytes.NewBufferString(script)
```
stolowski: Would it make sense to dump the script to a file in case something goes wrong and more insight is needed? The file could be removed if all goes well.

bboozzoo: It's logged to stderr when debugging is enabled. Would that work?

gadget/partition.go
```diff
@@ -0,0 +1,118 @@
...
+	for _, ps := range pv.PositionedStructure {
+		if ps.Type == "bare" || ps.Type == "mbr" {
+			continue
+		}
+
+		nonBareIdx++
+
+		start := asSector(ps.StartOffset)
+		size := asSector(ps.Size)
+		fmt.Fprintf(script, "%v : start=%v, size=%v", nonBareIdx, start, size)
```
zyga: The `sfdisk(8)` manual page doesn't recommend using the `device :` syntax unless one strictly has to. I would say we should drop it.

gadget/partition.go
bboozzoo: The script is now logged to stderr via logger.Noticef() on failure.

PR #7031, opened 2019-06-25
debian: re-enable systemd environment generator

mvo5:
This re-enables the systemd environment generator to fix bug
LP:#1771858. It also adds a break for the versions of systemd
that are not fixed.

This should be carefully tested as last time it broke the world because of a systemd bug.

### DIFF
--- a/packaging/ubuntu-16.04/control
+++ b/packaging/ubuntu-16.04/control
@@ -70,7 +70,7 @@ Depends: adduser,
          ${misc:Depends},
          ${shlibs:Depends}
 Replaces: ubuntu-snappy (<< 1.9), ubuntu-snappy-cli (<< 1.9), snap-confine (<< 2.23), ubuntu-core-launcher (<< 2.22), snapd-xdg-open (<= 0.0.0)
-Breaks: ubuntu-snappy (<< 1.9), ubuntu-snappy-cli (<< 1.9), snap-confine (<< 2.23), ubuntu-core-launcher (<< 2.22), snapd-xdg-open (<= 0.0.0), ${snapd:Breaks}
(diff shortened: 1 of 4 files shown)


xnox:
NACK until after https://bugs.launchpad.net/cloud-images/+bug/1791691 is fixed in bionic and cosmic.

bboozzoo:
labeled as blocked based on feedback from @xnox

rbalint:
@mvo5 @bboozzoo @xnox I'm fixing initramfs-tools, adding a Breaks: to next systemd upload, then with the Breaks: systemd (<< 237-3ubuntu10.25) this will be OK, to go in, too.

rbalint:
@mvo5 @bboozzoo @xnox it seems the initramfs fix is not needed, this snapd and systemd 237-3ubuntu10.24 work ok with and without initramfs. /usr/local/sbin:/usr/local/bin are present in the PATH.
I suggest moving ahead with uploading the snapd fix to bionic-proposed.

rbalint:
@mvo5 @bboozzoo @xnox 237-3ubuntu10.24 is released to updated, snapd can be updated in the archive, too.

xnox:
Also needs initramfs-tools tasks to be done on https://bugs.launchpad.net/ubuntu/+source/snapd/+bug/1814355

rbalint:
@xnox are you sure? Current initramfs worked ok for me.

mvo5:
<xnox> with new systemd which is in -updates, everything works correctly, and i cannot reproduce any bugs anymore. I.e. usr/local stays, the path is normal, /snapd gets appended, with or without initrd. so i think the systemd patches fixed all regressions, and there are no further changes needed to like initramfs-tools or anywhere else.

PR #7032, opened 2019-06-25
snap-confine: move seccomp load after permanent privilege drop

jdstrand:
Because snap-confine would load its seccomp profiles before permanently
dropping privileges, the seccomp profile needed to have rules for
setresuid() (glibc's implementation of seteuid()), setuid() and setgid()
to the dropped to user. Since traditionally we allowed use of all these
syscalls by default (letting AppArmor capability rules and DAC mediate
the access), this ordering has not been an issue. With the introduction
of privilege dropping in snaps, we no longer allow use of the bare
syscalls when the snap specifies it wants to privilege drop, causing
non-root snap commands to fail to run when snap-confine attempted to
drop privileges (this doesn't affect root snap commands since we already
allowed these calls in the privilege dropping seccomp snippet).

One approach to handling this would be for snap-confine to implicitly
add the following rules:

  setuid UID
  setgid GID
  setresuid -1 UID -1

where UID is calculated at runtime as the dropped-to user. Since this
must happen at runtime, snap-seccomp cannot be leveraged and we would
need to either attach a 3rd seccomp filter (the main one, global plus
this new one) or we would need to monkey patch the main profile prior to
loading into the kernel.

A cleaner approach is to instead rearrange the ordering of calls such
that we permanently privilege drop prior to loading the seccomp
profiles. Normally setuid/setgid not only permanently drops the UID/GID,
but also clears the capabilities bounding sets (see "Effect of user ID
changes on capabilities" in 'man capabilities'). To load a seccomp
profile, we need either CAP_SYS_ADMIN or PR_SET_NO_NEW_PRIVS. Since NNP
causes issues with AppArmor and exec transitions in certain snapd
interfaces, we will regain CAP_SYS_ADMIN temporarily after permanently
dropping privileges so we may then load the seccomp profiles, then drop
it after. This provides a small security benefit since we permanently
drop earlier in snap-confine and keep only the capabilities required to
load the seccomp policy.

This PR can land alone and is part of a series of PRs that break up https://github.com/snapcore/snapd/pull/6681


cmd/snap-confine/snap-confine.c
```diff
@@ -329,8 +331,12 @@ int main(int argc, char **argv)
 	// Who are we?
 	uid_t real_uid, effective_uid, saved_uid;
 	gid_t real_gid, effective_gid, saved_gid;
-	getresuid(&real_uid, &effective_uid, &saved_uid);
-	getresgid(&real_gid, &effective_gid, &saved_gid);
+	if (getresuid(&real_uid, &effective_uid, &saved_uid) != 0) {
```
zyga:
```suggestion
	if (getresuid(&real_uid, &effective_uid, &saved_uid) < 0) {
```

cmd/snap-confine/snap-confine.c
```diff
@@ -329,8 +331,12 @@ int main(int argc, char **argv)
...
+		die("getresuid failed");
+	}
+	if (getresgid(&real_gid, &effective_gid, &saved_gid) != 0) {
```
zyga:
```suggestion
	if (getresgid(&real_gid, &effective_gid, &saved_gid) < 0) {
```

cmd/snap-confine/snap-confine.c
```diff
@@ -414,18 +420,49 @@ int main(int argc, char **argv)
...
 	}
+
+	// Normally setuid/setgid not only permanently drops the UID/GID, but
+	// also clears the capabilities bounding sets (see "Effect of user ID
+	// changes on capabilities" in 'man capabilities'). To load a seccomp
+	// profile, we need either CAP_SYS_ADMIN or PR_SET_NO_NEW_PRIVS. Since
+	// NNP causes issues with AppArmor and exec transitions in certain
+	// snapd interfaces, keep CAP_SYS_ADMIN temporarily when we are
+	// permanently dropping privileges.
+	if (getresuid(&real_uid, &effective_uid, &saved_uid) != 0) {
```
zyga: Why are we re-probing this? It is already loaded in main on line 334 above?

cmd/snap-confine/snap-confine.c
```diff
@@ -414,18 +420,49 @@ int main(int argc, char **argv)
...
+		die("getresuid failed");
+	}
+	debug("ruid: %d, euid: %d, suid: %d",
+	      real_uid, effective_uid, saved_uid);
+	struct __user_cap_header_struct hdr = { _LINUX_CAPABILITY_VERSION_3, 0 };
```
zyga: This entire fragment looks correct. I've cross-checked it with the documentation of the system calls.

cmd/snap-confine/snap-confine.c
```diff
@@ -414,18 +420,49 @@ int main(int argc, char **argv)
...
+		debug("setting capabilities bounding set");
+		// clear all 32 bit caps but SYS_ADMIN, with none inheritable
+		data[0].effective = CAP_TO_MASK(CAP_SYS_ADMIN);
+		data[0].permitted = data[0].effective;
+		data[0].inheritable = 0;
+		// clear all 64 bit caps
+		data[1].effective = 0;
+		data[1].permitted = 0;
+		data[1].inheritable = 0;
+		if (capset(&hdr, data) != 0) {
```
zyga:
```suggestion
		if (capset(&hdr, data) < 0) {
```

cmd/snap-confine/snap-confine.c
```diff
@@ -438,6 +475,36 @@ int main(int argc, char **argv)
 		if (real_uid != 0 && (getgid() == 0 || getegid() == 0))
 			die("permanently dropping privs did not work");
 	}
+
+	// Now that we've permanently dropped, regain SYS_ADMIN
+	if (keep_sys_admin) {
+		debug("regaining SYS_ADMIN");
+		data[0].effective = CAP_TO_MASK(CAP_SYS_ADMIN);
+		data[0].permitted = data[0].effective;
+		if (capset(&hdr, data) != 0) {
```
zyga:
```suggestion
		if (capset(&hdr, data) < 0) {
```

cmd/snap-confine/snap-confine.c
```diff
@@ -438,6 +475,36 @@ int main(int argc, char **argv)
...
+		sc_apply_global_seccomp_profile();
+	}
+
+	// Even though we set inheritable to 0, let's clear SYS_ADMIN
+	// explicitly
+	if (keep_sys_admin) {
+		debug("clearing SYS_ADMIN");
+		data[0].effective = 0;
+		data[0].permitted = data[0].effective;
+		if (capset(&hdr, data) != 0) {
```
zyga:
```suggestion
		if (capset(&hdr, data) < 0) {
```

cmd/snap-confine/snap-confine.c
```diff
@@ -414,18 +420,49 @@ int main(int argc, char **argv)
...
+	// NNP causes issues with AppArmor and exec transitions in certain
+	// snapd interfaces, keep CAP_SYS_ADMIN temporarily when we are
+	// permanently dropping privileges.
+	if (getresuid(&real_uid, &effective_uid, &saved_uid) != 0) {
+		die("getresuid failed");
+	}
+	debug("ruid: %d, euid: %d, suid: %d",
+	      real_uid, effective_uid, saved_uid);
+	struct __user_cap_header_struct hdr = { _LINUX_CAPABILITY_VERSION_3, 0 };
+	struct __user_cap_data_struct data[2] = { {0} };
```
zyga: Nitpick: perhaps call this `cap_data` so that it stands out a little more throughout the file.

tests/main/snap-confine-drops-sys-admin/has-sys-admin.c
```diff
@@ -0,0 +1,52 @@
...
+	int res;
+	bool rc = false;
+	errno = 0;
+
+	cap_t caps;
+	cap_value_t cap;
+	cap_flag_value_t cap_flags_value;
+
+	if (cap_from_name(s, &cap) < 0) {
+		fprintf(stderr, "could not cap_from_name()\n");
```
zyga:
```suggestion
		fprintf(stderr, "cannot get capability index from capability name\n");
```

tests/main/snap-confine-drops-sys-admin/has-sys-admin.c
```diff
@@ -0,0 +1,52 @@
...
+		exit(2);
+	}
+
+	if ((caps = cap_get_proc()) == NULL) {
+		perror("could not get capabilities");
```
zyga:
```suggestion
		perror("cannot get process capabilities");
```

tests/main/snap-confine-drops-sys-admin/has-sys-admin.c
```diff
@@ -0,0 +1,52 @@
...
+		exit(2);
+	}
+
+	res = cap_get_flag(caps, cap, CAP_EFFECTIVE, &cap_flags_value);
+	cap_free(caps);
+	if (res < 0) {
+		fprintf(stderr, "could not cap_get_flag()\n");
```
zyga:
```suggestion
		fprintf(stderr, "cannot get value of capability flag\n");
```

tests/main/snap-confine-drops-sys-admin/has-sys-admin.c
```diff
@@ -0,0 +1,52 @@
...
+	if (cap_flags_value == CAP_SET)
+		rc = true;
+
+	return rc;
+}
+
+
+int main(void)
+{
+	char *cap = "cap_sys_admin";
```
zyga:
```suggestion
	const char *cap = "cap_sys_admin";
```

tests/main/snap-confine-drops-sys-admin/has-sys-admin.c
```diff
@@ -0,0 +1,52 @@
...
+	if (has_cap(cap))
+		printf("Can %s\n", cap);
+	else
+		printf("Cannot %s\n", cap);
+	exit(0);
```
zyga: You can just `return 0`, if there is a reason not to please add that as a comment.

tests/main/snap-confine-drops-sys-admin/has-sys-admin.c
```diff
@@ -0,0 +1,52 @@
...
+
+	return rc;
+}
+
+
+int main(void)
+{
+	char *cap = "cap_sys_admin";
+	if (has_cap(cap))
+		printf("Can %s\n", cap);
```
zyga: Nitpick: I find the word `Can` and `Cannot` below weird. Technically the process has a capability. At the same time this is not so essential. Feel free to reword or ignore as you see fit.

cmd/snap-confine/snap-confine.c
```diff
@@ -438,6 +475,36 @@ int main(int argc, char **argv)
...
+		data[0].permitted = data[0].effective;
+		if (capset(&hdr, data) != 0) {
+			die("capset regain failed");
+		}
+	}
+
+	// Now that we've dropped and regained SYS_ADMIN, we can load the
+	// seccomp profiles.
+	if (sc_apply_seccomp_profile_for_security_tag(invocation.security_tag)) {
+		/* If the process is not explicitly unconfined then load the global
```
bboozzoo: While we're changing this bit, should use `//` for comments to stay consistent with the rest of the code.

cmd/snap-confine/snap-confine.c
jdstrand: This falls into the category of https://github.com/snapcore/snapd/pull/7026#discussion_r297842904 where I find the '!= 0' clearer here (since 0 is the only measure of success).

cmd/snap-confine/snap-confine.c
jdstrand: Because that is 100 lines of code ago in this file and a bunch of callouts to sc_*. It's cheap and guarantees we know where we stand at this moment.

cmd/snap-confine/snap-confine.c
jdstrand: This falls into the category of https://github.com/snapcore/snapd/pull/7026#discussion_r297842904 where I find the '!= 0' clearer here (since 0 is the only measure of success).

cmd/snap-confine/snap-confine.c
jdstrand: This falls into the category of https://github.com/snapcore/snapd/pull/7026#discussion_r297842904 where I find the '!= 0' clearer here (since 0 is the only measure of success).

cmd/snap-confine/snap-confine.c
jdstrand: This falls into the category of https://github.com/snapcore/snapd/pull/7026#discussion_r297842904 where I find the '!= 0' clearer here (since 0 is the only measure of success).

cmd/snap-confine/snap-confine.c
jdstrand: This falls into the category of https://github.com/snapcore/snapd/pull/7026#discussion_r297842904 where I find the '!= 0' clearer here (since 0 is the only measure of success).

tests/main/snap-confine-drops-sys-admin/has-sys-admin.c
jdstrand: This falls into the category of https://github.com/snapcore/snapd/pull/7026#discussion_r297842904 where I find the '!= 0' clearer here (since 0 is the only measure of success).

jdstrand:
Thanks for the reviews! :)

PR #7034, opened 2019-06-26
cmd/snap-seccomp/syscalls: add io_uring syscalls

bboozzoo:
The snap-seccomp helper maintains a list of all know syscalls supported by
libseccomp. We catch when the internal list diverges from upstream libseccomp in
a tests/main/snap-seccomp-syscalls spread test.

Upstream libseccomp has been updated with io_uring syscalls, update the internal
list.


cmd/snap-seccomp/syscalls/syscalls.go
```diff
@@ -165,6 +165,9 @@ var SeccompSyscalls = []string{
 	"io_pgetevents",
 	"io_setup",
 	"io_submit",
+	"io_uring_enter",
+	"io_uring_register",
+	"io_uring_setup",
```
jdstrand: Perhaps it makes sense to update the comment in the spread test what to update (perhaps referencing https://forum.snapcraft.io/t/spread-test-failing-due-to-new-syscalls/12018/2 (your response)) so others know what to do. I would've been more than happy to create this PR myself.

PR #7035, opened 2019-06-26
interfaces/fuse-support: Adjust fstype rule

fnordahl:
A application may use fstype ``fuse`` when mounting a fuse
filesystem.

Adjust AppArmor rule to allow this.

Ref: https://forum.snapcraft.io/t/fuse-support-denied-failed-type-match/11676

### DIFF
--- a/interfaces/builtin/fuse_support.go
+++ b/interfaces/builtin/fuse_support.go
@@ -52,7 +52,7 @@ const fuseSupportConnectedPlugAppArmor = `
 capability sys_admin,
 
 # Allow mounts to our snap-specific writable directories
-# Note 1: fstype is 'fuse.<command>', eg 'fuse.sshfs'
+# Note 1: fstype is 'fuse' or 'fuse.<command>', eg 'fuse.sshfs'
 # Note 2: due to LP: #1612393 - @{HOME} can't be used in mountpoint
 # Note 3: local fuse mounts of filesystem directories are mediated by
 #         AppArmor. The actual underlying file in the source directory is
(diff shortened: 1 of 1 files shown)


chipaca:
Thank you for this!

Could you merge master and re-push, though? Some new syscalls mean we need to adjust some seccomp rules and the tests fail until that's done (and the fix is in master).

fnordahl:
Thanks for quick reviews! And sorry for basing on a out of date branch (d'oh!).  Updated.

chipaca:
the merge that included the fix only landed a couple of hours ago :-) thank you for updating!

jdstrand:
The reason why we want to require fstype is for future work where we can potentially mediate by specific fstypes. Granted, this is provided by the snap and there is nothing saying that a snap couldn't do 'fuse.innocuous' when it is is really doing 'fstype.sshfs' or something, but we may want to keep the door open for the fuse-control interface (which might need a snap declaration/etc and operate similar to personal-files where its use requires review and specific interface attributes-- all this is TBD atm though).

fnordahl:
Ah, got it @jdstrand, that does make sense. Sorry for the confusion with this PR then.

I'll have a look at feasibility of making ``libguestfs`` label it's fstype differently and then carrying a patch with the snap with intent to upstream.

jdstrand:
I'm going to close this PR for now then. Please reach out if this is in error or if we need to think through libguestfs more. Sorry again for the confusion, but thanks for jumping in with a PR! :)

PR #7039, opened 2019-06-27
cmd/snap-confine: exit from helper when parent dies

zyga:
When snap-confine executes it will fork a helper process. That process
establishes an elaborate dance to ensure both itself and the parent are
operating exactly as specified, so that no processes are left behind for
unbound amount of time. As a part of that dance the child pings the
parent to ensure it is still alive after establishing a notification
signal to be sent in case the parent dies. This is a race avoidance
mechanism, we set up the notification and then check if the parent is
alive by the time we are done.

In the case when the parent does go away we used to call abort(). On
some distributions this would trigger an unclean process termination
error report to be sent. One such example is [1]. Since the parent
process can be legitimately interrupted and killed, this should not
generate an error report. As such, perform clean exit in this specific
case.

[1] https://errors.ubuntu.com/problem/0056a844cdfd18814d7eeb470b97d836c1520207


cmd/snap-confine/ns-support.c
```diff
@@ -611,7 +611,7 @@ static void helper_main(struct sc_mount_ns *group, struct sc_apparmor *apparmor,
 		switch (errno) {
 		case ESRCH:
 			debug("parent process has terminated");
-			abort();
+			exit(0);
```
bboozzoo: Shouldn't this be `exit(<non-zero>)`?

zyga: I'm not sure ... we don't know the nature of the death of our parent. At the same time, apart from system services observing termination status of all the processes, nobody will see our termination code if snap-confine itself terminates.

mvo5: Maybe a small comment here that this is a very usual condition and why its ok to exit(0)?

jdstrand: I think it is correct for the child to exit(0) in this case for the reason @zyga mentioned. I do like @mvo5's idea of adding a small comment.

mvo5:
Thank you @zyga!

PR #7041, opened 2019-06-27
gadget: filesystem image writer

bboozzoo:
Add a helper for creating and populating filesystem images.

### DIFF
--- a/gadget/filesystemimage.go
+++ b/gadget/filesystemimage.go
@@ -45,8 +45,35 @@ type FilesystemImageWriter struct {
 }
 
 // PostStageFunc is called after the filesystem contents for the given structure
-// have been staged at a temporary location
+// have been staged at a temporary location, but before the filesystem image is
+// created. The function can be used to manipulate the staged data.
 type PostStageFunc func(rootDir string, ps *PositionedStructure) error
+
+// NewFilesystemImageWriter returns a writer capable of creating filesystem
+// images corresponding to the provided structure, with content from the given
+// content directory. A staging directory will be created in either, the
+// optionally provided work directory, or the default temp location.
+func NewFilesystemImageWriter(contentDir string, ps *PositionedStructure, workDir string) (*FilesystemImageWriter, error) {
+	if ps == nil {
+		return nil, fmt.Errorf("internal error: *PositionedStructure is nil")
+	}
+	if ps.IsBare() {
+		return nil, fmt.Errorf("internal error: structure has no filesystem")
+	}
+	if contentDir == "" {
+		return nil, fmt.Errorf("internal error: gadget content directory cannot be unset")
+	}
+	if _, ok := mkfsHandlers[ps.Filesystem]; !ok {
+		return nil, fmt.Errorf("internal error: filesystem %q has no handler", ps.Filesystem)
+	}
+
+	fiw := &FilesystemImageWriter{
+		contentDir: contentDir,
+		ps:         ps,
+		workDir:    workDir,
+	}
+	return fiw, nil
+}
 
 // Write creates the filesystem inside the provided image file and populates it
 // with data according to content declartion of the structure. Content data is
@@ -83,6 +110,8 @@ func (f *FilesystemImageWriter) Write(fname string, postStage PostStageFunc) err
 		}()
 	}
 
+	// use a mounted filesystem writer to populate the staging directory
+	// with contents of given structure
 	mrw, err := NewMountedFilesystemWriter(f.contentDir, f.ps)
 	if err != nil {
 		return fmt.Errorf("cannot prepare filesystem writer for %v: %v", f.ps, err)
@@ -104,31 +133,4 @@ func (f *FilesystemImageWriter) Write(fname string, postStage PostStageFunc) err
 	}
 
 	return nil
-}
-
-// NewFilesystemImageWriter returns a writer capable of creating filesystem
-// images corresponding to the provided positiioned structure, with content from
-// the given content directory. A staging directory will be created in
-// optionally prided work directory, otherwise the default temporary storage
-// location will be used.
-func NewFilesystemImageWriter(contentDir string, ps *PositionedStructure, workDir string) (*FilesystemImageWriter, error) {
-	if ps == nil {
-		return nil, fmt.Errorf("internal error: *PositionedStructure is nil")
-	}
-	if ps.IsBare() {
-		return nil, fmt.Errorf("internal error: structure has no filesystem")
-	}
-	if contentDir == "" {
-		return nil, fmt.Errorf("internal error: gadget content directory cannot be unset")
-	}
-	if _, ok := mkfsHandlers[ps.Filesystem]; !ok {
-		return nil, fmt.Errorf("internal error: filesystem %q has no handler", ps.Filesystem)
-	}
-
-	fiw := &FilesystemImageWriter{
-		contentDir: contentDir,
-		ps:         ps,
-		workDir:    workDir,
-	}
-	return fiw, nil
 }
--- a/gadget/filesystemimage.go
+++ b/gadget/filesystemimage.go
@@ -61,24 +61,24 @@ func (f *FilesystemImageWriter) Write(fname string, postStage PostStageFunc) err
 		return fmt.Errorf("size of image file %v is different from declared structure size %v", sz, f.ps.Size)
 	}
 
-	mkfs := mkfsHandlers[f.ps.Filesystem]
-	if mkfs == nil {
+	mkfsWithContent := mkfsHandlers[f.ps.Filesystem]
+	if mkfsWithContent == nil {
 		return fmt.Errorf("internal error: filesystem %q has no handler", f.ps.Filesystem)
 	}
 
-	where := filepath.Join(f.workDir, fmt.Sprintf("snap-stage-content-part-%04d", f.ps.Index))
-	if osutil.IsDirectory(where) {
-		return fmt.Errorf("cannot prepare staging directory %s: path exists", where)
+	stagingDir := filepath.Join(f.workDir, fmt.Sprintf("snap-stage-content-part-%04d", f.ps.Index))
+	if osutil.IsDirectory(stagingDir) {
+		return fmt.Errorf("cannot prepare staging directory %s: path exists", stagingDir)
 	}
 
-	if err := os.Mkdir(where, 0755); err != nil {
+	if err := os.Mkdir(stagingDir, 0755); err != nil {
 		return fmt.Errorf("cannot prepare staging directory: %v", err)
 	}
 
 	if os.Getenv("SNAP_DEBUG_IMAGE_NO_CLEANUP") == "" {
 		defer func() {
-			if err := os.RemoveAll(where); err != nil {
-				logger.Noticef("cannot remove filesystem staging directory %q: %v", where, err)
+			if err := os.RemoveAll(stagingDir); err != nil {
+				logger.Noticef("cannot remove filesystem staging directory %q: %v", stagingDir, err)
 			}
 		}()
 	}
@@ -88,17 +88,18 @@ func (f *FilesystemImageWriter) Write(fname string, postStage PostStageFunc) err
 		return fmt.Errorf("cannot prepare filesystem writer for %v: %v", f.ps, err)
 	}
 	// drop all contents to the staging directory
-	if err := mrw.Write(where, nil); err != nil {
+	if err := mrw.Write(stagingDir, nil); err != nil {
 		return fmt.Errorf("cannot prepare filesystem content: %v", err)
 	}
 
 	if postStage != nil {
-		if err := postStage(where, f.ps); err != nil {
+		if err := postStage(stagingDir, f.ps); err != nil {
 			return fmt.Errorf("post stage callback failed: %v", err)
 		}
 	}
 
-	if err := mkfs(fname, f.ps.Label, where); err != nil {
+	// create a filesystem with contents of the staging directory
+	if err := mkfsWithContent(fname, f.ps.Label, stagingDir); err != nil {
 		return fmt.Errorf("cannot create %q filesystem: %v", f.ps.Filesystem, err)
 	}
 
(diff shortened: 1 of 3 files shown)


gadget/filesystemimage.go
```diff
@@ -0,0 +1,133 @@
...
+type PostStageFunc func(rootDir string, ps *PositionedStructure) error
+
+// Write creates the filesystem inside the provided image file and populates it
+// with data according to content declartion of the structure. Content data is
+// staged in a temporary location. An optional post-stage helper function can be
+// used to manipulate the data before it is copied over to the image.
+func (f *FilesystemImageWriter) Write(fname string, postStage PostStageFunc) error {
+	st, err := os.Stat(fname)
+	if err != nil {
+		return fmt.Errorf("cannot stat image file: %v", err)
```
stolowski: How about `fmt.Errorf("cannot stat image file %q: %v", fname, err)` ?

gadget/filesystemimage.go
```diff
@@ -0,0 +1,133 @@
...
+		return fmt.Errorf("cannot prepare filesystem content: %v", err)
+	}
+
+	if postStage != nil {
+		if err := postStage(where, f.ps); err != nil {
+			return fmt.Errorf("post stage callback failed: %v", err)
+		}
+	}
+
+	if err := mkfs(fname, f.ps.Label, where); err != nil {
```
stolowski: Some wondering: is mkfs potentially the most critical and prone to fail op here? If so, would it make sense to move mkfs before everything else that can error out above?

gadget/filesystemimage.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil, fmt.Errorf("internal error: *PositionedStructure is nil")
+	}
+	if ps.IsBare() {
+		return nil, fmt.Errorf("internal error: structure has no filesystem")
+	}
+	if contentDir == "" {
+		return nil, fmt.Errorf("internal error: gadget content directory cannot be unset")
+	}
+	if _, ok := mkfsHandlers[ps.Filesystem]; !ok {
+		return nil, fmt.Errorf("internal error: filesystem %q is not supported", ps.Filesystem)
```
stolowski: Nitpick, this could use same error message as the mkfsHandlers check in `Write`, for consistency.

gadget/filesystemimage.go
```diff
@@ -0,0 +1,133 @@
...
+	st, err := os.Stat(fname)
+	if err != nil {
+		return fmt.Errorf("cannot stat image file: %v", err)
+	}
+	if sz := st.Size(); sz != int64(f.ps.Size) {
+		return fmt.Errorf("size of image file %v is different from declared structure size %v", sz, f.ps.Size)
+	}
+
+	mkfs := mkfsHandlers[f.ps.Filesystem]
+	if mkfs == nil {
```
stolowski: A remark/observation: In theory this is already guaranteed by `NewFilesystemImageWriter` at construction time, in practice  we're keeping PositionedStructure pointer and nothing prevents it from getting modified (and ending up with unsupported fs after ctor) since all members are public.

Also, afaict, this error case doesn't have a test atm.

gadget/filesystemimage.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// NewFilesystemImageWriter returns a writer capable of creating filesystem
+// images corresponding to the provided positiioned structure, with content from
+// the given content directory. A staging directory will be created in
+// optionally prided work directory, otherwise the default temporary storage
+// location will be used.
+func NewFilesystemImageWriter(contentDir string, ps *PositionedStructure, workDir string) (*FilesystemImageWriter, error) {
+	if ps == nil {
+		return nil, fmt.Errorf("internal error: *PositionedStructure is nil")
```
stolowski: No strong opinion here, but I think we haven't been using a "low level" message like this even for internal errors. Perhaps just "internal error: positioned structure cannot be nil" (based on existing: "internal error: assertion type cannot be nil").

gadget/filesystemimage.go
bboozzoo: No, mkfs creates a filesystem with contents of `where` directory (perhaps I should rename that and leave a comment there).

stolowski: Ah, right, makes sense then. Thanks.

gadget/filesystemimage.go
bboozzoo: There's one here: https://github.com/snapcore/snapd/pull/7041/files#diff-399ad2ddfb72083e771e374d24de15e4R259

gadget/filesystemimage.go
bboozzoo: `os` errors include the original path, this one would be ``cannot stat image file: /foo/bar/baz no such file or directory`

gadget/filesystemimage.go
stolowski: Right, missed that one because of the different error message ;)

gadget/filesystemimage.go
bboozzoo: I'd keep the error, it's in sync with errors returned by other ctors: with https://github.com/snapcore/snapd/blob/master/gadget/raw.go#L42-L44

gadget/filesystemimage.go
```diff
@@ -0,0 +1,134 @@
...
+// positioned structures.
+type FilesystemImageWriter struct {
+	contentDir string
+	ps         *PositionedStructure
+	workDir    string
+}
+
+// PostStageFunc is called after the filesystem contents for the given structure
+// have been staged at a temporary location
+type PostStageFunc func(rootDir string, ps *PositionedStructure) error
```
mvo5: What would be an example why this is needed?

gadget/filesystemimage.go
```diff
@@ -0,0 +1,134 @@
...
+// Write creates the filesystem inside the provided image file and populates it
+// with data according to content declartion of the structure. Content data is
+// staged in a temporary location. An optional post-stage helper function can be
+// used to manipulate the data before it is copied over to the image.
+func (f *FilesystemImageWriter) Write(fname string, postStage PostStageFunc) error {
+	st, err := os.Stat(fname)
+	if err != nil {
+		return fmt.Errorf("cannot stat image file: %v", err)
+	}
+	if sz := st.Size(); sz != int64(f.ps.Size) {
```
mvo5: Nice check

gadget/filesystemimage.go
```diff
@@ -0,0 +1,134 @@
...
+		return fmt.Errorf("size of image file %v is different from declared structure size %v", sz, f.ps.Size)
+	}
+
+	mkfsWithContent := mkfsHandlers[f.ps.Filesystem]
+	if mkfsWithContent == nil {
+		return fmt.Errorf("internal error: filesystem %q has no handler", f.ps.Filesystem)
```
mvo5: Slight wondering if there should be a single "supportFilesystems" struct with an optional mkfsHandler that is both used here and in vlidateVolumeStructure - but probably YAGNI and it could be done in a follow-up anyway.

gadget/filesystemimage.go
```diff
@@ -0,0 +1,134 @@
...
+
+	if os.Getenv("SNAP_DEBUG_IMAGE_NO_CLEANUP") == "" {
+		defer func() {
+			if err := os.RemoveAll(stagingDir); err != nil {
+				logger.Noticef("cannot remove filesystem staging directory %q: %v", stagingDir, err)
+			}
+		}()
+	}
+
+	mrw, err := NewMountedFilesystemWriter(f.contentDir, f.ps)
```
mvo5: Maybe a small comment why we do it this way? I mean, the "logical" way would be to create the filesystem first (mkfs) and then run the MountedFilesystemWriter on that created filesystem. Iirc we do it this way to avoid having to mount the filesystem, right? So maybe something like: `// create stage dir first so that mkfs can populate the filesystem and we avoid the need to be root to mount the filesystem` or something?

gadget/filesystemimage.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return fmt.Errorf("cannot prepare filesystem writer for %v: %v", f.ps, err)
+	}
+	// drop all contents to the staging directory
+	if err := mrw.Write(stagingDir, nil); err != nil {
+		return fmt.Errorf("cannot prepare filesystem content: %v", err)
+	}
+
+	if postStage != nil {
```
mvo5: I still wonder how this is used? Maybe an example what the post-stage callback typically does?

gadget/filesystemimage.go
```diff
@@ -0,0 +1,134 @@
...
+
+	return nil
+}
+
+// NewFilesystemImageWriter returns a writer capable of creating filesystem
+// images corresponding to the provided positiioned structure, with content from
+// the given content directory. A staging directory will be created in
+// optionally prided work directory, otherwise the default temporary storage
+// location will be used.
+func NewFilesystemImageWriter(contentDir string, ps *PositionedStructure, workDir string) (*FilesystemImageWriter, error) {
```
mvo5: (this is probably the worst nitpick ever, sorry!) I like it when the "constructor" is close to the struct definition. Mostly because when reading top to bottom its a bit of a more "natural" flow, i.e. I see what is initialized, validated and then I see how the other methods (like Write) operate on this data. But again, a nickpick^99.

gadget/filesystemimage_test.go
```diff
@@ -0,0 +1,375 @@
...
+	c.Assert(err, IsNil)
+
+	err = fiw.Write(img, cb)
+	c.Assert(err, IsNil)
+	c.Assert(cbCalled, Equals, true)
+	c.Assert(mkfsCalled, Equals, true)
+	// nothing left in temporary staging location
+	matches, err := filepath.Glob(s.work + "/*")
+	c.Assert(err, IsNil)
+	c.Assert(matches, HasLen, 0)
```
mvo5: Nice test!

gadget/filesystemimage_test.go
```diff
@@ -0,0 +1,375 @@
...
+	err = fiw.Write(img, cb)
+	c.Assert(err, ErrorMatches, "post stage callback failed: post stage exploded")
+}
+
+func (s *filesystemImageTestSuite) TestMkfsError(c *C) {
+	mkfsHappyFs := func(imgFile, label, contentsRootDir string) error {
+		return errors.New("will not mkfs")
+	}
+	restore := gadget.MockMkfsHandlers(map[string]gadget.MkfsFunc{
+		"happyfs": mkfsHappyFs,
```
mvo5: heh, this is "unhappyfs" now :)  (just kidding, I see it can't be changed without changing psTrivial as well)

gadget/filesystemimage_test.go
```diff
@@ -0,0 +1,375 @@
...
+	img := s.imgForPs(c, s.psTrivial)
+
+	err = fiw.Write(img, nil)
+	c.Assert(err, ErrorMatches, `cannot create "happyfs" filesystem: will not mkfs`)
+}
+
+func (s *filesystemImageTestSuite) TestFilesystemExtraCheckError(c *C) {
+	ps := &gadget.PositionedStructure{
+		VolumeStructure: &gadget.VolumeStructure{
+			Filesystem: "happyfs",
```
mvo5: Would it be easier to just use "ext4" here and then you can skip the mocking of mkfsHappyFs and MockMkfsHandlers? AFAICT you only need it because NewFilesystemImageWriter is also doing validation of the handler. And the handler is changed before the fiw.Write() call so that should be ok?

gadget/filesystemimage_test.go
```diff
@@ -0,0 +1,375 @@
...
+	ps.Filesystem = "foofs"
+
+	err = fiw.Write(img, nil)
+	c.Assert(err, ErrorMatches, `internal error: filesystem "foofs" has no handler`)
+}
+
+func (s *filesystemImageTestSuite) TestMountedWriterError(c *C) {
+	ps := &gadget.PositionedStructure{
+		VolumeStructure: &gadget.VolumeStructure{
+			Filesystem: "happyfs",
```
mvo5: Same as above?

gadget/filesystemimage_test.go
```diff
@@ -0,0 +1,375 @@
...
+			Size:       2 * gadget.SizeMiB,
+			Content:    []gadget.VolumeContent{},
+		},
+	}
+
+	mkfsHappyFs := func(imgFile, label, contentsRootDir string) error {
+		return errors.New("will not mkfs")
```
mvo5: Also this should probably be "unexpected call" (if kept), because the check tested happens before this is called, yes?

gadget/filesystemimage_test.go
```diff
@@ -0,0 +1,375 @@
...
+}
+
+func (s *filesystemImageTestSuite) TestBadWorkDirError(c *C) {
+	cb := func(rootDir string, cbPs *gadget.PositionedStructure) error {
+		return errors.New("unexpected call")
+	}
+	mkfsHappyFs := func(imgFile, label, contentsRootDir string) error {
+		return errors.New("unexpected call")
+	}
+	restore := gadget.MockMkfsHandlers(map[string]gadget.MkfsFunc{
```
mvo5: (nitpick^2) Given how often this is called I wonder if it would make sense to have it in SetUpTest (or maybe have a sub-suite that does it?).

bboozzoo: Went further. Added a sub suite with mocks for `happyfs`, as well as another sub-suite that verifies the default handlers for `ext4` and `vfat`.

gadget/filesystemimage_test.go
bboozzoo: Switcher to suite setup level mocks.

gadget/filesystemimage.go
bboozzoo: Moved

gadget/filesystemimage.go
bboozzoo: Comment added

gadget/filesystemimage.go
bboozzoo: Added a comment in code. The idea is to allow manipulation of the future root fs right after staging the contents in a directory, but before we create the filesystem image. Example use case is manipulation of a `system-boot` role partition, that pulls in contents of `image/boot/{grub,uboot}/` from a directory prepared by `snap prepare-image`, even though its structure level `content` in `gadget.yaml` refers only to the data inside `image/gadget/`.

gadget/filesystemimage.go
bboozzoo: Extended the comment

PR #7042, opened 2019-06-28
interfaces: add an interface granting access to AppStream metadata

jhenstridge:
AppStream is a data format used to describe packages and other components that can be installed on a Linux system as documented here:

https://www.freedesktop.org/software/appstream/docs/

In particular, Debian/Ubuntu systems publish AppStream collection metadata (aka DEP-11) about packages held in the repository.  Package managers can use this to display translated descriptions and extended metadata about packages to users.

This interface grants snaps permission to read the host system's AppStream metadata, and mounts it into the sandbox in locations where `libappstream` or `libappstream-glib` will search for it.

### DIFF
--- a/interfaces/builtin/appstream_metadata.go
+++ b/interfaces/builtin/appstream_metadata.go
@@ -49,6 +49,7 @@ const appstreamMetadataConnectedPlugAppArmor = `
 
 # Allow access to AppStream upstream metadata files
 /usr/share/metainfo/** r,
+/usr/share/appdata/** r,
 
 # Allow access to AppStream collection metadata
 /usr/share/app-info/** r,
@@ -61,6 +62,7 @@ const appstreamMetadataConnectedPlugAppArmor = `
 
 var appstreamMetadataDirs = []string{
 	"/usr/share/metainfo",
+	"/usr/share/appdata",
 	"/usr/share/app-info",
 	"/var/cache/app-info",
 	"/var/lib/app-info",
@@ -92,13 +94,7 @@ func (iface *appstreamMetadataInterface) AppArmorConnectedPlug(spec *apparmor.Sp
 }
 
 func (iface *appstreamMetadataInterface) MountConnectedPlug(spec *mount.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
-	for _, dir := range []string{
-		"/usr/share/metainfo",
-		"/usr/share/app-info",
-		"/var/cache/app-info",
-		"/var/lib/app-info",
-		"/var/lib/apt/lists",
-	} {
+	for _, dir := range appstreamMetadataDirs {
 		dir = filepath.Join(dirs.GlobalRootDir, dir)
 		if !osutil.IsDirectory(dir) {
 			continue
--- a/interfaces/builtin/appstream_metadata.go
+++ b/interfaces/builtin/appstream_metadata.go
@@ -20,10 +20,13 @@
 package builtin
 
 import (
+	"bytes"
+	"fmt"
 	"path/filepath"
 
 	"github.com/snapcore/snapd/dirs"
 	"github.com/snapcore/snapd/interfaces"
+	"github.com/snapcore/snapd/interfaces/apparmor"
 	"github.com/snapcore/snapd/interfaces/mount"
 	"github.com/snapcore/snapd/osutil"
 )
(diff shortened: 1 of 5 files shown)


robert-ancell:
A grep through the appstream-glib source shows it also can access `/usr/share/appdata`. Otherwise everything looks good to me!

jhenstridge:
I missed that path when reading the spec (it is mentioned separately as a legacy path). I see that there are still packages installing data to that location on Eoan though, so I've updated the PR accordingly.

On a related note, a version of `apparmor.Specification.AddUpdateNS` that took an `osutil.MountEntry` as an argument seems like it could reduce the need to repeat yourself in interfaces that add mounts.

interfaces/builtin/appstream_metadata.go
```diff
@@ -0,0 +1,120 @@
...
+# Allow access to AppStream collection metadata
+/usr/share/app-info/** r,
+/var/cache/app-info/** r,
+/var/lib/app-info/** r,
+
+# Apt symlinks the DEP-11 metadata to files in /var/lib/apt/lists
+/var/lib/apt/lists/*.yml.gz r,
+`
+
+var appstreamMetadataDirs = []string{
```
zyga: I'm somewhat worried about the number of mount entries this requires. This is will impact people's ability to use layouts.

jhenstridge: I don't expect a huge number of snaps to end up using this interface: it is only going to be of interest to package management tools, where the pros of access to the data outweigh the cons.

I had considered using `system-files` to punch holes to paths in `/var/lib/snapd/hostfs/...` as an alternative, but the mounts have a few benefits:

1. both libraries for reading AppStream metadata contain hard coded paths in the code:

    https://github.com/ximion/appstream/blob/master/src/as-pool.c#L95
    https://github.com/hughsie/appstream-glib/blob/master/libappstream-glib/as-store.c#L3326

2. On Ubuntu and Debian systems, the collection metadata files in `/var/lib/app-info/yaml` are absolute symlinks that won't resolve using the hostfs paths.

bboozzoo: > I'm somewhat worried about the number of mount entries this requires. This is will impact people's ability to use layouts.

you mean the potential for conflicts or just global limits on the number mounts?

zyga: Potential conflicts, we don't currently enforce any numeric limits.

tests/main/interfaces-appstream-metadata/task.yaml
```diff
@@ -0,0 +1,59 @@
...
+restore: |
+    rm -f /usr/share/metainfo/test1.metainfo.xml
+    rm -f /usr/share/appdata/test2.metainfo.xml
+    rm -f /usr/share/app-info/xmls/test1.xml
+    rm -f /var/cache/app-info/xmls/test2.xml
+    rm -f /var/lib/app-info/yaml/test3.yml.gz
+    rm -f /var/lib/apt/lists/test3.yml.gz
+
+execute: |
+    mkdir -p /usr/share/metainfo
```
zyga: This is something that would probably make more sense in prepare. All the way until "the plug is disconnected by default"

interfaces/builtin/appstream_metadata.go
jdstrand: > I'm somewhat worried about the number of mount entries this requires. This is will impact people's ability to use layouts.

It's conceivable that snapd could could check for conflicting layouts if the snap plugs this interface. This can also happen in the review-tools as an intermediate step to not block this PR. That said, IME, only the /var/lib/apt/lists is at all worrisome since the others are very clearly appstream.

One thought is that since @jhenstridge mentioned this is only probably for package management tools and we consider that type of snap to be highly trusted, we could give this interface an installation constraint and avoid the layouts issue in practice. The data itself doesn't warrant it, but we can add a code comment on why we have the installation constraint. This would not be onerous for legitimate users because they likely need packagekit-control (or some future, similarly powerful interface) which needs an installation request anyway. I'm not necessarily recommending this path, but like I said, it would avoid the layouts conflict issue in practice and could be removed at a future date when conflicts are detected.

interfaces/builtin/appstream_metadata.go
```diff
@@ -0,0 +1,120 @@
...
+	"/usr/share/metainfo",
+	"/usr/share/appdata",
+	"/usr/share/app-info",
+	"/var/cache/app-info",
+	"/var/lib/app-info",
+	"/var/lib/apt/lists",
```
jdstrand: Should this include support for other package managers or is this just due to some library that does this as a matter of course?

jhenstridge: As mentioned in source comments, we aren't exposing Apt specific metadata.  The problem is that Apt installs AppStream metadata to `/var/lib/apt/lists` and then symlinks it to `/var/lib/app-info`.  If we don't bind mount `/var/lib/apt/lists` into the sandbox, all we have is a bunch of dangling symlinks.

The AppArmor rules enforce that the only files in `/var/lib/apt/lists` that can actually be read are the AppStream `.yml.gz` files.

jdstrand: > As mentioned in source comments, we aren't exposing Apt specific metadata. The problem is that Apt installs AppStream metadata to `/var/lib/apt/lists` and then symlinks it to `/var/lib/app-info`. If we don't bind mount `/var/lib/apt/lists` into the sandbox, all we have is a bunch of dangling symlinks.
> 
> The AppArmor rules enforce that the only files in `/var/lib/apt/lists` that can actually be read are the AppStream `.yml.gz` files.

Sure. My point wasn't about dropping /var/lib/apt/lists, it was about does yum, dnf, pacman, etc do something similar and so should this be extended. "I don't know" is an acceptable answer-- these can come in subsequent PRs as needed.

jhenstridge: I haven't checked everywhere, but Debian/Ubuntu seem unusual in this respect.

For Fedora, the metadata is shipped as an RPM package installing directly to `/usr/share/app-info/xmls`: https://apps.fedoraproject.org/packages/appstream-data/

I'm not sure what OpenSUSE or Arch do, so I think the follow-up PRs would be a sensible way to move forward if what we have isn't sufficient.

interfaces/builtin/appstream_metadata.go
jhenstridge: I'd be happy to add a plug restriction to the interface: given that the only current user will need one for `packagekit-control` anyway, it isn't a huge burden.

As far as information disclosure goes, there's two types of info exposed:

1. collection metadata: this is metadata about packages available in the distro repository.
2. upstream metadata: this is metadata provided by installed packages.

While (1) only really leaks information about configured software repositories on the host,  (2) leaks information about (some of) the installed applications.  While we definitely shouldn't allow auto-connection by default, I'm not sure if this goes all the way to denying installation by default.

jdstrand: @zyga - can you comment on how concerned you are about the potential layout conflict? Personally, I don't think we should block this PR on this detection and this could come in future PRs. If you are quite concerned, we can cheaply use the installation constraint or add a review-tools check until the layout detection is in place.

interfaces/builtin/appstream_metadata.go
jdstrand: Thanks for looking. Based on your response a "wait and see" approach sounds fine to me.

interfaces/builtin/appstream_metadata.go
zyga: Given that this interface is to be used by specific applications only I'm no longer concerned. On my first read I didn't mentally scope it to a specific application and I assumed anyone would be interested in connecting it. With the additional information I'm +1 to merge it as well as to come up with better conflict detection, separately.

PR #7043, opened 2019-06-28
snap-userd-autostart: don't list as a startup application on the GUI

seb128:
Use NoDisplay=true so gnome-session-properties does list it as a startup program
We don't list the technical items/services in the GUI because it confuses users
and can lead to have them being turned off which creates problems.


zyga:
Hmm, the CLA checker is grumpy.

mvo5:
@seb128 Our CLA checker script is unhappy unfortunately. The easiest fix would be if you could just commit as "seb128@canonical.com" and force push. Thanks!

seb128:
k, forced pushed with a canonical email now

PR #7045, opened 2019-06-28
cmd: make fmt

zyga:


anonymouse64:
The go code is already checked for formatting in the tests I think, perhaps that's something to do for the c code as well in the tests?

zyga:
We cannot check because indent has version dependent behavior and is also not deterministic on some input. Until we switch entirely to clang format we just have to re format once in a while.

PR #7046, opened 2019-06-28
 cmd/libsnap: use fixed buffer for scanning infofile

zyga:
While working on unit tests I was worried that some attacker might
be able to point the parser at a large file and therefore exhaust the
memory pool, all while running as root. Since for our needs we can
certainly put a small cap on the memory pool used by the key=value
extractor I did just that.

In addition, the test coverage for infofile is now at 100%


zyga:
Closing to return later.

PR #7050, opened 2019-06-28
Retry: Handle DNSError when network message different

ardaguclu:
As mentioned below launchpad link; in some cases when network connection does not exist, snap install command returns unclear message. Thus, I added "server misbehaving" check to make it clear to the user when there is possible network connectivity problem and its unit test.

https://bugs.launchpad.net/snapd/+bug/1794006

### DIFF
--- a/httputil/retry.go
+++ b/httputil/retry.go
@@ -153,7 +153,9 @@ func isNetworkDown(err error) bool {
 	case *net.DNSError:
 		// on 16.04 we will not have SyscallError here, but DNSError, with
 		// no further details other than error message
-		return strings.Contains(lowerErr.Err, "connect: network is unreachable")
+		return strings.Contains(lowerErr.Err, "connect: network is unreachable") ||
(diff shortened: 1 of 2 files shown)


chipaca:
Is this something you are able to reproduce? Can you try to do so on 16.04, where we get a different error family for network errors?

ardaguclu:
On Ubuntu 18.04 and when the network connection is turned off, installation command returns; 

> cannot install "vlc": Post https://api.snapcraft.io/v2/snaps/refresh:
       dial tcp: lookup api.snapcraft.io: no such host

On Ubuntu 16.04 (running as lxd container);

> cannot install "vlc": persistent network error: Post
       https://api.snapcraft.io/v2/snaps/refresh: dial tcp: lookup
       api.snapcraft.io on 10.6.49.1:53: dial udp 10.6.49.1:53: connect:
       network is unreachable

I have not tested trying to install from localhost issued the launchpad link.

@chipaca

PR #7051, opened 2019-07-01
cmd/snap-apparmor-service: quit if there are no profiles

zyga:
We got a word-of-mouth bug report from Manjaro that
snapd-apparmor.service crashes when there are no apparmor profiles in
/var/lib/snapd/apparmor/profiles.

Looking at the shell script implementing the logic it is clear that in
absence of the actual profiles the script will execute with a single
file "name" of /var/lib/snapd/apparmor/profiles/* - a well known
"feature" of shell where glob expands to itself when it doesn't match
any actual files.

### DIFF
--- a/cmd/snapd-apparmor/snapd-apparmor
+++ b/cmd/snapd-apparmor/snapd-apparmor
@@ -81,6 +81,9 @@ case "$1" in
 		fi
 		# </copied-code>
 
+		if [ "$(find /var/lib/snapd/apparmor/profiles/ -type f | wc -l)" -eq 0 ]; then
+			exit 0
+		fi
 		for profile in /var/lib/snapd/apparmor/profiles/*; do
 			# Filter out profiles with names ending with ~, those are temporary files created by snapd.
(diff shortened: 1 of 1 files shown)


zyga:
@mvo5 unfortunately you are missing the same thing I missed when I wrote this originally:

```
for f in /non-existing/*; do echo "processing file: $f"; done
```

Glob expands to itself when it matches to nothing. This unfortunately requires careful guard code around instructions that expect actual files.

cmd/snapd-apparmor/snapd-apparmor
```diff
@@ -81,6 +81,9 @@ case "$1" in
 		fi
 		# </copied-code>
 
+		if [ "$(find /var/lib/snapd/apparmor/profiles/ -type f | wc -l)" -eq 0 ]; then
```
stolowski: '-maxdepth 0' ?

zyga: YAGNI, there are no directories there today.

PR #7058, opened 2019-07-03
interfaces/apparmor: use available apparmor features

zyga:
Historically snapd would fallback to "forced devmode", which is like
applying a very open and permissive confinement to all the systems that
do not implement full confinement. This is due to inability, at the
time, of ancient apparmor userspace, to even parse and compile the
regular snapd security template.

Over time we started allowing use of true partial confinement, that is
without the degradation to classic template, on openSUSE Tumbleweed.
That distribution was selected due to its rolling nature and the fact
that up-to-date apparmor userspace and kernel was available. The
experiment was further extended to cover Arch, along with the hardened
kernel that supports apparmor. This experiment was very successful,
showing that there are practical benefits in more complete sandbox
without any observable downsides.

Recently we reached a point where using the forced devmode fallback has
started to hurt the ecosystem, as people routinely develop snap packages
on Debian, openSUSE Leap or even on Ubuntu using mainline kernel. In
that situation they may not fully comprehend the snap interface system
and fail to use interfaces that are required under strict confinement.

While the complete journey towards explaining the state of confinement
is yet to be designed, we think it is important to use as many apparmor
features as are available, even if not all of the features we want are
present.

This patch will effectively start to enforce path based file protections
on Debian and openSUSE as well as any distribution running with apparmor
enabled. In practice it will shrink the perception delta between the
upstream kernel and the Ubuntu kernel down to fine-grained DBus
mediation and perhaps some of the network rules.

### DIFF
--- a/interfaces/apparmor/backend.go
+++ b/interfaces/apparmor/backend.go
@@ -502,19 +502,12 @@ func addUpdateNSProfile(snapInfo *snap.Info, opts interfaces.ConfinementOptions,
 
 func downgradeConfinement() bool {
 	kver := osutil.KernelVersion()
-	switch {
-	case release.DistroLike("opensuse-tumbleweed"):
-		if cmp, _ := strutil.VersionCompare(kver, "4.16"); cmp >= 0 {
-			// As a special exception, for openSUSE Tumbleweed which ships Linux
-			// 4.16, do not downgrade the confinement template.
-			return false
-		}
-	case release.DistroLike("arch", "archlinux"):
-		// The default kernel has AppArmor enabled since 4.18.8, the
-		// hardened one since 4.17.4
-		return false
-	}
-	return true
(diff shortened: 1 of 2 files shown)


zyga:
I looked at why this is red in spread on travis. We seem to be hitting a timeout. I ran this test manually without any issues. Perhaps it was not running before and now it does (fine grained apparmor checks?). In either case some more investigation is needed but the test is *passing* if you give it some more time.

zyga:
We decided not to enable this on existing releases and distributions to avoid the regression potential. During a call yesterday the action plan was decided as follows:

- enable this conditionally on Debian 11 and next openSUSE LEAP development release
- enable a smoke test that runs openSUSE Tumbleweed (which has apparmor enabled despite being partial) to report as strict, so that more test coverage is reported.
- discuss how to work on additional testing

PR #7060, opened 2019-07-03
.travis.yml: bump Go version back down to 1.9 for quick tests

chipaca:
Not _all_ distros are on 1.10 (looking at you, amz2)


chipaca:
bah humbug

PR #7061, opened 2019-07-03
cmd/libsnap: add sc_error_init_{simple,api_misuse}

zyga:
This is just syntactic sugar for the two common cases of error from the
internal library, without any specific error code and when arguments
provided to a function are incorrect.

### DIFF
--- a/cmd/libsnap-confine-private/error.h
+++ b/cmd/libsnap-confine-private/error.h
@@ -100,7 +100,7 @@ __attribute__((warn_unused_result,
 sc_error *sc_error_init_simple(const char *msgfmt, ...);
 
 /**
- * Initialize an api misuse error with formatted message.
+ * Initialize an API misuse error with formatted message.
  *
  * This is just syntactic sugar for sc_error_init(SC_LIBSNAP_DOMAIN,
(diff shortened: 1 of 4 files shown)


cmd/libsnap-confine-private/error.h
```diff
@@ -89,6 +89,26 @@ __attribute__((warn_unused_result,
...
+ *
+ * This is just syntactic sugar for sc_error_init(SC_LIBSNAP_ERROR,
+ * SC_UNSPECIFIED_ERROR, msgfmt, ...) which is repeated often.
+ **/
+__attribute__((warn_unused_result,
+	       format(printf, 1, 2) SC_APPEND_RETURNS_NONNULL))
+sc_error *sc_error_init_simple(const char *msgfmt, ...);
+
+/**
+ * Initialize an unspecified error with formatted message.
```
anonymouse64: `Initialize an api misuse error with formatted message.` ?

anonymouse64:
```suggestion
 * Initialize an api misuse error with formatted message.
```

bboozzoo: s/api/API/ even in @anonymouse64's suggestion

PR #7062, opened 2019-07-04
tests: re-enable ns-re-associate test

zyga:
This test was waiting for a kernel fix. The fix has landed but we didn't
notice that. Thanks to Pedronis for noticing.

### DIFF
--- a/tests/regression/lp-1644439/task.yaml
+++ b/tests/regression/lp-1644439/task.yaml
@@ -1,5 +1,11 @@
 summary: Regression test for https://bugs.launchpad.net/snap-confine/+bug/1644439
-
+# This test fails on Ubuntu 18.04 and later because snap-confine, along all of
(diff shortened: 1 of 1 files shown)


chipaca:
```
2019-07-04 10:36:02 Failed tasks: 2
    - google:ubuntu-18.04-64:tests/regression/lp-1644439
    - google:ubuntu-18.10-64:tests/regression/lp-1644439
```

zyga:
This fails because snap-confine, along all of snapd is built in a distribution from the future (ubuntu 18.04 for example) and the injected into a ubuntu 16.04-based chroot. This cannot work in general. We discussed this in the past and we'd have to change how we do our CI so that we can always build each branch against ubuntu 16.04 container and only then repackage and test on a given distribution.

mvo5:
Can we simply limit the test to 16.04 systems?

zyga:
@mvo5 done now, this passed on all the test systems locally.

PR #7065, opened 2019-07-04
tests: allow renumbering mount namespace identifiers

zyga:
The Linux kernel allows to bind-mount /proc/pid/ns/{mnt,...} on top of a
file to preserve a namespace object beyond the death of a process. Snapd
uses this feature extensively.

In the mount table, such preserved namespaces have a specific form:

    2290 2266 0:3 mnt:[4026532909] /run/snapd/ns/xkcd-webserver.mnt rw - nsfs nsfs rw

As one can see, the field "mnt:[...]" describes the ... root directory of
the file system that is exposed. Normally this attribute encodes the
subtree of the filesystem of the major:minor block device, that is visible
at the mount point. This abuse of the attribute is a bit odd, but not
unlike other kernel interfaces.

The number in the brackets is a kernel identifier that is not
deterministic in any way. For the purpose of writing tests that measure
the state of the mount table, we would like to have the option re-write
such numbers into something deterministic.

This patch does exactly that. When --renumber option is used such mount
identifiers are replaced by monotonically increasing numbers.

### DIFF
--- a/tests/lib/bin/mountinfo-tool
+++ b/tests/lib/bin/mountinfo-tool
@@ -396,6 +396,29 @@ def renumber_devices(entry, seen):
     entry.dev = alloc_n(entry.dev)
 
 
+def renumber_ns(entry, seen):
+    # type: (MountInfoEntry, Dict[Tuple[Text, int], int]) -> None
+    """renumber_mount_ns re-numbers mount namespace ID from .root_dir property."""
+
+    def alloc_n(ns_type, ns_id):
+        # type: (Text, int) -> int
+        key = (ns_type, ns_id)
+        try:
+            return seen[key]
+        except KeyError:
+            n = len(seen)
+            seen[key] = n
+            return n
+
+    if entry.fs_type != "nsfs":
+        return
+    match = re.match(r"^([a-z_]+):\[(\d+)\]$", entry.root_dir)
+    if match:
+        ns_type = match.group(1)
+        ns_id = int(match.group(2))
+        entry.root_dir = "{}:[{}]".format(ns_type, alloc_n(ns_type, ns_id))
+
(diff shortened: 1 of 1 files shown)


tests/lib/bin/mountinfo-tool
```diff
@@ -396,6 +396,29 @@ def renumber_devices(entry, seen):
...
+        try:
+            return seen[key]
+        except KeyError:
+            n = len(seen)
+            seen[key] = n
+            return n
+
+    if entry.fs_type != "nsfs":
+        return
+    match = re.match(r"^([a-z_]+):\[(\d+)\]$", entry.root_dir)
```
bboozzoo: aren't we expecting `r"^mnt:\[(\d+)\]$"`?

zyga: We only use the mount namespace but any namespace can be bound this way. The `fs_type` check makes sure we are catching only those things that are actual namespaces.

bboozzoo: Reading the kernel, it looks like we're only getting `mnt:[<inode>]` here, but I may be wrong.

zyga: Whatever it is, it is not documented as anything stable. The inode is from nsfs which is obviously memory-only.

zyga: What I meant by my reply to your initial question is that while the mount namespace uses `mnt:[...]` the other namespaces can be bound too and will yield their respective identifiers, so `pid:[...]` etc. I just did not want to special-case `mnt`.

bboozzoo: Yeah, let's keep the current code.

PR #7068, opened 2019-07-04
tests: refactor mountinfo-tool rewrite state

zyga:
The --renumber and --rename options used local state to assist various
normalization rules that work on making mount entries deterministic across
boots.

This patch makes that state explicit. This will be used in the upcoming
feature where we can inspect a mount namespace from the perspective of
a base namespace, keeping identifiers like block devices and peer groups
comparable across both rewritten tables.

### DIFF
--- a/tests/lib/bin/mountinfo-tool
+++ b/tests/lib/bin/mountinfo-tool
@@ -396,24 +396,31 @@ def renumber_devices(entry, seen):
     entry.dev = alloc_n(entry.dev)
 
 
-def rewrite_renumber(entries):
-    # type: (List[MountInfoEntry]) -> None
+class RewriteState(object):
+    """RewriteState holds state used in rewriting mount entries."""
+
+    def __init__(self):
+        # type: () -> None
+        self.seen_opt_fields = {}  # type: Dict[int, int]
+        self.seen_loops = {}  # type: Dict[int, int]
+        self.seen_snap_revs = {}  # type: Dict[Tuple[Text, Text], int]
+        self.seen_mount_ids = {}  # type: Dict[int, int]
+        self.seen_devices = {}  # type: Dict[Device, Device]
+
+
+def rewrite_renumber(entries, rs):
+    # type: (List[MountInfoEntry], RewriteState) -> None
     """rewrite_renumber applies all re-numbering helpers."""
-    seen_opt_fields = {}  # type: Dict[int, int]
-    seen_loops = {}  # type: Dict[int, int]
-    seen_snap_revs = {}  # type: Dict[Tuple[Text, Text], int]
-    seen_mount_ids = {}  # type: Dict[int, int]
-    seen_devices = {}  # type: Dict[Device, Device]
     for entry in entries:
-        renumber_mount_ids(entry, seen_mount_ids)
-        renumber_devices(entry, seen_devices)
-        renumber_snap_revision(entry, seen_snap_revs)
-        renumber_opt_fields(entry, seen_opt_fields)
-        renumber_loop_devices(entry, seen_loops)
+        renumber_mount_ids(entry, rs.seen_mount_ids)
+        renumber_devices(entry, rs.seen_devices)
+        renumber_snap_revision(entry, rs.seen_snap_revs)
+        renumber_opt_fields(entry, rs.seen_opt_fields)
+        renumber_loop_devices(entry, rs.seen_loops)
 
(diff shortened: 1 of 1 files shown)


tests/lib/bin/mountinfo-tool
```diff
@@ -396,24 +396,31 @@ def renumber_devices(entry, seen):
...
+        renumber_snap_revision(entry, rs.seen_snap_revs)
+        renumber_opt_fields(entry, rs.seen_opt_fields)
+        renumber_loop_devices(entry, rs.seen_loops)
 
 
-def rewrite_rename(entries):
-    # type: (List[MountInfoEntry]) -> None
+def rewrite_rename(entries, rw):
+    # type: (List[MountInfoEntry], RewriteState) -> None
     """rewrite_renameapplies all re-naming helpers."""
```
bboozzoo:
```suggestion
    """rewrite_rename applies all re-naming helpers."""
```

tests/lib/bin/mountinfo-tool
```diff
@@ -396,24 +396,31 @@ def renumber_devices(entry, seen):
...
+        renumber_mount_ids(entry, rs.seen_mount_ids)
+        renumber_devices(entry, rs.seen_devices)
+        renumber_snap_revision(entry, rs.seen_snap_revs)
+        renumber_opt_fields(entry, rs.seen_opt_fields)
+        renumber_loop_devices(entry, rs.seen_loops)
 
 
-def rewrite_rename(entries):
-    # type: (List[MountInfoEntry]) -> None
+def rewrite_rename(entries, rw):
```
bboozzoo: actual use will come in later patches?

zyga: Yes :)

tests/lib/bin/mountinfo-tool
zyga: Hmm, it doesn't look bad in vim, I wonder what's going on here.

zyga: Ahhh

tests/lib/bin/mountinfo-tool
```diff
@@ -419,27 +419,32 @@ def renumber_ns(entry, seen):
...
+    def __init__(self):
+        # type: () -> None
+        self.seen_opt_fields = {}  # type: Dict[int, int]
+        self.seen_loops = {}  # type: Dict[int, int]
+        self.seen_snap_revs = {}  # type: Dict[Tuple[Text, Text], int]
+        self.seen_mount_ids = {}  # type: Dict[int, int]
+        self.seen_devices = {}  # type: Dict[Device, Device]
+
+
+def rewrite_renumber(entries, rs):
```
stolowski: Any reason for not making all the `rewrite_*` functions into methods of RewriteState class?

zyga: They could either be methods on the mount entry or on the rewrite state. I think they are just functions that take two arguments. No strong feeling either way though.

tests/lib/bin/mountinfo-tool
```diff
@@ -419,27 +419,32 @@ def renumber_ns(entry, seen):
...
-        renumber_devices(entry, seen_devices)
-        renumber_snap_revision(entry, seen_snap_revs)
-        renumber_opt_fields(entry, seen_opt_fields)
-        renumber_loop_devices(entry, seen_loops)
-        renumber_ns(entry, seen_ns)
+        renumber_mount_ids(entry, rs.seen_mount_ids)
+        renumber_devices(entry, rs.seen_devices)
+        renumber_snap_revision(entry, rs.seen_snap_revs)
+        renumber_opt_fields(entry, rs.seen_opt_fields)
+        renumber_loop_devices(entry, rs.seen_loops)
```
stolowski: No `renumber_ns`, or am I missing something?

zyga: Oh, indeed, bad merge!

zyga: Fixed

stolowski: Should we have a test that ensure all renumbers* are run?

tests/lib/bin/mountinfo-tool
stolowski: Ok, I'd probably make them methods, but no hard opinion and not a blocker.

PR #7071, opened 2019-07-05
tests: replace various numeric mount options

zyga:
When comparing mount tables some options, such as number of inodes or
tmpfs size can be a problem as it depends on the amount of memory on the
system. To facilitate testing the --renumber option now handles this
case.

The embedded tests showcase the two instances where I saw this in
practice. It can be easily extended on demand.

### DIFF
--- a/tests/lib/bin/mountinfo-tool
+++ b/tests/lib/bin/mountinfo-tool
@@ -436,6 +436,42 @@ def renumber_ns(entry, seen):
         entry.root_dir = "{}:[{}]".format(ns_type, alloc_n(ns_type, ns_id))
 
 
+def renumber_mount_option(opt, seen):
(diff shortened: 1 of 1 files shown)


tests/lib/bin/mountinfo-tool
```diff
@@ -436,6 +436,42 @@ def renumber_ns(entry, seen):
...
+
+    if "=" in opt:
+        mount_opt_key, mount_opt_value = opt.split("=", 1)
+        # size, nr_inode: used by tmpfs
+        # fd, pipe_ino: used by binfmtmisc
+        if mount_opt_key in {"size", "nr_inodes", "fd", "pipe_ino"}:
+            return "{}={}".format(
+                mount_opt_key, alloc_n(mount_opt_key, mount_opt_value)
+            )
+    return opt
```
stolowski: Looks fine.

PR #7078, opened 2019-07-08
tests: change allocation for mount options

zyga:
Initially the renumbering of mount options would just allocate
consecutive numbers regardless of the mount option. For instance
size=100k, size=200k, fd=21 would get re-numbered as size=0, size=1,
fd=2. After using this mode for a while I noticed that data sets for
different systems sometimes differ by one more fringe option that causes
cascade of differences simply because the number pool is shared. For
example, suppose that after size=200k there was also size=300k. This
would cause the fd=21 to be remapped to fd=3 now.

To reduce this effect, allocate numbers for each mount option
separately. This way fd=21 is renumbered as fd=0 in both cases. This
makes comparison of data sets of the upcoming mount namespace test much
easier.

### DIFF
--- a/tests/lib/bin/mountinfo-tool
+++ b/tests/lib/bin/mountinfo-tool
@@ -453,7 +453,13 @@ def renumber_mount_option(opt, seen):
         try:
             return seen[key]
         except KeyError:
-            n = len(seen)
+            n = len(
+                {
+                    opt_value
+                    for opt_key, opt_value in seen.keys()
+                    if opt_key == mount_opt_key
+                }
+            )
             seen[key] = n
             return n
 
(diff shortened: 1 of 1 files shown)


tests/lib/bin/mountinfo-tool
```diff
@@ -453,7 +453,13 @@ def renumber_mount_option(opt, seen):
         try:
             return seen[key]
         except KeyError:
-            n = len(seen)
+            n = len(
+                {
+                    opt_value
+                    for opt_key, opt_value in seen.keys()
+                    if opt_key == mount_opt_key
```
stolowski: Why not just a helper dict (apart from `seen`) that maps mount_opt_key to a counter and increases it on every alloc? That would avoid iterating over the dict on every alloc and would make the intent more clear.

zyga: Because I used the same trick for the block device and because it's not a critical for performance. I'm happy to revisit after the patch set lands and we focus on reviewing the test this enables.

zyga:
NOTE: if you click on Travis status it is actually green there.

tests/lib/bin/mountinfo-tool
stolowski: Allright.

PR #7082, opened 2019-07-09
tests: allow reordering of rewrite operations

zyga:
When working with realistic mountinfo tables on systems booted with systemd it
is often the case that the order of certain mount operations its not
deterministic.  This defeats the purpose of rewrite (renumbering and renaming)
operations that is built to allow comparing mount tables.

To combat this introduce rewriting in non-sequential order. The idea is that we
can rewrite the mount table in the order established by, for example,
alphabetic ordering of mount points.

A common example of this issue is the non-deterministic ordering of special
file systems under /sys and /dev. For instance all of those are mounted
simultaneously and can result in different order from one boot to another.

    25 23 0:21 / /dev/shm rw,nosuid,nodev shared:22 - tmpfs tmpfs rw
    26 23 0:22 / /dev/pts rw,nosuid,noexec,relatime shared:23 - devpts devpts rw,gid=5,mode=620,ptmxmode=000
    45 23 0:19 / /dev/mqueue rw,nosuid,nodev,noexec,relatime shared:26 - mqueue mqueue rw
    48 23 0:43 / /dev/hugepages rw,relatime shared:29 - hugetlbfs hugetlbfs rw,pagesize=2M

In addition, for debugging, display ordering is decoupled from renumbering
ordering. I found it useful to look at the mount table as it really is ordered
while experimenting with various other settings.
    
### DIFF
--- a/tests/lib/bin/mountinfo-tool
+++ b/tests/lib/bin/mountinfo-tool
@@ -643,9 +643,10 @@ Known attributes, applicable for both filtering and display.
     except ValueError as exc:
         raise SystemExit(exc)
     entries = [MountInfoEntry.parse(line) for line in opts.file]
-    entries = [e for e in entries if matches(e, filters)]
-    rs = RewriteState()
 
+    # Build rewrite state based on reference tables. This way the entries
+    # we will display can be correlated to other tables.
+    rs = RewriteState()
     for ref in opts.refs:
         ref_entries = [MountInfoEntry.parse(line) for line in ref]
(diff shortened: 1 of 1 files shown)


zyga:
@anonymouse64 it changes the order of rewrite operations, you must use it together with `--renumber` to see the effects. My most practical use is to use `--renumber --rewrite-order mount_point`

zyga:
I'm merging this with one review, happy to revisit if anyone finds things to change.

PR #7086, opened 2019-07-10
many: rename snap info GetType() to Type() (4/4)

stolowski:
Rename snap.Info's `GetType()` to `Type()`. This was previously converted from `Type` member to `GetType() method to catch and refactor all references in the code; this brings the old name back (except for, it is now a method).

Based on #7083


chipaca:
There seems to have been some kind of a messup here, as this PR was somehow identical to #7083 (and merging that one has this one marked as merged). I suspect a mis-push, as there isn't anything in this PR that seems to do what it claims to do?

PR #7088, opened 2019-07-10
tests: manually stop the gvfsd-metadata process

sergiocazzolato:
Seems to be a race between the gvfsd-metadata which is being stopped and
the rm command. By using pkill we make sure the process is finished and
the dir can be removed.

This is the log:

++ ps -ef
...
root     21163     1  0 13:52 ?        00:00:00
/usr/lib/gvfs/gvfsd-metadata
...

++ pidof gvfsd-metadata
+ pid=
+ rm -rf
/home/gopath/src/github.com/snapcore/snapd/tests/main/interfaces-calendar-service/xdg
rm: cannot remove
'/home/gopath/src/github.com/snapcore/snapd/tests/main/interfaces-calendar-service/xdg/share/gvfs-metadata':
Directory not empty

### DIFF
--- a/tests/main/interfaces-calendar-service/task.yaml
+++ b/tests/main/interfaces-calendar-service/task.yaml
@@ -26,12 +26,10 @@ prepare: |
 restore: |
     kill "$(cat dbus-launch.pid)"
 
-    # In case the process gvfsd-metadata does not finish by itself, it is manually stopped
-    # The reason is that gvfsd-metadata locks the xdg/share/gvfs-metadata directory content
-    # producing an error when the xdg directory is removed.
-    if pid="$(pidof gvfsd-metadata)"; then
-        kill -9 "$pid" || true
-    fi
+    # We make sure the gvfsd-metadata process is stopped because it locks the 
+    # xdg/share/gvfs-metadata directory content producing an error when the 
+    # xdg directory is removed.
+    pkill gvfsd-metadata || true
     rm -rf "$XDG"
 
 execute: |
--- a/tests/main/interfaces-contacts-service/task.yaml
+++ b/tests/main/interfaces-contacts-service/task.yaml
@@ -25,12 +25,10 @@ restore: |
         kill "$(cat dbus-launch.pid)"
     fi
 
-    # In case the process gvfsd-metadata does not finish by itself, it is manually stopped
-    # The reason is that gvfsd-metadata locks the xdg/share/gvfs-metadata directory content
-    # producing an error when the xdg directory is removed.
-    if pid="$(pidof gvfsd-metadata)"; then
-        kill -9 "$pid" || true
-    fi
+    # We make sure the gvfsd-metadata process is stopped because it locks the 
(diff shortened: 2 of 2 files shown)


tests/main/interfaces-calendar-service/task.yaml
```diff
@@ -26,12 +26,10 @@ prepare: |
...
-    # In case the process gvfsd-metadata does not finish by itself, it is manually stopped
-    # The reason is that gvfsd-metadata locks the xdg/share/gvfs-metadata directory content
-    # producing an error when the xdg directory is removed.
-    if pid="$(pidof gvfsd-metadata)"; then
-        kill -9 "$pid" || true
-    fi
+    # We make sure the gvfsd-metadata process is stopped because it locks the 
+    # xdg/share/gvfs-metadata directory content producing an error when the 
+    # xdg directory is removed.
+    pkill gvfsd-metadata || true
```
bboozzoo: I think this should preserve KILL, so `pkill -9 gvfsd-metadata`.

tests/main/interfaces-contacts-service/task.yaml
```diff
@@ -25,12 +25,10 @@ restore: |
...
-    # In case the process gvfsd-metadata does not finish by itself, it is manually stopped
-    # The reason is that gvfsd-metadata locks the xdg/share/gvfs-metadata directory content
-    # producing an error when the xdg directory is removed.
-    if pid="$(pidof gvfsd-metadata)"; then
-        kill -9 "$pid" || true
-    fi
+    # We make sure the gvfsd-metadata process is stopped because it locks the 
+    # xdg/share/gvfs-metadata directory content producing an error when the 
+    # xdg directory is removed.
+    pkill gvfsd-metadata || true
```
bboozzoo: same here

tests/main/interfaces-calendar-service/task.yaml
bboozzoo: OTOH, this does not seem to address the problem. Rough guess, but `gvfsd-metadata` is probably writing to that directory. By the time `rm -rf ..` reaches `rmdir()` (and having removed entries that were inside), `gvfsd-metada` has probably created a new file inside, hence triggering `ENOTEMPTY`. Perhaps we ought to kill all gvfs processes instead, including `gvfsd`, so `pkill -9 gvfs` instead?

bboozzoo: Talked about this during standup yesterday. From digging into gvfs source code, it looks like all writes to metadata go through the gvfsd-metadata. That daemon is DBus activated. If there's any other gvfs process running (gvfsd or one of the monitors) they could try to access the metadata daemon and DBus will just resurrect it. Thus, I think a better solution is to make sure that the whole gvfs gang is stopped.

sergiocazzolato: @bboozzoo sure, I'll make that change. Thanks for reviewing!!

sergiocazzolato:
I could not reproduce this error anymore, I'll close the PR until the error appears again.

PR #7089, opened 2019-07-10
tests: don't preserve size= when rewriting mount tables

zyga:
When systemd mounts tmpfs it will compute and provide the maximum size
of the filesystem as a fraction of the available memory. This quantity
changes from execution to execution I've seen renumbering "failures" as
the computed quantity for /run/user/0 oscillated between 378180k and
378184k, resulting in either reuse of earlier renumbering of /run or an
allocation of a new value. To avoid this issue replace size= with
constant size=VARIABLE.

### DIFF
--- a/tests/lib/bin/mountinfo-tool
+++ b/tests/lib/bin/mountinfo-tool
@@ -922,7 +922,7 @@ class RenumberMountOptionsTests(unittest.TestCase):
         self.assertEqual(renumber_mount_option("fd=45", self.seen), "fd=1")
         self.assertEqual(renumber_mount_option("fd=21", self.seen), "fd=0")
 
-    def test_size_not_allocated(self):
+    def test_renumber_size_variable(self):
         # type: () -> None
         """
         size is special-cased and is not allocated because it is prone to fluctuations
--- a/tests/lib/bin/mountinfo-tool
+++ b/tests/lib/bin/mountinfo-tool
@@ -925,7 +925,7 @@ class RenumberMountOptionsTests(unittest.TestCase):
     def test_renumber_size_variable(self):
         # type: () -> None
         """
-        size is special-cased and is not allocated because it is prone to fluctuations
+        size is special-cased and always rewritten to the same value because it is prone to fluctuations
         """
         self.assertEqual(renumber_mount_option("size=100", self.seen), "size=VARIABLE")
         self.assertEqual(renumber_mount_option("size=200", self.seen), "size=VARIABLE")
--- a/tests/lib/bin/mountinfo-tool
+++ b/tests/lib/bin/mountinfo-tool
@@ -467,7 +467,9 @@ def renumber_mount_option(opt, seen):
         mount_opt_key, mount_opt_value = opt.split("=", 1)
         # size, nr_inode: used by tmpfs
         # fd, pipe_ino: used by binfmtmisc
-        if mount_opt_key in {"size", "nr_inodes", "fd", "pipe_ino"}:
+        if mount_opt_key == "size":
+            return "size=VARIABLE"
+        if mount_opt_key in {"nr_inodes", "fd", "pipe_ino"}:
             return "{}={}".format(
                 mount_opt_key, alloc_n(mount_opt_key, mount_opt_value)
             )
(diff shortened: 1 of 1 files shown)


tests/lib/bin/mountinfo-tool
```diff
@@ -911,34 +913,44 @@ class RenumberMountOptionsTests(unittest.TestCase):
         """
         numbers are allocated from subsets matching the key, this reduces delta.
         """
-        self.assertEqual(renumber_mount_option("size=100", self.seen), "size=0")
-        self.assertEqual(renumber_mount_option("size=100", self.seen), "size=0")
-        self.assertEqual(renumber_mount_option("size=200", self.seen), "size=1")
-        self.assertEqual(renumber_mount_option("size=100", self.seen), "size=0")
+        self.assertEqual(renumber_mount_option("pipe_ino=100", self.seen), "pipe_ino=0")
```
anonymouse64: so just to understand this change here, previously all the size options were rewritten in order of when they were seen, i.e. the first size option seen becomes size=0 and the next unique value becomes size=1, etc. 

And now, this specific test no longer tests the size parameter and instead tests that the pipe_ino parameter gets rewritten like that. 

The new size behavior is tested below with `test_size_not_allocated`.

tests/lib/bin/mountinfo-tool
```diff
@@ -911,34 +913,44 @@ class RenumberMountOptionsTests(unittest.TestCase):
...
+        self.assertEqual(renumber_mount_option("pipe_ino=100", self.seen), "pipe_ino=0")
+        self.assertEqual(renumber_mount_option("pipe_ino=200", self.seen), "pipe_ino=1")
+        self.assertEqual(renumber_mount_option("pipe_ino=100", self.seen), "pipe_ino=0")
         self.assertEqual(renumber_mount_option("fd=21", self.seen), "fd=0")
         self.assertEqual(renumber_mount_option("fd=21", self.seen), "fd=0")
         self.assertEqual(renumber_mount_option("fd=45", self.seen), "fd=1")
         self.assertEqual(renumber_mount_option("fd=21", self.seen), "fd=0")
 
+    def test_size_not_allocated(self):
```
anonymouse64: I think a more clear name for this function would be `test_renumber_size_variable` because the function currently always rewrites size to be variable and it's not clear what allocation refers to in this context if you're just looking at this function and not at the history or the function right above it.

```suggestion
    def test_renumber_size_variable(self):
```

tests/lib/bin/mountinfo-tool
```diff
@@ -911,34 +913,44 @@ class RenumberMountOptionsTests(unittest.TestCase):
...
+        # type: () -> None
+        """
+        size is special-cased and is not allocated because it is prone to fluctuations
```
anonymouse64: I think this would be more clear if written as:

```suggestion
        size is special-cased and always rewritten because it is prone to fluctuations
```

tests/lib/bin/mountinfo-tool
zyga: I moved the test for the size parameter to a new function, replacing size with another renumbered quantity, pipie_ino.

tests/lib/bin/mountinfo-tool
```diff
@@ -911,34 +913,44 @@ class RenumberMountOptionsTests(unittest.TestCase):
...
+        )
         self.assertEqual(
             renumber_mount_option("nr_inodes=1014347", self.seen), "nr_inodes=0"
         )
         # Option mode= is not renumbered.
         self.assertEqual(renumber_mount_option("mode=755", self.seen), "mode=755")
-        self.assertEqual(
-            self.seen, {("size", "4057388k"): 0, ("nr_inodes", "1014347"): 0}
-        )
+        self.assertEqual(self.seen, {("nr_inodes", "1014347"): 0})
```
bboozzoo: While at it, is `nr_inodes` stable across reboots?

zyga: Yeah, I haven't seen any issues with that. I think it depends on storage size, if anything.

bboozzoo: Digging into the kernel, looks like it's calculated here: https://elixir.bootlin.com/linux/latest/source/mm/shmem.c#L116 so indeed stable.

PR #7091, opened 2019-07-11
tests: measure properties of various  mount namespaces

zyga:
The mount namespace is tricky and changes were not measured in any
explicit way, apart from what various tests were doing inside that
namespace.

This test established select measurements of the:
 - host mount namespace
 - per snap mount namespace
 - per snap, per user mount namespace

In all cases only certain distributions are measured, this is because
their initial mount namespace differs, and this echoes into the two
per-snap mount namespaces. While some careful and perhaps tricky
selective comparison might be possible I chose to instead be explicit
and only run the test on some distributions. It is always easy to
support another distribution release and to use data used by this test
to compare one distribution with another.

The set of tested distributions contains:
 - ubuntu classic 16.04 and 18.04
 - ubuntu core 16 and 18

On each distribution the set of snap mount namespace measured contains:
 - a core16-based snap (aka without specific base)
 - a core18-based snap
 - a classic snap (only on classic systems)


zyga:
I force pushed to remove the qemu data set. We can add it back after the initial review.

tests/main/mount-ns/task.yaml
```diff
@@ -0,0 +1,112 @@
+summary: the shape of the mount namespace on classic systems for non-classic snaps
```
stolowski: Nitpick: we are not super consistent with this, but I think we generally start with capital letters in task summaries.

tests/main/mount-ns/task.yaml
```diff
@@ -0,0 +1,112 @@
...
+        su test -c "snap run test-snapd-mountinfo" >PER-USER-C7.raw.txt
+        sudo snap remove test-snapd-mountinfo
+    fi
+
+    echo "Collect mountinfo from core16-based, per-snap and per-snap, per-user mount namespaces"
+    snap pack test-snapd-mountinfo.core16
+    snap install --dangerous test-snapd-mountinfo_16_all.snap
+    snap connect test-snapd-mountinfo:mount-observe
+    su root -c "snap run test-snapd-mountinfo" >PER-SNAP-16.raw.txt
+    su test -c "snap run test-snapd-mountinfo" >PER-USER-16.raw.txt
```
stolowski: Nice.

tests/main/mount-ns/task.yaml
```diff
@@ -0,0 +1,112 @@
...
+    Lastly, on core systems the test is somewhat artificial as it also measures
+    changes performed by the test preparation process. Some of the mount
+    entries are not present on real core systems.
+
+    If you see this test randomly failing it may be because it has observed
+    state leaked by another test that ran on the same machine earlier in the
+    spread  execution chain.
+prepare: |
+    deterministic-mountinfo-tool() {
+        mountinfo-tool --renumber --rename --rewrite-order mount_point --display-order mount_point "$@"
```
stolowski: \o/

tests/main/mount-ns/task.yaml
```diff
@@ -0,0 +1,112 @@
...
+        diff -u "$SPREAD_BACKEND.$SPREAD_SYSTEM/PER-SNAP-C7.expected.txt" PER-SNAP-C7.deterministic.txt
+        diff -u "$SPREAD_BACKEND.$SPREAD_SYSTEM/PER-USER-C7.expected.txt" PER-USER-C7.deterministic.txt
+    fi
+restore: |
+    rm -f ./*.raw.txt ./*.deterministic.txt
+    rm -f test-snapd-mountinfo_c7_all.snap
+    rm -f test-snapd-mountinfo_16_all.snap
+    rm -f test-snapd-mountinfo_18_all.snap
+debug: |
+    for fname in ./*.deterministic.txt; do
```
stolowski: I fear this will generate too much noise and afaiu is not going to be super helpful. I think the failing diff output shoud be enough, no?

tests/main/mount-ns/task.yaml
zyga: Just propose a suggestion please. I'll gladly take it.

tests/main/mount-ns/task.yaml
zyga: This is actually more useful because you can take the full file and inspect it offline with tools better than just diff alone. It is also very useful for creating new data sets for things you don't have locally.

I agree it will be noisy if this fails often but I think we can reconsider after the initial painful phase.

tests/main/mount-ns/task.yaml
stolowski:
```suggestion
summary: The shape of the mount namespace on classic systems for non-classic snaps
```

tests/main/mount-ns/task.yaml
stolowski: Ok, I trust you on this since you've been dealing with these failures while working on the test. My main point is we're outputing here all of them, while showing the one that failed would be enough, right? Perhaps we should simply `rm -rf <name>.deterministic.txt`after every diff step, so we are immediately getting rid of sucesful ones and not show then in debug on failure?

zyga:
@stolowski the data files are real and I would recommend one to have a look at two aspects, though perhaps not in this PR as that would just add noise here:

Propagation settings, those can be seen as master:xxx, shared:xxx and similar. The interesting thing is to see if the settings are sensible between host, per-snap and per-user.

Construction of legacy (core16) and non-legacy (core18) mount namespaces. I would say there are some dragons there. I will attempt to explain that in subsequent PRs that change things.

tests/main/mount-ns/task.yaml
zyga: I would say that the interesting one is the HOST as that will show cruft leaking into the mount namespace. Non-determinism of the remaining ones seems to be handled well now and any differences there will be reflections of what HOST differs.

As I said, I'm happy to evolve this but I would keep it as-is for now, because having easy access to this even without `-debug` is very useful. I can change how the debug section works after the qemu data set is merged.

tests/main/mount-ns/task.yaml
```diff
@@ -0,0 +1,112 @@
...
+    This test measures the mount table of the host, of the mount namespace for
+    a simple core16-based snap for the per-user mount namespace of a simple
+    core16-base snap as well as for a simple core18 based snap and finally of a
+    simple snap using classic confinement.
+
+    The mount tables are obtained from /proc/pid/mountinfo interface. They are
+    then fed to mountinfo-tool with various determinism adjustment options and
+    compared to stock tables created when the test was first made.
+
+    Naturally occurring mount tables will differ from invocation to invocation
```
bboozzoo: I think you can drop this paragraph and update the previous one like this:
```
.. then fed to mountinfo-tool with determinism adjustment options (--rename, --renumber) and compared ...
```

tests/main/mount-ns/task.yaml
```diff
@@ -0,0 +1,112 @@
...
+    /snap/name/revision, which can be all started in parallel. Another one is
+    the set of control groups mounted in sysfs. Some mount options may be
+    created based on the amount of available memory. Some mount entries may
+    contain numeric IDs that are allocated and are hard to predict. Block
+    devices backing actual file systems may be on various disks, for instance
+    on /dev/vda or /dev/sdb. All of those are handled by mountinfo-tool
+    --rename and --renumber, along with sorting options that combat
+    non-deterministic mount order.
+
+    Unfortunately this test need separate data sets for google compute engine
```
bboozzoo: The initial mount namespace differs between test backends. This is captured by grouping the test data by backend type.

tests/main/mount-ns/task.yaml
```diff
@@ -0,0 +1,112 @@
...
+        diff -u "$SPREAD_BACKEND.$SPREAD_SYSTEM/PER-SNAP-C7.expected.txt" PER-SNAP-C7.deterministic.txt
+        diff -u "$SPREAD_BACKEND.$SPREAD_SYSTEM/PER-USER-C7.expected.txt" PER-USER-C7.deterministic.txt
+    fi
+restore: |
+    rm -f ./*.raw.txt ./*.deterministic.txt
+    rm -f test-snapd-mountinfo_c7_all.snap
+    rm -f test-snapd-mountinfo_16_all.snap
+    rm -f test-snapd-mountinfo_18_all.snap
+debug: |
+    for fname in ./*.deterministic.txt; do
```
bboozzoo: If you don't mind line numbers, the whole loop could be replaced by:
```
grep -n '' ./*.deterministic.txt
```
OTOH perhaps the output of `diff` is enough

zyga: About `grep`, yeah but ... nah, this is more readable IMO. As for `diff` I summarised my opinion in replies to Pawel. This is very useful for grabbing the full data set in case something goes south (diff context is not that useful. and shows only the first failure). I'm open to removing it once qemu is merged and we have "stable" passes.

zyga:
Some test is corrupting LXC filesystem, I'll hunt that down now.

zyga:
Naive look at things that touch `/var/lib/lxcfs` yields nothing. I will look a the test sequence and look at tests one by one.

zyga:
I think I finally got them all. The test should now pass. The helper PRs are proposed separately for evaluation. Please postpone merging until https://github.com/snapcore/snapd/pull/7138,  https://github.com/snapcore/snapd/pull/7139 and https://github.com/snapcore/snapd/pull/7141 are merged.

zyga:
Uh, I forgot the data sets for core 16 and core 18. Updating.

zyga:
I'll adjust wording in the follow-up PR.

PR #7093, opened 2019-07-11
seccomp/compiler: adjust test case names and comment for later changes

jdstrand:


bboozzoo:
I've edited the title so that our PR title checker does not complain.

PR #7096, opened 2019-07-12
cmd/snap: handle unsetting of config options with "!"

stolowski:
Handle exclamation mark after option name for unsetting it with snap set, e.g. `snap set mysnap foo!`.

This is done on the protocol/API level by passing nulls as option values and relies on already landed changes to config transaction internals.

Similiar change to `snapctl` as well as `snap unset` command will be proposed in a separate PRs.

### DIFF
--- a/cmd/snap/cmd_set.go
+++ b/cmd/snap/cmd_set.go
@@ -64,7 +64,7 @@ func init() {
 			// TRANSLATORS: This needs to begin with < and end with >
 			name: i18n.G("<conf value>"),
 			// TRANSLATORS: This should not start with a lowercase letter.
-			desc: i18n.G("Configuration value (key=value) or key!"),
+			desc: i18n.G("Set (key=value) or unset (key!) configuration value"),
 		},
 	})
 }
--- a/cmd/snap/cmd_set.go
+++ b/cmd/snap/cmd_set.go
@@ -74,7 +74,7 @@ func (x *cmdSet) Execute(args []string) error {
 	for _, patchValue := range x.Positional.ConfValues {
 		parts := strings.SplitN(patchValue, "=", 2)
 		if len(parts) == 1 && strings.HasSuffix(patchValue, "!") {
-			patchValues[patchValue[:len(patchValue)-1]] = nil
+			patchValues[strings.TrimSuffix(patchValue, "!")] = nil
 			continue
 		}
 		if len(parts) != 2 {
--- a/cmd/snap/cmd_set.go
+++ b/cmd/snap/cmd_set.go
@@ -41,6 +41,9 @@ snap's configuration hook returns successfully.
 Nested values may be modified via a dotted path:
 
     $ snap set author.name=frank
+
+Configuration option may be unset with exclamation mark:
+    $ snap set author!
 `)
 
 type cmdSet struct {
@@ -61,7 +64,7 @@ func init() {
 			// TRANSLATORS: This needs to begin with < and end with >
 			name: i18n.G("<conf value>"),
 			// TRANSLATORS: This should not start with a lowercase letter.
-			desc: i18n.G("Configuration value (key=value)"),
+			desc: i18n.G("Configuration value (key=value) or key!"),
 		},
 	})
 }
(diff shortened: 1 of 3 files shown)


tests/main/snap-set/task.yaml
```diff
@@ -46,6 +46,18 @@ execute: |
     fi
     snap get snapctl-hooks foo 2>&1 | MATCH 'snap "snapctl-hooks" has no "foo" configuration option'
 
+    echo "Set foo back"
+    snap set snapctl-hooks command=test-snapctl-get-foo foo=bar
+    if ! obtained=$(snap get snapctl-hooks foo); then
+        echo "Expected snap get to be able to retrieve set value"
+        exit 1
+    fi
+    [[ "$obtained" == "bar" ]]
```
zyga:
```suggestion
    test "$obtained" = "bar"
```

cmd/snap/cmd_set.go
```diff
@@ -70,6 +73,10 @@ func (x *cmdSet) Execute(args []string) error {
 	patchValues := make(map[string]interface{})
 	for _, patchValue := range x.Positional.ConfValues {
 		parts := strings.SplitN(patchValue, "=", 2)
+		if len(parts) == 1 && strings.HasSuffix(patchValue, "!") {
+			patchValues[patchValue[:len(patchValue)-1]] = nil
```
zyga: TrimSuffix
```suggestion
			patchValues[strings.TrimSuffix(patchValue, "!")] = nil
```

tests/main/snap-set/task.yaml
```diff
@@ -46,6 +46,18 @@ execute: |
     fi
     snap get snapctl-hooks foo 2>&1 | MATCH 'snap "snapctl-hooks" has no "foo" configuration option'
 
+    echo "Set foo back"
+    snap set snapctl-hooks command=test-snapctl-get-foo foo=bar
+    if ! obtained=$(snap get snapctl-hooks foo); then
```
zyga: I have not seen this syntax before. Interesting.

cmd/snap/cmd_set.go
```diff
@@ -61,7 +64,7 @@ func init() {
 			// TRANSLATORS: This needs to begin with < and end with >
 			name: i18n.G("<conf value>"),
 			// TRANSLATORS: This should not start with a lowercase letter.
-			desc: i18n.G("Configuration value (key=value)"),
+			desc: i18n.G("Configuration value (key=value) or key!"),
```
zyga:
```suggestion
			desc: i18n.G("Set (key=value) or unset (key!) configuration value"),
```

tests/main/snap-set/task.yaml
stolowski: Yep, I saw it in some other test the other day and borrowed it ;)

cmd/snap/cmd_set_test.go
```diff
@@ -98,6 +98,34 @@ func (s *SnapSuite) TestSnapSetIntegrationJson(c *check.C) {
 	c.Assert(err, check.IsNil)
 }
 
+func (s *SnapSuite) TestSnapSetIntegrationUnsetWithExclamationMark(c *check.C) {
+	// mock installed snap
+	snaptest.MockSnap(c, string(validApplyYaml), &snap.SideInfo{
```
mvo5: We don't need this AFAICT - we only hit the mock server which will not look at any snap or state. Removing it (here and in the other tests) seems to be fine, the tests are still passing. The same for "validApplyYaml".

cmd/snap/cmd_set_test.go
```diff
@@ -98,6 +98,34 @@ func (s *SnapSuite) TestSnapSetIntegrationJson(c *check.C) {
...
+		Revision: snap.R(1),
+	})
+
+	// and mock the server
+	s.mockSetConfigServer(c, nil)
```
mvo5: Would be nice to have a check here that ensures that the API endpoint was hit.

cmd/snap/cmd_set.go
```diff
@@ -41,6 +41,9 @@ snap's configuration hook returns successfully.
 Nested values may be modified via a dotted path:
 
     $ snap set author.name=frank
+
+Configuration option may be unset with exclamation mark:
+    $ snap set author!
```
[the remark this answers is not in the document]
mvo5: Silly me, I think this is not needed, the shell seems to be fine with that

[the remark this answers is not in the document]
pedronis: afaict the shell gets confused only if the ! has things after it that are not spaces

PR #7098, opened 2019-07-12
snapctl: handle unsetting of config options with "!"

stolowski:
Handle exclamation mark after option name for unsetting it with snapctl set, e.g. `snapctl set foo!`.
The unsetting logic is internally handled via nulls in the transaction logic (this change already landed some time ago).

This is a followup for #7096

### DIFF
--- a/tests/lib/snaps/snapctl-hooks/meta/hooks/configure
+++ b/tests/lib/snaps/snapctl-hooks/meta/hooks/configure
@@ -90,6 +90,14 @@ test_snapctl_foo_null() {
 	  fi
 }
 
+test_snapctl_unset() {
+	  echo "Unsetting an option"
+	  if ! snapctl set root.key2! ; then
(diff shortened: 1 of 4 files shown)


tests/lib/snaps/snapctl-hooks/meta/hooks/configure
```diff
@@ -90,6 +90,14 @@ test_snapctl_foo_null() {
 	  fi
 }
 
+test_snapctl_unset() {
+	  echo "Unsetting an option"
+	  if ! snapctl set root.key2! ; then
+	  echo "snapctl set unexpectedly failed when un-setting root.key2"
```
mvo5: Indent looks funky

mvo5: I fixed this indent. There is another funny one at the bottom but the entire file is messed up so I don't want to do too much there.

PR #7109, opened 2019-07-15
snap-confine: fallback gracefully on a cgroup v2 only system

mvo5:
On a cgroup v2 only system we cannot use certain features like
freezer or pid cgroups yet. In order to not die() when we run
on such systems this PR implements detection and skipping of
those.

This will have to be done "properly" but for now it ensures
that snaps continue to run on cgroup v2 only systems.

This obviously needs a fedora 31 spread machine. Mostly pushing it out to show what we can do.


cmd/libsnap-confine-private/cgroup-support.c
```diff
@@ -66,3 +67,18 @@ void sc_cgroup_create_and_join(const char *parent, const char *name, pid_t pid)
     }
     debug("moved process %ld to cgroup hierarchy %s/%s", (long)pid, parent, name);
 }
+
+static const char *cgroup_dir = "/sys/fs/cgroup";
+// from statfs(2)
+static const int CGROUP2_SUPER_MAGIC =  0x63677270;
```
anonymouse64: I assume that this isn't inherited from say `include/uapi/linux/magic.h` because it's not available on all the systems we need to compile on and also because we don't want to use external dependencies at all in snap-confine?

I see that this magic number is defined in include/uapi/linux/magic.h on bionic and disco at least (in `cat /usr/src/linux-headers-$(uname -r)/include/uapi/linux/magic.h  | grep CGROUP2`), but notably not on xenial. From my experience I find it more maintainable to inherit magic values like these from an upstream source whenever I can, but I understand if that's not desirable for snap-confine specifically.

cmd/snap-confine/snap-confine.c
```diff
@@ -574,8 +575,11 @@ static void enter_non_classic_execution_environment(sc_invocation * inv,
 
 	/** Populate and join the device control group. */
 	struct snappy_udev udev_s;
-	if (snappy_udev_init(inv->security_tag, &udev_s) == 0)
-		setup_devices_cgroup(inv->security_tag, &udev_s);
+	if (snappy_udev_init(inv->security_tag, &udev_s) == 0) {
+		if (sc_cgroup_is_v2()) {
```
anonymouse64: if we are disabling cgroups when on cgroupv2, shouldn't this be 

```suggestion
		if (!sc_cgroup_is_v2()) {
```

cmd/libsnap-confine-private/cgroup-support.c
bboozzoo: FWIW the value is correct https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/magic.h#L62

cmd/snap-confine/snap-confine.c
```diff
@@ -574,8 +575,11 @@ static void enter_non_classic_execution_environment(sc_invocation * inv,
 
 	/** Populate and join the device control group. */
 	struct snappy_udev udev_s;
-	if (snappy_udev_init(inv->security_tag, &udev_s) == 0)
-		setup_devices_cgroup(inv->security_tag, &udev_s);
+	if (snappy_udev_init(inv->security_tag, &udev_s) == 0) {
+		if (!sc_cgroup_is_v2()) {
```
bboozzoo: This will produce a generic cgroups v2 not supported warning. Perhaps we should inform which feature is actually getting disabled.

cmd/libsnap-confine-private/cgroup-support.c
zyga: Technical nitpick: it is safer to check if the macro is defined and only define it if it is *not*. This way you are safe from something else including the constant and ruining your day.
```
#ifndef CGRUOP2_SUPER_MAGIC
#define CGROUP2_SUPER_MAGIC 0x63677270
#endif 
```

cmd/libsnap-confine-private/cgroup-support.c
```diff
@@ -66,3 +67,18 @@ void sc_cgroup_create_and_join(const char *parent, const char *name, pid_t pid)
...
+
+bool sc_cgroup_is_v2() {
+    struct statfs buf;
+
+    int err = statfs(cgroup_dir, &buf);
+    if (err == 0 && buf.f_type == CGROUP2_SUPER_MAGIC) {
+       fprintf(stderr, "WARNING: cgroup v2 is not fully supported yet\n");
```
zyga: Nitpick: given that we call `sc_cgroup_is_v2` several times, could you please keep a `static int did_warn` flag to only print this once?

cmd/libsnap-confine-private/cgroup-support.c
```diff
@@ -66,3 +67,18 @@ void sc_cgroup_create_and_join(const char *parent, const char *name, pid_t pid)
...
     debug("moved process %ld to cgroup hierarchy %s/%s", (long)pid, parent, name);
 }
+
+static const char *cgroup_dir = "/sys/fs/cgroup";
+// from statfs(2)
+static const int CGROUP2_SUPER_MAGIC =  0x63677270;
+
+bool sc_cgroup_is_v2() {
+    struct statfs buf;
+
```
zyga: Can you please add a comment like this:
```
Detect if we are running in cgroup v2 unified mode (as opposed to hybrid or legacy)
The algorithm is described in https://systemd.io/CGROUP_DELEGATION.html
```

cmd/libsnap-confine-private/cgroup-support.h
```diff
@@ -30,4 +31,10 @@
  **/
 void sc_cgroup_create_and_join(const char *parent, const char *name, pid_t pid);
 
+/**
+ * sc_cgroup_is_v2() returns true if running on cgroups v2
+ *
+ **/
+bool sc_cgroup_is_v2(void);
```
zyga: +1 for defining the prototype correctly :)

cmd/libsnap-confine-private/cgroup-support.c
```diff
@@ -66,3 +67,18 @@ void sc_cgroup_create_and_join(const char *parent, const char *name, pid_t pid)
...
+    int err = statfs(cgroup_dir, &buf);
+    if (err == 0 && buf.f_type == CGROUP2_SUPER_MAGIC) {
```
zyga: I would prefer if we handle `EPERM` and die() (at least for now).

cmd/libsnap-confine-private/cgroup-support.h
mvo5: Haha, I read a while ago about the subtle differences between foo() and foo(void) and it makes me cry a bit

cmd/libsnap-confine-private/cgroup-support.c
mvo5: Is EPERM just an example? I ask because I don't see it in the statvfs manpage. I guess we can die() on anything but EFAULT or ENOENT?

zyga: Indeed, perhaps the logic should be swapped. It's okay to have ENOENT but anything else should be die() IMO.

cmd/libsnap-confine-private/cgroup-support.c
```diff
@@ -66,3 +67,24 @@ void sc_cgroup_create_and_join(const char *parent, const char *name, pid_t pid)
     }
     debug("moved process %ld to cgroup hierarchy %s/%s", (long)pid, parent, name);
 }
+
+static const char *cgroup_dir = "/sys/fs/cgroup";
+
+// from statfs(2)
+#ifndef CGRUOP2_SUPER_MAGIC
+#define CGROUP2_SUPER_MAGIC =  0x63677270;
```
bboozzoo:
```suggestion
#define CGROUP2_SUPER_MAGIC  0x63677270
```

bboozzoo:
This has been included in latest update of snapd in Fedora. I think we should land the change as is, at least snaps don't fail out of the box.

cmd/snap-confine/ns-support.c
```diff
@@ -486,7 +487,7 @@ static int sc_inspect_and_maybe_discard_stale_ns(int mnt_fd,
 		debug("preserved mount is not stale, reusing");
 		return 0;
 	case SC_DISCARD_SHOULD:
-		if (sc_cgroup_freezer_occupied(inv->snap_instance)) {
+                if (!sc_cgroup_is_v2() && sc_cgroup_freezer_occupied(inv->snap_instance)) {
```
jdstrand: I wonder if sc_cgroup_is_v2() we should always assume it is stale, but occupied? Or trigger SC_DISCARD_MUST. It is hard to know what to do, but my thinking is that perhaps we can at least provide dependable, broken behavior for snaps on v2 only systems, as opposed to the occasional odd error.

cmd/snap-confine/snap-confine.c
jdstrand: I'm conflicted on this. There is no point in udev tagging if there isn't a device cgroup, but on the other hand, not doing it is a larger runtime change. With this proposed change, snap-device-helper will bail early since `/sys/fs/cgroup/devices/snap.*` should not exist (only snap-confine creates it thus far) so we *should* be fine. I suspect a number of spread tests will need to omit cgroup v2 systems.

bboozzoo:
I'll be pushing some patches to address the comments.

cmd/snap-confine/ns-support.c
zyga: I think this is a useful compromise on the way towards better cgroup v2 support.

cmd/libsnap-confine-private/cgroup-support.c
bboozzoo: Added

cmd/snap-confine/ns-support.c
bboozzoo: Added a code path for this.

bboozzoo:
@jdstrand addressed your comments, can you take another look?

cmd/libsnap-confine-private/cgroup-support.c
jdstrand: Nitpick (and sorry for not suggesting this earlier), I think this would be a little easier to read:
```
    if (statfs(cgroup_dir, &buf) != 0) {
        if (errno == ENOENT) {
            return false;
        }
        die("cannot statfs %s", cgroup_dir);
    }
    if (buf.f_type == CGROUP2_SUPER_MAGIC) {
        if (!did_warn) {
            fprintf(stderr, "WARNING: cgroup v2 is not fully supported yet\n");
            did_warn = true;
        }
        return true;
    }
    return false;
```

bboozzoo:
@jdstrand discussed the concerns with @mvo5. For now I have tweaked the warning message to make it clear that confinement is lowered. I will be opening a PR with bits for cgroup detection in Go and then we can proceed with adding some info in system-key and making sure that /v2/system-info returns proper information about confinement.

mvo5:
:+1:

PR #7110, opened 2019-07-15
tests: new test to check the output after refreshing/reverting core

sergiocazzolato:
Ths change includes:
. New test to validate that after installing core or core18 snaps,
  it is triggered a reboot and the command would exit after the
  first phase of the installation reporting "snapd is about to
  reboot the system"
. snapd snap was remaining as part of the test after the prepare
  phase

### DIFF
--- a/tests/main/ubuntu-core-refresh/task.yaml
+++ b/tests/main/ubuntu-core-refresh/task.yaml
@@ -1,0 +1,61 @@
+summary: check that the ubuntu-core system is rebooted after the core snap is refreshed
+
+details: |
+    This test checks that when invoking a manual refresh/revert for core or core18 snaps,
+    it is triggered a reboot and the command would exit after the first phase of the installation
+    reporting "snapd is about to reboot the system"
+
+systems: [ubuntu-core-1*]
+
+execute: |
+    #shellcheck source=tests/lib/journalctl.sh
+    . "$TESTSLIB"/journalctl.sh
+    #shellcheck source=tests/lib/systems.sh
+    . "$TESTSLIB"/systems.sh
+
+    TARGET_SNAP_NAME=core
+    if is_core18_system; then
+        TARGET_SNAP_NAME=core18
+    fi    
+
+    # After installing a new version of the core/core18 snap the system is rebooted
+    if [ "$SPREAD_REBOOT" = 0 ]; then
+        snap list | awk "/^${TARGET_SNAP_NAME} / {print(\$3)}" > initialRev    
+
+        # use journalctl wrapper to grep only the logs collected while the test is running
+        if get_journalctl_log | MATCH "Waiting for system reboot"; then
+            echo "Already waiting for system reboot, exiting..."
+            exit 1
+        fi
+
+        # install new target snap
+        snap download "$TARGET_SNAP_NAME"
+        snap install --dangerous "${TARGET_SNAP_NAME}"_*.snap &> refresh.log
+        MATCH "snapd is about to reboot the system" < refresh.log
+
+        # use journalctl wrapper to grep only the logs collected while the test is running
+        while ! check_journalctl_log "Waiting for system reboot"; do
+            sleep 1
+        done
+
+        REBOOT
+    elif [ "$SPREAD_REBOOT" = 1 ]; then
+        # Check the current revision has changed
+        currRev=$(snap list | awk "/^${TARGET_SNAP_NAME} / {print(\$3)}") 
+        [ "$(cat initialRev)" !=  "$currRev" ]
+
+        # revert the target snap
+        snap revert "$TARGET_SNAP_NAME" &> revert.log
+        MATCH "snapd is about to reboot the system" < revert.log
+
+        # use journalctl wrapper to grep only the logs collected while the test is running
+        while ! check_journalctl_log "Waiting for system reboot" -b; do
+            sleep 1
+        done
+
+        REBOOT
+    elif [  "$SPREAD_REBOOT" = 2 ]; then
+        # Check the current revision is the same than the original
+        currRev=$(snap list | awk "/^${TARGET_SNAP_NAME} / {print(\$3)}") 
+        [ "$(cat initialRev)" ==  "$currRev" ]
+    fi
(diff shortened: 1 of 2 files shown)


mvo5:
The test in core16 fails after the first reboot it seems:
```
...
+ is_core18_system
+ [[ ubuntu-core-16-64 == ubuntu-core-18-* ]]
+ return 1
+ '[' 1 = 0 ']'
+ '[' 1 = 1 ']'
++ awk '/^core / {print($3)}'
++ snap list
+ currRev=x2
++ cat initialRev
+ '[' x1 '!=' x2 ']'
+ snap revert core
-----
travis_time:end:bee45b9d:start=1563231815339367325,finish=1563231860557337758,duration=45217970433
travis_fold:end:fold-bee45b9d
```
Its not at all clear why though :/ Maybe we need to "cat revert.log" in debug to see more details?

tests/main/ubuntu-core-refresh/task.yaml
```diff
@@ -0,0 +1,65 @@
+summary: check that the ubuntu-core system is rebooted after the core snap is refreshed
```
stolowski: Nitpick, summmaries start with capital letter.

tests/main/ubuntu-core-refresh/task.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+details: |
+    This test checks that when invoking a manual refresh/revert for core or core18 snaps,
+    it is triggered a reboot and the command would exit after the first phase of the installation
```
stolowski: Nitpick, "it is triggered a reboot" -> "a reboot is triggered"

tests/main/ubuntu-core-refresh/task.yaml
```diff
@@ -0,0 +1,65 @@
...
+            exit 1
+        fi
+
+        # install new target snap
+        snap download "$TARGET_SNAP_NAME"
+        snap install --dangerous "${TARGET_SNAP_NAME}"_*.snap &> refresh.log
+        MATCH "snapd is about to reboot the system" < refresh.log
+
+        # use journalctl wrapper to grep only the logs collected while the test is running
+        while ! check_journalctl_log "Waiting for system reboot"; do
```
stolowski: This has a potential for looping infinitely until kill timeout, there should be a limit for this loop. I think the comment should be changed too?

tests/main/ubuntu-core-refresh/task.yaml
```diff
@@ -0,0 +1,65 @@
...
+        # Check the current revision has changed
+        currRev=$(snap list | awk "/^${TARGET_SNAP_NAME} / {print(\$3)}") 
+        [ "$(cat initialRev)" !=  "$currRev" ]
+
+        # revert the target snap
+        snap revert "$TARGET_SNAP_NAME" &> revert.log
+        MATCH "snapd is about to reboot the system" < revert.log
+
+        # use journalctl wrapper to grep only the logs collected while the test is running
+        while ! check_journalctl_log "Waiting for system reboot" -b; do
```
stolowski: Same remark about infinite loop and the comment.

tests/main/ubuntu-core-refresh/task.yaml
sergiocazzolato: Nice, thanks!

tests/main/ubuntu-core-refresh/task.yaml
sergiocazzolato: done

tests/main/ubuntu-core-refresh/task.yaml
sergiocazzolato: done

tests/main/ubuntu-core-refresh/task.yaml
sergiocazzolato: done

tests/main/ubuntu-core-refresh/task.yaml
sergiocazzolato: done

sergiocazzolato:
@stolowski thanks for the review! All the comments have been addressed.

tests/main/ubuntu-core-refresh/task.yaml
```diff
@@ -52,11 +60,19 @@ execute: |
         snap revert "$TARGET_SNAP_NAME" &> revert.log
         MATCH "snapd is about to reboot the system" < revert.log
 
-        # use journalctl wrapper to grep only the logs collected while the test is running
-        while ! check_journalctl_log "Waiting for system reboot" -b; do
-            sleep 1
+        # Detect in the logs when the reboot can been triggered
```
stolowski: This reboot check loop and the check below could go into a helper function, but that's not critical.

tests/main/ubuntu-core-refresh/task.yaml
```diff
@@ -0,0 +1,81 @@
+summary: Check that the ubuntu-core system is rebooted after the core snap is refreshed
+
+details: |
+    This test checks that when invoking a manual refresh/revert for core or core18 snaps,
+    a reboot is triggered and the command would exit after the first phase of the installation
+    reporting "snapd is about to reboot the system"
+
+systems: [ubuntu-core-1*]
```
zyga:
```suggestion
systems: [ubuntu-core-*]
```

tests/main/ubuntu-core-refresh/task.yaml
```diff
@@ -0,0 +1,81 @@
...
+        snap list | awk "/^${TARGET_SNAP_NAME} / {print(\$3)}" > initialRev    
+
+        # use journalctl wrapper to grep only the logs collected while the test is running
+        if get_journalctl_log | MATCH "Waiting for system reboot"; then
+            echo "Already waiting for system reboot, exiting..."
+            exit 1
+        fi
+
+        # install new target snap
+        snap download "$TARGET_SNAP_NAME"
```
zyga: Can we just install the same revision we have? What's the point of installing a snap we downloaded from the network here?

zyga: It won't be cached if core/core18 is published at around the moment this runs. I think some tests already install the locally installed version of a snap this way so we could try to reuse that code.

I'm also worried about snapd figuring out that a snap provided as a file actually matches a known assertion that would give it a real revision.

PR #7121, opened 2019-07-17
tests: part2 making tests work on ubuntu-core-18

sergiocazzolato:
New models for core18 and 8 tests running on core-18.


tests/main/remodel/task.yaml
```diff
@@ -10,14 +10,25 @@ prepare: |
...
     rm -rf /var/lib/snapd/assertions/*
     rm -rf /var/lib/snapd/device
     rm -rf /var/lib/snapd/state.json
     mv /var/lib/snapd/seed/assertions/model model.bak
     cp "$TESTSLIB"/assertions/developer1.account /var/lib/snapd/seed/assertions
-    cp "$TESTSLIB"/assertions/developer1.account-key /var/lib/snapd/seed/assertions
-    cp "$TESTSLIB"/assertions/developer1-pc.model /var/lib/snapd/seed/assertions
+    cp "$TESTSLIB"/assertions/developer1.account-key /var/lib/snapd/seed/assertions    
+    if is_core18_system; then
+        cp "$TESTSLIB"/assertions/developer1-pc-18.model /var/lib/snapd/seed/assertions/developer1-pc.model
```
zyga: Is the file name significant? Can we just copy all the assertions over?

tests/main/remodel/task.yaml
```diff
@@ -61,35 +72,42 @@ execute: |
         exit
     fi
 
+    #shellcheck source=tests/lib/systems.sh
+    . "$TESTSLIB"/systems.sh
+    SNAP=test-snapd-tools
```
zyga: You could set SNAP to `test-snapd-tools-coreXX` to make this test "automatically" start doing stuff on core20. Not a blocker.

tests/main/remodel/task.yaml
sergiocazzolato: In this case the name is important because then we call to remodel using the model file name.

tests/main/remodel/task.yaml
sergiocazzolato: @zyga In fact, it is a really good idea, we could call systems and this give as the sufix depending on the core system. I'll create a following PR with this change which will impact on many tests. 
Thanks for the suggestion.

PR #7136, opened 2019-07-22
tests: part4 making tests work on ubuntu-core-18

sergiocazzolato:
This change includes a set of tests running on core18.


tests/main/remove-errors/task.yaml
```diff
@@ -1,18 +1,25 @@
...
     . "$TESTSLIB/names.sh"
+
+    BASE_SNAP=core
+    TARGET_SNAP=test-snapd-tools
+    if is_core18_system; then
+        BASE_SNAP=core18
+        TARGET_SNAP=test-snapd-tools-core18
+    fi
+
+    echo "Given a core snap is installed"
```
zyga:
```suggestion
    echo "Given a base snap, $BASE_SNAP, is installed"
```

tests/main/snap-auto-import-asserts-spools/task.yaml
```diff
@@ -17,7 +17,7 @@ prepare: |
     #shellcheck source=tests/lib/ramdisk.sh
     . "$TESTSLIB/ramdisk.sh"
     setup_ramdisk
-    mkfs.vfat /dev/ram0
+    mkfs.ext3 /dev/ram0
```
zyga: What is the motivation behind this change?

tests/main/snap-auto-mount/task.yaml
```diff
@@ -20,7 +23,13 @@ prepare: |
     #shellcheck source=tests/lib/ramdisk.sh
     . "$TESTSLIB/ramdisk.sh"
     setup_ramdisk
-    mkfs.vfat /dev/ram0
+
+    #We use different filesystems to cover both: fat and ext
+    if is_core18_system; then
+        mkfs.ext3 /dev/ram0
```
zyga: Can we use the same filesystem or is this a part of the test? Where is this coming from?

tests/main/snap-core-fixup/task.yaml
```diff
@@ -25,7 +25,7 @@ execute: |
     umount /boot/uboot
 
     echo "Now test with the real corrupted image"
-    unxz -c -d test.img.xz >test.img
+    tar zxvf test.img.tar.gz
```
zyga: What is the motivation behind this change? I would prefer to see this separately, with a comment explaining that, not bundled in a bigger "core18" change.

sergiocazzolato: @zyga this is because we don't have unxz on core18, so the alternative I found to make that test work on core18 is to use tar.

tests/main/snap-auto-mount/task.yaml
sergiocazzolato: @zyga Yes we can, but I discussed that with @mvo5 and the idea is to use both filesystems (fat is the most popular fs for this and ext3 because fat is not supported on core18 mkfs command).

tests/main/snap-auto-import-asserts-spools/task.yaml
sergiocazzolato: explaned below :)

tests/main/remove-errors/task.yaml
sergiocazzolato: done! thanks for the review!!!

tests/main/snap-auto-mount/task.yaml
sergiocazzolato: I'll add the explanation in the test

PR #7138, opened 2019-07-22
tests: remove lxd / lxcfs if pre-installed

zyga:
The new mount namespace test has data sets for google cloud images and
qemu images. Apart from some cgroup differences the main difference is
presence and absence of pre-installed LXD, including the lxcfs
filesystem.

Because of how our tests prepare and restore (in a non-obvious way) and
because chasing this now is ineffective, to ensure the mount namespace
looks the same across test execution, remove lxd if it is installed in
the initial image.

Tests that require lxd install and remove it already.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -529,6 +529,11 @@ prepare: |
         fi
     fi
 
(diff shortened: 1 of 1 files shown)


zyga:
@bboozzoo LXD is now a snap and handles this automatically. This is a remnant of the Debian package AFAIK.

PR #7141, opened 2019-07-22
tests: add mountinfo-tool --ref-x1000

zyga:
When looking at mount tables from the host and the per-snap or per-user
mount namespace it is nice to be easily able to distinguish numeric
identifiers as belonging to one or the other. In addition, if a new
mount entry is added or removed in one namespace it has cascading
effects on renumbered entries in derivative mount namespaces.

To combat both of those problems add the --ref-x1000 option. When used,
each time --ref is used to get deterministic renumbering and comparison
across mount tables belonging to different mount namespaces, the numeric
value of the mount_id, parent_id and peer group number is using ranges
0-999 for the host mount namespace, 1000-1999 for the per-snap mount
namespace and 2000-2999 for the per-user mount namespace.

This has nice consequence of really easy visual analysis of the mount
table propagation settings, since the peer group clearly belongs to a
given set.

### DIFF
--- a/tests/lib/bin/mountinfo-tool
+++ b/tests/lib/bin/mountinfo-tool
@@ -352,8 +352,8 @@ def renumber_snap_revision(entry, seen):
     entry.mount_point = "/".join(compose(parts, n))
 
 
-def renumber_opt_fields(entry, seen):
-    # type: (MountInfoEntry, Dict[int, int]) -> None
+def renumber_opt_fields(entry, seen, base_n):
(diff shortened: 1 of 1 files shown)


zyga:
@anonymouse64 if we ever hit 1000 mount namespace elements then we need to reconsider what we are doing. For that reason I think YAGNI applies for making this configurable. The testing process you described is correct. Addition of this option was triggered directly by the huge impact of the removal of `lxcfs` which made me worry about the long term maintenance cost of the mount namespace test. With this feature is should be much easier to maintain.

PR #7144, opened 2019-07-23
tests: measure mount namespaces on Ubuntu 14.04

zyga:
This patch contains additional data set for Ubuntu 14.04.

The 14.04 setup is a bit unusual as it is the only supported
distribution that boots without systemd as pid 1. The delegate systemd
doesn't set mount event sharing in the initial mount namespace like
systemd normally would. This is compensated for, for the /snap
hierarchy, by snap-confine.


zyga:
The test failed because something leaked a core revision: we have two core revisions mounted while the text expect one.

PR #7149, opened 2019-07-24
cmd: add snap model command; daemon: add /v2/model, /v2/model/serial REST APIs

anonymouse64:
* Add REST APIs for getting current model and serial assertions same as
  the assertions endpoint, but only returning the current model and 
  serial assertions
* Add snapd client methods for calling the REST API endpoints at 
  /v2/model and /v2/model/serial.
* Add snap model command to use the new snapd client methods and return
  either the raw assertion with --assertion or yamlish output of some of
  the fields from the assertion

TODO:

- [x] Write unit tests
- [x] Write spread tests
- [ ] Test with an actual remodel (how do I test this?)
- [x] Cleanup help message
- [x] Order the fields from `--verbose`?


client/model.go
```diff
@@ -43,3 +47,38 @@ func (client *Client) Remodel(b []byte) (changeID string, err error) {
...
+	response, err := client.raw("GET", path, q, nil, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to query current model assertion: %v", err)
+	}
+	defer response.Body.Close()
+	if response.StatusCode != 200 {
+		return nil, parseError(response)
+	}
+
+	dec := asserts.NewDecoder(response.Body)
```
pedronis: we could also ioutil.ReadAll and then use asserts.Decode. No strong preference.

client/model.go
```diff
@@ -43,3 +47,38 @@ func (client *Client) Remodel(b []byte) (changeID string, err error) {
 
 	return client.doAsync("POST", "/v2/model", nil, headers, bytes.NewReader(data))
 }
+
+// CurrentModelAssertion returns the current model assertion
+func (client *Client) CurrentModelAssertion() (asserts.Assertion, error) {
+	return currentAssertion(client, "/v2/model")
+}
```
pedronis: this should return the actual final type  *asserts.Model

client/model.go
```diff
@@ -43,3 +47,38 @@ func (client *Client) Remodel(b []byte) (changeID string, err error) {
...
+
+// CurrentSerialAssertion returns the current serial assertion
+func (client *Client) CurrentSerialAssertion() (asserts.Assertion, error) {
+	return currentAssertion(client, "/v2/model/serial")
+}
```
pedronis: same for *asserts.Serial

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,92 @@
...
+		assertion, err = x.client.CurrentModelAssertion()
+	}
+
+	if err != nil {
+		return err
+	}
+
+	// --assertion means output the raw assertion
+	if x.Assertion {
+		return asserts.NewEncoder(Stdout).Encode(assertion)
```
pedronis: I suspect doing asserts.Encode separely and then print might be slightly cleaner in this case

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	w := tabwriter.NewWriter(Stdout, 2, 2, 1, ' ', 0)
+
+	// --verbose means output all of the fields
+	if x.Verbose {
+		// TODO
+		headers := assertion.Headers()
```
pedronis: we need to suppress some internal assertion headers  (might need a list in asserts itself), like type authority-id format etc

regarding order (excluding those metadata fields) it's usually:

primary key headers  (list of their names is available  as theAssertion.Type().PrimaryKey)
other headers in lexicographical order  (we could have a helper in asserts to deal with that)

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,92 @@
...
+		for header, val := range headers {
+			fmt.Fprintf(w, "%s:\t%s\n", header, val)
```
pedronis: we will need actual yaml serialisation here, for example required-snaps is a list

daemon/api_model.go
```diff
@@ -69,3 +89,96 @@ func postModel(c *Command, r *http.Request, _ *auth.UserState) Response {
...
+		return BadRequest(err.Error())
+	}
+
+	devmgr := c.d.overlord.DeviceManager()
+	if devmgr == nil {
+		return InternalError("accessing device manager failed")
+	}
+
+	model, err := devmgr.Model()
+	if err != nil {
```
pedronis: we should special case state.ErrNoState, that means no model yet

daemon/api_model.go
```diff
@@ -69,3 +89,96 @@ func postModel(c *Command, r *http.Request, _ *auth.UserState) Response {
...
+		return BadRequest(err.Error())
+	}
+
+	devmgr := c.d.overlord.DeviceManager()
+	if devmgr == nil {
+		return InternalError("accessing device manager failed")
+	}
+
+	serial, err := devmgr.Serial()
```
pedronis: same

daemon/api_model.go
```diff
@@ -69,3 +89,96 @@ func postModel(c *Command, r *http.Request, _ *auth.UserState) Response {
...
+
+// getSerial gets the current serial assertion using the DeviceManager
+func getSerial(c *Command, r *http.Request, _ *auth.UserState) Response {
+	jsonResult, headersOnly, _, err := parseHeadersFormatOptionsFromURL(r.URL)
+	if err != nil {
+		return BadRequest(err.Error())
+	}
+
+	devmgr := c.d.overlord.DeviceManager()
+	if devmgr == nil {
```
pedronis: this should happen at all at this point (not sure we do this kind of checks in other places)

daemon/api_model.go
anonymouse64: What exactly should we do in the case of no model yet? Should we loop waiting for a model?

daemon/api_model.go
anonymouse64: Removed the check here in that case

cmd/snap/cmd_model.go
anonymouse64: Done

cmd/snap/cmd_model.go
anonymouse64: Done, I initially just remove:
- `authority-id`
- `format`
- `type`

what other ones should we remove from the output?

cmd/snap/cmd_model.go
anonymouse64: okay, done

client/model.go
anonymouse64: done

client/model.go
anonymouse64: done

client/model.go
anonymouse64: I left it as is, if you want me to change it to use asserts.Decode, let me know

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,134 @@
...
+	// --assertion means output the raw assertion
+	if x.Assertion {
+		_, err = Stdout.Write(asserts.Encode(assertion))
+		return err
+	}
+
+	w := tabwriter.NewWriter(Stdout, 2, 2, 1, ' ', 0)
+
+	primaryHeaderKeys := assertion.Type().PrimaryKey
+	sort.Strings(primaryHeaderKeys)
```
pedronis: this we don't want to do, the primary key headers have a well-defined order

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+
+		sort.Strings(allHeaders)
+		for _, headerName := range allHeaders {
+			header := allHeaderMaps[headerName]
+			if _, ok := primaryHeaderKeyMap[headerName]; ok {
+				continue
+			}
+			switch headerName {
+			case "authority-id", "format", "type":
```
pedronis: let's define for this

```
asserts.MetaHeaders = []string{ "type", "authority-id" ,"revision", "body-length", "sign-key-sha3-384"}
```

but we do want to show "revision" here, because that is relevant, not implicitly clear or noise

daemon/api_model.go
```diff
@@ -21,25 +21,45 @@ package daemon
...
 
-var modelCmd = &Command{
-	Path: "/v2/model",
-	POST: postModel,
-	// TODO: provide GET here too once we decided on the details of the API
-}
+var (
+	serialModelCmd = &Command{
+		Path: "/v2/model/serial",
+		GET:  getSerial,
```
chipaca: this should have a `UserOK: true`, as it's OK for a user to GET this endpoint.

daemon/api_model.go
```diff
@@ -21,25 +21,45 @@ package daemon
...
+	}
+	modelCmd = &Command{
+		Path: "/v2/model",
+		POST: postModel,
+		GET:  getModel,
```
chipaca: and this should have a `UserOK: true` as well

daemon/api_model.go
chipaca: in `cmd/snap` perhaps, or a clear error message. In the daemon, an appropriate 404 with an ad-hoc error kind.

daemon/api_model.go
```diff
@@ -69,3 +89,90 @@ func postModel(c *Command, r *http.Request, _ *auth.UserState) Response {
 	return AsyncResponse(nil, &Meta{Change: chg.ID()})
 
 }
+
+// helper for parsing url options into formatting option vars
+func parseHeadersFormatOptionsFromURL(url *url.URL) (
```
chipaca: this function can be improved significantly by:
* have it take a `url.Values` instead of the `url` directly, making it more obvious it's getting it from the Query and not the whole URL
* returning an ad-hoc struct instead of a bag of values that are hard to track (i.e. and e.g. `(modelOptions, error)`)

Furthermore, it _looks_ like all you're doing is checking `url.Query()["json"]`, for which the function looks like overkill.

daemon/api_model.go
```diff
@@ -69,3 +89,90 @@ func postModel(c *Command, r *http.Request, _ *auth.UserState) Response {
...
+	jsonResult bool,
```
chipaca: please don't format function return values like this. If they don't fit in one line you're doing something wrong :-)

daemon/api_model.go
```diff
@@ -69,3 +89,90 @@ func postModel(c *Command, r *http.Request, _ *auth.UserState) Response {
...
+				fallthrough
+			case "true":
+				jsonResult = true
+			default:
+				err = errors.New(`"json" query parameter when used must be set to "true" or "headers"`)
+				return jsonResult, headersOnly, headers, err
+			}
+			continue
+		}
+		headers[k] = q.Get(k)
```
chipaca: what use are these `headers`? you're not using them.

daemon/api_model.go
```diff
@@ -69,3 +89,90 @@ func postModel(c *Command, r *http.Request, _ *auth.UserState) Response {
...
+// getModel gets the current model assertion using the DeviceManager
+func getModel(c *Command, r *http.Request, _ *auth.UserState) Response {
+	jsonResult, headersOnly, _, err := parseHeadersFormatOptionsFromURL(r.URL)
+	if err != nil {
+		return BadRequest(err.Error())
+	}
+
+	devmgr := c.d.overlord.DeviceManager()
+
+	model, err := devmgr.Model()
```
chipaca: without locking state, this will panic every time.

You need to test this a lot more.

daemon/api_model.go
```diff
@@ -69,3 +89,90 @@ func postModel(c *Command, r *http.Request, _ *auth.UserState) Response {
...
+// getSerial gets the current serial assertion using the DeviceManager
+func getSerial(c *Command, r *http.Request, _ *auth.UserState) Response {
+	jsonResult, headersOnly, _, err := parseHeadersFormatOptionsFromURL(r.URL)
+	if err != nil {
+		return BadRequest(err.Error())
+	}
+
+	devmgr := c.d.overlord.DeviceManager()
+
+	serial, err := devmgr.Serial()
```
chipaca: ditto about locking state

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,134 @@
...
+			default:
+				// handle different types of the header value
```
chipaca: doing this like this produces really nasty output, that is not readable, does not look at how wide things need to be to not wrap in nasty ways, and is not valid yaml.
If you had run it even once you would've noticed immediately. As you haven't, I'm stealing all your cookies for the day.

This is what it currently looks like, once I got the daemon to not panic:
```
~$ snap model --verbose --serial
brand-id:   generic
model:      generic-classic
serial:     c2f87913-59fe-4b50-a3b3-f1806f7187de
device-key: AcbBTQRWhcGAARAA1XsFG+t9RAqOnoc6IubGzilRR/to79ihTU9ukOnOhORKtKyZHo/2lRKYKr13
MoM3dWn1U7tKGHsrjjOUSYK5v4TTbXPt99bygfMGG4mi3kQ2/1x62UVGgfqKHNnN1jtKohAQVHsp
psr9/4+WVyvxRAX80jM5w7lo2I44lNITx9O6d1oFGWrxstPjtQJs5IlYRhnvzFDP6mMHLKZFtofJ
o62j+sgnxlWswTlv6Fn9yh6n6yFiPHc8y/bsNHPnWvG2s+e7fRFPFmN17ZEYiDB/1ob/IbWCk+yu
Tr8eM+SvvkL8I3fRKJgPHtEJ4e3RRCHO4s6FtSxn4x1Ny//tU0l2s71ykeQyi5nW/kw3Qr8ALggG
l9Bqt2lsOaFh5jNg1bF/O7ZaCD4SvNSejcXkT+kMhhq7N95JT4N1Wg67CBqtAi0Uvqf4F5U2hFmJ
aaZ/twgfuK+irMCkjxTXBTbrHhJSoQdNPDR2aX5eJdip0dW+tpARwKR+swvOXybuNftnSGm/SwE6
yVBs4BU8++Wswgg1cR2UC9mogeqLmsL0O49CsW03n+KUi2nmzKXCS79UAtO43O0cIZBZp0f6etQ6
v2cstT9KybcRKb+4czzEFNxtLuIwhYgNyritg2A5Ols3mTobIzxzH+XiO1jT86QcoceYIMNt61tp
xBWjW1jBM3F4tz0AEQEAAQ==
device-key-sha3-384: Qxgo3bZNrgZoFsZf2R4bGAGm_1hlRlU5vL3VxWfUdygbc7FaZuW4is9MpEpRO2sk
sign-key-sha3-384:   wrfougkz3Huq2T_KklfnufCC0HzG7bJ9wP99GV0FF-D3QH3eJtuSRlQc2JhrAoh1
timestamp:           2018-04-14T13:58:20.487877Z
~$ 
```

particularly bad is that `device-key` behemoth in the middle of the thing, which breaks _everything_. Even the raw `--assertion` output gets this one better.

`cmd_info.go` has a bunch of helpers you might use to make things simpler. Reach out to me if they're too obscure (they might well be).

cmd/snap/cmd_model.go
anonymouse64: Fixed

cmd/snap/cmd_model.go
anonymouse64: Okay, done

client/model.go
```diff
@@ -43,3 +47,55 @@ func (client *Client) Remodel(b []byte) (changeID string, err error) {
...
+	assert, err := currentAssertion(client, "/v2/model")
+	if err != nil {
+		return nil, err
+	}
+	modelAssert, ok := assert.(*asserts.Model)
+	if !ok {
+		return nil, fmt.Errorf("invalid assertion type (%T) returned", assert)
+	}
+	return modelAssert, nil
+
```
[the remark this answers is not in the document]
anonymouse64: Removed

client/model.go
```diff
@@ -43,3 +47,55 @@ func (client *Client) Remodel(b []byte) (changeID string, err error) {
...
+	if response.StatusCode != 200 {
+		return nil, parseError(response)
+	}
+
+	dec := asserts.NewDecoder(response.Body)
+
+	// only decode a single assertion - we can't ever get more than a single
+	// assertion through these endpoints by design
+	assert, err := dec.Decode()
+	if err == io.EOF || err != nil {
```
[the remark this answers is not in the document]
anonymouse64: Yes, I adapted this from code that was decoding multiple assertions and as such handling io.EOF specially, but it's not needed here, I will just check `err != nil`.

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,134 @@
...
+		_, err = Stdout.Write(asserts.Encode(assertion))
+		return err
+	}
+
+	w := tabwriter.NewWriter(Stdout, 2, 2, 1, ' ', 0)
+
+	primaryHeaderKeys := assertion.Type().PrimaryKey
+	sort.Strings(primaryHeaderKeys)
+
+	primaryHeaderKeyMap := make(map[string]bool)
```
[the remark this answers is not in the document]
anonymouse64: This map ended up being unnecessary in my refactoring to address Samuele's comments

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,134 @@
...
+	for _, header := range primaryHeaderKeys {
+		primaryHeaderKeyMap[header] = true
+	}
+
+	for _, header := range primaryHeaderKeys {
```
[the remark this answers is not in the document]
anonymouse64: The previous loop ended up being unnecessary in my refactoring, but for clarity, I originally split it up because I was simply trying to create a simple map to test for existence of a header in the set of primary keys and didn't think about the fact that I already had a loop over the same list right below it.

daemon/api_model.go
anonymouse64: Yes, in 2.39 where I was testing this, it was locking inside Model, but that is not the case on master anymore.

daemon/api_model.go
anonymouse64: I thought that it wasn't okay for a user to do a POST on the /v2/model endpoint, as that would allow any user to remodel the device which AFAICT is undesirable.

daemon/api_model.go
anonymouse64: Done

cmd/snap/cmd_model.go
anonymouse64: I knew this was producing nasty output particularly around the device-key, but I tried looking through some of the helpers in cmd_info.go but it wasn't really clear to me how to use any of them without reading through all 748 lines of cmd_info.go, they all seem pretty specific to the machinery of snap info unfortunately...
I pushed this PR in the hope that someone like you would see this in my PR and go "don't be silly and try to do all this yaml formatting code yourself, just use XYZ yamlish helper package that does the right thing everywhere all the time:tm:", but it seems that is not the case :-|

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,134 @@
...
+		i := 0
+		for k := range allHeaderMaps {
+			allHeaders[i] = k
+			i++
+		}
+
+		sort.Strings(allHeaders)
+		for _, headerName := range allHeaders {
+			header := allHeaderMaps[headerName]
+			if _, ok := primaryHeaderKeyMap[headerName]; ok {
```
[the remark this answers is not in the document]
anonymouse64: As mentioned above this ended up being unnecessary in my refactoring

daemon/api_model.go
anonymouse64: This was modified from the api_asserts.go formatting option parsing, I was going to also implement using the helper there too where things like `headers` is used, but seem to have forgotten about using it there too. I have refactored as per your suggestions

daemon/api_model.go
anonymouse64: Fixed

daemon/api_model.go
anonymouse64: They are used in the /v2/assertions endpoint which I have refactored to use this helper

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,134 @@
...
+import (
+	"fmt"
+	"sort"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/jessevdk/go-flags"
+	"github.com/snapcore/snapd/asserts"
+	"github.com/snapcore/snapd/i18n"
+	yaml "gopkg.in/yaml.v2"
```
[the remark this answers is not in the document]
anonymouse64: Fixed

daemon/api_model.go
chipaca: UserOK etc only apply to GET; non-GET is always auth'ed

cmd/snap/cmd_model.go
anonymouse64: Alright, I think the formatting code is good, let me know if what I have now is okay

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,134 @@
...
+		return err
+	}
+
+	// --assertion means output the raw assertion
+	if x.Assertion {
+		_, err = Stdout.Write(asserts.Encode(assertion))
+		return err
+	}
+
+	w := tabwriter.NewWriter(Stdout, 2, 2, 1, ' ', 0)
```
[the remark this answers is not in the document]
anonymouse64: Done

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"github.com/jessevdk/go-flags"
+	"github.com/snapcore/snapd/asserts"
+	"github.com/snapcore/snapd/i18n"
+	yaml "gopkg.in/yaml.v2"
+)
+
+var (
+	shortModelHelp = i18n.G("Get the active model for this device")
+	longModelHelp  = i18n.G(`TODO`)
```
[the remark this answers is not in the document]
anonymouse64: Well I tried :man_shrugging:

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,134 @@
...
+	Assertion bool `long:"assertion"`
+}
+
+func init() {
+	cmd := addCommand("model",
+		shortModelHelp,
+		longModelHelp,
+		func() flags.Commander {
+			return &cmdModel{}
+		}, nil, []argDesc{},
```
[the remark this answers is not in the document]
anonymouse64: Okay, I think this is for the option help but it doesn't seem to do anything... What is the purpose of this snippet of code? I added it with the options for the --assertion, --verbose, and --serial options which seems correct

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,134 @@
...
+		assertion, err = x.client.CurrentSerialAssertion()
+	} else {
+		assertion, err = x.client.CurrentModelAssertion()
+	}
+
+	if err != nil {
+		return err
+	}
+
+	// --assertion means output the raw assertion
```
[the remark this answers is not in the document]
anonymouse64: IMHO help messages aren't replacements for code documentation

daemon/api_model.go
anonymouse64: Okay, thanks

cmd/snap/cmd_model.go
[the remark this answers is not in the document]
chipaca: agreed, sorry for implying otherwise

cmd/snap/cmd_model.go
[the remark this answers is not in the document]
chipaca: drop the `hidden` (we don't want this hidden) and you'll see it in `--help`

daemon/api_model.go
anonymouse64: Added for both

cmd/snap/cmd_model.go
chipaca: I'll take a detailed look in my morning :-)

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,187 @@
...
+			// for a device which has remodeled
+			if headerName != "revision" && stringInList(headerName, asserts.MetaHeaders) {
+				continue
+			}
+			if stringInList(headerName, primaryHeaderKeys) {
+				continue
+			}
+
+			switch v := header.(type) {
+			case int, uint, uint64, int64, float32, float64, string, bool:
```
pedronis: note: assertions use strings as their only scalar type.  See for example

https://github.com/snapcore/snapd/blob/master/asserts/headers.go#L274

asserts/asserts.go
```diff
@@ -37,6 +37,10 @@ const (
 	noAuthority typeFlags = iota + 1
 )
 
+// MetaHeaders is a list of headers in serial/model assertions which are about
```
pedronis: these are really all assertions

asserts/asserts.go
```diff
@@ -37,6 +37,10 @@ const (
...
+// the assertion itself
```
pedronis: missing period

asserts/asserts.go
```diff
@@ -37,6 +37,10 @@ const (
...
+var MetaHeaders = []string{"type", "authority-id", "revision", "body-length", "sign-key-sha3-384"}
```
pedronis: missing "format" after "type"

asserts/asserts.go
anonymouse64: Done

asserts/asserts.go
anonymouse64: Added

asserts/asserts.go
anonymouse64: Added

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,186 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	return w.Flush()
+}
+
+func stringInList(s string, list []string) bool {
```
chipaca: this is `strutil.ListContains`

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,186 @@
...
+	for _, elem := range list {
+		if elem == s {
+			return true
+		}
+	}
+	return false
+}
+
+func maybeWrapSingleYAMLKey(w io.Writer, key, val string, termWidth int) error {
```
chipaca: This has the problem of wrapping over-aggressively when things are already multi-line, as well as getting things wrong by counting `\t` as one character when it could be multiple.

I think with a little bit of knowledge of what exactly it's printing things can be made a lot better and simpler, and I don't think that knowledge is a bad thing. As this is ad-hoc for models and serials we can use this specificity. IMHO; maybe @pedronis knows more.

pedronis:
On Fri, Jul 26, 2019, 20:42 John R. Lenton <notifications@github.com> wrote:

> *@chipaca* commented on this pull request.
>
> The output is a lot better, now, but still needs work.
>
> The output for snap model, snap model --verbose, and snap model --serial
> is fine as is.
> The output of snap model --serial -verbose needs more work, IMHO.
>
> As @pedronis <https://github.com/pedronis> pointed out the type switch is
> unnecessary because the things are always strings, I think we should use
> that.
>

That's not what I said :) I said all scalar values are strings, we do have
potentially maps, list of maps, list of strings, "required-snaps" for
example is a list of strings.

Notice that the code I pointed to has a type switch, is just that the only
scalar type in it is string.

anonymouse64:
I added a new helper function which should make all of the output YAML compliant. The one thing I'm not clear on though after doing all of that is whether we should be stripping out the newlines from the device-key or not... It may look slightly cleaner but we are modifying the string and so it's not trivial (i.e. using direct comparison of the two strings) to compare the device-key we get from this endpoint if we re-wrapped device-key and the device-key we get from the assertions endpoint. In this PR, I actually opted for just entirely deleting all of the newlines in the serial assertion as that would make it the easiest to compare. It then looks like this:

```
 ./snap-exe model --serial --verbose
brand-id:    generic
model:       generic-classic
serial:      11111111-1111-1111-1111-111111111111
device-key:  "AcbBTQRWhcGAARAAqc9rys4ySRvoVs2E40yfjaDNiaXtsmazGQU9QMZ24DSGznu8kUBOxdXMzGoXJnD0A/abcde\
            4blfIvtpkYx9DTb4wTYYyN7Ex79VF2d8hlUtS1JvdS3dpp2CwzrJmvF1ii0DJj+0y7p0X4uxtzlQR/abcdefghi+\
            awZi6Z1l0HvIwDe3mP0u4pp2cYC0Y2qLGVh0JZBp5a6k1eM62hOiFzHOLEFL77bsLtPRgU+iiqGMyDffZsHrMJyv\
            MeHYHnoUircnNDeNAOLdv7UjmWPydUj5LJQNxuWK1HNFcnFhpnFv/NPqp+cJATThO/1wJC/zLS5qSV3B3uIRSglu\
            AOOsV+GwXiLyMgskcq/UVZU7KOgMiupopu+DYiXFIoKv2MJPt+UmNJbNCYJA92rm4YdyGXGubeUAK9pZITPS17Jo\
            R93oqYninuidQN6CFdTM5vLhMbRxmhmWzC3NmgX6b35yL0TZ4GFx/nLJ2tmly21EsWmmLxKc6kka5crgpbZOcR1G\
            Gs7AubgKVs0HlTe9UVLfygwZ6D2ZFQt/B7tgKvyAg+Y4AGL+8NEvrevgz42A4p8khIIw+slTwfTK/p46Bk/vfUPk\
            PmnW4aBGd7xFsTaSM9O1MdIwgR0TGPxQC2tEAh9kkmZ7Y652RfQ+gDXcCAP153tPkYoFl9Z6dL9lgUPUglOkAEQE\
            AAQ=="
device-key-sha3-384:  4hGV7gBuZPZ6F7YKmyRPRn_TM0whc2fLmXWO3FKTKP-JU-yI54tzCsmnWl3frnXM
timestamp:            2019-07-05T00:15:40.868474Z
```

and with a smaller terminal width:

```
$ ./snap-exe model --serial --verbose
brand-id:    generic
model:       generic-classic
serial:      11111111-1111-1111-1111-111111111111
device-key:  "AcbBTQRWhcGAARAAqc9rys4ySRvoVs2E40yfjaDNiaXtsmazGQU9QMZ24DSGznu8kUBOxdX\
            MzGoXJnD0A/DGddN4blfIvtpkYx9DTb4wTYYyN7Ex79VF2d8hlUtS1JvdS3dpp2CwzrJmvF1\
            ii0DJj+0y7p0X4uxtzlQR/yaxRXUwk7+awZi6Z1l0HvIwDe3mP0u4pp2cYC0Y2qLGVh0JZBp\
            5a6k1eM62hOiFzHOLEFL77bsLtPRgU+iiqGMyDffZsHrMJyvMeHYHnoUircnNDeNAOLdv7Uj\
            mWPydUj5LJQNxuWK1HNFcnFhpnFv/abcd+abcdefgh/1wJC/zLS5qSV3B3uIRSgluAOsV+Gw\
            XiLyMgskcq/UVZU7KOgMiupopu+DYiXFIoKv2MJPt+UmNJbNCYJA92rm4YdyGXGubeUAK9pZ\
            ITPS17JoR93oqYninuidQN6CFdTM5vLhMbRxmhmWzC3NmgX6b35yL0TZ4GFx/nLJ2tmly21E\
            sWmmLxKc6kka5crgpbZOcR1GGs7Aubgabc0HlTe9UVLfygwZ6D2ZFQt/B7tgKvyAg+abcde+\
            8NEvrevgz42A4p8khIIw+slTwfTK/abcde/vfUPkPmnW4aBGd7xFsTaSM9O1MdIwgR0TGPxQ\
            C2tEAh9kkmZ7Y652RfQ+gDXcCAP153tPkYoFl9Z6dL9lgUPUglOkAEQEAAQ=="
device-key-sha3-384:  "4hGV7gBuZPZ6F7YKmyRPRn_TM0whc2fLmXWO3FKTKP-JU-yI54tzCsmnWl3frn\
                     XM"
timestamp:  2019-07-05T00:15:40.868474Z

```

Additionally, is there any kind of specification for where snapd's "yamlish" deviates from proper yaml?

pedronis:
> Additionally, is there any kind of specification for where snapd's "yamlish" deviates from proper yaml?

We don't do consistent quoting I think, which means parsing it requires being type driven. I think in some cases the output cannot be parsed as YAML?  @chipaca might know more on this

chipaca:
> Additionally, is there any kind of specification for where snapd's "yamlish" deviates from proper yaml?

we call it yamlish because we're not promising it's yaml, but we try to make it so. But, it's intended mostly for human consumption with "it's yaml so you can parse it" being a secondary objective (which so far AFAIK we've achieved). On a value-by-value basis we shouldn't be afraid to munge the output a little bit to make it meet both objectives (i.e. be nice to humans, and be valid yaml).

What this means, for this case (and perhaps we need to revisit info as well but that can come later) is that:
* `device-key`, as it's a base64 string and base64 ignores whitespace, we should reflow it so it fits. It should also be last, if present, as when there's a single multi-line block having it last reads a lot better.
* `device-key-sha3-384` should be printed just before `device-key`, and it should break the tabs (i.e. call a `Flush` on the tabwriter) to not force everything else to align with it. Then, you could do a bit of fiddly work:
  1. if the terminal width is > 86, print it as normal:
    ```
             ₁         ₂         ₃         ₄         ₅         ₆         ₇         ₈         ₉
    ¹²³⁴⁵⁶⁷⁸⁹⁰¹²³⁴⁵⁶⁷⁸⁹⁰¹²³⁴⁵⁶⁷⁸⁹⁰¹²³⁴⁵⁶⁷⁸⁹⁰¹²³⁴⁵⁶⁷⁸⁹⁰¹²³⁴⁵⁶⁷⁸⁹⁰¹²³⁴⁵⁶⁷⁸⁹⁰¹²³⁴⁵⁶⁷⁸⁹⁰¹²³⁴⁵⁶⁷⁸⁹⁰
    device-key-sha3-384:  Qxgo3bZNrgZoFsZf2R4bGAGm_1hlRlU5vL3VxWfUdygbc7FaZuW4is9MpEpRO2sk
    ```
  2. if the terminal width is <= 86 but > 66, print it using a block:
    ```
             ₁         ₂         ₃         ₄         ₅         ₆         ₇         ₈
    ¹²³⁴⁵⁶⁷⁸⁹⁰¹²³⁴⁵⁶⁷⁸⁹⁰¹²³⁴⁵⁶⁷⁸⁹⁰¹²³⁴⁵⁶⁷⁸⁹⁰¹²³⁴⁵⁶⁷⁸⁹⁰¹²³⁴⁵⁶⁷⁸⁹⁰¹²³⁴⁵⁶⁷⁸⁹⁰¹²³⁴⁵⁶⁷⁸⁹⁰
    device-key-sha3-384: |
      Qxgo3bZNrgZoFsZf2R4bGAGm_1hlRlU5vL3VxWfUdygbc7FaZuW4is9MpEpRO2sk
    ```
  3. if the terminal width is <= 66, either give up and do (i) because 66 really is smaller than we actually support, or, quote-and-backslash. 
* `serial` could get printed with no wrapping helper: it *should* fit in any reasonable terminal (it needs 47 columns, or 48 in `--verbose` mode because of `timestamp`), especially as we're not forcing it to align with `device-key-sha3-384`. But, you *could* do the same (i)/(ii)/(iii) from above. For this first iteration I'd leave it simple, until we iterate on the smarter helper a bit more.

---

One thing I hadn't noticed is that you're printing `timestamp` in abs-time by default. You should be using `timeMixin` and its `fmtTime` helper for that.

anonymouse64:
Alright I've refactored this to take the manual approach of hard-coding the headers we currently support/format nicely. It now looks something like this:

```
$ SNAP_REEXEC=0 ./snap-exe model --verbose --abs-time
series:          16
brand-id:        memememememememe
model:           test-build
architecture:    amd64
base:            core18
gadget:          pc=18
kernel:          pc-kernel=18
timestamp:       45 days ago, at 04:40 UTC
required-snaps:  
  - core
  - jq
  - hello-world
  - edgexfoundry-device-thingy52-ijohnson
$ SNAP_REEXEC=0 ./snap-exe model --serial --verbose
brand-id:   generic
model:      generic-classic
serial:     c1163ba1-aef3-43c4-8635-e6976ca718d6
timestamp:  36 days ago, at 19:15 CDT
device-key-sha3-384: 4hGV7gBuZPZ6F7YKmyRPRn_thewho2fLmXWO3FKTKP-JU-yI54tzCsmnWl3frnXM
device-key: |
  AcbBTQRWhcGAARAAqc9rys4ySRvoVs2E40yfjaDNiaXtsmazGQU9QMZ24DSGznu8kUBOxdXMzGoXJnD0A/DGddN4blfIthewho
  9DTb4wTYYyN7Ex79VF2d8hlUtS1JvdS3dpp2CwzrJmvF1ii0DJj+0y7p0X4uxtzlQR/yaxRXUwk7+awZi6Z1l0HvIwDethewho
  pp2cYC0Y2qLGVh0JZBp5a6k1eM62hOiFzHOLEFL77bsLtPRgU+iiqGMyDffZsHrMJyvMeHYHnoUircnNDeNAOLdv7Ujmthewho
  5LJQNxuWK1HNFcnFhpnFv/NPqp+cJATThO/1wJC/zLS5qSV3B3uIRSgluAOOsV+GwXiLyMgskcq/UVZU7KOgMiupopu+thewho
  oKv2MJPt+UmNJbNCYJA92rm4YdyGXGubeUAK9pZITPS17JoR93oqYninuidQN6CFdTM5vLhMbRxmhmWzC3NmgX6b35yLthewho
  x/nLJ2tmly21EsWmmLxKc6kka5crgpbZOcR1GGs7AubgKVs0HlTe9UVLfygwZ6D2ZFQt/B7tgKvyAg+Y4AGL+8NEvrevthewho
  p8khIIw+slTwfTK/p46Bk/vfUPkPmnW4aBGd7xFsTaSM9O1MdIwgR0TGPxQC2tEAh9kkmZ7Y652RfQ+gDXcCAP153tPkthewho
  6dL9lgUPUglOkAEQEAAQ==
```
(the serial is from a classic machine and the model is from a core18 VM I built with a custom model using `required-snaps` to check the list formatting)

See the code for the new ordering of the keys, I put required-snaps at the bottom because it's a list of things for reasons similar to device-key being at the bottom.

I will save my generic strict YAML terminal-width respecting formatting code for another PR, I think the right place for that is to implement a `--strict-yaml` option which when set will output strict YAML formatting. 

I have also added support for checking when we don't have a serial assertion yet (i.e. ErrNoState), like this:

```
$ SNAP_REEXEC=0 ./snap-exe model --serial
device not ready - no assertion found
error: assertion not found
$ SNAP_REEXEC=0 ./snap-exe model --serial >/dev/null
error: assertion not found
$ SNAP_REEXEC=0 ./snap-exe model --serial 2>/dev/null
device not ready - no assertion found
```
Note that "device not ready - no assertion found" is on stdout, while error: assertion not found is on stderr and the exit code is 1. Please let me know if this is okay or should be refactored. 

Finally, I force-pushed the branch, even though there are a lot of reviews here because the commit history is getting awfully messy I think and most of the code will probably have to be re-reviewed anyways I think.

anonymouse64:
@degville could you review and/or help me word the help message here for this command? it looks like this currently:

```
$ ./snap-exe model --help
Usage:
  snap-exe model [model-OPTIONS]

The model command returns only the active model assertion information for this
device.

The --serial option allows printing the serial assertion instead of the model
assertion.

The --verbose option prints off more headers from the assertion.

The --assertion option prints off the raw assertion.

[model command options]
      --no-wait      Do not wait for the operation to finish but just print the change id.
      --abs-time     Display absolute times (in RFC 3339 format). Otherwise, display relative times up to 60 days, then YYYY-MM-DD.
      --serial       Print the serial assertion instead of the model.
      --verbose      Print more of the assertion fields.
      --assertion    Print the raw assertion.
```

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,229 @@
...
+
+	"github.com/snapcore/snapd/asserts"
+	"github.com/snapcore/snapd/client"
+	"github.com/snapcore/snapd/i18n"
+)
+
+var (
+	shortModelHelp = i18n.G("Get the active model for this device")
+	longModelHelp  = i18n.G(`
+The model command returns only the active model assertion information for this
```
degville: I'd drop *only* to make it a little clearer (I can see what you mean, but it could imply there's a command to return more than one assertion).

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,229 @@
...
+	"github.com/snapcore/snapd/i18n"
+)
+
+var (
+	shortModelHelp = i18n.G("Get the active model for this device")
+	longModelHelp  = i18n.G(`
+The model command returns only the active model assertion information for this 
+device.	
+
+The --serial option allows printing the serial assertion instead of the model
```
degville: If the text is mostly repeating the same information, we don't normally describe additional arguments outside of the command options section. Maybe change this to a more descriptive paragraph, even listing the headers that are included by default and where any additional headers come from, along the lines of:

By default, only a minimal set of headers are included in the output, but this can be expanded to include all of an assertion's headers. Similarly, the default model assertion can alternatively be replaced with the serial assertion.

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,229 @@
...
+
+func init() {
+	addCommand("model",
+		shortModelHelp,
+		longModelHelp,
+		func() flags.Commander {
+			return &cmdModel{}
+		}, timeDescs.also(waitDescs.also(map[string]string{
+			"assertion": i18n.G("Print the raw assertion."),
+			"verbose":   i18n.G("Print more of the assertion fields."),
```
degville: As above, it would be good if we could qualify *more* with the headers that are additionally returned - is it all the headers, or how have they been selected?

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,229 @@
...
+			"serial":    i18n.G("Print the serial assertion instead of the model."),
```
degville: This is pedantic, but I'd make it `model assertion.` at the end, just to make absolutely sure people don't think the command isn't returning an assertion by default.

client/model.go
```diff
@@ -43,3 +49,61 @@ func (client *Client) Remodel(b []byte) (changeID string, err error) {
...
+
+// CurrentModelAssertion returns the current model assertion
+func (client *Client) CurrentModelAssertion() (*asserts.Model, error) {
+	assert, err := currentAssertion(client, "/v2/model")
+	if err != nil {
+		return nil, err
+	}
+	modelAssert, ok := assert.(*asserts.Model)
+	if !ok {
+		return nil, fmt.Errorf("invalid assertion type (%T) returned", assert)
```
pedronis: we know it's an assertion though, so we can use assert.Type().Name instead of %T, also probably s/invalid/unexpected/

client/model.go
```diff
@@ -43,3 +49,61 @@ func (client *Client) Remodel(b []byte) (changeID string, err error) {
...
+
+// CurrentSerialAssertion returns the current serial assertion
+func (client *Client) CurrentSerialAssertion() (*asserts.Serial, error) {
+	assert, err := currentAssertion(client, "/v2/model/serial")
+	if err != nil {
+		return nil, err
+	}
+	serialAssert, ok := assert.(*asserts.Serial)
+	if !ok {
+		return nil, fmt.Errorf("invalid assertion type (%T) returned", assert)
```
pedronis: same

client/model.go
```diff
@@ -22,13 +22,19 @@ package client
...
+	"net/url"
+
+	"github.com/snapcore/snapd/asserts"
 )
 
 type remodelData struct {
 	NewModel string `json:"new-model"`
 }
 
+const ErrorKindAssertionNotFound = "assertion not found"
```
pedronis: mmh, this is not a ErrorKind as we usually intend them

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,229 @@
...
+func (x *cmdModel) Execute(args []string) error {
+	var assertion asserts.Assertion
+	var err error
+	if x.Serial {
+		assertion, err = x.client.CurrentSerialAssertion()
+	} else {
+		assertion, err = x.client.CurrentModelAssertion()
+	}
+
+	if err != nil && err.Error() == client.ErrorKindAssertionNotFound {
```
pedronis: as per above we need to rework a bit how we pass this error around to fit more with the rest

daemon/api_model.go
pedronis: yes, we want dedicated kind "device-assertion-not-available" I think, see response.go etc about how to go about introducing them

daemon/api_model.go
```diff
@@ -69,3 +90,73 @@ func postModel(c *Command, r *http.Request, _ *auth.UserState) Response {
 	return AsyncResponse(nil, &Meta{Change: chg.ID()})
 
 }
+
+// getModel gets the current model assertion using the DeviceManager
```
pedronis: these needs some unit tests

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,229 @@
...
+			headerValue, ok := allHeadersMap[headerName]
+			// make sure the header is in the map
+			if !ok {
+				continue
+			}
+
+			// switch on which header it is to handle some special cases
+			switch headerName {
+			// scalar values
+			case "architecture", "base", "classic", "display-name", "gadget", "kernel", "revision", "store":
```
pedronis: this could just be the default stanza, instead of listing these again

cmd/snap/cmd_model.go
pedronis: "Print all specific assertion fields" perhaps?

daemon/api_model.go
pedronis: Value would then be set to the assertion type "model"/"serial" possibly

cmd/snap/cmd_model.go
anonymouse64: Okay, removed.

cmd/snap/cmd_model.go
anonymouse64: Added, done.

cmd/snap/cmd_model.go
anonymouse64: That makes sense, I went with @pedronis's suggestion.

cmd/snap/cmd_model.go
anonymouse64: Okay, I took what you provided and slightly reworded it since we aren't printing all assertion headers with --verbose but rather all of the non-meta assertion headers (i.e. no headers that are about the assertion itself like the body length, etc.)

client/model.go
anonymouse64: Done

client/model.go
anonymouse64: Done

cmd/snap/cmd_model.go
anonymouse64: The reason I went with listing all of these here was so that we didn't inadvertently start printing a header key which we haven't vetted to print nicely and so if there was a new key not listed here in a model assertion, it may be printed in a not-nicely formatted way, or worse it may even not be a scalar string in which case the type assertion may fail. As such I think it's better to only support specific keys for now here.

daemon/api_model.go
anonymouse64: Okay I think I have implemented this as requested. There is now an additional errorKind in the daemon/response.go file which is used in the response we send if err is state.ErrNoState. The response we send is also a special crafted errorResult with Value set as requested, message same as what it was, and Kind as the new errorKind in daemon/response.go.

cmd/snap/cmd_model.go
anonymouse64: I think I have reworked this as requested to match other command/client/daemon patterns.

client/model.go
anonymouse64: I think I have reworked this as requested to match other command/client/daemon patterns.

daemon/api_model.go
anonymouse64: I added unit tests for getting the model without a model assertion and with a model assertion, and also that the endpoint returns the new model assertion after a remodel, however the last test doesn't seem to be passing, could you take a look at it and see what I should be doing in the mock to make the new model be taken into account in the device manager?

anonymouse64: I also couldn't find any examples of mocking serial assertions, any pointers for where I could look to mock a serial assertion for testing as well?

cmd/snap/cmd_model.go
pedronis: but we are not iterating over all the headers, we are iterating only over local niceOrdering no?

pedronis:
I let @chipaca do another pass before looking at this again

daemon/api_model.go
pedronis: looks for

```
serial, err := s.brands.Signing("my-brand").Sign(asserts.SerialType, map[string]interface{}{...
```

in overlord/managers_test.go for example

cmd/snap/cmd_model.go
anonymouse64: Actually yes you're right, sorry I was confused

daemon/api_model.go
anonymouse64: Alright, I dropped my test around /v2/model showing the new assertion, that ended up being a lot more complicated than I think is really necessary.
I also added new unit tests around getting a serial assertion so I think the test coverage should be adequate, let me know if you disagree.

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,228 @@
...
+	shortModelHelp = i18n.G("Get the active model for this device")
+	longModelHelp  = i18n.G(`
+The model command returns the active model assertion information for this 
+device.	
+
+By default, only the primary keys of the headers are included in the output, but
+this can be expanded to include all of an assertion's headers non-meta headers.
+
+Similarly, the default model assertion can alternatively be replaced with the 
+serial assertion.
```
chipaca: A nit: we're not replacing one assertion with the other :-) we provide information about one or the other.

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,228 @@
...
+`)
+
+	// this list is a "nice" "human" "readable" "ordering" of headers to print
+	// off, sorted in lexographical order with meta headers and primary key
+	// headers removed, and big nasty keys such as device-key-sha3-384 and
+	// device-key at the bottom
+	// it also contains both serial and model assertion headers, but we
+	// follow the same code path for both assertion types and some of the
+	// headers are shared between the two, so it still works out correctly
```
chipaca: <3

asserts/asserts.go
```diff
@@ -37,6 +37,17 @@ const (
 	noAuthority typeFlags = iota + 1
 )
 
+// MetaHeaders is a list of headers in assertions which are about the assertion
+// itself.
+var MetaHeaders = []string{
```
chipaca: I'd've made this a `[...]string` instead of a `[]string` but maybe I'm being awkward

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,228 @@
...
+	niceOrdering = []string{
```
chipaca: again I would've made this a `[...]string`, but again i suspect that's just me

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,228 @@
...
+		shortModelHelp,
+		longModelHelp,
+		func() flags.Commander {
+			return &cmdModel{}
+		}, timeDescs.also(waitDescs.also(map[string]string{
+			"assertion": i18n.G("Print the raw assertion."),
+			"verbose":   i18n.G("Print all specific assertion fields."),
+			"serial": i18n.G(
+				"Print the serial assertion instead of the model assertion."),
+		})),
```
chipaca: in general this is written with shorter `also`s, i.e. `timeDescs.also(waitDescs).also(...)`, but I guess it doesn't make much difference beyond having to count parens at the end?

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,228 @@
...
+		assertion, err = x.client.CurrentModelAssertion()
+	}
+
+	if err != nil && client.IsAssertionNotFoundError(err) {
+		// device is not registered yet - print helpful message to stdout and
+		// return err to set exit code non-zero
+		fmt.Fprintln(Stdout, "device not ready - no assertion found")
+		return err
+	} else if err != nil {
+		return err
```
chipaca: I think this if block would be easier to read as
```go
if err != nil {
        if client.IsAssertionNotFoundError(err) {
                fmt.Fprintln(...)
        }
        return err
}
```

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,228 @@
...
+			// long base64 key we can rewrap safely
+			case "device-key":
+				headerString, ok := headerValue.(string)
+				if !ok {
+					return errors.New("invalid type for \"device-key\" header")
+				}
+				// the string value here has newlines inserted as part of the
+				// raw assertion, but base64 doesn't care about whitespace, so
+				// split by newlines and re-wrap since base64 doesn't care about
+				// whitespace and we can make it prettier
```
chipaca: little bit of duplication in the language there

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,228 @@
...
+				headerString = strings.Join(strings.Split(headerString, "\n"), "")
+				fmt.Fprintln(w, "device-key: |")
+				wrapLine(w, []rune(headerString), "  ", termWidth)
+
+			// the default is all the rest of short scalar values, which all
+			// should be strings
+			default:
+				headerString, ok := headerValue.(string)
+				if !ok {
+					return fmt.Errorf("invalid type for \"%s\" header", headerName)
```
chipaca: `%q` rather than `\"%s\"` surely?

anonymouse64: ah yes you're right I always forget `%q` exists

cmd/snap/cmd_model.go
anonymouse64: reworded

cmd/snap/cmd_model.go
anonymouse64: ah sure that's easy enough to change

cmd/snap/cmd_model.go
anonymouse64: Okay, changed

cmd/snap/cmd_model.go
anonymouse64: yes that is simpler, thanks for the suggestion

cmd/snap/cmd_model.go
anonymouse64: reworded for clarity and succinctness

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,233 @@
...
+	var err error
+	if x.Serial {
+		assertion, err = x.client.CurrentSerialAssertion()
+	} else {
+		assertion, err = x.client.CurrentModelAssertion()
+	}
+
+	if err != nil {
+		if client.IsAssertionNotFoundError(err) {
+			// device is not registered yet - also print helpful message to
```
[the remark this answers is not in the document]
anonymouse64: I wasn't quite sure what the best thing to do here to indicate ErrNoState, at the minimum an error should be returned to make the exit code non-zero, but should we also output to stdout as well as stderr?

daemon/response.go
```diff
@@ -188,6 +188,8 @@ const (
 
 	errorKindDaemonRestart = errorKind("daemon-restart")
 	errorKindSystemRestart = errorKind("system-restart")
+
+	errorKindAssertionsNotFound = errorKind("assertion-not-found")
```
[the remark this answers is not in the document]
anonymouse64: Fixed

client/client.go
```diff
@@ -489,6 +491,18 @@ func IsInterfacesUnchangedError(err error) bool {
...
+// IsAssertionNotFoundError returns whether the given error means that the
+// assertion wasn't found and thus the device isn't ready/seeded.
+func IsAssertionNotFoundError(err error) bool {
+	e, ok := err.(*Error)
+	if !ok || e == nil {
+		return false
+	}
+
+	return e.Kind == ErrorKindNoAssertion
+
```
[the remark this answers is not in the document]
anonymouse64: Removed

client/client.go
```diff
@@ -456,6 +456,8 @@ const (
 
 	ErrorKindSystemRestart = "system-restart"
 	ErrorKindDaemonRestart = "daemon-restart"
+
+	ErrorKindNoAssertion = "assertion-not-found"
```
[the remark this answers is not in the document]
anonymouse64: I assume you meant `ErrorKindAssertionNotFound` and not `ErrKindAssertionNotFound` since I can't find anywhere else we do `ErrKind`, we always seem to do `ErrorKind`.

Fixed to `ErrorKindAssertionNotFound` anyways

cmd/snap/cmd_model.go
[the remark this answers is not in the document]
anonymouse64: Well I just went with your suggestion, if you want it to something else lmk and I'll refactor again

anonymouse64:
@pedronis I think the only thing missing from this PR are additional tests, is it okay to add those in a followup or do you want the tests in this PR before approving? The only things missing between your review and @chipaca is additional tests.

pedronis:
> @pedronis I think the only thing missing from this PR are additional tests, is it okay to add those in a followup or do you want the tests in this PR before approving? The only things missing between your review and @chipaca is additional tests.

As a general comment we do try to have extensive coverage (though not really 100%), also relatedly while I myself don't practice strict TDD in snapd, generally I think it's a good idea to write happy path and interesting paths tests as early as possible.

Now concretely here, I would say it would be good to add the daemon tests still here because that code is not exercised ATM.  The client and cmd/snap tests can be done in a follow up if preferred.

cmd/snap/cmd_model.go
chipaca: ... not a fan of the switch you used instead of this btw.

You could use the exact same pattern we'd agreed,
```go
if err != nil {
        if client.IsAssertionNotFoundError(err) {
                return fmt.Errorf(...)
        }
        return err
}
```
or even
```go
if err != nil {
        if client.IsAssertionNotFoundError(err) {
                err = fmt.Errorf(...)
        }
        return err
}
```
without it being as jarring as the `switch` is right now

daemon/api_model_test.go
```diff
@@ -251,3 +291,67 @@ func (s *apiSuite) TestGetModelHasSerialAssertion(c *check.C) {
...
+
+	// model assertion setup
+	d := s.daemonWithOverlordMock(c)
+	hookMgr, err := hookstate.Manager(d.overlord.State(), d.overlord.TaskRunner())
+	c.Assert(err, check.IsNil)
+	deviceMgr, err := devicestate.Manager(d.overlord.State(), hookMgr, d.overlord.TaskRunner(), nil)
+	c.Assert(err, check.IsNil)
+	d.overlord.AddManager(deviceMgr)
+	st := d.overlord.State()
+	st.Lock()
```
pedronis: best to do a defer st.Unlock()  as well when locking/unlocking over many checks, otherwise it's easy to get test failures that will panic on lock sanity instead of giving the actual problem

daemon/api_model_test.go
```diff
@@ -251,3 +291,67 @@ func (s *apiSuite) TestGetModelHasSerialAssertion(c *check.C) {
...
+	assertstatetest.AddMany(st, s.storeSigning.StoreAccountKey(""))
+	assertstatetest.AddMany(st, s.brands.AccountsAndKeys("my-brand")...)
+	s.mockModel(c, st, theModel)
+
+	// in case the name of the serial ever changes, just get it state in State
+	// currently it's hard-coded to always be serialserial
+	var authStateData auth.AuthState
+	err = st.Get("auth", &authStateData)
+	c.Assert(err, check.IsNil)
```
pedronis: this could have used d.overlord.DeviceManager().Serial() ?

cmd/snap/cmd_model.go
```diff
@@ -0,0 +1,232 @@
...
+			// switch on which header it is to handle some special cases
+			switch headerName {
+			// list of scalars
+			case "required-snaps":
+				headerIfaceList, ok := headerValue.([]interface{})
+				if !ok {
+					return invalidTypeErr
+				}
+				if len(headerIfaceList) == 0 {
+					fmt.Fprintf(w, "%s:\t[]\n", headerName)
```
pedronis: just omit it completely in this case

daemon/api_model_test.go
pedronis: ah, I'm not sure the puzzlement from this kind of code is worth the win not to have to change the test if we change the global setup (other things might likely go awry anyway)

daemon/api_model_test.go
```diff
@@ -249,28 +250,30 @@ func (s *apiSuite) TestGetModelHasSerialAssertion(c *check.C) {
...
 	c.Assert(err, check.IsNil)
 	assertstatetest.AddMany(st, serial)
+	devicestatetest.SetDevice(st, &auth.DeviceState{
+		Brand:  "my-brand",
+		Model:  "my-old-model",
+		Serial: "serialserial",
+	})
 
 	st.Unlock()
+	defer st.Lock()
```
anonymouse64: a comment around this would be helpful, as it is it reads rather confusingly if you don't remember that a few lines above we defer'd an unlock and also that go defers are LIFO

daemon/api_model_test.go
```diff
@@ -249,28 +250,30 @@ func (s *apiSuite) TestGetModelHasSerialAssertion(c *check.C) {
...
-	// currently it's hard-coded to always be serialserial
-	var authStateData auth.AuthState
-	err = st.Get("auth", &authStateData)
-	c.Assert(err, check.IsNil)
 	serial, err := s.brands.Signing("my-brand").Sign(asserts.SerialType, map[string]interface{}{
 		"authority-id":        "my-brand",
 		"brand-id":            "my-brand",
 		"model":               "my-old-model",
-		"serial":              authStateData.Device.Serial,
+		"serial":              "serialserial",
```
anonymouse64: a comment on the fact that this is hard-coded somewhere else and that's why we need it to be this would be helpful to future readers I think

daemon/api_model_test.go
```diff
@@ -310,39 +313,30 @@ func (s *apiSuite) TestGetModelJSONHasSerialAssertion(c *check.C) {
...
 		"device-key":          string(encDevKey),
 		"device-key-sha3-384": deviceKey.PublicKey().ID(),
 		"timestamp":           time.Now().Format(time.RFC3339),
 	}, nil, "")
 	c.Assert(err, check.IsNil)
 	assertstatetest.AddMany(st, serial)
+	devicestatetest.SetDevice(st, &auth.DeviceState{
```
anonymouse64: ah this is a good function, I didn't see it before, thanks :+1:

daemon/api_model_test.go
pedronis: maybe I should have chosen something else, we are setting it again below here, so it doesn't matter what is hard-coded to somewhere else, the test is self-contained now (at least wrt that)

daemon/api_model_test.go
pedronis: I understand the concern, but we do this kind of thing in many places,  I'm not sure we want a comment in all of them

anonymouse64: okay fair enough, it's just a bit difficult to track the state of a stack of `defer` statements and I haven't personally seen anywhere else where we have these multiple `defer` statements, but I'll take your word for it

pedronis:
This is ready to go in, as we discussed we should look at a follow up to rework the plain "snap model" and "snap model --serial" behavior, their output is very close now which means as a very similar pair they don't add much value. Also we are ignoring/not using the fact that both model and accounts (aka brands) have optionally nice display-names.

PR #7151, opened 2019-07-24
tests: remove local revision of core

zyga:
The snap-userd-reexec test re-installs the current revision of core to
measure some re-execution logic. It does not clean up after itself.
This patch fixes this by reverting core and removing the x1 revision.


chipaca:
Actually, before reverting core maybe you should check x1 is actually the current revision (revert also gets called when things fail, right?)

zyga:
Hmm, indeed, I'll make it smarter.

PR #7158, opened 2019-07-24
tests: part5 making tests work on ubuntu-core-18

sergiocazzolato:


tests/lib/boot.sh
```diff
@@ -11,12 +12,16 @@ bootenv() {
...
+        elif [ -s "$GRUBENV_FILE" ]; then
+            cat "$GRUBENV_FILE"
         else
             fw_printenv
         fi
     else
         if command -v "$GRUB_EDITENV" >/dev/null; then
             "$GRUB_EDITENV" list | grep "^$1"
+        elif [ -s "$GRUBENV_FILE" ]; then
+            grep "^$1" "$GRUBENV_FILE"
```
bboozzoo: Wonder whether it should be `grub "^$1=" "$GRUBENV_FILE"`, but looking how other query options fail to do it this way, I suppose this is fine.

tests/lib/boot.sh
```diff
@@ -11,12 +12,16 @@ bootenv() {
     if [ $# -eq 0 ]; then
         if command -v "$GRUB_EDITENV" >/dev/null; then
             "$GRUB_EDITENV" list
+        elif [ -s "$GRUBENV_FILE" ]; then
+            cat "$GRUBENV_FILE"
```
bboozzoo: Is grub-editenv unavailable in core18? @mvo5 do we need it?

mvo5: Unfortunately it is unavailable on core18 - we don't need it on core18, snapd has code to read/write grubenv. We could add something like "snap debug list-bootenv" or something. But I think its ok to just use `grep -v '^#' /boot/grub/grubenv ` - the format of the grubenv is pretty simple.

PR #7159, opened 2019-07-24
tests: add functions to make an abstraction for the snaps

sergiocazzolato:
This is the first part of the change to avoid deciding which snap to use
based on the core system which is being used.


tests/lib/systems.sh
```diff
@@ -27,3 +34,28 @@ is_ubuntu_14_system(){
     fi
     return 1
 }
+
+get_snap_for_system(){
+    snap=$1
+    system=${2:-$SPREAD_SYSTEM}
+
+    if [[ "$system" == ubuntu-core-18-* ]]; then
```
zyga: Perhaps use a switch statement here, it's a natural fit for this kind of code.

sergiocazzolato: Totally agree, thanks for the review

tests/lib/systems.sh
```diff
@@ -14,16 +14,55 @@ is_core18_system(){
...
+
 is_ubuntu_14_system(){
     if [[ "$SPREAD_SYSTEM" == ubuntu-14.04-* ]]; then
         return 0
     fi
     return 1
 }
+
+get_snap_for_system(){
+    snap=$1
```
zyga: If this is bash then we should be more careful with variables and use `local`. Alternatively, if you can, avoid the variables altogether, when it is natural to do so.

tests/lib/systems.sh
```diff
@@ -14,16 +14,55 @@ is_core18_system(){
...
+
+    case "$system" in
+        ubuntu-core-18-*)
+            echo "core18"
+            ;;
+        ubuntu-core-20-*)
+            echo "core20"
+            ;;
+        *)
+            echo "core"
```
zyga:
```suggestion
            echo "core"
            ;;
```

tests/lib/systems.sh
```diff
@@ -14,16 +14,55 @@ is_core18_system(){
...
+
+    case "$system" in
+        ubuntu-core-18-*)
+            echo "${snap}-core18"
+            ;;
+        ubuntu-core-20-*)
+            echo "${snap}-core20"
+            ;;
+        *)
+            echo "$snap"
```
zyga:
```suggestion
            echo "$snap"
            ;;
```

tests/lib/systems.sh
```diff
@@ -14,16 +14,55 @@ is_core18_system(){
     return 1
 }
 
+is_core20_system(){
+    if [[ "$SPREAD_SYSTEM" == ubuntu-core-20-* ]]; then
```
zyga: I wish there was a way to make this a one-liner without upsetting `set -e`.

tests/lib/systems.sh
sergiocazzolato: done

tests/lib/systems.sh
sergiocazzolato: done

tests/lib/systems.sh
sergiocazzolato: Sure, thanks for reviewing this.

sergiocazzolato:
@zyga could you make another round please?

tests/lib/systems.sh
```diff
@@ -14,16 +14,57 @@ is_core18_system(){
...
 is_ubuntu_14_system(){
     if [[ "$SPREAD_SYSTEM" == ubuntu-14.04-* ]]; then
         return 0
     fi
     return 1
 }
+
+get_snap_for_system(){
+    local snap=$1
+    local system=${2:-$SPREAD_SYSTEM}
```
zyga: I'm kind of -1 on the second argument unless there's a really good reason to have it.

sergiocazzolato: done

sergiocazzolato: done

PR #7163, opened 2019-07-25
tests: measure testbed for leaking mountinfo entries

zyga:
Tests can routinely leak system wide side effects, like mount tables.
This patch adds the `testebed-tool` executable that with two
sub-commands, one for setting baseline state and another for comparing
the current state to the baseline state.

The tool is applied in a very specific spot in the prepare/restore
logic, because that is where it seems to be mostly effective. The
location is chosen because prepare/restore is really mixed up in snapd
test suite and naive approach of setting the baseline after prepare and
comparing it to the current state after restore does not really
function.

### DIFF
--- a/tests/lib/bin/testbed-tool
+++ b/tests/lib/bin/testbed-tool
@@ -1,0 +1,69 @@
+#!/bin/sh -ex
+# This script is meant to monitor the state of the device under test.  It is a
+# convenient place to add detectors for leaky tests. Things that fit into this
+# category are:
+# - leftover mount points (now sporadically captured by tests/main/mount-ns)
+# - leftover processes (especially FUSE things or dbus-daemon)
+# - leftover packages
+# - altered kernel control knobs.
+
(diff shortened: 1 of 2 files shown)


tests/lib/bin/testbed-tool
```diff
@@ -0,0 +1,69 @@
...
+    D="$D_mountinfo"
+    if [ "${1:-}" = "--once" ] && [ -e "$D/baseline.raw.txt" ]; then
+        exit
+    fi
+    mkdir -p "$D"
+    cat /proc/self/mountinfo >"$D"/baseline.raw.txt
+}
+
+postprocess_mountinfo() {
+	# Rewrite mountinfo table using specific ordering to avoid some pitfalls:
```
zyga: Some tab-spaces SNAFU, I'll fix this before merging.

zyga:
Interesting, it found a whole set of issues on 18.04. I'll close this for now and focus on them.

PR #7164, opened 2019-07-25
test: enable and run mount-ns test as last

zyga:
Given that we have some confidence that tests don't leak mount resources
anymore, we can now run the mount namespace measurement test as the last
one in the group. This way if something does misbehave and "cruft"
accumulates in the host mount namespace the test will be able to capture
that. This approach can be contrasted with running the test first, where
it would surely pass but not have the effect of spotting tests that
clobber the environment.


zyga:
Given that there are more deviations detected on 18.04 systemd, I'll close this for now.

PR #7165, opened 2019-07-25
tests: restore cpuset clone_children clobbered by lxd

zyga:
The testbed-tool reported deviation on Ubuntu 18.04 was:

    -21 18 1:17 / /sys/fs/cgroup/cpuset rw,nosuid,nodev,noexec,relatime shared:21 - cgroup cgroup rw,cpuset
    +21 18 1:17 / /sys/fs/cgroup/cpuset rw,nosuid,nodev,noexec,relatime shared:21 - cgroup cgroup rw,cpuset,clone_children

Internally LXD + LXC enable the clone_children property on the cpuset
controller but nothing removes this change on LXD removal. This patch
fixes this.

Corresponding code in LXC:
https://github.com/lxc/lxc/blob/113ca429334073af2a969ded6f143065660c19d3/src/lxc/cgroups/cgfsng.c#L569

### DIFF
--- a/tests/main/lxd/task.yaml
+++ b/tests/main/lxd/task.yaml
@@ -31,6 +31,9 @@ restore: |
     lxd.lxc stop my-ubuntu --force
     lxd.lxc delete my-ubuntu
 
+    # LXD alters the configuration of the cpuset cgroup. Restore the original state.
+    echo 0 > /sys/fs/cgroup/cpuset/cgroup.clone_children
+
 debug: |
     # shellcheck source=tests/lib/journalctl.sh
     . "$TESTSLIB/journalctl.sh"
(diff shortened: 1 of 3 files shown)


tests/main/lxd/task.yaml
```diff
@@ -31,6 +31,9 @@ restore: |
     lxd.lxc stop my-ubuntu --force
     lxd.lxc delete my-ubuntu
 
+    # LXD alters the configuration of the cpuset cgroup. Restore the original state.
+    echo 0 > /sys/fs/cgroup/cpuset/cgroup.clone_children
```
bboozzoo: Should we rather save the current value in prepare and restore it here?

zyga: I was thinking about it but it's off by default so YAGNI. The monitoring tool will tell us if this is wrong.

PR #7168, opened 2019-07-25
tests: measure testbed for leaking mountinfo entries

zyga:
Tests can routinely leak system wide side effects, like mount tables.
This patch adds the `testbed-tool` executable that with two
sub-commands, one for setting baseline state and another for comparing
the current state to the baseline state.

The tool is applied in a very specific spot in the prepare/restore
logic, because that is where it seems to be mostly effective. The
location is chosen because prepare/restore is really mixed up in snapd
test suite and naive approach of setting the baseline after prepare and
comparing it to the current state after restore does not really
function.

### DIFF
--- a/tests/lib/bin/testbed-tool
+++ b/tests/lib/bin/testbed-tool
@@ -1,0 +1,138 @@
+#!/bin/sh -e
+# This script is meant to monitor the state of the device under test.  It is a
+# convenient place to add detectors for leaky tests. Things that fit into this
+# category are:
+# - leftover mount points (now sporadically captured by tests/main/mount-ns)
+# - leftover processes (especially FUSE things or dbus-daemon)
+# - leftover packages
+# - altered kernel control knobs.
+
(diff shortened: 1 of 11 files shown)


tests/lib/bin/testbed-tool
```diff
@@ -0,0 +1,69 @@
...
+    fi
+}
+
+case "${1:-}" in
+    set-baseline)
+        shift
+        measure_mountinfo "$@"
+        ;;
+    compare)
+        compare_mountinfo "$@"
```
zyga:
```suggestion
        shift
        compare_mountinfo "$@"
```

zyga:
I debugged the `clone_children` part of the failure as https://github.com/snapcore/snapd/pull/7435

zyga:
I will close this for now. The invarian-tool (that still needs to be renamed) is a better way to structure this. I will try to revive the idea later this week.

PR #7169, opened 2019-07-25
tests: remove core18 if installed by tests

zyga:
When tests run on ubuntu core, the automatic cleanup performed is
different and imperfect. In particular it is very easy to "leak"
additional base snaps, like core18.

Before we have a better mechanism available, remove core18 while
restoring a test case that did not have core18 before the test executed.


zyga:
Uh, I forgot those are not removable. Hmmm

zyga:
This is almost ready, my local tree is now clean on `ubuntu-core-16`. As soon as all the prerequisites have landed I will rebase and reopen this.

PR #7171, opened 2019-07-25
tests: improve how the system is restored when the upgrade-from-2.15 test fails

sergiocazzolato:
This fix is to make sure when the upgrade-from-2.15 test fails the
systems restores snapd and its dependencies properly.

To reproduce the error it is needed to add an "exit 1" at the begining
of the execution part, and the after this test failure all the following
tests will fail as well.

This fix was tested making the test fail before the old snapd is
installed and after that as well.

See this log as an example of the issue: https://api.travis-ci.org/v3/job/563445228/log.txt

### DIFF
--- a/tests/main/upgrade-from-2.15/task.yaml
+++ b/tests/main/upgrade-from-2.15/task.yaml
@@ -3,14 +3,18 @@ summary: Ensure upgrades from snapd 2.15 work
 systems: [ubuntu-16.04-64]
 
 prepare: |
-    dpkg --purge snapd
+    #shellcheck source=tests/lib/pkgdb.sh
+    . "$TESTSLIB/pkgdb.sh"
+    distro_purge_package snapd
 
 restore: |
-    # Just remove when it was installed
+    #shellcheck source=tests/lib/pkgdb.sh
+    . "$TESTSLIB/pkgdb.sh"
     if [ -e old_install_done ]; then
-        dpkg --purge snapd ubuntu-core-launcher snap-confine
+        echo "Ensure core is gone and we have ubuntu-core instead"
+        distro_purge_package snapd
     fi
-    apt install -y "${GOHOME}"/snapd_*.deb
+    distro_install_build_snapd
 
 execute: |
     #shellcheck source=tests/lib/systemd.sh
--- a/tests/main/upgrade-from-2.15/task.yaml
+++ b/tests/main/upgrade-from-2.15/task.yaml
@@ -6,7 +6,11 @@ prepare: |
     dpkg --purge snapd
 
 restore: |
-    dpkg --purge ubuntu-core-launcher snap-confine
+    # Just remove when it was installed
+    if [ -e old_install_done ]; then
+        dpkg --purge snapd ubuntu-core-launcher snap-confine
+    fi
+    apt install -y "${GOHOME}"/snapd_*.deb
 
 execute: |
     #shellcheck source=tests/lib/systemd.sh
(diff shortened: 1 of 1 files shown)


tests/main/upgrade-from-2.15/task.yaml
```diff
@@ -6,7 +6,11 @@ prepare: |
     dpkg --purge snapd
 
 restore: |
-    dpkg --purge ubuntu-core-launcher snap-confine
+    # Just remove when it was installed
+    if [ -e old_install_done ]; then
+        dpkg --purge snapd ubuntu-core-launcher snap-confine
+    fi
+    apt install -y "${GOHOME}"/snapd_*.deb
```
bboozzoo: Isn't this done in prepare suite each?

sergiocazzolato: @bboozzoo thanks for reviewing, snapd is just installed when the suite is prepared, then its state is restored in each suite preparation.

tests/main/upgrade-from-2.15/task.yaml
```diff
@@ -3,10 +3,18 @@ summary: Ensure upgrades from snapd 2.15 work
...
+    . "$TESTSLIB/pkgdb.sh"
+    distro_purge_package snapd
 
 restore: |
-    dpkg --purge ubuntu-core-launcher snap-confine
+    #shellcheck source=tests/lib/pkgdb.sh
+    . "$TESTSLIB/pkgdb.sh"
+    if [ -e old_install_done ]; then
+        echo "Ensure core is gone and we have ubuntu-core instead"
+        distro_purge_package snapd
```
zyga:
```suggestion
        distro_purge_package snapd
        rm old_install_done
```

tests/main/upgrade-from-2.15/task.yaml
```diff
@@ -3,10 +3,18 @@ summary: Ensure upgrades from snapd 2.15 work
 systems: [ubuntu-16.04-64]
 
 prepare: |
-    dpkg --purge snapd
+    #shellcheck source=tests/lib/pkgdb.sh
```
zyga: Is this relevant for the test? This test is specific to Ubuntu anyway.

tests/main/upgrade-from-2.15/task.yaml
```diff
@@ -3,10 +3,18 @@ summary: Ensure upgrades from snapd 2.15 work
...
 
 restore: |
-    dpkg --purge ubuntu-core-launcher snap-confine
+    #shellcheck source=tests/lib/pkgdb.sh
+    . "$TESTSLIB/pkgdb.sh"
+    if [ -e old_install_done ]; then
+        echo "Ensure core is gone and we have ubuntu-core instead"
+        distro_purge_package snapd
+    fi
+    distro_install_build_snapd
```
zyga: Just a nitpick, this function name should probably say `distro_install_built_snapd`.

sergiocazzolato: @zyga agree, I'll make that change in a following PR due it is being used in several tests. Thanks for reviewing!!

tests/main/upgrade-from-2.15/task.yaml
sergiocazzolato: It is done like this becasuse it is specific for ubuntu xenial

tests/main/upgrade-from-2.15/task.yaml
sergiocazzolato: it is not needed "rm -f old_install_done" due to the test directory is restored automatically.

PR #7172, opened 2019-07-25
tests: work around classic snap affecting the host

zyga:
The classic snap somehow ends up changing /dev/pts on the host. My hunch
is because it can mount anything and it actually _remounts_ an existing
mount point rather than mounting something new, therefore the
propagation settings don't protect us from this.

The bug is filed as https://github.com/snapcore/classic-snap/issues/32

### DIFF
--- a/tests/main/ubuntu-core-classic/task.yaml
+++ b/tests/main/ubuntu-core-classic/task.yaml
@@ -14,6 +14,9 @@ prepare: |
 
 restore: |
     rm -f /etc/sudoers.d/create-test
+    # Undo the change done by the classic snap.
+    # FIXME: https://github.com/snapcore/classic-snap/issues/32
+    mount devpts -t devpts /dev/pts -o remount,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000
 
 execute: |
     echo "Ensure classic can be installed"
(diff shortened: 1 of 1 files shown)


tests/main/ubuntu-core-classic/task.yaml
```diff
@@ -14,6 +14,9 @@ prepare: |
 
 restore: |
     rm -f /etc/sudoers.d/create-test
+    # Undo the change done by the classic snap.
+    # FIXME: https://github.com/snapcore/classic-snap/issues/32
+    mount devpts -t devpts /dev/pts -o remount,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000
```
bboozzoo: I'd try to catch the bad state, causing it to fail once a fix is implemented:

```
# Undo the change done by the classic snap.
# FIXME: https://github.com/snapcore/classic-snap/issues/32 the check will fail once a fix is in place
grep '/ /dev/pts' /proc/self/mountinfo | MATCH " rw,mode=666,ptmxmode=666"
# restore the original state
mount devpts -t devpts /dev/pts -o remount,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000
```

zyga: I don't want to introduce another reason for our tests to fail. I'm looking at the associated bug and hopefully can fix both sides, rendering this workaround void.

PR #7173, opened 2019-07-26
interfaces: support Tegra display drivers

alfonsosanchezbeato:
Support Tegra display drivers in x11/opengl interfaces


interfaces/builtin/opengl.go
```diff
@@ -91,10 +94,11 @@ unix (bind,listen) type=seqpacket addr="@cuda-uvmfd-[0-9a-f]*",
 @{PROC}/driver/prl_vtg rw,
 
 # /sys/devices
-/sys/devices/pci[0-9a-f]*/**/config r,
-/sys/devices/pci[0-9a-f]*/**/revision r,
-/sys/devices/pci[0-9a-f]*/**/{,subsystem_}device r,
-/sys/devices/pci[0-9a-f]*/**/{,subsystem_}vendor r,
+/sys/devices/**pci[0-9a-f]*/**/config r,
```
zyga: What is the ** for? what was the actual path you saw here?

interfaces/builtin/opengl.go
```diff
@@ -121,6 +125,7 @@ var openglConnectedPlugUDev = []string{
 	`KERNEL=="renderD[0-9]*"`,
 	`KERNEL=="nvhost-*"`,
 	`KERNEL=="nvmap"`,
+	`KERNEL=="tegra_dc_*"`,
```
zyga: Did you see any additional devices in /dev that snap-confine's Nvidia code needs to support?

interfaces/builtin/opengl.go
alfonsosanchezbeato: It is in the commit message:

/sys/devices/1003000.pcie-controller/pci0000:00/0000:00:02.0/*

interfaces/builtin/opengl.go
alfonsosanchezbeato: No, fortunately these devices do show in /sys and have proper udev properties.

interfaces/builtin/x11.go
```diff
@@ -71,6 +71,9 @@ unix (bind, listen, accept)
 unix (bind, listen, accept)
      type=stream
      addr="@/tmp/.ICE-unix/[0-9]*",
+unix (bind, listen, accept)
+     type=dgram
+     addr="@nvidia[0-9a-f]*",
```
jdstrand: This is in x11PermanentSlotAppArmor. Why does an X server provide a listening service on an abstract socket with @nvidia? This feels out of place here. Can you comment?

interfaces/builtin/opengl.go
```diff
@@ -83,6 +83,9 @@ unix (bind,listen) type=seqpacket addr="@cuda-uvmfd-[0-9a-f]*",
 /dev/nvhost-* rw,
 /dev/nvmap rw,
 
+# Tegra display driver
+/dev/tegra_dc_* rw,
```
jdstrand: What does this match in practice?

interfaces/builtin/opengl.go
```diff
@@ -91,10 +94,11 @@ unix (bind,listen) type=seqpacket addr="@cuda-uvmfd-[0-9a-f]*",
...
 # /sys/devices
-/sys/devices/pci[0-9a-f]*/**/config r,
-/sys/devices/pci[0-9a-f]*/**/revision r,
-/sys/devices/pci[0-9a-f]*/**/{,subsystem_}device r,
-/sys/devices/pci[0-9a-f]*/**/{,subsystem_}vendor r,
+/sys/devices/**pci[0-9a-f]*/**/config r,
+/sys/devices/**pci[0-9a-f]*/**/revision r,
+/sys/devices/**pci[0-9a-f]*/**/{,subsystem_}class r,
+/sys/devices/**pci[0-9a-f]*/**/{,subsystem_}device r,
+/sys/devices/**pci[0-9a-f]*/**/{,subsystem_}vendor r,
```
jdstrand: Can you adjust these to be somewhat finer with:
```
/sys/devices/{,*pcie-controller/}pci[0-9a-f]*/**/config r,
/sys/devices/{,*pcie-controller/}pci[0-9a-f]*/**/revision r,
/sys/devices/{,*pcie-controller/}pci[0-9a-f]*/**/{,subsystem_}class r,
/sys/devices/{,*pcie-controller/}pci[0-9a-f]*/**/{,subsystem_}device r,
/sys/devices/{,*pcie-controller/}pci[0-9a-f]*/**/{,subsystem_}vendor r,
```

alfonsosanchezbeato: Sure, will do

interfaces/builtin/opengl.go
alfonsosanchezbeato: `tegra_dc_ctrl`, `tegra_dc_0` and `tegra_dc_1` (control device plus additional devices per display you can connect).

interfaces/builtin/x11.go
alfonsosanchezbeato: The socket is created by Xorg when loading the nvidia driver. It is then used by open GL applications (see client rule in opengl interface: https://github.com/snapcore/snapd/blob/master/interfaces/builtin/opengl.go#L68). Without this rule, Xorg refuses to start, that is why I included it here - but there are reasons too to move this to the opengl interface. Please let me know which is your preference.

interfaces/builtin/opengl.go
alfonsosanchezbeato: I have created slightly more fine-grained rules.

chipaca:
@jdstrand ping

interfaces/builtin/x11.go
jdstrand: Based on your feedback, I think it is reasonable to keep it in the x11 slot policy since most applications won't need it. Please add a comment above this rule on why you added it. I suggest:

```
# On systems with Tegra drivers, X11 needs to create the socket for clients to
# use.
```

alfonsosanchezbeato: Comment added now

alfonsosanchezbeato:
Comment added now

PR #7182, opened 2019-07-26
tests: remove core18 if installed by tests

zyga:
When tests run on ubuntu core, the automatic cleanup performed is
different and imperfect. In particular it is very easy to "leak"
additional base snaps, like core18.

Before we have a better mechanism available, remove core18 while
restoring a test case that did not have core18 before the test executed.

### DIFF
--- a/tests/lib/prepare-restore.sh
+++ b/tests/lib/prepare-restore.sh
@@ -512,6 +512,16 @@ prepare_suite_each() {
     echo -n "$SPREAD_JOB " >> "$RUNTIME_STATE_PATH/runs"
     # shellcheck source=tests/lib/reset.sh
     "$TESTSLIB"/reset.sh --reuse-core
+
+    # If core18 was not installed at this stage then drop a marker that it
+    # should be removed at the end of the test, if present. This is a bit
+    # crufty and ideally we'd have a method of removing unused bases
+    # automatically.
+    if ! snap list core18 2>/dev/null >/dev/null; then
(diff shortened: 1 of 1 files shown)


tests/lib/prepare-restore.sh
```diff
@@ -543,6 +553,13 @@ restore_suite_each() {
         rm -rf "${PWD}.tar"
     fi
 
+    # The core18 snap may have been automatically installed by the test.
+    # If it was not present before the test then it should be removed.
+    if [ -e /run/spread/please-remove-core18 ]; then
```
anonymouse64: should this maybe be:

```suggestion
    if [ -f /run/spread/please-remove-core18 ]; then
```

I don't know why this file would "exist" but not be a regular file, but -f seems more natural to me.

zyga:
This is red because on some systems we still pull in more test snaps. I'm going though this list now.

zyga:
I think there are two aspects that make this fail:

- There are still some tests that keep a `base: core18` snap installed, which is not automatically removed on core systems.
- The leak detector is blocked by several other bugs which means it cannot be merged to easily identify the offending tests

In addition some tests leave snapd removed/disabled which makes the remove part fail. I need to re-consider how to approach this.

zyga:
The core18 situation should now be resolved, this might now pass.

zyga:
I don't think this is really necessary now. With the fix to snap removal this is redundant.

PR #7189, opened 2019-07-29
HACKING.md: update spread section, other updates

anonymouse64:
* The command on bionic and later is autopkgtest-buildvm-ubuntu-cloud and the adt-buildvm-ubuntu-cloud command is no longer available on bionic.
* Make the assumption a user is probably running bionic and later rather than xenial which is getting close to EOL now.
* The name of the machine that spread tries to use for xenial 64-bit, is ubuntu-16.04-64.img, not ubuntu-16.04.img, the latter will be ignored.
* The spread command example should just use the image the user would have downloaded rather than have spread try and fail to launch every other system defined in spread.yaml.
* Adding a `sudo apt update` in all lines where we install apt packages is a good habit for folks who might not be aware that their apt repos need to be kept up to date.
* Add download instructions for spread images at spread.zygoon.pl. (thanks Zygmunt for hosting these!)
* Split up the Running spread tests section into different subsections, for building VMs, downloading VMs, and finally running tests with the VMs.
* Make all lines under 60 lines.
* Specify supported Go versions with snapd.
* Download spread from the AWS link rather than the snap, because spread as a snap only tries to use $HOME/snap/spread/current/.spread, rather than $HOME/.spread which is easier to document and use here.
* Delete old, unused nvidia ./configure option

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -8,7 +8,8 @@ integration test framework for the integration/system level tests.
 
 ### Supported Go versions
 
-snapd is supported on Go 1.6 onwards.
+From snapd 2.38, snapd supports Go 1.9 and onwards. For earlier snapd 
+releases, snapd supports Go 1.6.
 
 ### Setting up a GOPATH
 
@@ -126,9 +127,11 @@ To run the various tests that we have to ensure a high quality source just run:
 This will check if the source format is consistent, that it builds, all tests
 work as expected and that "go vet" has nothing to complain.
 
-The source format follows the `gofmt -s` formating. Please run this on your sources files if `run-checks` complains about the format.
+The source format follows the `gofmt -s` formating. Please run this on your 
+source files if `run-checks` complains about the format.
 
-You can run individual test for a sub-package by changing into that directory and:
+You can run an individual test for a sub-package by changing into that 
+directory and:
 
     go test -check.f $testname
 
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -138,55 +140,87 @@ If a test hangs, you can enable verbose mode:
...
-    $ mv adt-xenial-amd64-cloud.img ubuntu-16.04.img
-    # For trusty
-    $ adt-buildvm-ubuntu-cloud -r trusty --post-command='sudo apt-get install -y --install-recommends linux-generic-lts-xenial && update-grub'
+To run the spread tests via QEMU you need to create VM images in the
+`~/.spread/qemu` directory:
+
+    $ mkdir -p ~/.spread/qemu
+    $ cd ~/.spread/qemu
+
+Assuming you are building on an Ubuntu 18.04 Bionic Beaver (or a later
```
degville: ~~an~~, also we normally write Ubuntu 18.04.2 LTS (Bionic Beaver) with the release name in brackets - same for Disco Dingo.
ie.
`Assuming you are building on Ubuntu 18.04 LTS (Bionic Beaver), or a later release like Ubuntu 19.04 (Disco Dingo), run the following to build a 64-bit Ubuntu 16.04 VM to run the spread test on:`

HACKING.md
```diff
@@ -138,55 +140,87 @@ If a test hangs, you can enable verbose mode:
...
+    $ cd ~/.spread/qemu
+
+Assuming you are building on an Ubuntu 18.04 Bionic Beaver (or a later 
+development release like Ubuntu 19.04 Disco Dingo), run the following to build a
+64-bit Ubuntu 16.04 VM to run the spread tests on:
+
+    $ autopkgtest-buildvm-ubuntu-cloud -r xenial
+    $ mv autopkgtest-xenial-amd64.img ubuntu-16.04-64.img
+
+To build a Ubuntu 15.04 Trusty Tahr based VM, use:
```
degville: To build *an* Ubuntu *14.04* LTS (Trust Tahr)...

HACKING.md
```diff
@@ -138,55 +140,87 @@ If a test hangs, you can enable verbose mode:
...
+
+    $ autopkgtest-buildvm-ubuntu-cloud -r trusty --post-command='sudo apt-get install -y --install-recommends linux-generic-lts-xenial && update-grub'
     $ mv adt-trusty-amd64-cloud.img ubuntu-14.04-64.img
 
+This is because we need at least 4.4+ kernel for snapd to run on trusty, which
```
degville: maybe *Trusty* (capital T)...

HACKING.md
```diff
@@ -138,55 +140,87 @@ If a test hangs, you can enable verbose mode:
...
+instead of `autopkgtest-buildvm-ubuntu-cloud` (the latter replaced the former in
+18.04):
 
-And you can run the tests via:
+    $ adt-buildvm-ubuntu-cloud -r xenial
+    $ mv adt-xenial-amd64-cloud.img ubuntu-16.04-64.img
+
+#### Downloading spread VM images
+
+Alternatively, you can instead of building the QEMU images manually, you may
```
degville: ~~you can~~

HACKING.md
zyga: If you are not on those systems, pointing developers to spread.zygoon.pl could be an alternative.

HACKING.md
anonymouse64: Done

HACKING.md
anonymouse64: Changed to Trusty

HACKING.md
anonymouse64: Okay, I changed all instances to match that pattern, but do we need to add the .2 ? At some point there will be an 18.04.3, and 18.04.4, etc and it seems clear that just saying 18.04 refers to all of those. Also, for clarity dev releases like Ubuntu 19.04 (Disco Dingo) don't have LTS in them, but do they have a "Dev" name in there somewhere?

HACKING.md
```diff
@@ -1,18 +1,19 @@
...
+and we use the [spread](https://github.com/snapcore/spread) integration test
+framework for the integration/system level tests.
 
 ## Development
 
 ### Supported Go versions
 
-snapd is supported on Go 1.6 onwards.
+Starting with snapd 2.38, Go 1.9 onwards is supported. For older releases of
+snapd, Go 1.6 and onwards is supported.
```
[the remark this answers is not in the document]
anonymouse64: Okay fixed

HACKING.md
anonymouse64: Fixed

HACKING.md
```diff
@@ -126,9 +126,11 @@ To run the various tests that we have to ensure a high quality source just run:
 This will check if the source format is consistent, that it builds, all tests
 work as expected and that "go vet" has nothing to complain.
 
-The source format follows the `gofmt -s` formating. Please run this on your sources files if `run-checks` complains about the format.
+The source format follows the `gofmt -s` formating. Please run this on your
+sources files if `run-checks` complains about the format.
 
-You can run individual test for a sub-package by changing into that directory and:
+You can run individual test for a sub-package by changing into that directory
```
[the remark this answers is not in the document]
anonymouse64: Fixed

HACKING.md
```diff
@@ -126,9 +126,11 @@ To run the various tests that we have to ensure a high quality source just run:
 This will check if the source format is consistent, that it builds, all tests
 work as expected and that "go vet" has nothing to complain.
 
-The source format follows the `gofmt -s` formating. Please run this on your sources files if `run-checks` complains about the format.
+The source format follows the `gofmt -s` formating. Please run this on your
+sources files if `run-checks` complains about the format.
```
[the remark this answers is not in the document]
anonymouse64: Fixed

anonymouse64:
@zyga okay I dropped the line width changes to otherwise unchanged lines and only changed the content around spread testing. I will propose a followup PR making everything less than 80 characters, but I still applied @degville's proposed typo fixes to those lines here anyways.

HACKING.md
degville: I think it's ok just to mark the LTS releases as I think readers will be used to seeing *LTS* and may question their memory if its missing. I normally only mention the distro name first time (eg. Bionic Beaver) within a text, and only then if the text exists in relative isolation to other docs (as this does). It's mainly to reduce the mental friction of the same distros being referred to in lots of different ways.

anonymouse64: Okay, in that case I'll probably drop the duplicated distro names

anonymouse64:
merging master here in the hopes that unbreaks the tests here...

PR #7190, opened 2019-07-29
tests: set GOTRACEBACK=1 when running tests

zyga:
We're seeing panics in unit test code when invoked in travis. It seems
the test code is passing over the 10 minute watermark built into "go
test". With this patch we'll see what was going on at the time of the
failure.

Forum: https://forum.snapcraft.io/t/12513


bboozzoo:
Go test will stop and dump a backtrace once tests go over 10 minutes run time. Sadly, the tests hit the travis timeout of 10 minutes without any output before go test gets to act.

PR #7191, opened 2019-07-29
tests: remove installed snap on restore

zyga:
The test snap keeps the core18 snap busy, thus preventing removal in the
suite restore section.

### DIFF
--- a/tests/main/core18-configure-hook/task.yaml
+++ b/tests/main/core18-configure-hook/task.yaml
@@ -30,3 +30,8 @@ execute: |
     fi
 
     test -e /var/snap/test-snapd-with-configure-core18/common/configure-ran
+
+restore: |
+    if snap list test-snapd-with-configure-core18; then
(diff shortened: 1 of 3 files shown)


tests/main/core18-configure-hook/task.yaml
```diff
@@ -30,3 +30,8 @@ execute: |
     fi
 
     test -e /var/snap/test-snapd-with-configure-core18/common/configure-ran
+
+restore: |
+    if snap list test-snapd-with-configure-core18; then
+        snap remove test-snapd-with-configure-core18
+    fi
```
chipaca: `snap remove` of an uninstalled snap succeeds, so the `if` isn't strictly necessary (which is why the other restores that don't do this check are OK).

zyga: Ahh, that's neat. I didn't know that. I'll do a pass to simplify this.

PR #7192, opened 2019-07-29
tests: initial change to retry on external files download

sergiocazzolato:
This change is to retry downloads to avoid errors caused by connectivity
issues.


tests/lib/remote.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+download_file(){
```
zyga: Can you please make `download_file` a new executable in `tests/lib/bin`. This way it has extremely predictable behavior and does not need to be implemented in shell anymore. It also means you don't have to shell-source anything to get access to it, it is always on path.

zyga: Also having that, we could really use python for it, dropping the need to use curl entirely.

sergiocazzolato: @zyga sure, nice idea! Thanks for reviewing!!

zyga: Just a follow up suggestion, since we have `mountinfo-tool` and are about to get `testbed-tool`, could you please call this new one `download-tool`

tests/lib/bin/download-file
```diff
@@ -0,0 +1,76 @@
...
+
+    print('Failed to download file from {} after {} attempts'.format(
+        url, MAX_ATTEMPTS))
+    exit(1)
+
+
+if __name__ == '__main__':
+    args_parser = ArgumentParser("Parse input parameters")
+    args_parser.add_argument('url')
+    args_parser.add_argument('-p', '--print-file', action='store_true')
```
anonymouse64: this option name is a little confusing, currently if this option is set, then the output is not printed to a file, but rather the file downloaded is printed to stdout. I think a better name for this would be `--print-body`, as to me that's clear what's being printed and also clear that it's not going to a file anywhere.

tests/lib/bin/download-file
```diff
@@ -0,0 +1,76 @@
...
+import os
+try:
+    from urlparse import urlsplit
+except ImportError:
+    from urllib.parse import urlsplit
+
+
+MAX_ATTEMPTS = 5
+CONNECTION_TIMEOUT = 10
+READ_TIMEOUT = 5
```
anonymouse64: it would be nice if there was an easy way to override these parameters from the command line, especially for a developer trying to debug if a particular endpoint is just really slow. Unfortunately I have seem some internet connections with really low latency that while slow would eventually pass if I could have configured a timeout like this was just a little bit higher and be a little bit more tolerant *cough* *cough* hotel WiFi

tests/lib/bin/download-file
```diff
@@ -0,0 +1,76 @@
...
+
+
+def get_url_filename(url):
+    return os.path.basename(urlsplit(url)[2])
+
+
+def download_file(url, print_file=False, output_name='', output_dir=''):
```
mvo5: Silly question (maybe missed stuff because of my vacation) - curl has --retry and --retry-delay and --connect-timeout so I wonder a bit about the benefits of python here? The code is simple but doing in python means we need to think about things like how to get the data on disk. The current code seems to be reading it into memory first and then writes it on disk which may be a problem for big files. Anyway, I guess I wonder if we could get away with using curl here. Having the abstraction of bin/download-file is nice though because it means we can replace the implementation if we want to.

sergiocazzolato:
Close untils we define if we should move that to an external tool or keep it using cuerl

PR #7194, opened 2019-07-30
packaging: use %systemd_user_* macros to enable session agent socket according to presets

jhenstridge:
As mentioned by @Conan-Kudo in comments on PR #6954 after merge, we shouldn't be shipping symlinks in `sockets.target.wants` to enable the socket for Fedora/OpenSUSE.  Instead we should use the `%systemd_*` macros to enable the socket at install time according to the preset policy.

### DIFF
--- a/packaging/fedora/snapd.spec
+++ b/packaging/fedora/snapd.spec
@@ -60,6 +60,7 @@
 %global import_path     %{provider_prefix}
 
 %global snappy_svcs     snapd.service snapd.socket snapd.autoimport.service snapd.seeded.service
+%global snappy_user_svcs snapd.session-agent.socket
 
 # Until we have a way to add more extldflags to gobuild macro...
 %if 0%{?fedora}
@@ -727,7 +728,6 @@ popd
 %{_unitdir}/snapd.seeded.service
 %{_userunitdir}/snapd.session-agent.service
 %{_userunitdir}/snapd.session-agent.socket
-%{_userunitdir}/sockets.target.wants/snapd.session-agent.socket
 %{_datadir}/dbus-1/services/io.snapcraft.Launcher.service
 %{_datadir}/dbus-1/services/io.snapcraft.Settings.service
 %{_datadir}/polkit-1/actions/io.snapcraft.snapd.policy
@@ -803,6 +803,7 @@ popd
 %sysctl_apply 99-snap.conf
 %endif
 %systemd_post %{snappy_svcs}
+%systemd_user_post %{snappy_user_svcs}
 # If install, test if snapd socket and timer are enabled.
 # If enabled, then attempt to start them. This will silently fail
 # in chroots or other environments where services aren't expected
@@ -815,6 +816,7 @@ fi
 
 %preun
 %systemd_preun %{snappy_svcs}
+%systemd_user_preun %{snappy_user_svcs}
 
 # Remove all Snappy content if snapd is being fully uninstalled
 if [ $1 -eq 0 ]; then
@@ -823,6 +825,7 @@ fi
 
 %postun
 %systemd_postun_with_restart %{snappy_svcs}
+%systemd_user_postun %{snappy_user_svcs}
 
 %if 0%{?with_selinux}
 %triggerun -- snapd < 2.39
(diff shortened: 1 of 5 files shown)


Conan-Kudo:
Preset request filed for Fedora: https://bugzilla.redhat.com/show_bug.cgi?id=1734371

packaging/fedora/snapd.spec
```diff
@@ -600,6 +601,8 @@ pushd ./data
               SYSTEMDSYSTEMUNITDIR="%{_unitdir}" \
               SNAP_MOUNT_DIR="%{_sharedstatedir}/snapd/snap" \
               SNAPD_ENVIRONMENT_FILE="%{_sysconfdir}/sysconfig/snapd"
+# Don't package the sockets.target.wants symlink
```
zyga: It would be good to explain why.

packaging/opensuse/snapd.spec
```diff
@@ -260,6 +261,8 @@ popd
 		LIBEXECDIR=%{_libexecdir} \
 		SYSTEMDSYSTEMUNITDIR=%{_unitdir} \
 		SNAP_MOUNT_DIR=%{snap_mount_dir}
+# Don't package the sockets.target.wants symlink
```
zyga: Ditto

tests/main/snap-session-agent-socket-activation/task.yaml
```diff
@@ -12,10 +12,16 @@ environment:
     USER_RUNTIME_DIR: /run/user/${TEST_UID}
 
 prepare: |
+    # Ensure that snapd.session-agent.socket is enabled.  This may not
+    # be the case on distributions where presets have been used to
+    # disable it.
+    systemctl --user --global enable snapd.session-agent.socket
```
zyga: Please store a marker file and add a corresponding restore code to undo this.

jhenstridge: I think I've got this part implemented correctly now.  Things were a little confusing, since `is-enabled` was returning false on the Ubuntu targets where we'd enabled the socket via a symlink in `/usr/lib/systemd/user`.  It seems this command only considers symlinks in `/etc/systemd/user`.

This was also the cause of the ubuntu-core-16 test failure: systemctl thought the unit was disabled, so tried to create a new symlink to enable it, which failed due to `/etc/systemd/user` being read-only on that release.

Since hard shipping the `sockets.target.wants` symlink in the package seems to be a Debian-ism, I've removed it from the `data/systemd-user` Makefile and added it to the .deb packaging instead.  According to [Debian bug #764678](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=764678), debhelper 11.5 has support for enabling user units the standard way.  So we might be able to remove it from there too eventually.

packaging/fedora/snapd.spec
```diff
@@ -803,6 +803,9 @@ popd
 %sysctl_apply 99-snap.conf
 %endif
 %systemd_post %{snappy_svcs}
+%if !(0%{?rhel} && 0%{?rhel} < 8)
+%systemd_user_post %{snappy_user_svcs}
```
Conan-Kudo: Why is this not being executed on EL7? The macro exists there...

packaging/fedora/snapd.spec
```diff
@@ -815,6 +818,9 @@ fi
 
 %preun
 %systemd_preun %{snappy_svcs}
+%if !(0%{?rhel} && 0%{?rhel} < 8)
+%systemd_user_preun %{snappy_user_svcs}
```
Conan-Kudo: Why is this not being executed on EL7? The macro exists there...

packaging/fedora/snapd.spec
```diff
@@ -823,6 +829,9 @@ fi
 
 %postun
 %systemd_postun_with_restart %{snappy_svcs}
+%if !(0%{?rhel} && 0%{?rhel} < 8)
+%systemd_user_postun %{snappy_user_svcs}
```
Conan-Kudo: Why is this not being executed on EL7? The macro exists there...

packaging/fedora/snapd.spec
jhenstridge: I was getting some failures on one of the RHEL7 derivatives (I think Amazon Linux 2?) earlier in this branch's development, and adding this conditional seemed to fix it.  I've just removed the conditionals and all the tests are passing so I'm not sure what was going on there.

I guess I jumped to the conclusion that the user instance of systemd was non-functional since it didn't seem to be running when I booted up the CentOS 7 live GNOME image.

jhenstridge:
As far as other distros go, it looks like we'll need to get a similar preset policy change in OpenSUSE:

https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines#Enabling_systemd_unit_files

For Arch, it looks like our package leaves the main snapd daemon disabled on install, so perhaps it is just a documentation change to say that the session agent socket unit will also need to be enabled.

PR #7196, opened 2019-07-30
packaging: use passthrough for type:snapd

stolowski:
Temporarily use `passthrough:` to set `type: snapd` when building snapd snap with snapcraft 2.x, until failures with snapcraft 3.x and lxd on travis (PR #7092) are understood and fixed.

Note, the code that magically sets `TypeSnapd` on `infoSkeletonFromSnapYaml` is still present and will be removed in a followup once new snapd snap is in edge, because only then we can avoid hack in `repack_snapd_snap_with_deb_content()` test helper.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -10,8 +10,9 @@ description: |
 version: set-by-version-script-thxbye
 version-script: |
   ./mkversion.sh --output-only
-#FIXME: enable once snapcraft understands this
-#type: snapd
+#FIXME: get rid of passthrough once snapcraft 3.x + lxd can build snapd
+passthrough:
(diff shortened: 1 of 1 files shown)


stolowski:
This is not needed after all I think, I found way to move forward with #7092, closing.

stolowski:
Re-opening as #7092 needs a fix in snapocraft after all.

mvo5:
I would like to block this until 2.41 is branched (probably today after the standup). Just to ensure that 2.41 gets no hickups due to this change (its already a bit delayed).

stolowski:
Closing as #7092 should land soon.

PR #7197, opened 2019-07-30
usersession: track connections to session agent for exit on idle and peer credential checks

jhenstridge:
This PR adds a connection tracker to the user session agent.  This serves two purposes:

1. have the daemon exit after a period of inactivity.
2. drop connections when the peer is not root or the owner of the session agent.

As well as reducing resource usage, (1) also means new versions of the session agent will eventually take over on upgrades.  If permissions on `$XDG_RUNTIME_DIR` are correct (2) should never be triggered, but it makes sense to be paranoid for a service like this.


jhenstridge:
I've disabled building the `snap userd` suite of commands on MacOS, since the use of `SO_PEERCRED` is not supported on that platform.  Given that the MacOS support is meant to be just enough to run `snap download` and `snap pack`, I don't think it is worth trying to make that small part of the session agent conditional.

zyga:
Partial review from last night

usersession/agent/session_agent.go
```diff
@@ -113,12 +183,39 @@ func (s *SessionAgent) Start() {
...
+Loop:
+	for {
+		select {
+		case <-s.tomb.Dying():
+			break Loop
+		case <-timer.C:
+			s.tomb.Kill(nil)
+			break Loop
+		case <-s.idle.change:
+			newIdle := s.idle.isIdle()
```
mvo5: Do we (unit) test this part of the code?

usersession/agent/session_agent.go
```diff
@@ -79,6 +84,63 @@ func (c *Command) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	}
+	return nil, fmt.Errorf("cannot get peer credentials for connection %v", conn)
+}
+
+func (it *idleTracker) trackConn(conn net.Conn, state http.ConnState) {
+	// Perform peer credentials check
+	if state == http.StateNew {
+		ucred, err := getUcred(conn)
+		uid := uint32(sys.Geteuid())
+		if err != nil || (ucred.Uid != 0 && ucred.Uid != uid) {
```
anonymouse64: I would feel better about this check if we also had a spread test for this which directly tested if non-root can pass this check

usersession/agent/session_agent.go
```diff
@@ -79,6 +84,63 @@ func (c *Command) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+	it.mu.Lock()
+	oldActive := len(it.active)
+	if state == http.StateNew || state == http.StateActive {
+		it.active[conn] = struct{}{}
+	} else {
+		delete(it.active, conn)
+	}
+	newActive := len(it.active)
+	it.mu.Unlock()
```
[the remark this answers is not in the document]
jhenstridge: The aim was to release the lock before sending on the channel to avoid a deadlock.  But thinking about things more, I think things are still open to deadlocking during shutdown: if `exitOnIdle()` kills the tomb, it won't be reading from the `idleTracker` channel.  That could cause a deadlock if a connection came in between then and when `http.Server.Shutdown()` is called.  I'll have to think about this a bit more.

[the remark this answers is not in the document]
mvo5: Thanks for this comment! Maybe we can add a small comment in the code like: `// release the lock before sending to the channel to avoid deadlock` or similar (or whatever is appropriate after you rethought this bit :)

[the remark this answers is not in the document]
jhenstridge: Okay, so I've changed the way things work now:

1. `idleTracker` now has a `lastActive` field holding a timestamp that is updated each time the number of active connections changes to 0.
2. a new `idleTracker.idleDuration` method returns the length of time server has been idle, or 0 if there are active connections.
3. `exitOnIdle` now only selects on the tomb and the timer.  When the timer expires, we exit if the idle duration is greater than or equal to the timeout.  If it is less, we reset the timer appropriately (e.g. if the timeout is 30 seconds and the server has been idle for 5 seconds, we reset the timer for 25 seconds).

This gets rid of the synchronisation problem, and probably results in fewer wakeups overall.

usersession/agent/session_agent.go
jhenstridge: I've extended the spread test to make a curl call as the test user.  I'm not sure we've got an obvious non-`root`, non-`test` user to make use of in spread tests though.

usersession/agent/session_agent.go
jhenstridge: I've updated the `TestExitOnIdle` test to verify that the timeout is correctly extended when new requests are made.

usersession/agent/session_agent.go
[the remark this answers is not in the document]
jhenstridge: I've updated the code to log when either (a) getUcred returns an error, or (b) we detect a different user connecting to the socket, so these cases should end up in the journal now.  I also updated the test to verify the logging.

Both cases are definitely worth logging, since they would indicate a pretty serious misconfiguration.

usersession/agent/session_agent.go
```diff
@@ -25,6 +25,8 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"sync"
+	sys "syscall"
```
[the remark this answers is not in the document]
jhenstridge: I adopted this from the `daemon/ucrednet.go` code.  Looking at the tree this seems to be the only place the pattern is used, so I've removed the alias here.

usersession/agent/session_agent_test.go
```diff
@@ -103,3 +110,72 @@ func (s *sessionAgentSuite) TestDying(c *C) {
...
+
+	makeRequest := func() {
+		response, err := s.client.Get("http://localhost/v1/session-info")
+		c.Assert(err, IsNil)
+		defer response.Body.Close()
+		c.Check(response.StatusCode, Equals, 200)
+	}
+	makeRequest()
+	time.Sleep(25 * time.Millisecond)
+	makeRequest()
```
pedronis: how long does it take from Start to here, just wondering if 100ms is ok or low for a overloaded test machine

usersession/agent/session_agent_test.go
```diff
@@ -103,3 +110,72 @@ func (s *sessionAgentSuite) TestDying(c *C) {
...
+	logger.SetLogger(log)
+
+	sa, err := agent.New()
+	c.Assert(err, IsNil)
+	sa.Start()
+	defer sa.Stop()
+
+	// Connections from other users are dropped.
+	uid := uint32(syscall.Geteuid())
+	restore := agent.MockUcred(&syscall.Ucred{Uid: uid + 1}, nil)
```
pedronis: you want to call MockUcred once before starting things and change the result of a fixed mock function under some locking. As it is this test is not "go test -race" clean. there might be other issues, the output is fairly noisy atm

jhenstridge: I've split the test into three, calling MockUcred before the goroutines are spawned.

The other data races are on access to the buffer we are logging to.  This feels a bit like a shortcoming of our `logger.MockLogger` API: while all writes to the logger are protected by `logger.lock`, there is no way for calling code to acquire that mutex when inspecting the contents of the returned log buffer.  If a test involves multiple goroutines creating log messages, there doesn't seem to be a race free way to check it while the goroutines are still running.

pedronis: Would introducing something like this help (haven't tried it)?:

```
diff --git a/logger/logger.go b/logger/logger.go
index 313e5b4dc..2bc631342 100644
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -102,6 +102,15 @@ func MockLogger() (buf *bytes.Buffer, restore func()) {
        }
 }
 
+// WithLoggerLock invokes f with the global logger lock, useful for
+// tests involving goroutines with MockLogger.
+func WithLoggerLock(f func()) {
+       lock.Lock()
+       defer lock.Unlock()
+
+       f()
+}
+
 // SetLogger sets the global logger to the given one
 func SetLogger(l Logger) {
        lock.Lock()
```

usersession/agent/session_agent_test.go
jhenstridge: The two requests plus the sleep generally take 27-27.5ms on my system, so about 1ms per request.  I haven't noticed any runs exceeding the idle timeout yet, but I suppose we could increase it.  That adds to the runtime of the test suite though.

usersession/agent/session_agent_test.go
jhenstridge: That seems cleaner than shutting down the session agent before checking the log message.  I've added this helper (along with a short test) to the PR.

usersession/agent/session_agent_test.go
pedronis: let's push it then to 150ms  (a factor of ~5 pessimism)

usersession/agent/session_agent_test.go
```diff
@@ -103,3 +109,98 @@ func (s *sessionAgentSuite) TestDying(c *C) {
...
+	case <-agent.Dying():
+	case <-time.After(2 * time.Second):
+		c.Fatal("agent did not exit after idle timeout expired")
+	}
+	elapsed := time.Now().Sub(startTime)
+	if elapsed < 125*time.Millisecond || elapsed > 175*time.Millisecond {
+		// The idle timeout should have been extended when we
+		// issued a second request after 25ms.
+		c.Errorf("Expected ellaped time close to 125 ms, but got %v", elapsed)
+	}
```
jdstrand: This is a good test. Wondering if 50ms is long enough to die for if something went wrong on a very taxed system...

usersession/agent/session_agent_test.go
```diff
@@ -103,3 +109,98 @@ func (s *sessionAgentSuite) TestDying(c *C) {
...
+}
+
+func (s *sessionAgentSuite) TestConnectFromOtherUser(c *C) {
+	logbuf, restore := logger.MockLogger()
+	defer restore()
+
+	// Mock connections to appear to come from a different user ID
+	uid := uint32(syscall.Geteuid())
```
jdstrand: osutil/sys Geteuid() instead?

usersession/agent/session_agent.go
```diff
@@ -79,6 +85,70 @@ func (c *Command) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	if uconn, ok := conn.(*net.UnixConn); ok {
+		f, err := uconn.File()
+		if err != nil {
+			return nil, err
+		}
+		defer f.Close()
+		return sysGetsockoptUcred(int(f.Fd()), syscall.SOL_SOCKET, syscall.SO_PEERCRED)
+	}
+	return nil, fmt.Errorf("expected a net.UnixConn, but got a %T", conn)
+}
```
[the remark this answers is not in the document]
jhenstridge: It would probably make sense to factor out the common code into `netutil`.  We can't import `daemon` here because it pulls in a lot of unwanted dependencies (in particular, dependencies with init functions or global initialisers that would change the behaviour of the `snap` command line tool).

PR #7199, opened 2019-07-31
overlord/snapstate: keep current track if only risk is specified

stolowski:
Keep current track (if set) if only the risk level is specified - this affects refresh and switch operations. That means that e.g. if a snap is tracking `2.0/stable` then `snap refresh --candidate some-snap` will refresh and switch to 2.0/candidate. 

The change is explained here: https://forum.snapcraft.io/t/behavior-change-risk-only-channel-specifications-will-not-switch-track/11769


stolowski:
The is one small problem with snap cli and `switch` command as the output message shown by switch command uses channel name as provided via the commandline args, the switch op api call doesn't return actual details of the channel used. So, if - say - track 3.0 is pinned then switch does the right think internally but the client reports:
```
$ sudo snap switch --edge lxd
"lxd" switched to the "edge" channel
``
instead of the expected `"lxd" switched to the "3.0/edge" channel`.

The cli output is correct with `snap refresh` because the api returns details of the refreshed snaps, including actual channels:
```
$ sudo snap refresh --edge lxd
lxd (3.0/edge) git-81b81b9 from Canonical✓ refreshed
```

TBD what's the best course of action with this detail.

File: overlord/snapstate/snapstate_test.go [hunk 1]
```diff
@@ -14547,6 +14548,12 @@ func (s *snapmgrTestSuite) TestResolveChannelPinnedTrack(c *C) {
 		// risk only defaults to pinned kernel track
 		{snap: "kernel", new: "stable", exp: "17/stable", kernelTrack: "17"},
 		{snap: "kernel", new: "edge", exp: "17/edge", kernelTrack: "17"},
+		// risk only with regular snap defaults to pinned track
+		{snap: "some-snap", new: "stable", exp: "3.0/stable", snapChannel: "3.0/edge"},
+		{snap: "some-snap", new: "4.0/stable", exp: "4.0/stable", snapChannel: "3.0/edge"},
```
bboozzoo: Could use additional test case with the `latest` track:
```
{snap: "some-snap", new: "latest/stable", exp: "latest/stable", snapChannel: "3.0/edge"},
```

stolowski: Added.

PR #7202, opened 2019-07-31
tests: sync journal log before start the test

sergiocazzolato:
Sporadically some logs were being logged late (as part of the next
test). As the catalog-update test is checking if the refresh is verbose
or not, some messages could affect that check as in the following error:

+ echo 'Catalog update is doing verbose http logging (it should not).'
Catalog update is doing verbose http logging (it should not).
+ exit 1

The idea is to sync and flush the logs before start the new test as we
are doing before query the logs.

See full test error:
https://paste.ubuntu.com/p/DvpmHKcHtH/


tests/lib/prepare-restore.sh
```diff
@@ -513,6 +513,7 @@ prepare_suite_each() {
     # shellcheck source=tests/lib/reset.sh
     "$TESTSLIB"/reset.sh --reuse-core
     # Reset systemd journal cursor.
+    sync_journalctl_log
```
zyga: Would it be more sensible to move that functionality into `start_new_journalctl_log`?  This way there are fewer functions to keep track of.

tests/lib/journalctl.sh
```diff
@@ -65,3 +64,8 @@ get_journalctl_log_from_cursor(){
         journalctl "$@" --cursor "$cursor"
     fi
 }
+
+sync_journalctl_log(){
+    journalctl --flush || true
```
zyga: When can those two commands fail and why should we ignore the failure?

sergiocazzolato: it is because depending on the system sometime --flush or --sync could not be supported

tests/lib/prepare-restore.sh
sergiocazzolato: Thanks for the review!!! 
change done

PR #7203, opened 2019-08-01
i18n, vendor, packaging: drop github.com/ojii/gettext.go, use github.com/snapcore/go-gettext

bboozzoo:
The github.com/ojii/gettext.go library causes various issues with Go tools. Because of the name ending with `*.go`, Go tools are confused and the package is not go gettable. At the same time, the package has been dropped from Fedora, as the packaging infrastructure helpers break because of the import name.

I am trying to get github.com/snapcore/go-gettext packaged for Fedora. The i18n support is disabled for Debian Sid. So we might as well just switch to using the snapcore fork.


bboozzoo:
Marked as blocked since golang-github-snapcore-gettext is not available in Fedora yet.

Conan-Kudo:
@bboozzoo I've sponsored you in as a packager in Fedora, so you should be able to proceed further.

Conan-Kudo:
@mvo5 This needs to be cherry-picked into 2.40, as it blocks my ability to update snapd in Fedora entirely.

bboozzoo:
The `golang-github-snapcore-gettext` package was submitted for F30 and F29. Unblocking the PR as it should be possible to build it in a day or two.

PR #7204, opened 2019-08-01
cmd/snap: use showDone helper with 'snap switch'

stolowski:
Use `showDone` helper to display actual channel we switched to, rather than the requested one. This makes no difference right now, but it prepares `snap switch` for the upcoming change in channel switching semantics with regard to track and risk level, and also makes it internally consistent with the implementation of other snap ops.

### DIFF
--- a/cmd/snap/cmd_snap_op.go
+++ b/cmd/snap/cmd_snap_op.go
@@ -1007,7 +1007,7 @@ func (x cmdSwitch) Execute(args []string) error {
 	// some duplication between this and the two other switch-summarisers...
 	// in this one, we have three boolean things to check, meaning 2³=8 possibilities
 	// of which 3 are errors (which is why we look at this before running it)
-	if x.Cohort != "" && x.LeaveCohort {
+	if switchCohort && x.LeaveCohort {
 		// this one counts as two (no channel filter)
 		return fmt.Errorf(i18n.G("cannot specify both --cohort and --leave-cohort"))
 	}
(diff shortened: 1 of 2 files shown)


cmd/snap/cmd_snap_op.go
```diff
@@ -979,34 +1001,19 @@ func (x cmdSwitch) Execute(args []string) error {
 	name := string(x.Positional.Snap)
 	channel := string(x.Channel)
 
-	var msg string
+	switchCohort := x.Cohort != ""
+	switchChannel := x.Channel != ""
+
 	// some duplication between this and the two other switch-summarisers...
 	// in this one, we have three boolean things to check, meaning 2³=8 possibilities
 	// of which 3 are errors (which is why we look at this before running it)
```
chipaca: I think the comment needs adjusting to point out that the cases are split now into error ones (checked here) and non-error ones (checked in showDone)

stolowski: Done.

cmd/snap/cmd_snap_op.go
```diff
@@ -979,34 +1003,19 @@ func (x cmdSwitch) Execute(args []string) error {
...
-	// in this one, we have three boolean things to check, meaning 2³=8 possibilities
-	// of which 3 are errors (which is why we look at this before running it)
 	switchCohort := x.Cohort != ""
 	switchChannel := x.Channel != ""
-	switch {
-	case switchCohort && x.LeaveCohort:
+
+	// we have three boolean things to check, meaning 2³=8 possibilities
+	// of which 3 are errors (which is why we look at the errors first).
+	// the remaining 5 cases are handled by showDone.
```
bboozzoo: I'd use `remaining 5 valid cases ...`

PR #7205, opened 2019-08-01
rfc: introduce confinement options failsafe flag

zyga:
This patch introduces a new internal confinement option flag called failsafe
mode. When a security setup fails it is attempted again in failsafe mode. This
can be used to mitigate impact of both known and unknown issues in the
interface system that, otherwise, might result in total system failure.

The first consumer of the failsafe mode is the mount backend. When the flag is
asserted the mount namespace will be discarded ahead of the setup. This allows
to nullify effects of corrupted state of a mount namespace preventing changes
from being applied.

This is both a workaround and a reliability idea. I'm sharing it with Ondra for
some experiments.

### DIFF
--- a/interfaces/backend.go
+++ b/interfaces/backend.go
@@ -63,6 +63,10 @@ type ConfinementOptions struct {
 	JailMode bool
 	// Classic flag switches the core snap "chroot" off.
 	Classic bool
+	// FailSafeMode flag switches the security backend to failsafe mode.
+	// In failsafe mode some extra things happen. For instance the mount
+	// backend will discard the mount namespace and construct them anew.
+	FailSafeMode bool
 }
 
 // SecurityBackendOptions carries extra flags that affect initialization of the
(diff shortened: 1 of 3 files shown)


stolowski:
That's an interesting idea, one general remark - should we issue warnings (as: `snap warnings`) if something like this is detected? This of course depends on whether we expect user to take any actions, such as investigate logs, file a bug report etc. I'd say it probably makes sense, the scenearios we are trying to mitigate here are uncommon / unexpected, so it may make sense to ask user to do something.

zyga:
@stolowski I was thinking about poking the error tracker. In the case where it is a classic device or a core device with an active user we can also issue a warning.

zyga:
I would like to discuss this with @pedronis and @mvo5 tomorrow. I think we need something like this for reliability in face of unknown failures, like we experienced recently with some of our field engineers. The branch is not really meant for landing but as a discussion point.

jdstrand:
I think the concept of failsafe mode is interesting and useful, though it isn't clear when it should be applied or if it should be applied unconditionally (thinking about the scenario where something always fails the first time, then falls back to failsafe mode and works, but we never see that it happened). Based on other comments in this PR, it seems like this is understood (eg, warnings vs error tracker vs ...).

PR #7206, opened 2019-08-02
overlord/snapstate: use reflect to improve snapstate.Get

chipaca:
Based on [some profiling I did ages ago][1], loading a single element
from a json object using reflect to create an ad-hoc parser is more
efficient in both memory and speed than our current approach of using
a map[string]*json.RawMessage.

This implements this style of loader for one of the places where it
would help, in snapstate.

[1]: https://gist.github.com/chipaca/10ecdf364db962c9b5ee590f0f610b2a

### DIFF
--- a/overlord/snapstate/snapstate.go
+++ b/overlord/snapstate/snapstate.go
@@ -26,7 +26,9 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"reflect"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


overlord/snapstate/snapstate.go
```diff
@@ -2184,32 +2186,31 @@ func CurrentInfo(st *state.State, name string) (*snap.Info, error) {
...
+	}})
+	v := reflect.New(t)
+
+	err := st.Get("snaps", v.Interface())
 	if err != nil {
 		return err
 	}
-	raw, ok := snaps[name]
-	if !ok {
+	found := v.Elem().Field(0).Interface().(*SnapState)
```
anonymouse64: should this have a check to make sure the type assertion doesn't panic?

chipaca: not in this case, because a few lines up it defined the value to be a `*SnapState`.

chipaca:
Closing this non-essential PR until we get our PR count down.

PR #7208, opened 2019-08-02
cmd/snap, data/completion: improve completion for 'snap debug'

chipaca:
Two small changes:
1. actually support tab completion of debug commands:
   without this change completion of debug was broken, always
   suggesting more debug commands
2. tag the argument of validate-seed as a file, and required
   without this change, when giving it no argument it tried to
   validate a file called "" which failed with a curious error
   message. Also completion didn't know what to offer.

### DIFF
--- a/cmd/snap/cmd_debug_validate_seed.go
+++ b/cmd/snap/cmd_debug_validate_seed.go
@@ -27,8 +27,8 @@ import (
 
 type cmdValidateSeed struct {
 	Positionals struct {
-		SeedYamlPath string `positional-arg-name:"<seed-yaml-path>"`
-	} `positional-args:"true"`
+		SeedYamlPath flags.Filename `positional-arg-name:"<seed-yaml-path>"`
+	} `positional-args:"true" required:"true"`
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


data/completion/snap
```diff
@@ -38,15 +38,20 @@ _complete_snap() {
...
 
     # Only split on newlines
     local IFS=$'\n'
 
     # now we pass _just the bit that's being completed_ of the command
     # to snap for it to figure it out. go-flags isn't smart enough to
     # look at COMP_WORDS etc. itself.
-    COMPREPLY=($(GO_FLAGS_COMPLETION=1 snap "$command" "$cur"))
+    if [ "$command" = "debug" ]; then
+        command="${words[2]}"
```
zyga: Do you need anything like `${foo:-}` in case `foo` is undefined?

cmd/snap/cmd_debug_validate_seed.go
```diff
@@ -27,8 +27,8 @@ import (
 
 type cmdValidateSeed struct {
 	Positionals struct {
-		SeedYamlPath string `positional-arg-name:"<seed-yaml-path>"`
-	} `positional-args:"true"`
+		SeedYamlPath flags.Filename `positional-arg-name:"<seed-yaml-path>"`
+	} `positional-args:"true" required:"true"`
```
stolowski: \o/

data/completion/snap
chipaca: After writing and debugging and testing the change, I realised that we already checked for that (and bail out early)! So, no.

PR #7213, opened 2019-08-06
cmd/snap-confine: implement snap-device-helper internally

zyga:
This patch begins the cleanup of snap-confine's handling of udev
tagging. The script snap-device-helper had two roles:

1) Invoked by udev to dynamically add and remove device group ACLs
in response to udev hotplug events.
2) Invoked by snap-confine to configure the device cgroup.

The second role is discarded. Instead snap-confine now implements the
relevant logic itself. This "logic" is literally:

    dprintf(devices_allow_fd, "%c %u:%u", block_device ? 'b' : 'c',
	dev_major, dev_minor);

Doing so in C allows us to spare countless fork + exec combos, which
also fork for various parts of the script, making snap executing a
little bit faster.

In addition, since the snap-device-helper script is now only invoked by
udev in the initial mount namespace, we no longer have to worry about
the script interpreter. As such the packaging for some distributions is
adjusted to disable our previous workarounds.

This step opens the path towards cgroup v2 support, although much more
work is still required there.


zyga:
I'm closing this, based on further research I have changed my approach.

PR #7214, opened 2019-08-06
interfaces/network-setup-control: allow dbus netplan apply messages

anonymouse64:
This patch adds dbus accesses to allow a strictly confined `netplan apply` to call out to a D-Bus activated service that netplan will ship outside of confinement, and the service will call `netplan apply` in an unconfined manner, thus allowing it to run `systemctl` arbitrarily. 

The spread test is mostly done, the snap installed, `test-snapd-netplan-apply` is one I registered and built myself from upstream netplan master branch, with some small additions to make it strict. I have included the snapcraft.yaml source for this snap in the test dir and will request that the snap be transferred to the Canonical account.

Additionally, the spread system needs to have the D-Bus service setup outside of confinement for netplan to communicate with, so what I do in the spread test is uninstall the netplan D-Bus service if it exists, and replace it with a fake python service which just logs calls to the Apply member and always returns True. Since it has to be a system service, I was unable to get it working with the `dbus-launch` and overwriting `DBUS_SYSTEM_BUS_ADDRESS` with `DBUS_SESSION_BUS_ADDRESS` because when the service would be started, it wouldn't have permissions to request the io.netplan.Netplan name. I think this could have been resolved with some kind of policy configuration file, but I figured it was easier to just install the fake one and then remove it later than try to get it working with `dbus-launch`. Currently the spread tests are just set to run on all ubuntu classic systems except 14.04. I am not sure how to test Core yet since the rootfs would be read-only from the core snap and make it hard to overwrite the installed netplan service.

See also upstream netplan PR's: https://github.com/CanonicalLtd/netplan/pull/99 and https://github.com/CanonicalLtd/netplan/pull/93.

One additional question I had was whether we should add policy to allow executing just the `netplan` binary from the core snap? We will need to ship netplan in the core snap to ship the unconfined D-Bus service anyways, so it makes sense to me to allow applications to use the version of netplan provided by the core snap rather than have all of these network configuration agent snaps ship their own duplicated version of netplan (which could also some day get out of sync with the D-Bus service). I think yes, but I left it out of the policy for now...

### DIFF
--- a/interfaces/builtin/network_setup_control.go
+++ b/interfaces/builtin/network_setup_control.go
@@ -53,7 +53,8 @@ dbus (send)
     bus=system
     interface=io.netplan.Netplan
     path=/io/netplan/Netplan
-    member=Apply,
+	member=Apply
+	peer=(label=unconfined),
 
 `
 
--- a/interfaces/builtin/network_setup_control.go
+++ b/interfaces/builtin/network_setup_control.go
@@ -46,9 +46,7 @@ const networkSetupControlConnectedPlugAppArmor = `
 /run/udev/rules.d/ rw,                 # needed for cloud-init
 /run/udev/rules.d/[0-9]*-netplan-* rw,
 
-capability net_admin,
-
-#include <abstractions/dbus-session-strict>
+#include <abstractions/dbus-strict>
 
 # Allow use of NetPlan Apply API, used to apply network configuration
 dbus (send)
--- a/interfaces/builtin/network_setup_control.go
+++ b/interfaces/builtin/network_setup_control.go
@@ -54,8 +54,8 @@ capability net_admin,
 dbus (send)
     bus=system
     interface=io.netplan.Netplan
-	path=/io/netplan/Netplan
-	member=Apply,
+    path=/io/netplan/Netplan
+    member=Apply,
 
 `
 
--- a/interfaces/builtin/network_setup_control.go
+++ b/interfaces/builtin/network_setup_control.go
@@ -45,6 +45,18 @@ const networkSetupControlConnectedPlugAppArmor = `
 
 /run/udev/rules.d/ rw,                 # needed for cloud-init
 /run/udev/rules.d/[0-9]*-netplan-* rw,
+
+capability net_admin,
+
+#include <abstractions/dbus-session-strict>
+
+# Allow use of NetPlan Apply API, used to apply network configuration
+dbus (send)
+    bus=system
+    interface=io.netplan.Netplan
+	path=/io/netplan/Netplan
+	member=Apply,
+
 `
 
 func init() {
--- a/tests/main/netplan-apply/fake-netplan-apply-service.py
+++ b/tests/main/netplan-apply/fake-netplan-apply-service.py
@@ -19,7 +19,7 @@ class NetplanApplyService(dbus.service.Object):
                          out_signature="b")
     def Apply(self):
         # log that we were called and always return True
-        with open(self._logfile, "a") as fp:
+        with open(self._logfile, "a+") as fp:
             fp.write("Apply called\n")
         return True
 
--- a/tests/main/netplan-apply/fake-netplan-apply-service.py
+++ b/tests/main/netplan-apply/fake-netplan-apply-service.py
@@ -1,0 +1,50 @@
+#!/usr/bin/python3
+
+from gi.repository import GLib
+import dbus.mainloop.glib
+import dbus.service
+import sys
+
+BUS_NAME = "io.netplan.Netplan"
+OBJECT_PATH = "/io/netplan/Netplan"
+DOC_IFACE = "io.netplan.Netplan"
+
+
+class NetplanApplyService(dbus.service.Object):
+    def __init__(self, connection, object_path, logfile):
+        super(NetplanApplyService, self).__init__(connection, object_path)
+        self._logfile = logfile
+
+    @dbus.service.method(dbus_interface=DOC_IFACE, in_signature="",
+                         out_signature="b")
+    def Apply(self):
+        # log that we were called and always return True
+        with open(self._logfile, "a") as fp:
+            fp.write("Apply called\n")
+        return True
+
+
+def main(argv):
+    logfile = argv[1]
+    dbus.mainloop.glib.DBusGMainLoop(set_as_default=True)
+    main_loop = GLib.MainLoop()
+
+    bus = dbus.SystemBus()
+    # Make sure we quit when the bus shuts down
+    bus.add_signal_receiver(
+        main_loop.quit, signal_name="Disconnected",
+        path="/org/freedesktop/DBus/Local",
+        dbus_interface="org.freedesktop.DBus.Local")
+
+    NetplanApplyService(bus, OBJECT_PATH, logfile)
+
+    # Allow other services to assume our bus name
+    dbus.service.BusName(
+        BUS_NAME, bus, allow_replacement=True, replace_existing=True,
+        do_not_queue=True)
+
+    main_loop.run()
+
+
+if __name__ == '__main__':
+    sys.exit(main(sys.argv))
--- a/tests/main/netplan-apply/fake-netplan-apply-service.py
+++ b/tests/main/netplan-apply/fake-netplan-apply-service.py
@@ -12,7 +12,7 @@ DOC_IFACE = "io.netplan.Netplan"
 
 class NetplanApplyService(dbus.service.Object):
     def __init__(self, connection, object_path, logfile):
-        super(NetplanApplyService, self).__init__(connection, object_path)
+        super().__init__(connection, object_path)
         self._logfile = logfile
 
     @dbus.service.method(dbus_interface=DOC_IFACE, in_signature="",
--- a/tests/main/netplan-apply/task.yaml
+++ b/tests/main/netplan-apply/task.yaml
@@ -15,42 +15,33 @@ prepare: |
     . "$TESTSLIB"/snaps.sh
     snap install test-snapd-netplan-apply --edge
 
-    echo "temporarily installing netplan D-Bus service until SRU is done..."
-    # for now, until netplan's D-Bus has been SRU'd to the netplan.io debian 
-    # package, we will just manually install the netplan D-Bus service units
-    # into the machine from the snap and install netplan.io from the archive
-    apt update
-
-    # the netplan package was renamed nplan -> netplan.io in bionic
-    case "$SPREAD_SYSTEM" in
-        ubuntu-16*)
-            NETPLAN_APT_PKG=nplan;;
-        *)
-            NETPLAN_APT_PKG=netplan.io;;
-    esac
-
-    apt install -y $NETPLAN_APT_PKG
-
-    NETPLAN_SNAP=/snap/test-snapd-netplan-apply/current
-
-    mkdir -p /lib/netplan
-    cp $NETPLAN_SNAP/lib/netplan/netplan-dbus \
-        /lib/netplan/netplan-dbus
-    chmod +x /lib/netplan/netplan-dbus
-    mkdir -p /usr/share/dbus-1/system-services
-    cp $NETPLAN_SNAP/share/dbus-1/system-services/io.netplan.Netplan.service \
-        /usr/share/dbus-1/system-services/io.netplan.Netplan.service
-    mkdir -p /usr/share/dbus-1/system.d
-    cp $NETPLAN_SNAP/share/dbus-1/system.d/io.netplan.Netplan.conf \
-        /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
-
-restore: |
-    echo "uninstalling netplan D-Bus service"
-    rm -f /usr/lib/netplan/netplan-dbus
+    # ensure that the service file doesn't exist before running
     rm -f /usr/share/dbus-1/system-services/io.netplan.Netplan.service
     rm -f /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
 
+restore: |
+    echo "uninstalling netplan D-Bus service"
+    rm -f /usr/share/dbus-1/system-services/io.netplan.Netplan.service
+    rm -f /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
+    snap remove test-snapd-netplan-apply
+
 execute: |
+    # install the dbus configuration file for our fake netplan system dbus service
+    mkdir -p /usr/share/dbus-1/system.d 
+    mkdir -p /usr/share/dbus-1/system-services
+    cp $(pwd)/io.netplan.Netplan.conf /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
+
+    # install the fake netplan system dbus service
+    echo "Make the netplan service activatable"
+    cat << EOF > /usr/share/dbus-1/system-services/io.netplan.Netplan.service
+    [D-BUS Service]
+    Name=io.netplan.Netplan
+    Exec=$(pwd)/fake-netplan-apply-service.py $(pwd)/dbus-netplan-apply.log
+    User=root
+    EOF
+
+    touch dbus-netplan-apply.log
+
     echo "The interface is disconnected by default"
     snap connections test-snapd-netplan-apply | MATCH 'network-setup-control +test-snapd-netplan-apply:network-setup-control +- +-'
 
@@ -60,6 +51,9 @@ execute: |
         exit 1
     fi
 
+    echo "The D-Bus service was not activated"
+    not MATCH "Apply called" < dbus-netplan-apply.log
+
     echo "When the interface is connected"
     snap connect test-snapd-netplan-apply:network-setup-control
 
@@ -68,3 +62,6 @@ execute: |
         echo "Unexpected error running netplan apply"
         exit 1
     fi
+
+    echo "And the D-Bus service was activated"
+    MATCH "Apply called" < dbus-netplan-apply.log
--- a/tests/main/netplan-apply/task.yaml
+++ b/tests/main/netplan-apply/task.yaml
@@ -39,6 +39,7 @@ prepare: |
     Name=$NETPLAN
     Exec=$(pwd)/fake-netplan-apply-service.py $(pwd)/dbus-netplan-apply.log
     User=root
+    AssumedAppArmorLabel=unconfined
     EOF
 
     touch dbus-netplan-apply.log
--- a/tests/main/netplan-apply/task.yaml
+++ b/tests/main/netplan-apply/task.yaml
@@ -23,16 +23,6 @@ prepare: |
         fi      
     done
 
-restore: |
-    # restore the dbus service file and policy config file if the backup 
-    # exists
-    for f in system-services/io.netplan.Netplan.service system.d/io.netplan.Netplan.conf; do
-        if [ -f /usr/share/dbus-1/$f.backup ]; then
-            mv /usr/share/dbus-1/$f.backup /usr/share/dbus-1/$f
-        fi      
-    done
-
-execute: |
     # install the dbus configuration file for our fake netplan system dbus service
     mkdir -p /usr/share/dbus-1/system.d 
     mkdir -p /usr/share/dbus-1/system-services
@@ -49,6 +39,16 @@ execute: |
 
     touch dbus-netplan-apply.log
 
+restore: |
+    # restore the dbus service file and policy config file if the backup 
+    # exists
+    for f in system-services/io.netplan.Netplan.service system.d/io.netplan.Netplan.conf; do
+        if [ -f /usr/share/dbus-1/$f.backup ]; then
+            mv /usr/share/dbus-1/$f.backup /usr/share/dbus-1/$f
+        fi      
+    done
+
+execute: |
     echo "The interface is disconnected by default"
     snap connections test-snapd-netplan-apply | MATCH 'network-setup-control +test-snapd-netplan-apply:network-setup-control +- +-'
 
--- a/tests/main/netplan-apply/task.yaml
+++ b/tests/main/netplan-apply/task.yaml
@@ -3,6 +3,9 @@ summary: Ensure that netplan apply works with network-setup-control
 details: |
     Netplan apply is used to apply network configuration to the system
 
+environment:
+    NETPLAN: io.netplan.Netplan
+
 # run on all classic ubuntu LTS and current dev systems 16.04+
 systems: 
     - ubuntu-16*
@@ -17,22 +20,23 @@ prepare: |
 
     # backup the dbus service file and policy config if they exist before 
     # executing
-    for f in system-services/io.netplan.Netplan.service system.d/io.netplan.Netplan.conf; do
+    for f in system-services/$NETPLAN.service system.d/$NETPLAN.conf; do
         if [ -f /usr/share/dbus-1/$f ]; then
             mv /usr/share/dbus-1/$f /usr/share/dbus-1/$f.backup
         fi      
     done
 
-    # install the dbus configuration file for our fake netplan system dbus service
+    # install the dbus policy config file and service unit for our fake netplan
+    # system dbus service
+    echo "Install the netplan D-Bus activatable system service"
     mkdir -p /usr/share/dbus-1/system.d 
     mkdir -p /usr/share/dbus-1/system-services
-    cp io.netplan.Netplan.conf /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
-
-    # install the fake netplan system dbus service
-    echo "Make the netplan service activatable"
-    cat << EOF > /usr/share/dbus-1/system-services/io.netplan.Netplan.service
+    cp $NETPLAN.conf /usr/share/dbus-1/system.d/$NETPLAN.conf
+    # generate the service file here so that we can referece the python file and
+    # the log file in this directory
+    cat << EOF > /usr/share/dbus-1/system-services/$NETPLAN.service
     [D-BUS Service]
-    Name=io.netplan.Netplan
+    Name=$NETPLAN
     Exec=$(pwd)/fake-netplan-apply-service.py $(pwd)/dbus-netplan-apply.log
     User=root
     EOF
@@ -40,9 +44,16 @@ prepare: |
     touch dbus-netplan-apply.log
 
 restore: |
-    # restore the dbus service file and policy config file if the backup 
-    # exists
-    for f in system-services/io.netplan.Netplan.service system.d/io.netplan.Netplan.conf; do
+    # kill the dbus service if it is running 
+    set +e
+    netplanDBusSvcPID=$(pgrep --full fake-netplan-apply-service.py)
+    set -e
+    if [ -n "$netplanDBusSvcPID" ]; then
+        kill -9 "$netplanDBusSvcPID"
+    fi
+
+    # restore the dbus service file and policy config file if the backup exists
+    for f in system-services/$NETPLAN.service system.d/$NETPLAN.conf; do
         if [ -f /usr/share/dbus-1/$f.backup ]; then
             mv /usr/share/dbus-1/$f.backup /usr/share/dbus-1/$f
         fi      
--- a/tests/main/netplan-apply/task.yaml
+++ b/tests/main/netplan-apply/task.yaml
@@ -29,7 +29,7 @@ execute: |
     # install the dbus configuration file for our fake netplan system dbus service
     mkdir -p /usr/share/dbus-1/system.d 
     mkdir -p /usr/share/dbus-1/system-services
-    cp $(pwd)/io.netplan.Netplan.conf /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
+    cp "$(pwd)/io.netplan.Netplan.conf" /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
 
     # install the fake netplan system dbus service
     echo "Make the netplan service activatable"
--- a/tests/main/netplan-apply/task.yaml
+++ b/tests/main/netplan-apply/task.yaml
@@ -15,21 +15,28 @@ prepare: |
     . "$TESTSLIB"/snaps.sh
     snap install test-snapd-netplan-apply --edge
 
-    # ensure that the service file doesn't exist before running
-    rm -f /usr/share/dbus-1/system-services/io.netplan.Netplan.service
-    rm -f /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
+    # backup the dbus service file and policy config if they exist before 
+    # executing
+    for f in system-services/io.netplan.Netplan.service system.d/io.netplan.Netplan.conf; do
+        if [ -f /usr/share/dbus-1/$f ]; then
+            mv /usr/share/dbus-1/$f /usr/share/dbus-1/$f.backup
+        fi      
+    done
 
 restore: |
-    echo "uninstalling netplan D-Bus service"
-    rm -f /usr/share/dbus-1/system-services/io.netplan.Netplan.service
-    rm -f /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
-    snap remove test-snapd-netplan-apply
+    # restore the dbus service file and policy config file if the backup 
+    # exists
+    for f in system-services/io.netplan.Netplan.service system.d/io.netplan.Netplan.conf; do
+        if [ -f /usr/share/dbus-1/$f.backup ]; then
+            mv /usr/share/dbus-1/$f.backup /usr/share/dbus-1/$f
+        fi      
+    done
 
 execute: |
     # install the dbus configuration file for our fake netplan system dbus service
     mkdir -p /usr/share/dbus-1/system.d 
     mkdir -p /usr/share/dbus-1/system-services
-    cp "$(pwd)/io.netplan.Netplan.conf" /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
+    cp io.netplan.Netplan.conf /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
 
     # install the fake netplan system dbus service
     echo "Make the netplan service activatable"
--- a/tests/main/netplan-apply/task.yaml
+++ b/tests/main/netplan-apply/task.yaml
@@ -46,11 +46,12 @@ prepare: |
 restore: |
     # kill the dbus service if it is running 
     set +e
-    netplanDBusSvcPID=$(pgrep --full fake-netplan-apply-service.py)
-    set -e
-    if [ -n "$netplanDBusSvcPID" ]; then
-        kill -9 "$netplanDBusSvcPID"
+    if [ -n "$(pgrep --full fake-netplan-apply-service.py)" ]; then
+        for pid in $(pgrep --full fake-netplan-apply-service.py); do
+            kill -9 "$pid"
+        done
     fi
+    set -e
 
     # restore the dbus service file and policy config file if the backup exists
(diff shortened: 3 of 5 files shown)


interfaces/builtin/network_setup_control.go
```diff
@@ -45,6 +45,18 @@ const networkSetupControlConnectedPlugAppArmor = `
...
+
+capability net_admin,
+
+#include <abstractions/dbus-session-strict>
+
+# Allow use of NetPlan Apply API, used to apply network configuration
+dbus (send)
+    bus=system
+    interface=io.netplan.Netplan
+	path=/io/netplan/Netplan
```
zyga: The spacing here is a bit odd, tabs-vs-spaces?

interfaces/builtin/network_setup_control.go
```diff
@@ -45,6 +45,18 @@ const networkSetupControlConnectedPlugAppArmor = `
...
+	member=Apply,
+
```
zyga: Do you need to provide `peer=unconfined` here? I recall it matters when the method activates the other side.

tests/main/netplan-apply/task.yaml
```diff
@@ -0,0 +1,73 @@
+summary: Ensure that netplan apply works with network-setup-control
+
+details: |
+    Netplan apply is used to apply network configuration to the system
+
+# run on all classic ubuntu LTS and current dev systems 16.04+
```
zyga: Can we inverse the selection, is there a place where it does not work?

tests/main/netplan-apply/task.yaml
```diff
@@ -0,0 +1,73 @@
...
+systems: 
+    - ubuntu-16*
+    - ubuntu-18*
+    - ubuntu-19*
+    - ubuntu-20*
+
+debug: |
```
zyga: Leftover?

tests/main/netplan-apply/task.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+restore: |
+    echo "uninstalling netplan D-Bus service"
+    rm -f /usr/lib/netplan/netplan-dbus
+    rm -f /usr/share/dbus-1/system-services/io.netplan.Netplan.service
+    rm -f /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
+
+execute: |
+    echo "The interface is disconnected by default"
+    snap interfaces -i network-setup-control | MATCH -- '^- +test-snapd-netplan-apply:network-setup-control'
```
bboozzoo: New tests should probably use `snap connection`:
```
snap connections test-snapd-netplan-apply | MATCH 'network-setup-control +test-snapd-netplan-apply:network-setup-control +- +-
```

tests/main/netplan-apply/task.yaml
```diff
@@ -0,0 +1,73 @@
...
+    echo "temporarily installing netplan D-Bus service until SRU is done..."
+    # for now, until netplan's D-Bus has been SRU'd to the netplan.io debian 
+    # package, we will just manually install the netplan D-Bus service units
+    # into the machine from the snap and install netplan.io from the archive
+    apt update
+
+    # the netplan package was renamed nplan -> netplan.io in bionic
+    case "$SPREAD_SYSTEM" in
+        ubuntu-16*)
+            NETPLAN_APT_PKG=nplan;;
```
bboozzoo: s/nplan/netplan/ ?

tests/main/netplan-apply/task.yaml
```diff
@@ -0,0 +1,73 @@
...
+        *)
+            NETPLAN_APT_PKG=netplan.io;;
+    esac
+
+    apt install -y $NETPLAN_APT_PKG
```
bboozzoo: I think the typical way to test this is to write a little dbus service in python and run it for the test via systemd-run. Take a look at tests/main/document-portal-activation for examples.

Then `test-snapd-netplan-apply` could a bit of Python code that calls `io.netplan.Netplan.Apply()`, which fails predictably until we connect the interface.

tests/main/netplan-apply/task.yaml
zyga: I booted xenial in multipass and:
```
multipass@venerated-newt:~$ apt-cache show nplan
N: Can't select versions from package 'nplan' as it is purely virtual
N: No packages found
multipass@venerated-newt:~$ apt-cache show netplan
N: Unable to locate package netplan
E: No packages found
multipass@venerated-newt:~$ cat /etc/os-release 
NAME="Ubuntu"
VERSION="16.04.6 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.6 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```

Something is off here?

bboozzoo: `netplan.io` then?

zyga: Ah, sorry, I didn't have the cache at all:
```

multipass@venerated-newt:~$ apt-cache show netplan
Package: netplan
Priority: optional
Section: universe/misc
Installed-Size: 155
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: Petter Reinholdtsen <pere@debian.org>
Architecture: amd64
Source: plan
Version: 1.10.1-5
Replaces: plan (<< 1.7.1-1)
Depends: adduser (>= 3.21), libc6 (>= 2.15)
Suggests: plan
Conflicts: plan (<< 1.7.1-1)
Filename: pool/universe/p/plan/netplan_1.10.1-5_amd64.deb
Size: 29448
MD5sum: c7330e5b7d6a3faf47252a045ee2ad3b
SHA1: 23a20803bb8cc67aa8e025c58e71b4456ed4b157
SHA256: c5dbdc212adea7545709e2f7c1b0e4b6c34c6ef4956328876a8cc0077f204caa
Description-en: network server for `plan'
 Plan is a schedule planner based on X/Motif.
 .
 Netplan adds to plan multiuser capability using an IP server.
 .
 WARNING: the best level of authentication offered by netplan in
 this version is identd.  That's quite weak, so watch the manpage and
 tune the config carefully.
Description-md5: f65dc3e49b68b13ca3e8f766701f07c2
Homepage: http://www.bitrot.de/plan.html
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Origin: Ubuntu

multipass@venerated-newt:~$ apt-cache show nplan
Package: nplan
Architecture: amd64
Version: 0.32~16.04.6
Multi-Arch: foreign
Priority: standard
Section: universe/net
Origin: Ubuntu
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Installed-Size: 130
Depends: libc6 (>= 2.4), libglib2.0-0 (>= 2.16.0), libuuid1 (>= 2.16), libyaml-0-2, systemd (>= 229-4ubuntu20), python3, python3-yaml
Suggests: network-manager | wpasupplicant
Conflicts: netplan
Breaks: network-manager (<< 1.2.2-0ubuntu4~)
Filename: pool/universe/n/nplan/nplan_0.32~16.04.6_amd64.deb
Size: 41868
MD5sum: 4c43c1c1b6ad10d4e3293ff2a1a67b49
SHA1: 65c7a0344df8e0c566b3d5c5a407529913fafd77
SHA256: 46198816f72a65aaf6f24f853fadf06a978c97d3e961b0b2ea00acb136acd540
Description-en: YAML network configuration abstraction for various backends
 netplan reads YAML network configuration files which are written
 by administrators, installers, cloud image instantiations, or other OS
 deployments. During early boot it then generates backend specific
 configuration files in /run to hand off control of devices to a particular
 networking daemon.
 .
 Currently supported backends are networkd and NetworkManager.
Description-md5: 171990596b9adbbde53590d783767a6d
```

anonymouse64: Right as @zyga discovered, netplan in xenial is something else, we want `nplan` in xenial and `netplan.io` in bionic and later.

tests/main/netplan-apply/task.yaml
anonymouse64: It does not work on ubuntu 14.04 and all non-ubuntu systems, but I wanted to avoid having to enumerate all non-ubuntu systems in case another distro gets added because then this test will fail there and someone will have to go and add that distro to the blacklist here.

anonymouse64: As such, I think it's preferable to just enumerate the systems it does work on, but if you feel strongly about it I can invert the selection

interfaces/builtin/network_setup_control.go
anonymouse64: yes I will fix this up

interfaces/builtin/network_setup_control.go
anonymouse64: I thought that would be needed as well, but strangely enough the D-Bus rule doesn't work with `peer=(label=unconfined)`. If I had to guess I think it's because the "other side" in this case isn't the unconfined daemon running netplan apply but rather the system bus. It's very possible it's something else however.

tests/main/netplan-apply/task.yaml
anonymouse64: I was trying to avoid writing artificial tests which just test the policy, because it's important here not that specifically this D-Bus rule works, but that the total operation of `netplan apply` actually works. While your suggestion would probably be a good enough test, there's possibilities that it could be a bug in the test showing a false positive and the real netplan doesn't work but our artificial test does. It also would inform us more quickly of changes/bugs in upstream netplan that would break snap applications.

tests/main/netplan-apply/task.yaml
anonymouse64: Thanks I'll use that instead

tests/main/netplan-apply/task.yaml
anonymouse64: Removed

tests/main/netplan-apply/task.yaml
bboozzoo: I agree that it probably is ok to use the real netplan on the caller side. However, I think that the daemon side, should rather be a mocked service in case netplan (the daemon) does something awkward and we lose connectivity with the node.

bboozzoo:
@sergiocazzolato http2 protocol error popped up here:
```
Aug 07 00:41:28 aug070019-732258 snapd[684]: logger.go:67: DEBUG: > "GET /api/v1/snaps/download/PMrrV4ml8uWuEUDBT8dSGnKUYbevVhc4_4238.snap HTTP/1.1\r\nHost: api.snapcraft.io\r\nUser-Agent: snapd/1337.2.40 (series 16; classic; testing) ubuntu/16.04 (i386) linux/4.4.0-157-generic\r\nAccept: \r\nSnap-Cdn: cloud-name=\"gce\" region=\"us-east1\" availability-zone=\"us-east1-b\"\r\nX-Device-Authorization: Macaroon root=\"MDAxZWxvY2F0aW9uIGFwaS5zbmFwY3JhZnQuaW8KMDAxZWlkZW50aWZpZXIgZGV2aWNlLXNlc3Npb24KMDA0MGNpZCBhcGkuc25hcGNyYWZ0LmlvfHZhbGlkX3NpbmNlfDIwMTktMDgtMDdUMDA6MjY6MDYuNDAzMzMwCjAwNDZjaWQgYXBpLnNuYXBjcmFmdC5pb3xzZXNzaW9ufGE2YTNmODkxLWMxOWEtNDFhMi1hZDIwLTZlZDRjMTY4ODkxMgowMDliY2lkIGFwaS5zbmFwY3JhZnQuaW98ZGV2aWNlfHsiYXV0aG9yaXR5IjogImdlbmVyaWMiLCAic2VyaWFsIjogIjA0YmI2MjJkLWI0NDktNDUyNS05YWM2LTA5ZTkwNDhmODM1OCIsICJtb2RlbCI6ICJnZW5lcmljLWNsYXNzaWMiLCAiYnJhbmQiOiAiZ2VuZXJpYyJ9CjAwMjBjaWQgYXBpLnNuYXBjcmFmdC5pb3xzdG9yZXwKMDAyZnNpZ25hdHVyZSCwMuNKOEa66tqucchMwpZ_Pv1eSWO5uAzetWSNku-UGgo\"\r\nX-Ubuntu-Architecture: i386\r\nX-Ubuntu-Classic: true\r\nX-Ubuntu-Series: 16\r\nX-Ubuntu-Wire-Protocol: 1\r\nAccept-Encoding: gzip\r\n\r\n"
Aug 07 00:41:29 aug070019-732258 snapd[684]: logger.go:74: DEBUG: < "HTTP/1.1 302 FOUND\r\nContent-Length: 517\r\nContent-Type: text/html; charset=utf-8\r\nDate: Wed, 07 Aug 2019 00:41:29 GMT\r\nLocation: https://fastly.cdn.snapcraft.io/download-origin/fastly/PMrrV4ml8uWuEUDBT8dSGnKUYbevVhc4_4238.snap?token=1565150400_38982e5c841e4de27d9b13a4ac4b5b5bd4f8cdd0\r\nServer: gunicorn/19.7.1\r\nSnap-Store-Version: 15\r\nX-Request-Id: 23E5411FA5E80A325CD101BB5D4A1E3830FF8200\r\nX-Vcs-Revision: c0e0e44\r\nX-View-Name: snapdevicegw.webapi_download.snap_download\r\n\r\n"
Aug 07 00:41:29 aug070019-732258 snapd[684]: logger.go:67: DEBUG: > "GET /download-origin/fastly/PMrrV4ml8uWuEUDBT8dSGnKUYbevVhc4_4238.snap?token=1565150400_38982e5c841e4de27d9b13a4ac4b5b5bd4f8cdd0 HTTP/1.1\r\nHost: fastly.cdn.snapcraft.io\r\nUser-Agent: snapd/1337.2.40 (series 16; classic; testing) ubuntu/16.04 (i386) linux/4.4.0-157-generic\r\nAccept: \r\nReferer: https://api.snapcraft.io/api/v1/snaps/download/PMrrV4ml8uWuEUDBT8dSGnKUYbevVhc4_4238.snap\r\nSnap-Cdn: cloud-name=\"gce\" region=\"us-east1\" availability-zone=\"us-east1-b\"\r\nX-Device-Authorization: Macaroon root=\"MDAxZWxvY2F0aW9uIGFwaS5zbmFwY3JhZnQuaW8KMDAxZWlkZW50aWZpZXIgZGV2aWNlLXNlc3Npb24KMDA0MGNpZCBhcGkuc25hcGNyYWZ0LmlvfHZhbGlkX3NpbmNlfDIwMTktMDgtMDdUMDA6MjY6MDYuNDAzMzMwCjAwNDZjaWQgYXBpLnNuYXBjcmFmdC5pb3xzZXNzaW9ufGE2YTNmODkxLWMxOWEtNDFhMi1hZDIwLTZlZDRjMTY4ODkxMgowMDliY2lkIGFwaS5zbmFwY3JhZnQuaW98ZGV2aWNlfHsiYXV0aG9yaXR5IjogImdlbmVyaWMiLCAic2VyaWFsIjogIjA0YmI2MjJkLWI0NDktNDUyNS05YWM2LTA5ZTkwNDhmODM1OCIsICJtb2RlbCI6ICJnZW5lcmljLWNsYXNzaWMiLCAiYnJhbmQiOiAiZ2VuZXJpYyJ9CjAwMjBjaWQgYXBpLnNuYXBjcmFmdC5pb3xzdG9yZXwKMDAyZnNpZ25hdHVyZSCwMuNKOEa66tqucchMwpZ_Pv1eSWO5uAzetWSNku-UGgo\"\r\nX-Ubuntu-Architecture: i386\r\nX-Ubuntu-Classic: true\r\nX-Ubuntu-Series: 16\r\nX-Ubuntu-Wire-Protocol: 1\r\nAccept-Encoding: gzip\r\n\r\n"
Aug 07 00:41:29 aug070019-732258 snapd[684]: logger.go:74: DEBUG: < "HTTP/2.0 200 OK\r\nContent-Length: 21356544\r\nAccept-Ranges: bytes\r\nAccept-Ranges: bytes\r\nAge: 0\r\nCache-Control: max-age=86400\r\nContent-Disposition: attachment; filename=PMrrV4ml8uWuEUDBT8dSGnKUYbevVhc4_4238.snap\r\nContent-Type: application/octet-stream\r\nDate: Wed, 07 Aug 2019 00:41:29 GMT\r\nEtag: cfdd32db-a047-4631-9e38-d5da65f60c50\r\nLast-Modified: Wed, 07 Aug 2019 00:35:16 GMT\r\nServer: TwistedWeb/14.0.2\r\nStrict-Transport-Security: max-age=2592000\r\nVary: Accept-Encoding\r\nVia: 1.1 varnish\r\nVia: 1.1 varnish\r\nX-Bzr-Revision-Number: 7464\r\nX-Cache: MISS, MISS\r\nX-Cache-Hits: 0, 0\r\nX-Request-Id: XUocw38AAQEAACli3@MAAABo1\r\nX-Served-By: cache-lcy19245-LCY, cache-dca17753-DCA\r\nX-Timer: S1565138489.235017,VS0,VE111\r\n\r\n"
Aug 07 00:41:29 aug070019-732258 snapd[684]: retry.go:136: DEBUG: Not retrying: http.http2StreamError{StreamID:0x1, Code:0x1, Cause:error(nil)}
Aug 07 00:41:29 aug070019-732258 snapd[684]: store.go:1398: DEBUG: download of "https://api.snapcraft.io/api/v1/snaps/download/PMrrV4ml8uWuEUDBT8dSGnKUYbevVhc4_4238.snap" failed: http.http2StreamError{StreamID:0x1, Code:0x1, Cause:error(nil)}
Aug 07 00:41:29 aug070019-732258 snapd[684]: task.go:337: DEBUG: 2019-08-07T00:41:29Z ERROR stream error: stream ID 1; PROTOCOL_ERROR
+ case "$SPREAD_SYSTEM" in
```

I tried to repeat the request via curl, bit the token seems expired already. Perhaps you can ping the store team to find out what happened around the time frame from the log.

tests/main/netplan-apply/task.yaml
anonymouse64: Okay, that makes more sense. I will adjust it.

anonymouse64: Okay, this has been implemented

tests/main/netplan-apply/task.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+    # ensure that the service file doesn't exist before running
+    rm -f /usr/share/dbus-1/system-services/io.netplan.Netplan.service
+    rm -f /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
+
+restore: |
+    echo "uninstalling netplan D-Bus service"
+    rm -f /usr/share/dbus-1/system-services/io.netplan.Netplan.service
+    rm -f /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
+    snap remove test-snapd-netplan-apply
```
sergiocazzolato: snaps are automatically removed during reset, not needed to make it on restore

anonymouse64: ah okay I'll remove this then

tests/main/netplan-apply/task.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - ubuntu-19*
+    - ubuntu-20*
+
+prepare: |
+    #shellcheck source=tests/lib/snaps.sh
+    . "$TESTSLIB"/snaps.sh
+    snap install test-snapd-netplan-apply --edge
+
+    # ensure that the service file doesn't exist before running
+    rm -f /usr/share/dbus-1/system-services/io.netplan.Netplan.service
```
sergiocazzolato: Perhasps you could rename/backup these files and restore them during the restore, so you make sure the configuration will be the same than before the test is executed.

tests/main/netplan-apply/task.yaml
```diff
@@ -0,0 +1,67 @@
...
+    echo "uninstalling netplan D-Bus service"
+    rm -f /usr/share/dbus-1/system-services/io.netplan.Netplan.service
+    rm -f /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
+    snap remove test-snapd-netplan-apply
+
+execute: |
+    # install the dbus configuration file for our fake netplan system dbus service
+    mkdir -p /usr/share/dbus-1/system.d 
+    mkdir -p /usr/share/dbus-1/system-services
+    cp $(pwd)/io.netplan.Netplan.conf /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
```
sergiocazzolato: just:
cp io.netplan.Netplan.conf
or 
cp "$PWD/io.netplan.Netplan.conf"

anonymouse64: Fixed

tests/main/netplan-apply/task.yaml
anonymouse64: Done

tests/main/netplan-apply/task.yaml
```diff
@@ -0,0 +1,74 @@
...
+    mkdir -p /usr/share/dbus-1/system.d 
+    mkdir -p /usr/share/dbus-1/system-services
+    cp io.netplan.Netplan.conf /usr/share/dbus-1/system.d/io.netplan.Netplan.conf
+
+    # install the fake netplan system dbus service
+    echo "Make the netplan service activatable"
+    cat << EOF > /usr/share/dbus-1/system-services/io.netplan.Netplan.service
+    [D-BUS Service]
+    Name=io.netplan.Netplan
+    Exec=$(pwd)/fake-netplan-apply-service.py $(pwd)/dbus-netplan-apply.log
```
zyga: We may have issues with this on SELinux systems, due to tag based permissions. At the same time I don't have a workaround I can present.

tests/main/netplan-apply/task.yaml
```diff
@@ -0,0 +1,74 @@
...
+    User=root
+    EOF
+
+    touch dbus-netplan-apply.log
```
zyga: Do you need to touch it? The python side should be enough to create the file.

tests/main/netplan-apply/task.yaml
```diff
@@ -0,0 +1,74 @@
...
+    # restore the dbus service file and policy config file if the backup 
+    # exists
+    for f in system-services/io.netplan.Netplan.service system.d/io.netplan.Netplan.conf; do
+        if [ -f /usr/share/dbus-1/$f.backup ]; then
+            mv /usr/share/dbus-1/$f.backup /usr/share/dbus-1/$f
+        fi      
+    done
+
+execute: |
+    # install the dbus configuration file for our fake netplan system dbus service
```
zyga: This feels like prepare section, all the way until `snap connections`.

interfaces/builtin/network_setup_control.go
zyga: CC @jdstrand -- correct me if I'm wrong, do we need to duplicate this rule for peer=unconfined or is this good as-is?

tests/main/netplan-apply/task.yaml
anonymouse64: Okay, moved to prepare section

tests/main/netplan-apply/task.yaml
anonymouse64: Yes, otherwise I would need to adjust the bash below to be slightly more complicated because in the first case we test that the service wasn't activated when called without network-setup-control, so we need the file to be there for us to check the content of it. 

If you prefer I could switch to checking if the file exists and then also checking that it doesn't contain "Apply called" and then wouldn't need the `touch`.

tests/main/netplan-apply/fake-netplan-apply-service.py
```diff
@@ -0,0 +1,50 @@
...
+import sys
+
+BUS_NAME = "io.netplan.Netplan"
+OBJECT_PATH = "/io/netplan/Netplan"
+DOC_IFACE = "io.netplan.Netplan"
+
+
+class NetplanApplyService(dbus.service.Object):
+    def __init__(self, connection, object_path, logfile):
+        super(NetplanApplyService, self).__init__(connection, object_path)
```
bboozzoo:
```suggestion
        super().__init__(connection, object_path)
```

tests/main/netplan-apply/task.yaml
```diff
@@ -0,0 +1,74 @@
...
+    Exec=$(pwd)/fake-netplan-apply-service.py $(pwd)/dbus-netplan-apply.log
+    User=root
+    EOF
+
+    touch dbus-netplan-apply.log
+
+restore: |
+    # restore the dbus service file and policy config file if the backup 
+    # exists
+    for f in system-services/io.netplan.Netplan.service system.d/io.netplan.Netplan.conf; do
```
mvo5: Do we need to stop the dbus servce here too?

anonymouse64: Yes that's a good idea. I changed it to do as such now.

anonymouse64:
For some reason I don't understand, on xenial the first invocation of netplan-apply via the D-Bus fails as expected on the snap side (and on the service side doesn't result in the method being called), but still ends up activating the service on the other side:

```
Aug 13 03:57:37 autopkgtest audit[22854]: AVC apparmor="DENIED" operation="open" profile="snap.test-snapd-netplan-apply.netplan" name="/home/gopath/src/github.com/snapcore/snapd/tests/main/netplan-apply/" pid=22854 comm="python3" requested_mask="r" denied_mask="r" fsuid=0 ouid=12345
Aug 13 03:57:38 autopkgtest audit[22878]: AVC apparmor="DENIED" operation="open" profile="snap.test-snapd-netplan-apply.netplan" name="/proc/1/environ" pid=22878 comm="busctl" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
Aug 13 03:57:38 autopkgtest audit[22878]: AVC apparmor="DENIED" operation="open" profile="snap.test-snapd-netplan-apply.netplan" name="/proc/1/sched" pid=22878 comm="busctl" requested_mask="r" denied_mask="r" fsuid=0 ouid=0
Aug 13 03:57:38 autopkgtest audit[22878]: AVC apparmor="DENIED" operation="capable" profile="snap.test-snapd-netplan-apply.netplan" pid=22878 comm="busctl" capability=12  capname="net_admin"
Aug 13 03:57:38 autopkgtest dbus[586]: [system] Activating service name='io.netplan.Netplan' (using servicehelper)
Aug 13 03:57:38 autopkgtest dbus[586]: [system] Successfully activated service 'io.netplan.Netplan'
Aug 13 03:57:38 autopkgtest audit[586]: USER_AVC pid=586 uid=107 auid=4294967295 ses=4294967295 msg='apparmor="DENIED" operation="dbus_method_call"  bus="system" path="/io/netplan/Netplan" interface="io.netplan.Netplan" member="Apply" mask="send" name="io.netplan.Netplan" pid=22878 label="snap.test-snapd-netplan-apply.netplan" peer_pid=22880 peer_label="unconfined"
                                         exe="/usr/bin/dbus-daemon" sauid=107 hostname=? addr=? terminal=?'
```

This results in having multiple versions of the same python service running:

```
qemu:ubuntu-16.04-64 .../tests/main/netplan-apply# pgrep --full --list-full fake-netplan-apply-service.py
22880 /usr/bin/python3 /home/gopath/src/github.com/snapcore/snapd/tests/main/netplan-apply/fake-netplan-apply-service.py /home/gopath/src/github.com/snapcore/snapd/tests/main/netplan-apply/dbus-netplan-apply.log
22939 /usr/bin/python3 /home/gopath/src/github.com/snapcore/snapd/tests/main/netplan-apply/fake-netplan-apply-service.py /home/gopath/src/github.com/snapcore/snapd/tests/main/netplan-apply/dbus-netplan-apply.log
```

and indeed if I run it again with the interface disconnected the python service is started up again as a separate process. 

Is this a bug with systemd and/or dbus on xenial? I don't see the same behavior on bionic. 

Additionally, should I kill all versions of the service in the restore section, effectively working around that bug or should I somehow figure out how to make the test not create multiple processes? I've chosen to do that now, but just wanted to see if that's okay to work around the issue or if I shouldn't be working around that issue in the tests. 

Finally, I also notice that every invocation of `busctl call` from netplan ends up creating a new process for the io.netplan.Netplan D-Bus service, essentially re-activating it every time. This smells like a bug but I'm not sure how to fix it, would we have to change something in the D-Bus systemd service unit file?

interfaces/builtin/network_setup_control.go
jdstrand: This interface is currently about exposing netplan apply from the host system to the snap in some fashion. netplan is part of classic/core and runs unconfined (since it isn't supplied as a snap). The types of rules we use in this sistuation are: `dbus (send) ... peer=(label=unconfined),` or sometimes `dbus (send) ... peer=(name=somthing, label=unconfined),`. I would prefer that we use `peer=(label=unconfined)` and understand why that isn't working. Adjusting it and trying and looking at the logged policy violations is the first step.

interfaces/builtin/network_setup_control.go
```diff
@@ -45,6 +45,18 @@ const networkSetupControlConnectedPlugAppArmor = `
 
 /run/udev/rules.d/ rw,                 # needed for cloud-init
 /run/udev/rules.d/[0-9]*-netplan-* rw,
+
+capability net_admin,
```
jdstrand: Why is net_admin needed if we are talking to netplan over DBus?

interfaces/builtin/network_setup_control.go
```diff
@@ -45,6 +45,18 @@ const networkSetupControlConnectedPlugAppArmor = `
...
+
+#include <abstractions/dbus-session-strict>
```
jdstrand: This should be `#include <abstractions/dbus-strict>` since below you have specified `bus=system`.

tests/main/netplan-apply/task.yaml
jdstrand: FYI, I also felt inverting it made sense, but will defer to others.

jdstrand:
> One additional question I had was whether we should add policy to allow executing just the `netplan` binary from the core snap? We will need to ship netplan in the core snap to ship the unconfined D-Bus service anyways, so it makes sense to me to allow applications to use the version of netplan provided by the core snap rather than have all of these network configuration agent snaps ship their own duplicated version of netplan (which could also some day get out of sync with the D-Bus service). I think yes, but I left it out of the policy for now...

This sounds reasonable to add so long as there is a code comment stating what is expected to work and what is not (ie, I don't know how many other 'apply'-like commands there are that might cause trouble. If we expose it in the policy, we'll need to understand what is and isn't expected to work). Feel free to do this in a followup PR since you have so many reviews on this as it is.

interfaces/builtin/network_setup_control.go
anonymouse64: It's unclear to me why net_admin ends up being necessary, but this is the denial I get: 

```
Aug 14 16:21:57 ian-disco-desktop audit[21051]: AVC apparmor="DENIED" operation="capable" profile="snap.test-snapd-netplan-apply.netplan" pid=21051 comm="busctl" capability=12  capname="net_admin"
Aug 14 16:21:57 ian-disco-desktop kernel: audit: type=1400 audit(1565817717.617:3584): apparmor="DENIED" operation="capable" profile="snap.test-snapd-netplan-apply.netplan" pid=21051 comm="busctl" capability=12  capname="net_admin"
```

`busctl` is being called by netplan here: https://github.com/CanonicalLtd/netplan/blob/5d22e9d22c4a3724d27b80b0cd9b898ae8f59d2b/netplan/cli/commands/apply.py#L55-L60

Note that after more testing, this ends up not being necessary, but busctl tries for it every time.

interfaces/builtin/network_setup_control.go
anonymouse64: Okay, changed.

interfaces/builtin/network_setup_control.go
anonymouse64: So I debugged this a bit more and discovered that actually for some reason including `peer=(label=unconfined),` is preventing the _initial_ activation, and not subsequent D-Bus calls. If I have not yet activated the netplan apply command over D-Bus, then I get `Access Denied` from busctl when running from inside a snap:

```
$ sudo test-snapd-netplan-apply.netplan apply
Access denied
```

and I see these denials:

```
Aug 14 16:37:38 ian-disco-desktop kernel: audit: type=1400 audit(1565818658.939:125): apparmor="DENIED" operation="capable" profile="snap.test-snapd-netplan-apply.netplan" pid=6739 comm="busctl" capability=12  capname="net_admin"
Aug 14 16:37:38 ian-disco-desktop kernel: audit: type=1107 audit(1565818658.939:126): pid=1572 uid=103 auid=4294967295 ses=4294967295 msg='apparmor="DENIED" operation="dbus_method_call"  bus="system" path="/io/netplan/Netplan" interface="io.netplan.Netplan" member="Apply" mask="send" name="io.netplan.Netplan" pid=6739 label="snap.test-snapd-netplan-apply.netplan"
                                           exe="/usr/bin/dbus-daemon" sauid=103 hostname=? addr=? terminal=?'
```

(note that I removed sys_admin as a capability here, not sure if that's needed for initial activation, it doesn't seem to make a difference)

However, if I activate it outside of confinement, using the exact same busctl command netplan uses internally, it then works from inside confinement:

```
$ sudo busctl call --system io.netplan.Netplan /io/netplan/Netplan io.netplan.Netplan Apply
b true
$ sudo test-snapd-netplan-apply.netplan apply
$ echo $?
0
```

anonymouse64:
> This sounds reasonable to add so long as there is a code comment stating what is expected to work and what is not

Yes I think I will add it to the base template as that will simplify folks using it as they won't have to build and ship their own netplan. But as you say I will do it in a followup PR. Additionally, I have been told that the netplan team is adding additional netplan capabilities to be accessible over D-Bus, so it's conceivable the list of what is expected to work will expand from just `apply` and `generate`.

jdstrand:
> > This sounds reasonable to add so long as there is a code comment stating what is expected to work and what is not
> 
> Yes I think I will add it to the base template as that will simplify folks using it since they won't have to build and ship their own netplan. But as you say I will do it in a followup PR. Additionally, I have been told that the netplan team is adding additional netplan capabilities to be accessible over D-Bus, so it's conceivable the list of what is expected to work will expand from just `apply` and `generate`.

I wouldn't add `netplan` to the base template but instead network-setup-control (and possibly network-setup-observe if it makes sense).

interfaces/builtin/network_setup_control.go
jdstrand: Ah, right. netplan needs to adjust its service file in /usr/share/dbus-1/system-services to have: `AssumedAppArmorLabel=unconfined` (see https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1749000 for related bug).

anonymouse64:
> I wouldn't add netplan to the base template but instead network-setup-control (and possibly network-setup-observe if it makes sense).

Okay, we can discuss more in the PR when I open it after this is merged

interfaces/builtin/network_setup_control.go
anonymouse64: Okay, adding that to the service activate the first time. I will push up here a commit adding the unconfined peer label.

anonymouse64:
For reference, I decided not to propose adding `usr/sbin/netplan` to the rules for the `network-setup-{observe,control}` interfaces because it's a python app and so there are a number of other accesses that would need to be allowed as well, so rather than set that all up, I think it's easier to just have consumer snaps build/bundle their own version of netplan instead of trying to maintain all the permissions that python3 needs to run things from the base snap.

PR #7216, opened 2019-08-06
interfaces/misc: updates for k8s 1.15 (and greengrass test)

jdstrand:
microk8s and upstream k8s has moved to containerd and made various other changes. In updating the policy, discovered that docker-support and kubernetes-support were uses the Getter for UsesPtraceTrace so adjust to use Setter instead and update tests accordingly (adding extra test to greengrass-support in passing).

* interfaces/kubernetes-support: update for k8s 1.15 
* interfaces/docker-support: updates for newer containerd 
* interfaces/docker,kubernetes-support: use SetUsesPtraceTrace()
* interfaces/greengrass-support: adjust testsuite to verify UsesPtraceTrace

### DIFF
--- a/interfaces/builtin/docker_support.go
+++ b/interfaces/builtin/docker_support.go
@@ -625,7 +625,7 @@ func (iface *dockerSupportInterface) AppArmorConnectedPlug(spec *apparmor.Specif
 	if !release.OnClassic {
 		spec.AddSnippet(dockerSupportConnectedPlugAppArmorCore)
 	}
-	spec.UsesPtraceTrace()
+	spec.SetUsesPtraceTrace()
 	return nil
 }
 
--- a/interfaces/builtin/docker_support.go
+++ b/interfaces/builtin/docker_support.go
@@ -46,17 +46,16 @@ const dockerSupportBaseDeclarationSlots = `
 `
 
 const dockerSupportConnectedPlugAppArmorCore = `
-# these accesses are necessary for Ubuntu Core 16 and 18, likely due to the version 
-# of apparmor or the kernel which doesn't resolve the upper layer of an 
-# overlayfs mount correctly
-# the accesses show up as runc trying to read from
+# These accesses are necessary for Ubuntu Core 16 and 18, likely due to the
+# version of apparmor or the kernel which doesn't resolve the upper layer of an
+# overlayfs mount correctly the accesses show up as runc trying to read from
 # /system-data/var/snap/docker/common/var-lib-docker/overlay2/$SHA/diff/
 /system-data/var/snap/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}/common/{,**/} rwl,
 /system-data/var/snap/{@{SNAP_NAME},@{SNAP_INSTANCE_NAME}}/@{SNAP_REVISION}/{,**/} rwl,
 `
 
 const dockerSupportConnectedPlugAppArmor = `
-# Description: allow operating as the Docker daemon. This policy is
+# Description: allow operating as the Docker daemon/containerd. This policy is
 # intentionally not restrictive and is here to help guard against programming
 # errors and not for security confinement. The Docker daemon by design requires
 # extensive access to the system and cannot be effectively confined against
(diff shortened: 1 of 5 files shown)


jdstrand:
FYI @ktsakalozos

jdstrand:
@anonymouse64 - fyi re the docker-support updates.

interfaces/builtin/docker_support.go
```diff
@@ -157,7 +172,8 @@ ptrace (read, trace) peer=docker-default,
 # needed by runc for mitigation of CVE-2019-5736
 # For details see https://bugs.launchpad.net/apparmor/+bug/1820344
 / ix,
-/bin/runc rix,
+/bin/runc ixr,
+/pause ixr,
```
zyga: `/pause` is an interesting path. If you know how it happens to work I think a comment would be useful here.

interfaces/builtin/docker_support.go
```diff
@@ -609,7 +625,7 @@ func (iface *dockerSupportInterface) AppArmorConnectedPlug(spec *apparmor.Specif
 	if !release.OnClassic {
 		spec.AddSnippet(dockerSupportConnectedPlugAppArmorCore)
 	}
-	spec.UsesPtraceTrace()
+	spec.SetUsesPtraceTrace()
```
zyga: Ouch, that's a bug!

interfaces/builtin/kubernetes_support.go
```diff
@@ -67,17 +67,23 @@ profile systemd_run (attach_disconnected,mediate_deleted) {
   # Common rules for kubernetes use of systemd_run
   #include <abstractions/base>
 
-  /usr/bin/systemd-run r,
+  /{,usr/}bin/systemd-run rm,
   owner @{PROC}/@{pid}/stat r,
   owner @{PROC}/@{pid}/environ r,
   @{PROC}/cmdline r,
+  @{PROC}/sys/kernel/osrelease r,
```
zyga: The surface of kernel interfaces is daunting.

interfaces/builtin/kubernetes_support.go
```diff
@@ -197,14 +203,14 @@ func (iface *kubernetesSupportInterface) AppArmorConnectedPlug(spec *apparmor.Sp
 	case "kubelet":
 		systemd_run_extra = kubernetesSupportConnectedPlugAppArmorKubeletSystemdRun
 		snippet += kubernetesSupportConnectedPlugAppArmorKubelet
-		spec.UsesPtraceTrace()
+		spec.SetUsesPtraceTrace()
```
zyga: More bugs?

interfaces/builtin/kubernetes_support.go
```diff
@@ -197,14 +203,14 @@ func (iface *kubernetesSupportInterface) AppArmorConnectedPlug(spec *apparmor.Sp
...
 	case "kubeproxy":
 		snippet += kubernetesSupportConnectedPlugAppArmorKubeproxy
 	default:
 		systemd_run_extra = kubernetesSupportConnectedPlugAppArmorKubeletSystemdRun
 		snippet += kubernetesSupportConnectedPlugAppArmorKubelet
 		snippet += kubernetesSupportConnectedPlugAppArmorKubeproxy
-		spec.UsesPtraceTrace()
+		spec.SetUsesPtraceTrace()
```
zyga: And here?

interfaces/builtin/kubernetes_support.go
```diff
@@ -67,17 +67,23 @@ profile systemd_run (attach_disconnected,mediate_deleted) {
...
 
+  # ptrace 'trace' is coarse and not required for using the systemd private
+  # socket, and while the child profile omits 'capability sys_ptrace', skip
+  # for now since it isn't strictly required.
+  deny ptrace trace peer=unconfined,
   /run/systemd/private rw,
-  /bin/true ixr,
-  ptrace trace peer=unconfined,
+
+  /{,usr/}bin/true ixr,
```
anonymouse64: not a blocker, but out of curiosity, why is this in the policy in kubernetes-support to begin with? shouldn't something as common as  `/bin/true` be in the base apparmor template?

Also FWIW, /bin/true exists in both core and core18, so we could add it to the base template I think

interfaces/builtin/docker_support.go
jdstrand: Yep, here and below. I recall that there was a change in the API for the Getter and the Setter during the review. Guess we missed it :\

interfaces/builtin/kubernetes_support.go
jdstrand: @anonymouse64 - it is already in the base template. This is in the systemd_run which the child profile that doesn't use that template. The apparmor base abstraction could conceivably use it, as you mention, but we would need to carry it here regardless until everyone upgraded to the new apparmor.

interfaces/builtin/kubernetes_support.go
jdstrand: Yep

interfaces/builtin/kubernetes_support.go
jdstrand: Yep

PR #7218, opened 2019-08-07
tests: measure behavior of the device cgroup

zyga:
This test checks the interaction of systemd and snap-confine with
regards to the device cgroup. Since the 2.0 release, snap-confine is
using the device cgroup to constrain access to character and block
devices. Effectively, snap-confine is stealing the process away from
systemd by moving it from the cgroup managed by systemd to one managed
by snapd.

Note that this model can only work in cgroup v1, where a process can at
a given time inhabit multiple separate cgroups.

Not all processes us the snapd-made device cgroup. If a snap is placed
in devmode or uses no interfaces that would toggle udev to tag a device
to add to the cgroup, then no additional cgroup is created or
configured.

This model lead to an overlooked conflict between how systemd and snapd
configure a process that started out without udev tagged devices, and
thus without a snapd-managed cgroup, is connected to an interface that
relies on tagging for precise confinement.

### DIFF
--- a/tests/main/cgroup-devices/task.sh
+++ b/tests/main/cgroup-devices/task.sh
@@ -71,6 +71,8 @@ echo "Verify the constraints imposed by the device cgroup made by snapd"
 # NOTE: the actual permissions may drift over time. We just care about the fact
 # that there *are* some constraints here now and there were none before.
 test 'c 1:3 rwm' = "$(head -n 1 "/sys/fs/cgroup/devices/snap.test-snapd-service.test-snapd-service/devices.list")"
+# The device cgroup made by snapd is, currently, still empty.
+test -z "$(cat "/sys/fs/cgroup/devices/snap.test-snapd-service.test-snapd-service/cgroup.procs")"
 
 echo "Restart the test service"
 # See the comment for the similar code above.
--- a/tests/main/cgroup-devices/task.sh
+++ b/tests/main/cgroup-devices/task.sh
@@ -8,19 +8,30 @@ echo "Extract the PID of the main process tracked by systemd"
 pid=$(systemctl show snap.test-snapd-service.test-snapd-service.service --property=ExecMainPID | cut -d = -f 2)
 
 echo "Extract the device cgroup of the main process"
-effective_device_cgroup=$(grep devices < "/proc/$pid/cgroup" | cut -d : -f 3)
+initial_device_cgroup=$(grep devices < "/proc/$pid/cgroup" | cut -d : -f 3)
+
 # Initially, because there are no udev tags corresponding to this application,
 # snap-confine is not moving the process to a new cgroup. As such the service
-# runs in the cgroup created by systemd.
-test "$effective_device_cgroup" = /system.slice/snap.test-snapd-service.test-snapd-service.service
+# runs in the cgroup created by systemd, which varies across version of systemd.
+case "$initial_device_cgroup" in
+    /)
+        ;;
+    /system.slice)
+        ;;
+    /system.slice/snap.test-snapd-service.test-snapd-service.service)
+        ;;
+    *)
+        echo "Unexpected initial device cgroup: $initial_device_cgroup"
+        exit 1
+esac
 
 echo "Ensure that the claim of the process is consistent with the claim of the cgroup"
 # This is just a sanity check.
-MATCH "$pid" < "/sys/fs/cgroup/devices/$effective_device_cgroup/cgroup.procs"
+MATCH "$pid" < "/sys/fs/cgroup/devices/$initial_device_cgroup/cgroup.procs"
 
 echo "Verify the constraints imposed by the device cgroup made by systemd"
 # This may change over time as it is governed by systemd.
-test 'a *:* rwm' = "$(cat "/sys/fs/cgroup/devices/$effective_device_cgroup/devices.list")"
+test 'a *:* rwm' = "$(cat "/sys/fs/cgroup/devices/$initial_device_cgroup/devices.list")"
 
 echo "Connect the joystick interface"
 snap connect test-snapd-service:joystick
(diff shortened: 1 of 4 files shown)


tests/main/cgroup-devices/test-snapd-service/bin/service
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+rm -f "$SNAP_COMMON/remove-me"
```
sergiocazzolato: Why this snap is defined as part of the test instead of adding that one to tests/lib/snaps ?

tests/main/cgroup-devices/task.yaml
```diff
@@ -0,0 +1,8 @@
+summary: measuring basic properties of device cgroup
+# Ubuntu 14.04 uses a custom build of an older systemd that does not
+# seem to use device cgroups for services. The test is invalid there.
+systems: [-ubuntu-14.04-*]
+execute: ./task.sh
```
sergiocazzolato: Why there is a .sh file instead of adding the body in the execute part of the test?

tests/main/cgroup-devices/task.yaml
```diff
@@ -0,0 +1,8 @@
...
+restore: |
+    rm -f test-snapd-service_1.0_all.snap
```
sergiocazzolato: The generic restore for each task makes a clean up on the test dir and removes all the installed snaps. I thinks these lines are not needed.

sergiocazzolato:
Thanks for adding this test!!! I left a comment and 2 questions inline.

tests/main/cgroup-devices/test-snapd-service/bin/service
zyga: Because that doesn't scale. This snap has single purpose. It is not meant to be reused and modified and extended. I looked at the "simple" service snaps and they were anything but. I just chose not to share.

tests/main/cgroup-devices/task.yaml
zyga: Because it is equivalent except my editor knows what I'm doing then.

tests/main/cgroup-devices/task.yaml
zyga: Then they don't hurt and I can easily re-run this in debug shell. Eventually I'd like to have tests clean up after themselves and not rely on the magic global restore to fix things. It is more costly and doesn't really as reliably as test writers assume (see my work on removing snaps on core tests where they otherwise linger).

chipaca:
@sergiocazzolato were your concerns addressed?

zyga:
This is failing on Amazon, it seems that the version of systemd there is has different behaviour than the one used elsewhere:
```
2019-08-13 10:01:23 Error executing google:amazon-linux-2-64:tests/main/cgroup-devices : 
-----
+ ./task.sh
+ echo 'Install a test service'
Install a test service
+ snap pack test-snapd-service
built: test-snapd-service_1.0_all.snap
+ snap install --dangerous ./test-snapd-service_1.0_all.snap
test-snapd-service 1.0 installed
+ echo 'Extract the PID of the main process tracked by systemd'
Extract the PID of the main process tracked by systemd
++ cut -d = -f 2
++ systemctl show snap.test-snapd-service.test-snapd-service.service --property=ExecMainPID
+ pid=10534
+ echo 'Extract the device cgroup of the main process'
Extract the device cgroup of the main process
++ cut -d : -f 3
++ grep devices
+ effective_device_cgroup=/
+ test / = /system.slice/snap.test-snapd-service.test-snapd-service.service
```

The service was not placed into a device cgroup by systemd.

I'll investigate that now.

zyga:
This should be sorted out now. I added detection that skips the test on older versions of systemd.

zyga:
The version check is wrong, need to try this on the two system when this fails and tweak it.

tests/main/cgroup-devices/task.sh
```diff
@@ -0,0 +1,90 @@
...
+
+echo "Verify the constraints imposed by the device cgroup made by systemd"
+# This may change over time as it is governed by systemd.
+test 'a *:* rwm' = "$(cat "/sys/fs/cgroup/devices/$initial_device_cgroup/devices.list")"
+
+echo "Connect the joystick interface"
+snap connect test-snapd-service:joystick
+
+echo "Refresh the value of the main pid and the effective device cgroup after snap connect"
+# NOTE: As of snapd 2.40 the PID and cgroup are expected to be the same as before.
```
bboozzoo: we could have a check for this too

tests/main/cgroup-devices/task.sh
```diff
@@ -0,0 +1,90 @@
...
+test 'a *:* rwm' = "$(cat "/sys/fs/cgroup/devices/$updated_device_cgroup/devices.list")"
+
+echo "Run /bin/true via snap-confine, so that we create the device cgroup made by snapd"
+snap run --shell test-snapd-service -c /bin/true
+
+echo "Verify the constraints imposed by the device cgroup made by snapd"
+# NOTE: the actual permissions may drift over time. We just care about the fact
+# that there *are* some constraints here now and there were none before.
+test 'c 1:3 rwm' = "$(head -n 1 "/sys/fs/cgroup/devices/snap.test-snapd-service.test-snapd-service/devices.list")"
+
```
bboozzoo: There should be no processes in the group at this time, so we could check that as well.

tests/main/cgroup-devices/task.sh
```diff
@@ -0,0 +1,90 @@
...
+pid=$(systemctl show snap.test-snapd-service.test-snapd-service.service --property=ExecMainPID | cut -d = -f 2)
+
+echo "Extract the device cgroup of the main process"
+initial_device_cgroup=$(grep devices < "/proc/$pid/cgroup" | cut -d : -f 3)
+
+# Initially, because there are no udev tags corresponding to this application,
+# snap-confine is not moving the process to a new cgroup. As such the service
+# runs in the cgroup created by systemd, which varies across version of systemd.
+case "$initial_device_cgroup" in
+    /)
```
bboozzoo: AMZN2?

zyga: Indeed, check the commit message here for details https://github.com/snapcore/snapd/pull/7218/commits/5dd2b2cf39ebe03603c76629e1fe0494b79afaf5

zyga:
@bboozzoo added your suggestions in the last two patches.

PR #7219, opened 2019-08-07
devicestate/firstboot: check for missing bases early

stolowski:
Check for missing bases and error out early in firstboot seeding.


overlord/devicestate/firstboot.go
```diff
@@ -76,6 +76,24 @@ func installSeedSnap(st *state.State, sn *snap.SeedSnap, flags snapstate.Flags,
 	return snapstate.InstallPath(st, &sideInfo, path, "", sn.Channel, flags)
 }
 
+func checkSeedBases(snaps []*snap.Info) error {
```
anonymouse64: it would be good if some of the checks that `snap debug validate-seed` does would share code with this (or vice versa). See https://github.com/mvo5/snappy/blob/master/image/image.go#L792 and https://github.com/snapcore/snapd/pull/7105

image/image.go
```diff
@@ -789,6 +789,32 @@ func copyLocalSnapFile(snapPath, targetDir string, info *snap.Info) (dstPath str
...
+	var errs []error
+	for _, info := range snapInfos {
+		// ensure base is available
+		if info.Base != "" && info.Base != "none" {
+			if _, ok := snapInfos[info.Base]; !ok {
+				errs = append(errs, fmt.Errorf("cannot use snap %q: base %q is missing", info.InstanceName(), info.Base))
+			}
+		}
+		// ensure core is available
+		if info.Base == "" && info.SnapType == snap.TypeApp && info.InstanceName() != "snapd" {
```
zyga: I would split this out into info.EffectiveBase()

overlord/devicestate/firstboot.go
```diff
@@ -238,6 +243,14 @@ func populateStateFromSeedImpl(st *state.State, tm timings.Measurer) ([]*state.T
...
 		infos = append(infos, info)
 		infoToTs[info] = ts
+		allSnapInfos[info.InstanceName()] = info
+	}
+
+	// validate that all snaps have bases
+	errs := image.CheckBasesAndProviders(allSnapInfos)
+	if errs != nil {
+		// only report the first error encountered
+		return nil, errs[0]
```
mvo5: I wonder if we should do something similar to https://github.com/snapcore/snapd/pull/7219/files#diff-da07f373046a0a24e67b18699dc15a70L845 here, i.e. build an error that contains all the individual issues? Maybe we can make the result from "CheckBaseAndProviders" something like: 
```go
type MultiErr struct {
  Header string
  errs []errors
}
func (e MultiErr) Add(errs []error) { ... }
func (e MultiErr) Error() string {
  var buf bytes.Buffer
  for _, err := range e.errs {
	fmt.Fprintf(&buf, "\n- %s", err)
   }
   return fmt.Sprintf("%s:%s", e.Header, buf.Bytes())
}
```
but looking at it it (after I typed it out) it seems like this should probably a separate PR and get a :+1: from @pedrons first :)

stolowski: Right, I was considering this and I'm happy to reconsider. The reason I did it this way is `prepare-image` validates all aspects of the image and reports all errors at once and this is useful summary for the user when assembling an image. However, when the image is already booted I thought it's fine to report just one error which is in-line with error reporting we normally do in snapd.

pedronis: yes, I don't think it's worth the complexity in seeding code itself.

overlord/devicestate/firstboot.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/snapcore/snapd/asserts/snapasserts"
 	"github.com/snapcore/snapd/dirs"
 	"github.com/snapcore/snapd/i18n"
+	"github.com/snapcore/snapd/image"
```
pedronis: I don't think we want to import in this direction though

overlord/devicestate/firstboot.go
pedronis: I misread the issue, I'm fine either way given this will end up in snap/ code.

image/image.go
pedronis: as long as we find at least a couple other places that can use it

overlord/devicestate/firstboot.go
stolowski: Moved CheckBasesAndProviders to snap package (and renamed).

snap/validate.go
```diff
@@ -951,3 +951,43 @@ func ValidateSystemUsernames(info *Info) error {
...
+			if err := plug.Attr("default-provider", &dprovider); err == nil && dprovider != "" {
+				cps = append(cps, dprovider)
+			}
+		}
+	}
+	return cps
+}
+
+// ValidateBasesAndProviders checks that all bases/default-providers are part of the seed
+func ValidateBasesAndProviders(snapInfos map[string]*Info) []error {
```
pedronis: now that it's here, this probably needs some unit tests

stolowski: Added (+ we have the existing tests in image/validate_seed_test.go).

snap/validate.go
```diff
@@ -951,3 +951,43 @@ func ValidateSystemUsernames(info *Info) error {
...
+	var errs []error
+	for _, info := range snapInfos {
+		// ensure base is available
+		if info.Base != "" && info.Base != "none" {
+			if _, ok := snapInfos[info.Base]; !ok {
+				errs = append(errs, fmt.Errorf("cannot use snap %q: base %q is missing", info.InstanceName(), info.Base))
+			}
+		}
+		// ensure core is available
+		if info.Base == "" && info.SnapType == TypeApp && info.InstanceName() != "snapd" {
```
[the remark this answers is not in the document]
stolowski: Yes, we started setting `type:snapd` recently, previously it was `app`.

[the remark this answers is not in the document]
anonymouse64: okay, then this check is still needed. thanks for clarifying

PR #7223, opened 2019-08-08
recovery: update fde-utils

cmatsuoka:
Update ubuntu-core-fde-utils to the latest version, and refactor
the recovery code to use the new API. Also remove unused
recovery code.

### DIFF
--- a/recovery/recovery.go
+++ b/recovery/recovery.go
@@ -31,6 +31,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/chrisccoulson/go-tpm2"
 	"github.com/chrisccoulson/ubuntu-core-fde-utils"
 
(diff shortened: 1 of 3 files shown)


zyga:
@cmatsuoka unit tests seem to fail, required packaging changes missing?

cmatsuoka:
We're building with go 1.9.7 and go-tpm2 uses `strings.Builder` which was introduced in go 1.10. I'll ask @chrisccoulson to change go-tpm2 to be buildable with go 1.9.

xnox:
Given this targets uc20, shouldn't we aim to use newer go? Like 1.12?

cmatsuoka:
We're using different versions in tests, including newer ones, and the 1.9 series.

PR #7224, opened 2019-08-09
xdgopenproxy: update test API to match upstream

zyga:
The upstream github.com/godbus/dbus has changed and our tests no longer compile
because the dbus.BusObject interface has gained additional methods to implement.
This patch adds a dummy implementation of the missing methods.

Fixes: https://bugs.launchpad.net/snappy/+bug/1839498


zyga:
Eh, it seems we need to update the vendorized dependency as well. Let's see if this works.

PR #7225, opened 2019-08-09
tests: don't repeat checks

zyga:
We were repeating journal content checks. I'm not convinced we actually
need to do so, let's see if we can drop that and simplify the related
code.

### DIFF
--- a/tests/lib/journalctl.sh
+++ b/tests/lib/journalctl.sh
@@ -33,23 +33,15 @@ check_journalctl_ready(){
 }
 
 check_journalctl_log(){
-    expression=$1
+    expression="$1"
     shift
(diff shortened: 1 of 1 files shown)


stolowski:
As discussed last Friday, it's probably best to consider landing this after current drama with tests is over and they stabilize, so we don't introduce new unknowns.

mvo5:
Setting to "blocked" per Pawels comment.

zyga:
@sergiocazzolato I'm closing this but please consider taking this patch instead of the approach you took in your PR. I think we don't need the repeated checks anymore and I would rather have that than the change to logging.

zyga:
Aha,  there is something interesting I learned here. Closing for now.

PR #7226, opened 2019-08-09
tests: mountinfo-tool fail if there are no matches

zyga:
This is useful for using mountinfo-tool /path/to/mountpoint
as a quick way to check if something is mounted.

### DIFF
--- a/tests/lib/bin/mountinfo-tool
+++ b/tests/lib/bin/mountinfo-tool
@@ -613,6 +613,8 @@ Known attributes, applicable for both filtering and display.
     parent_id:      identifier of parent mount point
     dev:            major:minor numbers of the mounted device
     root_dir:       subtree of the mounted filesystem exposed at mount_point
+
+The exit code indicates if any mount point matched the query.
     """,
         formatter_class=RawTextHelpFormatter,
     )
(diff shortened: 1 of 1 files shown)


tests/lib/bin/mountinfo-tool
```diff
@@ -749,7 +749,8 @@ Known attributes, applicable for both filtering and display.
         raise SystemExit(
             "--one requires exactly one match, found {}".format(len(entries))
         )
-
+    if len(entries) == 0:
```
stolowski: Could you please add a comment here of what's the use case (i.e. what you explained in PR description). As as it may look like an error case that's missing a proper message.

zyga: Sure, the idea is that it allows you to write scripts like:
```
if mountinfo-tool /some/path; then
   # stuff is mounted, do something
fi
```

I'll add a comment in the code.

stolowski: Yes, sure, sorry for not being clear: I understand this change (PR description is clear), it is just about adding a comment for anyone who looks at this chunk of code later.

tests/lib/bin/mountinfo-tool
```diff
@@ -749,7 +749,8 @@ Known attributes, applicable for both filtering and display.
...
+        raise SystemExit(1)
```
mvo5: How about making the output describe why it fails? Something like `raise SystemExit("no mount entry found for given input")` (or something along the inlines, maybe we should even include the inputs?

zyga: I kind of like the "unix tools don't talk much" approach. I'm not opposed to adding this behind a verbose flag. This PR was motivated by a test cleanup that used `mountinfo-tool` as a test condition so I don't technically need the verbosity here.

PR #7227, opened 2019-08-09
tests: Enable verbose http2

sergiocazzolato:
This is to have extra information about the protocol error downloading snaps

PR #7232, opened 2019-08-12
tests/lib/prepare: temporarily disable HTTP/2 in the tests

bboozzoo:
Because of HTTP2 issues we keep seeing during snap download, disable it to
unblock submissions to master.

Details provided here:
[1]. https://github.com/snapcore/snapd/pull/7227
[2]. https://forum.snapcraft.io/t/snap-download-failures-with-protocol-error/12677

### DIFF
--- a/tests/lib/prepare.sh
+++ b/tests/lib/prepare.sh
@@ -63,7 +63,9 @@ setup_systemd_snapd_overrides() {
 [Unit]
 $START_LIMIT_INTERVAL
 [Service]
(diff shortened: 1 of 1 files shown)


mvo5:
I'm not sure this is the right way forward, I did a quick google search for PROTOCOL_ERROR and it seems like our users are hitting this for real (e.g. https://forum.manjaro.org/t/opera-snap-wont-install/54127).

bboozzoo:
Let's try the retry path first. We can always revisit this one.

PR #7233, opened 2019-08-12
interfaces/firewall-control: add nf_nat_* to kmod plug

alfonsosanchezbeato:
The nf_nat_* kernel modules contain rules to be able to do NAT for some
commonly used protocols.

### DIFF
--- a/interfaces/builtin/firewall_control.go
+++ b/interfaces/builtin/firewall_control.go
@@ -158,6 +158,12 @@ var firewallControlConnectedPlugKmod = []string{
 	"br_netfilter",
 	"ip6table_filter",
 	"iptable_filter",
+	"nf_nat_ftp",
+	"nf_nat_irc",
+	"nf_nat_sip",
+	"nf_nat_tftp",
+	"nf_nat_pptp",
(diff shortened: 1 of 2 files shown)


alfonsosanchezbeato:
> This makes me uncomfortable. I acknowledge that the modules don't autoload, but most systems are not routing firewalls and so loading these by default doesn't make a lot of sense. I would prefer a `snap set` mechanism for core that gadgets/admins/etc could use to configure the device to load these by default. Assuming this mechanism is generalized, it would allow a convenient way to ensure everything is in place for the device's intended purpose.

Maybe having a separate interface would be an option too?

Anyway, although NetworkManager loads these modules by default when creating a shared connection, I do not think we have a concrete use case for this at the moment. So there is no urgency to have a solution for this yet.

zyga:
Do we expect more iteration in this PR or shall we close it while the ideas are explored?

alfonsosanchezbeato:
> Do we expect more iteration in this PR or shall we close it while the ideas are explored?

It is fine for me if you want to close for the moment.

PR #7234, opened 2019-08-12
tests: make sure snapd is built against new gorilla mux

stolowski:
Some time ago a new version of gorilla/mux was introduced in vendor (commit `be4fc4d117c255cadd697a93f9f94e49c708f2c3`), however our current code succesfully compiles against the old version, and then fails in very confusing ways at runtime (experienced recently by @anonymouse64  and me), probably due to some changes in semantics of the mux API. When built against old mux, snapd & snap ops fail complaining about empty (missing) snap name argument.

This PR adds a very simple test that fails to build if one forgets to update his vendor tree - it uses one of the new mux API methods:
`./api_test.go:7355:12: router.Name undefined (type *mux.Router has no field or method Name)`

### DIFF
--- a/daemon/api.go
+++ b/daemon/api.go
@@ -2425,3 +2425,14 @@ func getWarnings(c *Command, r *http.Request, _ *auth.UserState) Response {
 
 	return SyncResponse(ws, nil)
 }
+
+// dummyGorillaMuxVersionCheck is not used and acts as a quick check against old gorilla mux.
+func dummyGorillaMuxVersionCheck() {
+	// this code will not build on older version of gorilla mux (i.e. gorilla mux revision
+	// 0eeaf8392f5b04950925b8a69fe70f110fa7cbfc; the new gorilla was introduced in snapd
+	// with commit be4fc4d117c255cadd697a93f9f94e49c708f2c3) and it is here to ensure
+	// snapd is built with refreshed vendor (i.e. new gorilla). snapd would happily build
+	// against old gorilla mux but would fail in confusing ways at runtime.
+	router := mux.NewRouter()
+	_ = router.Name("foo")
+}
--- a/daemon/api_test.go
+++ b/daemon/api_test.go
@@ -41,6 +41,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/gorilla/mux"
 	"golang.org/x/crypto/sha3"
 	"gopkg.in/check.v1"
 	"gopkg.in/tomb.v2"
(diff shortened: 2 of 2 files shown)


stolowski:
> This looks good to me and matches my findings with [#7100 (comment)](https://github.com/snapcore/snapd/pull/7100#issuecomment-520037041).
> 
> However, I agree with @mvo5 that this should be a runtime check that prevents even building a snapd with too old of a version of mux, because sometimes folks doing development might not even be running unit tests before trying to use their locally built version of snapd (i.e. a developer who is working on writing unit tests 😉 and knows they are broken for some reason).

I can move this code to actual snapd, somewhere around api.go, but I wanted to avoid having no-op code in the real codebase. I think failing unit tests should be enough of an indication of a problem, and running unit tests is one of the first things to check if something doesn't work as expected when hacking on the code.

anonymouse64:
I think while yes, more seasoned and experienced developers may know to check out the unit tests when something is fishy, it's not a completely obvious place to go looking when you have issues with a successfully building binary and in cases like these where the error condition is quite subtle and complicated having somewhere obvious at build-time helps block ever trying to use this still broken version of snapd. I think it's good to try and avoid the [survivor bias](https://en.wikipedia.org/wiki/Survivorship_bias) logic of "I thought to look at the unit tests so therefore other developers would think to look at the unit tests too".

stolowski:
@anonymouse64 fair enough, moved it into api.go, now it would make the build fail (both on test and when building snapd).

daemon/api.go
```diff
@@ -2425,3 +2425,14 @@ func getWarnings(c *Command, r *http.Request, _ *auth.UserState) Response {
...
+
+// dummyGorillaMuxVersionCheck is not used and acts as a quick check against old gorilla mux.
+func dummyGorillaMuxVersionCheck() {
+	// this code will not build on older version of gorilla mux (i.e. gorilla mux revision
+	// 0eeaf8392f5b04950925b8a69fe70f110fa7cbfc; the new gorilla was introduced in snapd
+	// with commit be4fc4d117c255cadd697a93f9f94e49c708f2c3) and it is here to ensure
+	// snapd is built with refreshed vendor (i.e. new gorilla). snapd would happily build
+	// against old gorilla mux but would fail in confusing ways at runtime.
+	router := mux.NewRouter()
+	_ = router.Name("foo")
```
zyga: This fails on Fedora, we must update our dependencies there, it seems:
```
src/github.com/snapcore/snapd/daemon/api.go:2437:12: router.Name undefined (type *mux.Router has no field or method Name)
```

zyga: Ping, we need to do something about this.

stolowski:
I'm inclined to close this PR. The check is mostly for our convienience to detect if local build was made against outdated vendor where both new gorilla and old github.com/gorilla/context are present. FC30 has slightly older gorilla-mux that doesn't pass the API check I proposed, but that's just an artifical check and not a real issue. A new gorilla-mux is present in rawhide, but our use case will not justify a backport to FC30 I think.

stolowski:
Closing per my earlier comment.

PR #7236, opened 2019-08-12
packaging/ubuntu-16.04: provide a way to reset snapd without purging

mwhudson:
There is sometimes a need to "reset" snapd without going to the full
extent of purging the package. Because it's hard for postrm to execute
code outside of itself (because that code will have been removed before
it is been invoked!) it's more practical to have a way to invoke the
reset code that lives in postrm without having the generic debhelper
postrm damage the installation by e.g. removing all the apparmor rules.

That's all a long winded justification for a very simple code change: if
SNAPD_RESET_HACK is set in the environment, exit before getting to the
debhelper-inserted code. Then a reset is as simple as:

SNAPD_RESET_HACK=1 /var/lib/dpkg/info/snapd.postrm purge

### DIFF
--- a/packaging/ubuntu-16.04/snapd.postrm
+++ b/packaging/ubuntu-16.04/snapd.postrm
@@ -126,4 +126,6 @@ if [ "$1" = "purge" ]; then
     rm -rf /var/lib/snapd
 fi
 
+[ -n "${SNAPD_RESET_HACK}" ] && exit 0
+
 #DEBHELPER#


mwhudson:
> I'm -0 on a specific environment variable. Could we use the `snap-mgmt` script and just call that instead of purging the package?

Well if we can get a better way of doing the equivalent work that would be grand (I've been a bit guilty of proposing a solution vs explaining a need here I guess). It basically has to be inlined into the postrm though as I understand it.

zyga:
@mwhudson perhaps I misunderstand but if you are after "wipe snapd state without removing the package" then we can use `snap-mgmt` script for that. If I misunderstood then I'm not going to block this patch as-is, I was just looking for something reusing another existing part.

mwhudson:
Ah OK, so I didn't realize snap-mgmt was a thing :) Is it a tested and supported thing? It does look like just what I need...

bboozzoo:
> Ah OK, so I didn't realize snap-mgmt was a thing :) Is it a tested and supported thing? It does look like just what I need...

Yeah, we use in on Fedora and Arch. The script being part of the package cannot be involved in post remove step, so we do in pre-remove instead.

mvo5:
From what I hear "snap-mgmt" is ok for this use-case. So I will close this PR. Please reopen if I misunderstood.

PR #7241, opened 2019-08-13
tests: remove snap holding core18

zyga:


chipaca:
How is this different from any other snap, that would get cleaned up by the usual state restore?

zyga:
@chipaca this snap is using core18, we tend not to remove core18 because of snaps like this, which makes core18 leak from test to test, which is triggering the mount leak detector I'm trying to land.

zyga:
I think this is redundant in face of https://github.com/snapcore/snapd/pull/7244

PR #7242, opened 2019-08-13
tests: show why sbuild failed

zyga:
The sbuild test seems to fail sometimes, but not due to timeout that
kills the test. When that happens test logs contain absolutely nothing
useful.

Maciej suggested that we can look at the tail of the sbuild logs to do
so. In a practical run invoked with --shell-after there are two log
files:

    snapd_2.40-2_amd64-2019-08-13T09:46:14Z.build
    snapd_2.40-2_amd64.build -> snapd_2.40-2_amd64-2019-08-13T09:46:14Z.build

I don't want to be in the business of extracting that version so I just
take the one with the Z.build suffix.

### DIFF
--- a/tests/main/sbuild/task.yaml
+++ b/tests/main/sbuild/task.yaml
@@ -29,6 +29,10 @@ restore: |
     rm -f /etc/schroot/chroot.d/sid-amd64-sbuild-*
 
 debug: |
+    # There there's a log file and a symbolic link pointing to it.
+    # The non-symlink has a time-stamp and we can match on the "Z" timezone
+    # marker to find it.
+    tail -n 100 *Z.build
     cat <<EOM
     Use release-tools/debian-package-builder to interactively fix build
     issues. The debug shell created there shows the true layout of the source
--- a/tests/main/sbuild/task.yaml
+++ b/tests/main/sbuild/task.yaml
@@ -32,7 +32,7 @@ debug: |
     # There there's a log file and a symbolic link pointing to it.
     # The non-symlink has a time-stamp and we can match on the "Z" timezone
     # marker to find it.
-    tail -n 100 *Z.build
+    test "$(find . -maxdepth 1 -name '*Z.build' | wc -l)" -ge 1 &&  tail -n 100 *Z.build
     cat <<EOM
     Use release-tools/debian-package-builder to interactively fix build
     issues. The debug shell created there shows the true layout of the source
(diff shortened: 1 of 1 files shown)


tests/main/sbuild/task.yaml
```diff
@@ -29,6 +29,10 @@ restore: |
     rm -f /etc/schroot/chroot.d/sid-amd64-sbuild-*
 
 debug: |
+    # There there's a log file and a symbolic link pointing to it.
+    # The non-symlink has a time-stamp and we can match on the "Z" timezone
+    # marker to find it.
+    tail -n 100 *Z.build
```
chipaca: and if there isn't a log file?

zyga: Then the test will fail anyway.

chipaca: yes, but so will the debug, and then you can't `-debug`?

bboozzoo:
```suggestion
    test "$(find . -maxdepth 1 -name '*Z.build' | wc -l)" -ge 1 &&  tail -n 100 *Z.build
```

zyga: Do you mean that this would prevent a `-debug` shell?

tests/main/sbuild/task.yaml
```diff
@@ -29,6 +29,10 @@ restore: |
     rm -f /etc/schroot/chroot.d/sid-amd64-sbuild-*
 
 debug: |
+    # There there's a log file and a symbolic link pointing to it.
+    # The non-symlink has a time-stamp and we can match on the "Z" timezone
+    # marker to find it.
+    test "$(find . -maxdepth 1 -name '*Z.build' | wc -l)" -ge 1 &&  tail -n 100 *Z.build
```
zyga:
```suggestion
    test "$(find . -maxdepth 1 -name '*Z.build' | wc -l)" -ge 1 &&  tail -n 100 ./*Z.build
```

tests/main/sbuild/task.yaml
```diff
@@ -29,6 +29,10 @@ restore: |
     rm -f /etc/schroot/chroot.d/sid-amd64-sbuild-*
 
 debug: |
+    # There there's a log file and a symbolic link pointing to it.
```
anonymouse64:
```suggestion
    # Test that there's a log file and a symbolic link pointing to it.
```

PR #7243, opened 2019-08-13
tests: be less verbose when checking service status

zyga:
Our classic system test reset code is very chatty, as it will attempt,
up to 20 times, to check the status of snapd.service. In each time it
really only cares about a particular property but it doesn't hide the
output from systemctl status. When debugging other issues I found that I
was going through pages of this output so here's a small way to make the
test log shorter.

### DIFF
--- a/tests/lib/reset.sh
+++ b/tests/lib/reset.sh
@@ -20,10 +20,10 @@ reset_classic() {
 
     echo "Ensure the service is active before stopping it"
     retries=20
-    systemctl status snapd.service snapd.socket || true
-    while systemctl status snapd.service snapd.socket | grep "Active: activating"; do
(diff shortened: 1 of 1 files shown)


tests/lib/reset.sh
```diff
@@ -119,7 +119,7 @@ reset_all_snap() {
                 ;;
             *)
                 # make sure snapd is running before we attempt to remove snaps, in case a test stopped it
-                if ! systemctl status snapd.service snapd.socket; then
+                if ! systemctl status snapd.service snapd.socket >/dev/null; then
```
bboozzoo: We should probably be using `systemctl is-active ...` but this is good enough

zyga: +1 but I think it's absent on old systemd version.

PR #7247, opened 2019-08-14
interfaces/mount: forward SNAPD_DEBUG to ns tools

zyga:
When debugging snapd with SNAPD_DEBUG=1 it would be nice to be able to
forward that setting to snap-discard-ns and snap-update-ns. This patch
makes that happen.

### DIFF
--- a/interfaces/mount/ns.go
+++ b/interfaces/mount/ns.go
@@ -21,6 +21,7 @@ package mount
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 	"path/filepath"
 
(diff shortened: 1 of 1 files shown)


interfaces/mount/ns.go
```diff
@@ -44,6 +45,10 @@ func runNamespaceTool(toolName, snapName string) ([]byte, error) {
 			return nil, err
 		}
 		cmd := exec.Command(toolPath, snapName)
+		if osutil.GetenvBool("SNAPD_DEBUG") {
+			cmd.Env = os.Environ()
+			cmd.Env = append(cmd.Env, "SNAPD_DEBUG=1")
```
mvo5: Why do we need to append this environment after we just "copied" os.Environ() above which also has this flag? Also, why is this needed at all? The docs for exec.Command() say:
```
// If Env is nil, the new process uses the current process's
// environment.
```
and it seems like the original code was not doing anything to cmd.Env which means its nil. Or am I missing something?

zyga: On second thought this is a good question. Just to be clear, this patch did make a difference, I just no longer feel like I understand why exactly. I'll investigate.

zyga: I understand what was going on here. I'll close this branch and open a slightly different one. The key is to feed the stdout/stderr to both the error (capture it) as well as to actual stdout/stderr so that it ends up in the journal for recovery.

PR #7249, opened 2019-08-14
packaging/fedora: build on RHEL8

bboozzoo:
Tiny patches that make snapd build on RHEL8.

@Conan-Kudo can you take a look?


bboozzoo:
Let's wait for feedback from @Conan-Kudo before landing this.

mvo5:
Please remove "blocked" when its ready for merging.

packaging/fedora/snapd.spec
```diff
@@ -66,7 +66,8 @@
 # buildmode PIE triggers external linker consumes -extldflags
 %define gobuild_static(o:) go build -buildmode pie -compiler gc -tags=rpm_crashtraceback -ldflags "${LDFLAGS:-} -B 0x$(head -c20 /dev/urandom|od -An -tx1|tr -d ' \\n') -extldflags '%__global_ldflags -static'" -a -v -x %{?**};
 %endif
-%if 0%{?rhel} == 7
+# These macros are not defined in RHEL 7 & 8
+%if 0%{?rhel} >= 7
```
Conan-Kudo: The RHEL 7 macro should be left alone. Instead, we should change the `%if 0%{?fedora}` conditional for `%gobuild_static` to `%if 0%{?fedora} || 0%{?rhel} >= 8`.

packaging/fedora/snapd.spec
```diff
@@ -182,7 +183,6 @@ BuildRequires:  libtool
 BuildRequires:  gcc
 BuildRequires:  gettext
 BuildRequires:  gnupg
-BuildRequires:  indent
```
Conan-Kudo: Since we're going to be building with EPEL8, if this doesn't exist, someone should request the EPEL 8 branch so that it exists.

Conan-Kudo: Actually, `indent` is already being introduced in EPEL8: https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2019-891ee8a7be

bboozzoo: It's not needed for building snapd. I don't declare that in Arch either.

packaging/fedora/snapd.spec
bboozzoo: Ok, will do

packaging/fedora/snapd.spec
Conan-Kudo: I'm pretty sure it's used in the tests for `snap-confine`?

bboozzoo: It's disabled because indent output was unstable across versions.

Conan-Kudo: Okay then...

bboozzoo:
@mvo5 can cherry-pick this to 2.40 as well?

PR #7252, opened 2019-08-14
interfaces/network-manager: allow using org.freedesktop.DBus.ObjectManager

alfonsosanchezbeato:
interfaces/network-manager: allow using org.freedesktop.DBus.ObjectManager

The org.freedesktop.DBus.ObjectManager is implemented in newer versions
of NetworkManager and allows having a full view of current NM objects
with just one DBus call. It is used by nmcli.


alfonsosanchezbeato:
It looks like the test failure is about something with the gpio interface which is unrelated to the change.

zyga:
@jdstrand please merge with your +1

interfaces/builtin/network_manager.go
```diff
@@ -257,6 +257,13 @@ dbus (receive, send)
     path=/org/freedesktop/NetworkManager{,/**}
     peer=(label=###PLUG_SECURITY_TAGS###),
 
+# NM implements org.freedesktop.DBus.ObjectManager too
```
jdstrand: Can you modify these comments to be:
```
# Later versions of NetworkManager implement org.freedesktop.DBus.ObjectManager for
# clients to easily obtain all (and be alerted to added/removed) objects from the service.
```

mvo5: I did this now, once spread is green I will merge.

PR #7254, opened 2019-08-14
cmd/snap-update-ns: fix pair of bugs affecting refresh of snap with layouts

zyga:
This branch contains two bug fixes, one helper method and a spread test for a
bug that would manifest itself only when you were refreshing a snap with a
particular layout twice.

Please refer to commit messages from individual patches for details, but to set
the context, the bug was caused by the trespassing detector reporting incorrect
trespassing for two separate reasons.

https://bugs.launchpad.net/snapd/+bug/1808821


cmd/snap-update-ns/utils.go
```diff
@@ -156,7 +156,7 @@ func MkPrefix(base string, perm os.FileMode, uid sys.UserID, gid sys.GroupID, rs
 		// Keep closing the previous descriptor as we go, so that we have the
 		// last one handy from the MkDir below.
 		defer sysClose(fd)
-		fd, err = MkDir(fd, iter.CurrentBase(), iter.CurrentCleanName(), perm, uid, gid, rs)
+		fd, err = MkDir(fd, iter.CurrentCleanBase(), iter.CurrentCleanName(), perm, uid, gid, rs)
```
pedronis: this looks right

strutil/pathiter.go
```diff
@@ -85,6 +85,14 @@ func (iter *PathIterator) CurrentBase() string {
 	return iter.path[:iter.left]
 }
 
+// CurrentCleanBase returns the same value as CurrentBase with the right slash trimmed.
+func (iter *PathIterator) CurrentCleanBase() string {
```
pedronis: should we keep CurrentBase() at all, is not used now and seem a nuisance?

tests/regression/lp-1808821/task.yaml
```diff
@@ -0,0 +1,16 @@
...
+    one is that Keep changes were not tracked in filesystem
+    assumptions, therefore did not participate in trespassing
+    detection correctly. A mount namespace that chose to reuse a tmpfs
+    (a keep mount change) would subsequently trigger a trespassing
+    violation for a path that was directly under the mounted tmpfs.
+    The second bug was that path creation code, which used a helper to
+    create prefix, passed a directory name of the prefix with the
+    trailing slash, which would not match the restrictive trespassing
+    detection code.
+execute: ./task.sh
```
pedronis: a bit unclear why is done this way?

cmd/snap-update-ns/utils_test.go
```diff
@@ -136,6 +136,21 @@ func (s *utilsSuite) TestSecureMkdirAllLevel3(c *C) {
...
+	s.as.AddChange(&update.Change{Action: update.Mount, Entry: osutil.MountEntry{Dir: "/path", Type: "tmpfs", Name: "tmpfs"}})
+	s.sys.InsertFstatfsResult(`fstatfs 3 <ptr>`, syscall.Statfs_t{Type: update.SquashfsMagic})
+	s.sys.InsertFstatResult(`fstat 3 <ptr>`, syscall.Stat_t{})
+	s.sys.InsertFstatfsResult(`fstatfs 4 <ptr>`, syscall.Statfs_t{Type: update.TmpfsMagic})
+	s.sys.InsertFstatResult(`fstat 4 <ptr>`, syscall.Stat_t{})
+	s.sys.InsertFault(`mkdirat 3 "path" 0755`, syscall.EEXIST)
+	s.sys.InsertFault(`mkdirat 3 "to" 0755`, syscall.EEXIST)
+	rs := s.as.RestrictionsFor("/path/to/something")
+	// Trespassing detector checked "/path", not "/path/" (which would not match).
+	c.Assert(update.MkdirAll("/path/to/something", 0755, 123, 456, rs), IsNil)
```
jdstrand: Does it make sense to include the RCalls checks here?

strutil/pathiter.go
```diff
@@ -85,6 +85,14 @@ func (iter *PathIterator) CurrentBase() string {
...
+	if iter.left > 0 && iter.path[iter.left-1] == '/' && iter.path[:iter.left] != "/" {
```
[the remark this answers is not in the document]
pedronis: I'm nor sure what is the question, your example has CurrentBase while the code here is the new CurrentCleanBase ?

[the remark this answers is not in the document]
pedronis: I have added a test with such a middle segment fwiw

cmd/snap-update-ns/utils_test.go
pedronis: done

strutil/pathiter.go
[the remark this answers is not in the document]
jdstrand: I meant CurrentCleanBase. Thanks for the added test.

cmd/snap-update-ns/utils_test.go
jdstrand: Thanks!

cmd/snap-update-ns/change.go
```diff
@@ -392,6 +392,7 @@ func (c *Change) lowLevelPerform(as *Assumptions) error {
 		}
 		return err
 	case Keep:
+		as.AddChange(c)
```
mvo5: I really know little about this code so my comment maybe silly - but when I read this I was a bit surprised. It reads as "Keep:" but the first thing it does is adding a change which seems strange. Maybe a comment? The commit comment explains it but a bit. Maybe something like: `// keep track of reused entries for the trespassing algorithm` ?

strutil/pathiter.go
zyga: It can be dropped if there is no compelling reason to keep it.

tests/regression/lp-1808821/task.yaml
zyga: Mainly because it lets me run things in debug shell conveniently. There is some side improvement from getting IDE to understand the syntax of the test.

PR #7259, opened 2019-08-15
tests: just build snapd commands on go-build test

sergiocazzolato:
It's building lib packages atm, and the idea of the test is just to
build snapd commands.

### DIFF
--- a/tests/cross/go-build/task.yaml
+++ b/tests/cross/go-build/task.yaml
@@ -52,6 +52,6 @@ prepare: |
 
 execute: |
     cd /tmp/cross-build/src/github.com/snapcore/snapd
-    for cmd in $( find ./cmd -mindepth 2 -maxdepth 2 -name \*.go -printf '%h\n' | sort -u ); do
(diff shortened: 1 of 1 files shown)


tests/cross/go-build/task.yaml
```diff
@@ -52,6 +52,6 @@ prepare: |
 
 execute: |
     cd /tmp/cross-build/src/github.com/snapcore/snapd
-    for cmd in $( find ./cmd -mindepth 2 -maxdepth 2 -name \*.go -printf '%h\n' | sort -u ); do
+    for cmd in $( find ./cmd -mindepth 2 -maxdepth 2 -name main.go -printf '%h\n' | sort -u ); do
```
bboozzoo: We should (ab)use `go list` and bash and look at the actual package name rather than arbitrary files:
```
go list -f '{{.ImportPath}} {{.Name}}' ./cmd/... | \
    while read pkg main; do if [[ "$main" == "main" ]]; then go build -v  "${pkg}"; fi ; done
```

zyga: Can we just `go build ./...` everything and avoid the magic? Are there any downsides?

PR #7262, opened 2019-08-15
tests: use a different archive based on the spread backend on go-build test

sergiocazzolato:
This is to speed up the test and avoid timeouts during execution on travis

### DIFF
--- a/tests/cross/go-build/task.yaml
+++ b/tests/cross/go-build/task.yaml
@@ -36,7 +36,7 @@ prepare: |
 
     UBUNTU_ARCHIVE="http://gce.clouds.archive.ubuntu.com/ubuntu/"
     if [ "$SPREAD_BACKEND" = "google" ]; then
-        UBUNTU_ARCHIVE="http://gce.clouds.archive.ubuntu.com/ubuntu/"
+        UBUNTU_ARCHIVE="http://$(cloud-id -l | cut -f2).gce.archive.ubuntu.com/ubuntu/"
     fi
 
     mv /etc/apt/sources.list /etc/apt/sources.list.orig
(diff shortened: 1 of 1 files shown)


tests/cross/go-build/task.yaml
```diff
@@ -34,12 +34,17 @@ prepare: |
     cp -ar "$PROJECT_PATH" /tmp/cross-build/src/github.com/snapcore
     chown -R test:12345 /tmp/cross-build
 
+    UBUNTU_ARCHIVE="http://gce.clouds.archive.ubuntu.com/ubuntu/"
+    if [ "$SPREAD_BACKEND" = "google" ]; then
+        UBUNTU_ARCHIVE="http://gce.clouds.archive.ubuntu.com/ubuntu/"
```
chipaca: no, AFAICT that's wrong; that archive (if it is an archive) is quite a lot further away than the one I think we want.
Try this:
```suggestion
        UBUNTU_ARCHIVE="http://$(cloud-id -l | cut -f2).gce.archive.ubuntu.com/ubuntu/"
```
the one you put is about as far away as archive.u.c (~30ms), whereas this method gets you something that's ~.5ms away.

sergiocazzolato: @chipaca nice, change done, thanks for the comment!!

tests/cross/go-build/task.yaml
```diff
@@ -34,12 +34,17 @@ prepare: |
     cp -ar "$PROJECT_PATH" /tmp/cross-build/src/github.com/snapcore
     chown -R test:12345 /tmp/cross-build
 
+    UBUNTU_ARCHIVE="http://gce.clouds.archive.ubuntu.com/ubuntu/"
```
chipaca:
```suggestion
    UBUNTU_ARCHIVE="http://archive.ubuntu.com/ubuntu/"
```

PR #7272, opened 2019-08-19
interfaces/bluez: enable communication between bluetoothd and meshd via dbus

jrigling:


interfaces/builtin/bluez_test.go
```diff
@@ -178,6 +178,7 @@ func (s *BluezInterfaceSuite) TestAppArmorSpec(c *C) {
 	c.Assert(spec.SecurityTags(), DeepEquals, []string{"snap.consumer.app"})
 	c.Assert(spec.SnippetForTag("snap.consumer.app"), testutil.Contains, "peer=(name=org.bluez, label=unconfined)")
 	c.Assert(spec.SnippetForTag("snap.consumer.app"), testutil.Contains, "peer=(name=org.bluez.obex, label=unconfined)")
+	c.Assert(spec.SnippetForTag("snap.consumer.app"), testutil.Contains, "peer=(name=org.bluez.mesh, label=unconfined)")
```
jhenstridge: This appears to be testing for changes to the AppArmor connected plug snippet that haven't been included in this PR.

mvataleu: Thanks, added rules for AppArmor

michallowasrzechonek-silvair:
This is needed to allow a newest (yet unreleased) addition to BlueZ work on Ubuntu Core, namely bluetooth-meshd daemon. As with 'regular' bluetoothd, the new daemon communicates with applications over DBus.

API is described in detail in https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/doc/mesh-api.txt

jhenstridge:
From a quick skim of the docs, it sounds like this will need some additional rules for the plug side too.  In addition to allowing applications to make calls to meshd, it is necessary to allow meshd to make certain method calls to the application.

In particular:
1. method calls with the `org.bluez.mesh.Application1` interface to an application supplied object path passed to the org.bluez.mesh.Network1.Join` method.
2. `org.freedesktop.DBus.ObjectManager` on the app supplied object path: both receiving method calls and sending signals.
3. `org.bluez.mesh.Attention1` method calls on the app supplied object path
4. `org.bluez.mesh.Provisioner1` method calls on the app supplied object path
5. `org.bluez.mesh.ProvisionAgent1` method calls on some object path below the one supplied by the app.
6. `org.bluez.mesh.Element1` method calls on any object paths below the one supplied by the app.
7. property access for the above interfaces on all the above object paths.

On the slot side, meshd will need to be able to make all these calls and receive all these signals.

michallowasrzechonek-silvair:
Please note that object paths are freely definable, i.e. they don't need to be nested within 'main' application object,. Even though most apps would probably implement it in that manner, it's not mandatory.

jhenstridge:
Sure.  I mentioned object paths because that is one of the elements the AppArmor rules can match on.  Since the API doesn't define the client-side object paths, the rules would need to allow the relevant APIs to be provided by any object path.

This means we'd need to be particularly careful with the rules, so that e.g. when we allow the plug side to send `ObjectManager` signals we don't inadvertantly also let it send `ObjectManager` method calls to arbitrary unconfined daemons (and similar for the D-Bus properties interface).

mvataleu:
Hello everyone. Сould you tell me a way how to get detailed logs from your CI server due to integrational tests execution? As far I can see that integration tests fail on CI server but seems it passes locally (even on spread). Thanks in advance.

anonymouse64:
Hi @mvataleu I was unable to reproduce the failed test error with either a local QEMU spread run or with one on the google cloud, so I restarted the integration tests for you. Hopefully it should pass now. The failing unit test (github.com/snapcore/snapd/cmd/snap) doesn't seem to have anything to do with your changes here.

To answer your question, the most detailed logs you can get are by going to the Travis build job and clicking on the failed build and going to the "Raw log", I don't think we save more than that.

PR #7274, opened 2019-08-19
tests: change cgroups so that LXD doesn't have to

zyga:
This patch sets the "clone_children" attribute on the cpuset cgroup
controller, early in the spread project-wide preparation code, in
anticipation of LXD needing this setting.

LXD alters cgroup configuration in a one-off way. Attempts to
reconfigure them back were only partially successful and also block on
a patch to spread that cannot be merged soon.

Instead of waiting, let's turn the problem around. If cgroups are
configured in a way that LXD expects up front then installing and using
LXD doesn't change anything. This allows us to keep using our mount leak
detector straight away.

### DIFF
--- a/tests/lib/prepare-restore.sh
+++ b/tests/lib/prepare-restore.sh
@@ -465,6 +465,13 @@ prepare_project() {
     RateLimitBurst=0
 EOF
     systemctl restart systemd-journald.service
+
+    # Re-configure cgroups in a way that LXD would so that
+    # installation, use and removal of LXD does not leave any changes
+    # in the system.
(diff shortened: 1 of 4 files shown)


zyga:
@bboozzoo yes but those will come with extending the mount leak detector to other systems. I know that on fedora there were several other changes. This is sufficient to enable the detector on Ubuntu 16.04. I'll know what other changes are needed for our set of spread systems in a few hours.

mvo5:
I checked the documentation in https://www.kernel.org/doc/Documentation/cgroup-v1/cgroups.txt and this looks fine.

PR #7285, opened 2019-08-20
httputil: improve http2 PROTOCOL_ERROR detection

mvo5:
This cleans up the http2 protocol error detection in the retry
code and also adds a test.


httputil/retry_test.go
```diff
@@ -484,3 +487,21 @@ func (s *retrySuite) TestRetryOnTemporaryDNSfailureNotGo19(c *C) {
...
 }
+
+func (s *retrySuite) TestRetryOnHttp2ProtocolErrors(c *C) {
+	n := 0
+	doRequest := func() (*http.Response, error) {
+		n++
+		return nil, &url.Error{
+			Op:  "Get",
+			URL: "http://...",
+			Err: http2.StreamError{Code: http2.ErrCodeProtocol},
```
stolowski: ❤️

mvo5:
Setting to blocked as the amount of vendoring needed is a bit worrying - OTOH we have a golang-golang-x-net-dev package in debian that contains all of golang.org/x/net. So *maybe* its fine if such a package is available in the other distros as well.

mvo5:
Set to squash merge so that *if* we merge it we can at least easily revert if it causes (dependency) problems down the line.

PR #7286, opened 2019-08-20
tests: use images:ubuntu/ in lxd tests

mvo5:
We had a bunch of download speed issues with the ubuntu: hosted
images. As suggested by Stephane this PR switches to the community
hosted "images:ubuntu/$image" server which apparently is mirrored
more widely. This should fix the timeout problems we are seeing
right now.


zyga:
Curiously the test now fails:
```
error: system does not fully support snapd: cannot mount squashfs image using
       "fuse.snapfuse": ----- mount: wrong fs type, bad option, bad superblock
       on /tmp/sanity-squashfs-740076654,

       missing codepage or helper program, or other error

       In some cases useful info is found in syslog - try
       dmesg | tail or so.
```

Could it be that those images are somehow different?

mvo5:
This fails right now because the community images do not ship the "fuse" package by default. Maybe it can be installed by default (cc @stgraber)? In any case I think we need to add a new sanity check that checks for "is_container && !have("/dev/fuse") -> please install fuse". The current error is very confusing. It also means we probably need to keep the test with the official images because we need to make sure they work and if they are different from the other images changing this test would be a regression :(

stgraber:
Installing fuse by default in our images seems reasonable.
I would also say that snapd should depend on or at least recommend fuse if it's going to require it to function.

mvo5:
Thanks @stgraber ! Your suggestion makes a lot of sense, I added fuse to the recommends now.

packaging/ubuntu-16.04/control
```diff
@@ -71,7 +71,7 @@ Depends: adduser,
          ${shlibs:Depends}
 Replaces: ubuntu-snappy (<< 1.9), ubuntu-snappy-cli (<< 1.9), snap-confine (<< 2.23), ubuntu-core-launcher (<< 2.22), snapd-xdg-open (<= 0.0.0)
 Breaks: ubuntu-snappy (<< 1.9), ubuntu-snappy-cli (<< 1.9), snap-confine (<< 2.23), ubuntu-core-launcher (<< 2.22), snapd-xdg-open (<= 0.0.0), ${snapd:Breaks}
-Recommends: gnupg
+Recommends: gnupg, fuse
```
zyga: Can you please make sure we recommend fuse on packaging for other systems as well?

zyga:
Spread test is failing:
```
+ lxd.lxc exec my-ubuntu -- su -c '/snap/bin/test-snapd-tools.echo from-the-inside' ubuntu
snap-confine has elevated permissions and is not confined but should be. Refusing to continue to avoid permission escalation attacks
```

mvo5:
The failure is a bit strange, I ran this in spread with "--debug" and indeed, inside the lxd container there is no apparmor profile loaded for "us.lib.snapd.snap-confine":
```
root@my-ubuntu:~# apparmor_status 
apparmor module is loaded.
14 profiles are loaded.
14 profiles are in enforce mode.
   /snap/core/7396/usr/lib/snapd/snap-confine
   /snap/core/7396/usr/lib/snapd/snap-confine//mount-namespace-capture-helper
   snap-update-ns.core
   snap-update-ns.test-snapd-tools
   snap.core.hook.configure
   snap.test-snapd-tools.block
   snap.test-snapd-tools.cat
   snap.test-snapd-tools.cmd
   snap.test-snapd-tools.echo
   snap.test-snapd-tools.env
   snap.test-snapd-tools.fail
   snap.test-snapd-tools.head
   snap.test-snapd-tools.sh
   snap.test-snapd-tools.success
0 profiles are in complain mode.
0 processes have profiles defined.
0 processes are in enforce mode.
0 processes are in complain mode.
0 processes are unconfined but have a profile defined.
```
However this *should* be loaded from the snapd postinst. And indeed if I run the postinst agai with "configure" the profile gets loaded:
```
root@my-ubuntu:~# apparmor_status 
apparmor module is loaded.
16 profiles are loaded.
16 profiles are in enforce mode.
   /snap/core/7396/usr/lib/snapd/snap-confine
   /snap/core/7396/usr/lib/snapd/snap-confine//mount-namespace-capture-helper
   /usr/lib/snapd/snap-confine
   /usr/lib/snapd/snap-confine//mount-namespace-capture-helper
...
```
and with that things work normally. This is puzzling. I also noticed in the failure log that apparmor and systemd get upgraded in the same dpkg/apt run so this issue might be some generic upgrade bug that is just triggered in the minimal lxd env.

zyga:
@mvo5 the apparmor systemd unit has some checks for running inside privileged containers and just doesn't start up in this case. Snapd does not contain such checks (but it should) and loads its profiles, corrupting the host. In reality snapd does not operate in privileged containers without damaging the host's ability to use snaps. I think this is what you are observing.

Can we somehow diff the container configuration?

stgraber:
@zyga hmm, LXD does use apparmor namespacing even for privileged containers.
```
stgraber@castiana:~$ lxc launch ubuntu:18.04 c2 -c security.privileged=true
Creating c2
Starting c2     
stgraber@castiana:~$ lxc exec c2 bash
root@c2:~# cat /proc/self/attr/current 
unconfined
root@c2:~# cat /sys/kernel/security/apparmor/.ns_name 
lxd-c2_<var-lib-lxd>
root@c2:~# 

```

mvo5:
Closing this until there is time to investigate why the test fails and we end up with no apparmor profile inside the lxd container.

PR #7287, opened 2019-08-20
hookstate/ctlcmd: snapctl unset command

stolowski:
Support `snapctl unset` command for removing config options from hooks.


overlord/hookstate/ctlcmd/unset_test.go
```diff
@@ -0,0 +1,128 @@
...
+	tr = config.NewTransaction(s.mockContext.State())
+	c.Check(tr.Get("test-snap", "foo", &value), ErrorMatches, `snap "test-snap" has no "foo" configuration option`)
+	c.Check(tr.Get("test-snap", "bar", &value), ErrorMatches, `snap "test-snap" has no "bar" configuration option`)
+	c.Check(tr.Get("test-snap", "baz", &value), IsNil)
+	c.Check(value, Equals, "c")
+}
+
+func (s *unsetSuite) TestUnsetRegularUserForbidden(c *C) {
+	state := state.New(nil)
+	state.Lock()
```
mvo5: (nitpick^2) the \n here is a bit unusually (to me), I would put it more like:
```
state := ...

state.Lock()
do_stuff()
state.Unlock()

...
```
but not really an issue.

overlord/hookstate/ctlcmd/unset_test.go
```diff
@@ -0,0 +1,128 @@
...
+	s.mockContext, err = hookstate.NewContext(task, task.State(), setup, s.mockHandler, "")
+	c.Assert(err, IsNil)
+}
+
+func (s *unsetSuite) TestInvalidArguments(c *C) {
+	_, _, err := ctlcmd.Run(s.mockContext, []string{"unset"}, 0)
+	c.Check(err, ErrorMatches, "unset which option.*")
+}
+
+func (s *unsetSuite) TestUnsetMany(c *C) {
```
mvo5: Should we have a test that checks for a single value unset too?

overlord/hookstate/ctlcmd/unset_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	task := state.NewTask("test-task", "")
+	setup := &hookstate.HookSetup{Snap: "test-snap", Revision: snap.R(1), Hook: "hook"}
+
+	state.Unlock()
+
+	mockHandler := hooktest.NewMockHandler()
+	mockContext, err := hookstate.NewContext(task, task.State(), setup, mockHandler, "")
+	c.Assert(err, IsNil)
+	_, _, err = ctlcmd.Run(mockContext, []string{"unset", "key"}, 1000)
```
mvo5: Could we reuse s.mockContext here? It seems we don't do anything different there?

overlord/hookstate/ctlcmd/unset_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	task := state.NewTask("test-task", "")
+	setup := &hookstate.HookSetup{Snap: "test-snap", Revision: snap.R(1), Hook: "hook"}
+
+	state.Unlock()
+
+	mockHandler := hooktest.NewMockHandler()
+	mockContext, err := hookstate.NewContext(task, task.State(), setup, mockHandler, "")
+	c.Assert(err, IsNil)
+	_, _, err = ctlcmd.Run(mockContext, []string{"unset", "-h"}, 1000)
```
mvo5: Same question as above about reusing s.mockContext

overlord/hookstate/ctlcmd/unset_test.go
stolowski: Good point.  I removed it alltogether though, it was a duplication of SetUpTest after all...

overlord/hookstate/ctlcmd/unset_test.go
stolowski: Definately no harm in having it, added.

overlord/hookstate/ctlcmd/unset_test.go
stolowski: Good point, removed. Thanks!

overlord/hookstate/ctlcmd/unset_test.go
stolowski: Yep, removed as well.

overlord/hookstate/ctlcmd/unset.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	context := s.context()
+	if context == nil {
+		return fmt.Errorf("cannot unset without a context")
+	}
+
+	context.Lock()
+	tr := configstate.ContextTransaction(context)
+	context.Unlock()
```
zyga: I'm not entirely sure what lock/unlock protects against but should the rest of the transaction be protected by the lock? Did you mean to `defer context.Unlock()` here?

overlord/hookstate/ctlcmd/unset.go
```diff
@@ -0,0 +1,74 @@
...
+
+var shortUnsetHelp = i18n.G("Remove configuration options")
+var longUnsetHelp = i18n.G(`
+The unset command removes the provided configuration options as requested.
+
+$ snapctl unset name address
+
+All configuration changes are persisted at once, and only after the
+snap's configuration hook returns successfully.
+
```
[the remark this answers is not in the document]
stolowski: This is implied by the nature of hooks that are executed in the context of their snap and the help string is consistent with the rest of `snapctl` commands. Maybe rewording it to be more explicit is a good idea, but I'd reserve this for a followup where all commands (get/set/unset) would be updated at once.

overlord/hookstate/ctlcmd/unset.go
stolowski: `configstate.ContextTransaction(context)` needs a context lock because it does `Get(..)` on the state to take a snapshot of snap's config. Lock is not needed afterwards for individual transaction `Set`s, it is acquired again only on `Commit` (which happens outside).

PR #7288, opened 2019-08-20
tests: cleanup "snap_daemon" user in system-usernames-install-twice

mvo5:
When a snap is installed that uses the snap_daemon user it needs
cleanup so that the state of the (test) system is not altered. We have
(slightly) different code in the two tests, consolidate it into a single
helper.


tests/lib/users.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
zyga: Could we please roll this into a binary in `tests/lib/bin`, this way:
- we can cluster and reuse functionality easily
- we can decouple the state of the calling shell from this, deciding corectness is easier
- we can reimplement it in any language without changing tests
- it's on path, no need to source it

I would suggest the name `user-tool` since it matches the `-tool` suffix used by other helpers. I can happily make the changes in this PR to make that happen.

I would propose:
```
user-tool remove-with-group <NAME>
```

mvo5: sounds good, go for it!

PR #7295, opened 2019-08-21
tests: spam test logs less while waiting for systemd unit to stop

zyga:
The systemd_stop_units function checks, up to 20 times, for the given
units to move out of "activating" phase before stopping them. At each
attempt it would print the full status of the unit, making logs longer.

This patch changes that so only the fainal failed status is printed.

### DIFF
--- a/tests/lib/systemd.sh
+++ b/tests/lib/systemd.sh
@@ -67,7 +67,6 @@ systemd_stop_units() {
         if systemctl is-active "$unit"; then
             echo "Ensure the service is active before stopping it"
             retries=20
-            systemctl status "$unit" || true
             while systemctl status "$unit" | grep -q "Active: activating"; do
                 if [ $retries -eq 0 ]; then
                     echo "$unit unit not active"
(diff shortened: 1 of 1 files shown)


tests/lib/systemd.sh
```diff
@@ -68,9 +68,10 @@ systemd_stop_units() {
             echo "Ensure the service is active before stopping it"
             retries=20
             systemctl status "$unit" || true
-            while systemctl status "$unit" | grep "Active: activating"; do
+            while systemctl status "$unit" | grep -q "Active: activating"; do
```
bboozzoo: Hm this should probably do `systemctl show -p ActiveState "$unit" | grep -q 'ActiveState=activating'` or
 ```
while [[ "$(systemctl is-active "$unit"" = "activating" ]]; then
...
done
```
Either way, the proposed fix works too.

zyga: Yeah, there's some simplification possible but for the sake of not checking Ubuntu 14.04 I went for the simplest thing possible.

PR #7298, opened 2019-08-21
tests: clean up after NFS tests

zyga:
The test was leaking two mount points: /var/lib/nfs/rpc_pipefs and
/proc/fs/nfsd. In addition because of how the test operates, files
placed in the current working directory can get "lost" somehow. As a
precaution I moved all of the helper files to /tmp.

Note that the entire test is disabled on Fedora, CentOS, openSUSE and
Ubuntu Core so there is still some "dead code" in terms of switch
statements that are not executed. I tried to do my best to keep the dead
code as useful, to the future developers, as I could.

There are small unrelated changes as well, like how /home is unmounted
or a FIXME note about test clobbering a file on Fedora.

### DIFF
--- a/tests/main/nfs-support/task.yaml
+++ b/tests/main/nfs-support/task.yaml
@@ -128,7 +128,7 @@ execute: |
             systemctl restart nfs-kernel-server
             ;;
         fedora-*)
-            # Enable udp protocol for nfs on fedora which is disable by default          
+            # Enable udp protocol for nfs on fedora which is disabled by default
             sed -i -e 's/RPCNFSDARGS=.*/RPCNFSDARGS="--udp"/g' /etc/sysconfig/nfs
             # FIXME: this is not restored anywhere.
             systemctl restart nfs
--- a/tests/main/nfs-support/task.yaml
+++ b/tests/main/nfs-support/task.yaml
@@ -19,16 +19,29 @@ prepare: |
     . "$TESTSLIB/snaps.sh"
     install_local test-snapd-sh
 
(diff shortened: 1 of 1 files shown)


tests/main/nfs-support/task.yaml
```diff
@@ -41,6 +54,46 @@ restore: |
     systemctl reset-failed snapd.service
     systemctl start snapd.service
 
+    # Depending on OS do cleanup appropriate for the host.
+    case "$SPREAD_SYSTEM" in
+        ubuntu-14.04-*)
```
mvo5: Why is ubuntu-16.04, 18.04 etc missing here?

mvo5: Maybe a comment here for our future selfs that may have the same question?

zyga: I'll add a comment but the basic answer is that the cleanup is split between two chunks (stop, unmount, start-if-needed). Not all systems need the changes, specifically Ubuntu 16.04 and beyond have NFS preinstalled and working and we don't need to stop it.

zyga: Can you look again?

tests/main/nfs-support/task.yaml
```diff
@@ -75,8 +128,9 @@ execute: |
             systemctl restart nfs-kernel-server
             ;;
         fedora-*)
-            # Enable udp protocol for nfs on fedora which is disable by default           
+            # Enable udp protocol for nfs on fedora which is disable by default
```
bboozzoo:
```suggestion
            # Enable udp protocol for nfs on fedora which is disabled by default
```

PR #7299, opened 2019-08-21
sanity: report proper errror when fuse is needed but not available

mvo5:
During the spread run for PR#7286 we noticed that for some systems
our sanity check in lxd is not helpful. When a container does
not have fuse at all we try to mount the fusesnap filesystem
but because there is no fuse the error message is generic and
confusing. To fix this this PR adds a check if fuse is needed
and if the fuse mount binary is missing.


tests/main/lxd-no-fuse/task.yaml
```diff
@@ -0,0 +1,43 @@
+summary: Check that we give a useful error when fuse is missing in lxd
+
+# we just need a single system to verify this
+systems: [ubuntu-18.04-64]
```
zyga: Can you please mirror changes to the LXD tests recently introduced here? Alternatively delegate this to me, so that we do proper cleanup.

osutil/squashfs/fstype.go
```diff
@@ -60,6 +60,11 @@ func MockUseFuse(r bool) func() {
 	return func() { useFuse = oldUseFuse }
 }
 
+// NeedsFuse returns true if the given system needs fuse to mount snaps
+func NeedsFuse() bool {
+	return useFuse()
```
zyga: I think it would be nice to rename useFuse to `NeedsFuse` and call it a day.

tests/main/lxd-no-fuse/task.yaml
mvo5: I pick option (2), please push whatever is needed for proper cleanup.

tests/main/lxd-no-fuse/task.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+    echo "Setting up proxy for lxc"
+    if [ -n "${http_proxy:-}" ]; then
+        lxd.lxc config set core.proxy_http "$http_proxy"
+    fi
+    if [ -n "${https_proxy:-}" ]; then
+        lxd.lxc config set core.proxy_https "$http_proxy"
+    fi
+
+    . /etc/os-release
```
bboozzoo: Is this needed?

mvo5: Its not needed, nice catch!

tests/main/lxd-no-fuse/task.yaml
zyga: I didn't notice this is just for Ubuntu 18.04, this is covered by the general fixes I did that affect the entire machine so the test doesn't need anything extra. LGTM

PR #7300, opened 2019-08-21
interfaces/network-{control,manager}: allow 'k' on /run/resolvconf/**

jdstrand:
References:
https://bugs.launchpad.net/snapd/+bug/1840873/comments/2

PR #7301, opened 2019-08-21
interfaces/network-{control,manager}: allow 'k' on /run/resolvconf/** - 2.41

jdstrand:
https://github.com/snapcore/snapd/pull/7300 for 2.41


mvo5:
Our pr-title check is a bit silly, this is fixed with https://github.com/snapcore/snapd/pull/7305

tsunghanliu:
I've applied the changes to the system has this issue, and this PR fixes the issue and  the system works well.

PR #7303, opened 2019-08-21
interfaces/wayland,x11: allow reading an Xwayland Xauth file. Fixes LP: #1840925.

oSoMoN:
(see https://gitlab.gnome.org/GNOME/mutter/merge_requests/626)


oSoMoN:
> I believe a corresponding rule is needed in the x11 interface in x11ConnectedPlugAppArmor

Done.

PR #7306, opened 2019-08-21
overlord/configstate: sort patch keys to have deterministic order with snap set

stolowski:
Sort patch keys by path "depth" to have deterministic order of multiple set ops with `snap set`.

`snap set snap-name foo='{"key1":"a"}' foo.key2="b"` currently has random outcome, because patch values in our API are carried as a map (in this case the map has "foo" and "foo.key2" keys and their corresponding values) and are processed in random order of the keys, so one value may override the other. Some time ago we decided not to change the API, but ensure the order is deterministic and always the same.

### DIFF
--- a/overlord/configstate/export_test.go
+++ b/overlord/configstate/export_test.go
@@ -20,3 +20,4 @@
 package configstate
 
 var NewConfigureHandler = newConfigureHandler
+var SortPatchKeys = sortPatchKeys
--- a/overlord/configstate/helpers.go
+++ b/overlord/configstate/helpers.go
@@ -1,0 +1,40 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2016 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package configstate
+
+import (
+	"sort"
+	"strings"
+)
+
+func sortPatchKeys(patch map[string]interface{}) []string {
+	depths := make(map[string]int, len(patch))
+	var keys []string
+	for k := range patch {
+		parts := strings.Split(k, ".")
+		depths[k] = len(parts)
+		keys = append(keys, k)
+	}
+
+	sort.Slice(keys, func(i, j int) bool {
+		return depths[keys[i]] < depths[keys[j]]
+	})
+	return keys
+}
--- a/overlord/configstate/helpers.go
+++ b/overlord/configstate/helpers.go
@@ -25,11 +25,13 @@ import (
 )
 
 func sortPatchKeys(patch map[string]interface{}) []string {
+	if len(patch) == 0 {
+		return nil
+	}
 	depths := make(map[string]int, len(patch))
-	var keys []string
+	keys := make([]string, 0, len(patch))
 	for k := range patch {
-		parts := strings.Split(k, ".")
-		depths[k] = len(parts)
+		depths[k] = strings.Count(k, ".")
 		keys = append(keys, k)
 	}
 
--- a/overlord/configstate/helpers_test.go
+++ b/overlord/configstate/helpers_test.go
@@ -1,0 +1,45 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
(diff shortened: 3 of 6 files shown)


overlord/configstate/helpers.go
```diff
@@ -0,0 +1,40 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2016 Canonical Ltd
```
zyga:
```suggestion
 * Copyright (C) 2019 Canonical Ltd
```

overlord/configstate/helpers.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"sort"
+	"strings"
+)
+
+func sortPatchKeys(patch map[string]interface{}) []string {
+	depths := make(map[string]int, len(patch))
+	var keys []string
+	for k := range patch {
+		parts := strings.Split(k, ".")
```
zyga: I think using `strings.Count` might be a bit cheaper here.

overlord/configstate/helpers.go
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package configstate
+
+import (
+	"sort"
+	"strings"
+)
+
+func sortPatchKeys(patch map[string]interface{}) []string {
```
zyga: Naming nitpick:
```suggestion
func sortPatchKeysByDepth(patch map[string]interface{}) []string {
```

overlord/configstate/helpers_test.go
```diff
@@ -0,0 +1,45 @@
...
+func (s *miscSuite) TestSortPatchKeys(c *C) {
+	patch := map[string]interface{}{
+		"a.b.c":         0,
+		"a":             0,
+		"a.b.c.d":       0,
+		"q.w.e.r.t.y.u": 0,
+		"f.g":           0,
+	}
+
+	keys := configstate.SortPatchKeys(patch)
```
zyga:
```suggestion
	keys := configstate.SortPatchKeysByDepth(patch)
```

overlord/configstate/hooks.go
```diff
@@ -112,8 +112,9 @@ func (h *configureHandler) Before() error {
 		}
 	}
 
-	for key, value := range patch {
-		if err := tr.Set(instanceName, key, value); err != nil {
+	patchKeys := sortPatchKeys(patch)
```
zyga:
```suggestion
	patchKeys := sortPatchKeysByDepth(patch)
```

overlord/configstate/export_test.go
```diff
@@ -20,3 +20,4 @@
 package configstate
 
 var NewConfigureHandler = newConfigureHandler
+var SortPatchKeys = sortPatchKeys
```
zyga:
```suggestion
var SortPatchKeysByDepth = sortPatchKeysByDepth
```
```suggestion
var SortPatchKeys = sortPatchKeys
```

tests/main/snap-set/task.yaml
```diff
@@ -71,3 +71,11 @@ execute: |
         exit 1
     fi
     [[ "$obtained" == *"error from within configure hook"* ]]
+
+    echo "Test that the 'snap set' order is deterministic"
+    for _ in $(seq 50); do
+        snap set snapctl-hooks command=noop one!
+        snap set snapctl-hooks one.two=2 one='{"three":3}'
+        snap get snapctl-hooks -l one.two | MATCH "one.two .* 2"
```
zyga: Why `.*` in the MATCH pattern?

stolowski: Because of the potentially changing number of spaces. It's probably fixed for these crafted data, but I thought I woudn't bother with whitespace.

overlord/configstate/helpers.go
stolowski: Indeed, I thought `Split` would make the intent more clear, and the amounts of data we're dealing with here is negligible; I can change this if you think it's better to have Count?

overlord/configstate/helpers.go
```diff
@@ -0,0 +1,40 @@
...
+	"sort"
+	"strings"
+)
+
+func sortPatchKeysByDepth(patch map[string]interface{}) []string {
+	depths := make(map[string]int, len(patch))
+	var keys []string
+	for k := range patch {
+		parts := strings.Split(k, ".")
+		depths[k] = len(parts)
```
bboozzoo: You can do `strings.Count(k, ".")` and avoid extra allocations.

overlord/configstate/helpers.go
```diff
@@ -0,0 +1,40 @@
...
+		keys = append(keys, k)
+	}
+
+	sort.Slice(keys, func(i, j int) bool {
+		return depths[keys[i]] < depths[keys[j]]
```
bboozzoo: If we don't care about caching 'dot' counts, this could be further simplified:
```
sort.Slice(keys, func(i, j int) bool {
    return strings.Count(keys[i], ".") < strings.Count(keys[j], ".")
})
```

stolowski: I'd actually cache counts, repeating Count or Split here would be wasteful.

overlord/configstate/helpers.go
stolowski: Ok, due to popular demand (@zyga suggested this as well) I'm changing this to `Count` :)

overlord/configstate/helpers.go
stolowski: Changed, per similiar suggestion from Maciej.

tests/main/snap-set/task.yaml
zyga: On second though should you check for `[ ]*`? That would make the intent clear.

PR #7307, opened 2019-08-22
mkversion.sh: fix version from git checkouts

mvo5:
The mkversion.sh script is used during the build of the snapd snap.
Today its a bit naive and just uses "git describe" there to
generate the version number. The downside of this is that for
pre-releases like 2.41~pre1 we may end up with a version number
lower than the version in debian/changelog if the pre release
does not have a git tag. We historically don't tag pre/rc
releases and doing so would be a bit messy (IMO) so we need
a less naive approach to generate the version number.

This PR changes the mkversion.sh script to use something
similar to what a packaging recipe does to construct a version.

On the current master this will yield:
```
2.41~pre1+git905.g5e58678
```
instead of (the output before this fix):
```
2.40+git905.g5e58678f94
```

### DIFF
--- a/mkversion.sh
+++ b/mkversion.sh
@@ -47,7 +47,9 @@ fi
 # at this point we maybe in _build/src/github etc where we have no
 # debian/changelog (dh-golang only exports the sources here)
 # switch to the real source dir for the changelog parsing
-version_from_changelog="$(cd "$PKG_BUILDDIR"; dpkg-parsechangelog --show-field Version)";
+if command -v dpkg-parsechangelog >/dev/null; then
+    version_from_changelog="$(cd "$PKG_BUILDDIR"; dpkg-parsechangelog --show-field Version)";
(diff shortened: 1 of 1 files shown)


mvo5:
@Conan-Kudo This should still work because the parts that require git are "guarded". The first one by "command -v git" and the second one only runs if "$version_from_user" is empty which is not the case for fedora releases, iirc/afaict the snapd.spec file sets the version explicitly. Or am I missing something?

Conan-Kudo:
@mvo5 I think I was missing something... I forgot how `command` works!

Conan-Kudo:
@mvo5 Will this work properly with snapshot rpm package builds? I think we've relied on the scheme to figure out versioning...

mvo5:
@Conan-Kudo I think so, there should be no change here if the dpkg-parsechangelog is not available.

PR #7308, opened 2019-08-22
tests: add new "user-tool" helper and use in system-user tests

mvo5:
This is a followup on #7288 - instead of cleaning up using a
shell function (which is a bit cumbersome to use, i.e. it needs
to be sources and shellcheck annotated) this just creates a
helper binary.


zyga:
@mvo5 my follow-up is https://github.com/snapcore/snapd/pull/7309

PR #7309, opened 2019-08-22
tests: re-implement user tool in python

zyga:
Now that user-tool is an executable and not a shell function it can be
implemented in something more appropriate. There are no unit tests but
it is black and mypy clean.

### DIFF
--- a/tests/lib/bin/user-tool
+++ b/tests/lib/bin/user-tool
@@ -1,42 +1,55 @@
-#!/bin/bash
+#!/usr/bin/env any-python
+from __future__ import print_function, absolute_import, unicode_literals
 
-set -e
+import argparse
+import os
+import subprocess
 
-# remove_user_with_group remove the given username from the system and also
-# cleans the group
-remove_user_with_group() {
-    REMOVE_USER="$1"
+# Define MYPY as False and use it as a conditional for typing import. Despite
+# this declaration mypy will really treat MYPY as True when type-checking.
+# This is required so that we can import typing on Python 2.x without the
+# typing module installed. For more details see:
+# https://mypy.readthedocs.io/en/latest/common_issues.html#import-cycles
+MYPY = False
+if MYPY:
+    from typing import Text
 
-    if [ -z "$REMOVE_USER" ]; then
-        echo "please provide a username to remove"
-        return 1
-    fi
 
-    if [ -e /var/lib/extrausers/passwd ]; then
-        userdel --extrausers --force "${REMOVE_USER}" || true
-    else
-        userdel --force --remove "${REMOVE_USER}" || true
-        # some systems do not set "USERGROUPS_ENAB yes" so we need to cleanup
+def remove_user_with_group(user_name):
+    # type: (Text) -> None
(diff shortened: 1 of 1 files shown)


zyga:
@mvo5 do we lose anything over not having shell?

tests/lib/bin/user-tool
```diff
@@ -1,42 +1,55 @@
...
-    if [ -e /var/lib/extrausers/passwd ]; then
-        userdel --extrausers --force "${REMOVE_USER}" || true
-    else
-        userdel --force --remove "${REMOVE_USER}" || true
-        # some systems do not set "USERGROUPS_ENAB yes" so we need to cleanup
+def remove_user_with_group(user_name):
+    # type: (Text) -> None
+    """remove the user and group with the same name, if present."""
+    if os.path.exists("/var/lib/extrausers/passwd"):
+        subprocess.call(["userdel", "--extrausers", "--force", user_name])
```
mvo5: I think as a followup this should also use "--remove" here. I know the original shell did not but that is a bug  in this silly shell.

tests/lib/bin/user-tool
```diff
@@ -1,42 +1,55 @@
...
-        exit 1
-esac
+        proc = subprocess.Popen(["groupdel", "-h"], stdout=subprocess.PIPE)
+        out, _ = proc.communicate()
+        if b"force" in out:
+            subprocess.call(["groupdel", "-f", user_name])
+        else:
+            subprocess.call(["groupdel", user_name])
+    # Ensure the user user really got deleted
+    if subprocess.call(["getent", "passwd", user_name]) == 0:
```
mvo5: I guess we could use check_call() here?

tests/lib/bin/user-tool
```diff
@@ -1,42 +1,55 @@
...
+    # type: () -> None
+    parser = argparse.ArgumentParser()
+    sub = parser.add_subparsers()
+    cmd = sub.add_parser(
+        "remove-with-group", description="Remove system user and group, if present"
+    )
+    cmd.set_defaults(func=lambda ns: remove_user_with_group(ns.user))
+    cmd.add_argument("user", help="name of the user and group to remove")
+    ns = parser.parse_args()
+    ns.func(ns)
```
mvo5: This is a bit confusing but probably shows I'm just rusty :/

tests/lib/bin/user-tool
zyga: I used the "manual" version to have a more precise output in case of error.

tests/lib/bin/user-tool
zyga: This takes the apparmor argument namespace, extracts the `func` argument from it, which is  a function responsible for the command that was selected, runs that function passing the namespace argument back (so that the function can have access to command line arguments). That's all.

tests/lib/bin/user-tool
zyga: I'll make that so.

zyga: This is now https://github.com/snapcore/snapd/pull/7328

PR #7315, opened 2019-08-22
store, image, cmd: make 'snap download' leave partials

chipaca:
Without this change, a network error during a 'snap download' will
remove the .partial file.

While this is probably OK for snapd, as it's the simpler & safer
route¹, it's not OK for 'snap download'; the .partial should be left
so that a following 'snap download' will resume, as the user can see
and remove the file if needed.

So, this change adds a LeavePartialOnError to store.DownloadOptions,
exports it through image.DownloadOptions, and sets that to true in
cmd_download.

Also note, it explicitly makes sure it doesn't leave _empty_ partial
files, as those would just be noise.

----

1. We might want to revisit what we do in snapstate so that there also
we can resume, but it needs careful consideraiton of available space
and removing stale partial files, which is one of the hard problems.

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -1380,17 +1380,11 @@ func (s *Store) Download(ctx context.Context, name string, targetPath string, do
 		if cerr := w.Close(); cerr != nil && err == nil {
 			err = cerr
 		}
-		if err != nil {
-			doRemove := dlOpts == nil || !dlOpts.LeavePartialOnError
-			if !doRemove {
-				// also do remove if the file is empty
-				if fi != nil && fi.Size() == 0 {
-					doRemove = true
-				}
-			}
-			if doRemove {
-				os.Remove(w.Name())
-			}
+		if err == nil {
+			return
+		}
+		if dlOpts == nil || !dlOpts.LeavePartialOnError || fi == nil || fi.Size() == 0 {
+			os.Remove(w.Name())
 		}
 	}()
 	if resume > 0 {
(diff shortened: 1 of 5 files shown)


store/store.go
```diff
@@ -1375,11 +1376,21 @@ func (s *Store) Download(ctx context.Context, name string, targetPath string, do
...
 	}
 	defer func() {
+		fi, _ := w.Stat()
 		if cerr := w.Close(); cerr != nil && err == nil {
 			err = cerr
 		}
 		if err != nil {
-			os.Remove(w.Name())
+			doRemove := dlOpts == nil || !dlOpts.LeavePartialOnError
+			if !doRemove {
```
anonymouse64: at the sake of being nitpicky, this would read a little bit easier to me if it was:

```go
			if !doRemove && fi != nil && fi.Size() == 0 {
				// also do remove if the file is empty
				doRemove = true
			}
```

it's just kind of confusing to initially see

```go
if !thing {

}
if thing {

}
```

chipaca: Yeah, I struggled a bit with that one.
I could also write it as
```go
		if err != nil && (dlOpts == nil || !dlOpts.LeavePartialOnError || (fi != nil && fi.Size() == 0)) {
			os.Remove(w.Name())
		}
```
wdyt?

anonymouse64: IMHO the `()` is a bit more confusing there to see which conditions combine with which

anonymouse64: that looks best actually, no nested paranthesis and only `||` between the checks

PR #7316, opened 2019-08-22
tests: add unit tests for cmd_whoami

ardaguclu:
Added unit tests for cmd_whoami. It would be great adding tests for logged users email address verification either. But logging in cmd_whoami might be out of responsibility of it. If there is builtin mocked logging mechanism, it will be great.


ardaguclu:
Thanks for your feedback and help. I made additional changes according to your comments.

cmd/snap/cmd_whoami_test.go
```diff
@@ -0,0 +1,54 @@
+package main_test
```
zyga: The copyright header is missing here.

cmd/snap/cmd_whoami_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (s *SnapSuite) TestWhoamiEmptyAuthFile(c *C) {
+	s.Login(c)
+
+	wErr := osutil.AtomicWriteFile(s.AuthFile, []byte(``), 0600, 0)
+	c.Assert(wErr, IsNil)
+
+	_, err := snap.Parser(snap.Client()).ParseArgs([]string{"whoami"})
+	c.Assert(err, NotNil)
+	c.Check(err.Error(), Equals, "EOF")
```
zyga: Just a style suggestion, we rarely call `.Error`, on error objects. Instead we almost always use `ErrorMatches` for that.

```suggestion
	c.Check(err, ErrorMatches, "EOF")
```

cmd/snap/cmd_whoami_test.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func (s *SnapSuite) TestWhoamiEmptyAuthFile(c *C) {
+	s.Login(c)
+
+	wErr := osutil.AtomicWriteFile(s.AuthFile, []byte(``), 0600, 0)
+	c.Assert(wErr, IsNil)
+
+	_, err := snap.Parser(snap.Client()).ParseArgs([]string{"whoami"})
+	c.Assert(err, NotNil)
```
zyga: You can drop the non-nil assertion check if you use `ErrorMatches`

cmd/snap/cmd_whoami_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+	_, err := snap.Parser(snap.Client()).ParseArgs([]string{"whoami"})
+	c.Assert(err, IsNil)
+	c.Check(s.Stdout(), Equals, "email: -\n")
+}
+
+func (s *SnapSuite) TestWhoamiExtraParamError(c *C) {
+	_, err := snap.Parser(snap.Client()).ParseArgs([]string{"whoami", "test"})
+	c.Assert(err, NotNil)
+	c.Check(err.Error(), Equals, "too many arguments for command")
```
zyga:
```suggestion
	c.Check(err, ErrorMatches, "too many arguments for command")
```

cmd/snap/cmd_whoami_test.go
```diff
@@ -0,0 +1,54 @@
...
+	_, err := snap.Parser(snap.Client()).ParseArgs([]string{"whoami"})
+	c.Assert(err, IsNil)
+	c.Check(s.Stdout(), Equals, "email: hello@mail.com\n")
+
+	s.Logout(c)
+}
+
+func (s *SnapSuite) TestWhoamiNotLoggedInUser(c *C) {
+	s.RedirectClientToTestServer(func(w http.ResponseWriter, r *http.Request) {
+		c.Errorf("whoami does not hit the API, it just reads ~/.snap/auth.json (which is mocked in the tests)")
```
zyga: I think it's fine to just say `panic("unexpected call to snapd API")` here

cmd/snap/cmd_whoami_test.go
```diff
@@ -0,0 +1,54 @@
...
+	})
+
+	s.Login(c)
+
+	_, err := snap.Parser(snap.Client()).ParseArgs([]string{"whoami"})
+	c.Assert(err, IsNil)
+	c.Check(s.Stdout(), Equals, "email: hello@mail.com\n")
+
+	s.Logout(c)
```
zyga: It is unclear what this logout is for. If it is to clean up some state it should be paired with `s.Login` call above in the style of `defer s.Logout(c)`

cmd/snap/cmd_whoami_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/snapcore/snapd/osutil"
+
+	. "gopkg.in/check.v1"
+
+	snap "github.com/snapcore/snapd/cmd/snap"
+)
+
+func (s *SnapSuite) TestWhoamiLoggedInUser(c *C) {
+	s.RedirectClientToTestServer(func(w http.ResponseWriter, r *http.Request) {
+		c.Errorf("whoami does not hit the API, it just reads ~/.snap/auth.json (which is mocked in the tests)")
```
zyga: Same remark about the error in API server as below.

cmd/snap/cmd_whoami_test.go
ardaguclu: Changed.

cmd/snap/cmd_whoami_test.go
ardaguclu: Added.

cmd/snap/cmd_whoami_test.go
```diff
@@ -0,0 +1,69 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2016 Canonical Ltd
```
zyga:
```suggestion
 * Copyright (C) 2019 Canonical Ltd
```

cmd/snap/cmd_whoami_test.go
```diff
@@ -0,0 +1,69 @@
...
+func (s *SnapSuite) TestWhoamiExtraParamError(c *C) {
+	_, err := snap.Parser(snap.Client()).ParseArgs([]string{"whoami", "test"})
+	c.Check(err, ErrorMatches, "too many arguments for command")
+}
+
+func (s *SnapSuite) TestWhoamiEmptyAuthFile(c *C) {
+	s.Login(c)
+	defer s.Logout(c)
+
+	wErr := osutil.AtomicWriteFile(s.AuthFile, []byte(``), 0600, 0)
```
zyga: You can reuse error variables. We have some checks that verify that an assignment is not _ineffectual_, that is, assigned but then never used. This allows you to write code like:
```
err := foo()
c.Assert(err, IsNil)
err = bar()
c.Assert(bar, ErrorMatches, `...`)
```
without having to invent variable names for each err.

cmd/snap/cmd_whoami_test.go
ardaguclu: Changed.

cmd/snap/cmd_whoami_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (s *SnapSuite) TestWhoamiEmptyAuthFile(c *C) {
+	s.Login(c)
+	defer s.Logout(c)
+
+	err := osutil.AtomicWriteFile(s.AuthFile, []byte(``), 0600, 0)
+	c.Assert(err, IsNil)
+
+	_, err = snap.Parser(snap.Client()).ParseArgs([]string{"whoami"})
+	c.Check(err, ErrorMatches, "EOF")
```
mvo5: (nitpick) this would be io.EOF I think - but thats fine :)

PR #7321, opened 2019-08-23
tests: restore dpkg selections after upgrade-from-2.15 test

zyga:
While investigating other test issue I briefly looked at how tests
affect the set of installed packages and noticed a huge spike in the
upgrade-from-2.15 test. While I don't have the data on hand anymore I
have a patch that I wrote at the time, to save and restore dpkg
selections.

### DIFF
--- a/tests/lib/bin/apt-tool
+++ b/tests/lib/bin/apt-tool
@@ -1,0 +1,36 @@
+#!/usr/bin/python3
+
+import sys
+
+import apt
+
+
+def checkpoint():
+    pkgs = set()
+    cache = apt.Cache()
+    for pkg in cache:
+        if pkg.is_installed:
+            pkgs.add(pkg.name)
+    for name in sorted(pkgs):
+        print(name)
+
+
+def restore(fname):
+    desired = set([line.strip() for line in open(fname)])
+            
+    cache = apt.Cache()
+    for pkg in cache:
+        if pkg.is_installed and not pkg.name in desired:
+            print("removing", pkg)
+            pkg.mark_delete(auto_fix=False)
+        if not pkg.is_installed and pkg.name in desired:
+            print("installing", pkg)
+            pkg.mark_install(auto_fix=False, auto_install=False)
+    cache.commit()
+
+
+if __name__ == "__main__":
+    if sys.argv[1] == "checkpoint":
+        checkpoint()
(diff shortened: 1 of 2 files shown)


tests/main/upgrade-from-2.15/task.yaml
```diff
@@ -16,6 +17,9 @@ restore: |
     fi
     distro_install_build_snapd
 
+    dpkg --set-selections < selections
+    rm -f selections
```
mvo5: I'm not super sure what the intension of this is but as its written now nothing will happen, i.e. dpkg will not perform any installs/removes. If you want to perform actions based on this you need to do a `apt-get dselect-upgrade -y`. But even then I'm not sure this is actually what you want, most likely nothing will happen still because new packages will not be marked as "deinstall".

zyga:
We discussed this with @mvo5 on IRC and not only is the approach wrong (dpkg --set-selections doesn't do what I though it does) it requires some more apt love to work correctly. Closing for now, we'll iterate elsewhere.

tests/main/upgrade-from-2.15/task.yaml
```diff
@@ -3,6 +3,7 @@ summary: Ensure upgrades from snapd 2.15 work
 systems: [ubuntu-16.04-64]
 
 prepare: |
+    apt-tool checkpoint > installed.pkgs
```
zyga: Nitpick: taking an argument with filename vs printing the output is inconsistent.

tests/lib/bin/apt-tool
```diff
@@ -0,0 +1,36 @@
...
+            print("installing", pkg)
+            pkg.mark_install(auto_fix=False, auto_install=False)
+    cache.commit()
+
+
+if __name__ == "__main__":
+    if sys.argv[1] == "checkpoint":
+        checkpoint()
+    elif sys.argv[1] == "restore":
+        restore(sys.argv[2])
```
stolowski: else `exit(1)` to ensure we don't miss typos in the tests?

zyga: I'll follow up with argparse-based command line handling so that this part is accounted for. If you agree with that we can land and Ill iterate.

stolowski: Sure.

PR #7336, opened 2019-08-26
tests: add debug section to interfaces-contacts-service

mvo5:
The cleanup of this tests fails sometimes in opensuse-tumbleweed-64
when trying to "rm -rf $XDG" - so something is writing this when
rm runs.

To understand a bit better what is happening this PR adds a list
of the processes and the content of XDG to the debug output.

### DIFF
--- a/tests/main/interfaces-contacts-service/task.yaml
+++ b/tests/main/interfaces-contacts-service/task.yaml
@@ -20,6 +20,12 @@ environment:
     XDG_DATA_HOME: $XDG/share
     XDG_CACHE_HOME: $XDG/cache
 
+debug: |
+    echo "Output process to see what might write to $XDG"
+    ps uafx
+    echo "Show what is in $XDG"
+    ls -alR "$XDG"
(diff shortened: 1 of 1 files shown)


tests/main/interfaces-contacts-service/task.yaml
```diff
@@ -20,6 +20,12 @@ environment:
     XDG_DATA_HOME: $XDG/share
     XDG_CACHE_HOME: $XDG/cache
 
+debug: |
+    echo "Output process to see what might write to $XDG"
+    ps uafx
+    echo "Show what is in $XDG"
+    ls -alR "$XDG"
```
stolowski: Would it be useful to show `lsof` output here (possibly filtered by grep "$XDG")?

mvo5: Probably, I hope we can remove this debug stanza again soon once we figure out what is going on with the failure. But a good idea

PR #7340, opened 2019-08-26
tests: adding support for arm devices on ubuntu-core-device-reg test

sergiocazzolato:
This is because the test is failing on arm devices. 

Error: 
+ snap known serial
+ MATCH 'model: pc'
type: serial
authority-id: canonical
brand-id: canonical
model: pi2

### DIFF
--- a/tests/main/ubuntu-core-device-reg/task.yaml
+++ b/tests/main/ubuntu-core-device-reg/task.yaml
@@ -5,6 +5,8 @@ summary: |
 systems: [ubuntu-core-1*]
 
 execute: |
+    #shellcheck source=tests/lib/names.sh
+    . "$TESTSLIB"/names.sh
     #shellcheck source=tests/lib/systems.sh
     . "$TESTSLIB"/systems.sh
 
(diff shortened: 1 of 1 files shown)


tests/main/ubuntu-core-device-reg/task.yaml
```diff
@@ -5,6 +5,8 @@ summary: |
 systems: [ubuntu-core-1*]
 
 execute: |
+    #shellcheck source=tests/lib/names.sh
+    . "$TESTSLIB"/names.sh
```
mvo5: Silly question, why do we need this input?

tests/main/ubuntu-core-device-reg/task.yaml
```diff
@@ -25,10 +27,12 @@ execute: |
     echo "Check we have a serial"
     snap known serial|MATCH "authority-id: canonical"
     snap known serial|MATCH "brand-id: canonical"
-    if is_core18_system; then
```
zyga: Slight preference for switch-case here.

tests/main/ubuntu-core-device-reg/task.yaml
mvo5: Aha, right, sorry, I overlooked this!

PR #7343, opened 2019-08-26
tests: moving tests to nightly suite

sergiocazzolato:
Moving tests to nightly test suite:
 . sbuild
 . interfaces-openvswitch
 . install-snaps

Also the restriction for systems is removed.


tests/nightly/interfaces-openvswitch/task.yaml
```diff
@@ -16,19 +15,20 @@ details: |
...
     . "$TESTSLIB/pkgdb.sh"
-
+    
     echo "Given openvswitch is installed"
     distro_install_package --no-install-recommends openvswitch-switch
 
     # Ensure the openvswitch service is started which isn't the case by
     # default on all distributions
-    systemctl enable --now openvswitch
+    if find / -name openvswitch-switch.service | MATCH "openvswitch-switch.service"; then
```
zyga: Isn't this a bit brute-force? Can we use something that systemd offers to check if a service exists?

stolowski: I agree with zyga,`systemctl status..` should help here.

PR #7344, opened 2019-08-27
snapstate: use strutil.VersionCompare() in checkVersion()

mvo5:
The code in checkVersion() to compare version numbers is
hand written. We actually don't need this because the snapd
versions are following the debian version numbering schema
so we can comapre them using "strutil.VersionCompare()".

### DIFF
--- a/overlord/snapstate/check_snap.go
+++ b/overlord/snapstate/check_snap.go
@@ -22,7 +22,6 @@ package snapstate
 import (
 	"fmt"
 	"regexp"
-	"strconv"
 	"strings"
 
 	"github.com/snapcore/snapd/arch"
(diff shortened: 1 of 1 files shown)


mvo5:
@stolowski This is covered via TestCheckSnapAssumes in check_snap_test.go. I think the coverage is ok but happy to add more if needed.

stolowski:
> @stolowski This is covered via TestCheckSnapAssumes in check_snap_test.go. I think the coverage is ok but happy to add more if needed.

Ah, indeed, I looked there but it wasn't immediately clear, thank you, it's fine then.

mvo5:
Looking at this again this actually changes the semantic slightly, I will close this and add another PR with a test that shows it.

PR #7347, opened 2019-08-27
gadget: do not error on gadget refreshes with multiple volumes

mvo5:
The new gadget asserts update code will check early if the gadget
contains multiple volumes. This is not supported yet and the old
code did error. However this has the unintended consequence that
any gadget update fails now when the gadget has multiple volumes
even if no gadget asset update will happen.

For now the code will just log a warning that multi-volume
gadget asset updates are not supported and continue like
before.

### DIFF
--- a/gadget/update.go
+++ b/gadget/update.go
@@ -57,6 +57,15 @@ type GadgetData struct {
 // rollback directory. Should the apply step fail, the modified data is
 // recovered.
 func Update(old, new GadgetData, rollbackDirPath string) error {
+	// TODO: support multi-volume gadgets. But for now we simply
+	//       do not do any gadget updates on those. We cannot error
+	//       here because this would break refreshes of gadgets even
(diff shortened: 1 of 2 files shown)


gadget/update.go
```diff
@@ -57,6 +57,15 @@ type GadgetData struct {
 // rollback directory. Should the apply step fail, the modified data is
 // recovered.
 func Update(old, new GadgetData, rollbackDirPath string) error {
+	// TODO: support multi-volume gadgets. But for now we simply
+	//       do not do any gadget updates on those. We cannot error
+	//       here because this would break refreshes of gadgets even
+	//       when they don't require any updates.
+	if len(new.Info.Volumes) != 1 || len(old.Info.Volumes) != 1 {
```
zyga:
```suggestion
	if len(new.Info.Volumes) > 1 || len(old.Info.Volumes) > 1 {
```

PR #7349, opened 2019-08-27
 snap: revert PR#7204 as breaks `snap switch` output

mvo5:
The PR#7204 has the unintended consequence that on "snap switch"
a misleading message is printed:
```
$ snap switch test-snapd-tools --edge
"test-snapd-tools" switched to the "edge" channel

Channel edge for test-snapd-tools is closed; temporarily forwarding to stable.
```

So this PR adds a test and reverts the PR#7204 so that we can
release 2.41.


stolowski:
I proposed #7357 which should fix it.

mvo5:
Closing in favor of  #7357

PR #7351, opened 2019-08-27
tests: retry checking until the written file on desktop-portal-filechooser

sergiocazzolato:
The test is failing sporadically when checking the text in the
/tmp/file-to-write.txt file. The text is written be sometimes takes a
bit of time making the test fail because of this.

### DIFF
--- a/tests/main/desktop-portal-filechooser/task.yaml
+++ b/tests/main/desktop-portal-filechooser/task.yaml
@@ -52,15 +52,21 @@ execute: |
 
     echo "The confined application can write files via the portal"
     [ ! -f /tmp/file-to-write.txt ]
+
+    # TODO: workaround for testing on 18.04
     # The python code does open(path, 'w'), which attempts to truncate the
     # file if it exists. Then in fuse handlers inside document-portal, the
     # code path for when the inode exists and the caller requested O_TRUNC,
     # returns -ENOSYS, resulting in OSError:
     # [Errno 38] Function not implemented on the Python side
-    # make sure the file does not exist
-    rm -f /tmp/file-to-write.txt
+    # Creating the file beforehand prevents that.
+    touch /tmp/file-to-write.txt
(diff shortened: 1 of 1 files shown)


tests/main/desktop-portal-filechooser/task.yaml
```diff
@@ -59,7 +59,13 @@ execute: |
     touch /tmp/file-to-write.txt
     chown test:test /tmp/file-to-write.txt
     as_user test-snapd-portal-client save-file "from-sandbox"
```
zyga: Since this is not running in the background I think it is highly fishy that a write can finish without being visible here.

CC @jhenstridge for ideas.

jhenstridge: This is surprising to me.  Looking at the underlying code, a close of the fuse file descriptor should also close the underlying file.  Is it just one particular distro release we're seeing the failure on?

zyga: I don't have the data on hand to answer. I will record the failures as I see them starting from now,.

sergiocazzolato: @jhenstridge @zyga 
These are the last failures during the last 2 weeks for this test and all of them are happening on ubuntu-18.04-64:
https://travis-ci.org/snapcore/snapd/jobs/578664831#L3446
https://travis-ci.org/snapcore/snapd/jobs/578176116#L6804
https://travis-ci.org/snapcore/snapd/jobs/576845090#L1288
https://travis-ci.org/snapcore/snapd/jobs/575899118#L6118

sergiocazzolato: 2 errors today always on ubuntu-18.04-64:
https://travis-ci.org/snapcore/snapd/jobs/580112858#L8998
https://travis-ci.org/snapcore/snapd/jobs/580073815#L6135

PR #7353, opened 2019-08-27
tests: simplify interfaces-account-control test

zyga:
The test used some weird way to run the commands, using their full
path instead of directly or via "snap run snap.app".

### DIFF
--- a/tests/main/interfaces-account-control/task.yaml
+++ b/tests/main/interfaces-account-control/task.yaml
@@ -27,11 +27,8 @@ restore: |
     snap remove "$TSNAP"
 
 execute: |
-    #shellcheck source=tests/lib/dirs.sh
-    . "$TESTSLIB"/dirs.sh
-
-    "$SNAP_MOUNT_DIR"/bin/"$TSNAP".useradd --extrausers alice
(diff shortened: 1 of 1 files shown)


tests/main/interfaces-account-control/task.yaml
```diff
@@ -27,11 +27,8 @@ restore: |
     snap remove "$TSNAP"
 
 execute: |
-    #shellcheck source=tests/lib/dirs.sh
-    . "$TESTSLIB"/dirs.sh
-
-    "$SNAP_MOUNT_DIR"/bin/"$TSNAP".useradd --extrausers alice
-    echo alice:password | "$SNAP_MOUNT_DIR"/bin/"$TSNAP".chpasswd
```
stolowski: Oh wow 🙀

tests/main/interfaces-account-control/task.yaml
```diff
@@ -27,11 +27,8 @@ restore: |
...
+    snap run "$TSNAP".useradd --extrausers alice
```
mvo5: Silly question - why do we have the "snap run" here at all? Could we simply use `"$TSNAP".useradd` here?

PR #7357, opened 2019-08-27
cmd/snap: fix snap switch message

stolowski:
Do not to display the "Channel .. for .. is closed; temporarily forwarding to .." message with `snap switch`.

In the PR #7204 I refactored snap switch slightly to reuse existing `showDone` helper and missed the fact that this helper also prints this message under certain circumstances, apart from all the specific cases based on snap ops. With this change the output for `snap switch` will be as before #7204, except the new track/semantics are handled (thanks to `showDone` logic).

Thanks @mvo5 for the spread test.

### DIFF
--- a/cmd/snap/cmd_snap_op.go
+++ b/cmd/snap/cmd_snap_op.go
@@ -369,16 +369,15 @@ func showDone(cli *client.Client, names []string, op string, opts *client.SnapOp
 				msg = fmt.Sprintf(i18n.G("%q left the cohort"), snap.Name)
 			}
 			fmt.Fprintln(Stdout, msg)
-			// go to next snap, skipping the temporary forward logic below for "switch" op.
-			continue
 		default:
 			fmt.Fprintf(Stdout, "internal error: unknown op %q", op)
 		}
-		// XXX: this makes sense only for certain operations (refresh?), but not for "switch"
-		if snap.TrackingChannel != snap.Channel && snap.Channel != "" {
-			if sameRisk, err := isSameRisk(snap.TrackingChannel, snap.Channel); err == nil && !sameRisk {
-				// TRANSLATORS: first %s is a channel name, following %s is a snap name, last %s is a channel name again.
-				fmt.Fprintf(Stdout, i18n.G("Channel %s for %s is closed; temporarily forwarding to %s.\n"), snap.TrackingChannel, snap.Name, snap.Channel)
+		if op == "install" || op == "refresh" || op == "revert" {
+			if snap.TrackingChannel != snap.Channel && snap.Channel != "" {
+				if sameRisk, err := isSameRisk(snap.TrackingChannel, snap.Channel); err == nil && !sameRisk {
+					// TRANSLATORS: first %s is a channel name, following %s is a snap name, last %s is a channel name again.
+					fmt.Fprintf(Stdout, i18n.G("Channel %s for %s is closed; temporarily forwarding to %s.\n"), snap.TrackingChannel, snap.Name, snap.Channel)
+				}
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


cmd/snap/cmd_snap_op.go
```diff
@@ -369,9 +369,12 @@ func showDone(cli *client.Client, names []string, op string, opts *client.SnapOp
 				msg = fmt.Sprintf(i18n.G("%q left the cohort"), snap.Name)
 			}
 			fmt.Fprintln(Stdout, msg)
+			// go to next snap, skipping the temporary forward logic below for "switch" op.
```
mvo5: I think we should expand the existing unit test a wee bit, something like:
```
diff --git a/cmd/snap/cmd_snap_op_test.go b/cmd/snap/cmd_snap_op_test.go
index 54212caad5..d95eab704e 100644
--- a/cmd/snap/cmd_snap_op_test.go
+++ b/cmd/snap/cmd_snap_op_test.go
@@ -1890,7 +1890,9 @@ func (s *SnapOpSuite) TestSwitchHappy(c *check.C) {
        rest, err := snap.Parser(snap.Client()).ParseArgs([]string{"switch", "--beta", "foo"})
        c.Assert(err, check.IsNil)
-       c.Check(s.Stdout(), check.Matches, `(?sm).*"foo" switched to the "beta" channel`)
+       c.Check(s.Stdout(), check.Equals, `"foo" switched to the "beta" channel
+
+`)
        // ensure that the fake server api was actually hit
        c.Check(s.srv.n, check.Equals, s.srv.total)
```

cmd/snap/cmd_snap_op.go
```diff
@@ -369,9 +369,12 @@ func showDone(cli *client.Client, names []string, op string, opts *client.SnapOp
...
+			continue
```
mvo5: I'm really sitting on the fence here - I don't really like the "continue" here as it feels a bit magic. I wonder if something like: http://paste.ubuntu.com/p/kGHWBFjJ97/ would be simpler? Maybe @chipaca has an opinion?

cmd/snap/cmd_snap_op.go
```diff
@@ -369,9 +369,12 @@ func showDone(cli *client.Client, names []string, op string, opts *client.SnapOp
...
 		default:
 			fmt.Fprintf(Stdout, "internal error: unknown op %q", op)
 		}
+		// XXX: this makes sense only for certain operations (refresh?), but not for "switch"
```
chipaca: Hmm. It only makes sense for install and refresh... right now for example `revert` could also trigger it and there's no "forwarding" going on there either.
As we're wanting to refactor it asap, probably before default channels, I think it makes sense to keep this particular change minimal. As such, adding a check for `op == "install" || op == "refresh"` before we check whether we need to print the notice would be best, even if introducing helper functions would be cleaner.

cmd/snap/cmd_snap_op.go
chipaca: I'm not sure I see why your suggestion is better.
In any case, ideally checks to Stdout would use the wrapping-aware checkers from testutil, but I'm not sure it's a win for this PR.

mvo5: This one is using "equals" vs "matches", I think this is important because the old test did not catch the regression because the contains was still fine but an equals would have caught this.

cmd/snap/cmd_snap_op.go
mvo5: Sounds good to me, so lets go with that!

stolowski: Thanks for the suggestions, done, although note, there was a unit test for revert that expected the "fowarding" message. I assume the test was wrong then @chipaca ?

chipaca: yeah, I think it was wrong. From a particular point of view it was technically correct, but there's no point in bringing that to the attention of the user at that time as it has nothing to do with what snapd just did on their behalf. Also `revert` doesn't check the store so there's no way of knowing if the "forwarding" is still a thing at that point. So, yeah, no.

stolowski: Good, thanks for clarification.

PR #7358, opened 2019-08-27
cmd/system-shutdown: Rename die() -> die_reboot() to fix build failures with LTO enabled

dcermak:
`void die(const char *msg)` from `cmd/system-shutdown/system-shutdown-utils.h` clashes with `void die(const char *fmt, ...)` from `cmd/libsnap-confine-private/utils.h`, which turns into an error when you build snap with link time optimization enabled.

This patch renames `die` to `die_reboot`, to make it clearer what it does and to resolve this ambiguity.

PR #7362, opened 2019-08-27
cmd: unify die() across C programs

zyga:
Historically we inherited die() from ubuntu-core-launcher. It underwent a few
small changes but otherwise retained its function for print-and-quit helper.
This was true until system-shutdown was first implemented. There die existed as
a function with the same purpose but different implementation. The changes were
motivated by desire to fit the environment in which system-shutdown executes:
it runs as the last PID of the system and should not just exit. Therefore the
implementation called sync and reboot. The output of the message was also
formatted differently to fit other messages printed by the system at the same
time.

At some point, system-shutdown started to use all of libsnap-confine-private,
not just a single module. This presented a problem because now die() was
defined twice. Oddly, this went unnoticed until
https://github.com/snapcore/snapd/pull/7358 was opened, trying to resolve the
problem.

After some inspection I decided to implement the following solution. There is
only one die(), it always does what it says on the tin. Die is now implemented
with a new function, sc_panic(), which is just like die but can be tweaked in
two ways: the way the error is printed and the way the process is exited. This
is exactly what system-shutdown needed to change.

The panic module comes with simple tests and documentation.

### DIFF
--- a/cmd/libsnap-confine-private/panic.c
+++ b/cmd/libsnap-confine-private/panic.c
@@ -1,0 +1,67 @@
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "panic.h"
+
+#include <errno.h>
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+static sc_panic_exit_fn panic_exit_fn = NULL;
+static sc_panic_msg_fn panic_msg_fn = NULL;
+
+void sc_panic(const char *fmt, ...) {
+    va_list ap;
+    va_start(ap, fmt);
+    sc_panicv(fmt, ap);
+    va_end(ap);
+}
+
+void sc_panicv(const char *fmt, va_list ap) {
+    int errno_copy = errno;
+
+    if (panic_msg_fn != NULL) {
+        panic_msg_fn(fmt, ap, errno_copy);
+    } else {
+        vfprintf(stderr, fmt, ap);
+        if (errno != 0) {
+            fprintf(stderr, ": %s\n", strerror(errno_copy));
+        } else {
+            fprintf(stderr, "\n");
+        }
+    }
+
+    if (panic_exit_fn != NULL) {
+        panic_exit_fn();
+    }
+    exit(1);
+}
+
+sc_panic_exit_fn sc_set_panic_exit_fn(sc_panic_exit_fn fn) {
+    sc_panic_exit_fn old = panic_exit_fn;
+    panic_exit_fn = fn;
+    return old;
+}
+
+sc_panic_msg_fn sc_set_panic_msg_fn(sc_panic_msg_fn fn) {
+    sc_panic_msg_fn old = panic_msg_fn;
+    panic_msg_fn = fn;
+    return old;
+}
--- a/cmd/system-shutdown/system-shutdown.c
+++ b/cmd/system-shutdown/system-shutdown.c
@@ -46,7 +46,7 @@ static void show_error(const char *fmt, va_list ap, int errno_copy)
 	fprintf(stderr, "\n");
 }
 
-static void sync_and_reboot(void)
+static void sync_and_halt(void)
 {
 	sync();
 	reboot(RB_HALT_SYSTEM);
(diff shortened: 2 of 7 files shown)


cmd/libsnap-confine-private/panic.c
```diff
@@ -0,0 +1,67 @@
...
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+static sc_panic_exit_fn panic_exit_fn = NULL;
+static sc_panic_msg_fn panic_msg_fn = NULL;
+
+void sc_panic(const char *fmt, ...) {
```
mvo5: Quick drive-by: maybe some else than "panic"? Panic smells a lot like a go panic() which will behave differently AFAICT (print stacktrace etc).

zyga: Well, the fact that it prints a stack trace isn't the essential feature. We could also print one if that would make you feel better. I'm somewhat -1 on this precisely because of error + panic parity that C and Go can have.

mvo5: I'm not arguing in favour of adding a stacktrace to the C code. Just that its confusing to use the same name for something different without any advantage in doing so (at least I don't see one).

zyga: I argue that the principal function of `sc_panic` is the same as `panic`, to abort execution without retaining regular control flow.

mvo5:
Thanks for the PR - the solution looks interesting. But I think its likely that the shutdown utility will go away with UC20 and its also really rarely touched (last real change ~1y ago). So I would prefer if  we would go with the simpler fix in #7358. I think the use-case(s) we have does not justfiy the extra code/API.

zyga:
@mvo5 that fix is incorrect, we should reboot, not exit in `system-shutdown`

mvo5:
One critical piece of information that I was missing when I suggested to go with 7358 is that the helpers in "libsnapconfine-private" may call "die()" which will cause a program exit. But when die() is called in the context of system-shutdown we need it to behave differently. So I need to recheck this with fresh eyes - sorry that I did not realize this before.

cmd/libsnap-confine-private/panic.c
```diff
@@ -0,0 +1,67 @@
...
+            fprintf(stderr, ": %s\n", strerror(errno_copy));
+        } else {
+            fprintf(stderr, "\n");
+        }
+    }
+
+    if (panic_exit_fn != NULL) {
+        panic_exit_fn();
+    }
+    exit(1);
```
bboozzoo: Probably a stretch, but if it's a panic, perhaps it'd be useful to be able to chose via some flags whether to die cleanly (like now) or abort() and potentially have a core dump generated.

mvo5: Fwiw, I still don't like panic() as it sounds too close to go and abort()

bboozzoo: terminate perhaps?

zyga: @bboozzoo but that's the whole point of this feature - you can, you can exit from your own handler. This exit function is to ensure that under no circumstances the `noreturn` nature of the function is guaranteed.

zyga: FYIW: I'll stick my guns to calling it panic ;-)

bboozzoo: @mvo5 how do the proposals for far sound?

cmd/system-shutdown/system-shutdown.c
```diff
@@ -33,9 +33,30 @@
...
+	fprintf(stderr, "snapd system-shutdown helper: ");
+	fprintf(stderr, "*** ");
+	vfprintf(stderr, fmt, ap);
+	if (errno_copy != 0) {
+		fprintf(stderr, ": %s", strerror(errno_copy));
+	}
+	fprintf(stderr, "\n");
+}
+
+static void sync_and_reboot(void)
```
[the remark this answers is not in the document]
zyga: Huh, good point. I didn't notice we halt vs reboot.

cmd/system-shutdown/system-shutdown.c
```diff
@@ -33,9 +33,30 @@
 
 #include "system-shutdown-utils.h"
 #include "../libsnap-confine-private/string-utils.h"
+#include "../libsnap-confine-private/panic.h"
+
+static void show_error(const char *fmt, va_list ap, int errno_copy)
+{
+	fprintf(stderr, "snapd system-shutdown helper: ");
+	fprintf(stderr, "*** ");
```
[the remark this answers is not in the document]
zyga: What do you mean by "show kmsg"? I think the output is identical (look at what kmsg does).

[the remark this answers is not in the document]
stolowski: Right, I naively assumed kmsg(..) writes to `/dev/kmsg` because... it's named kmsg() :). It's fine then.

cmd/system-shutdown/system-shutdown.c
```diff
@@ -33,9 +33,30 @@
...
+static void sync_and_reboot(void)
+{
+	sync();
+	reboot(RB_HALT_SYSTEM);
+}
 
 int main(int argc, char *argv[])
 {
+	sc_set_panic_msg_fn(show_error);
+	sc_set_panic_exit_fn(sync_and_reboot);
```
zyga:
```suggestion
	sc_set_panic_exit_fn(sync_and_halt);
```

cmd/system-shutdown/system-shutdown.c
zyga:
```suggestion
static void sync_and_halt(void)
```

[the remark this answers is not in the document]
zyga: Fixed now.

PR #7367, opened 2019-08-28
snap, cmd/snap: support (but warn) using deprecated multi-slash channel

chipaca:
if `$FOO` was `""`.

of behaviour.

break our users. So...

will come in a separate PR.


cmd/snap/cmd_snap_op.go
```diff
@@ -252,10 +252,20 @@ func (mx *channelMixin) setChannelFromCommandline() error {
...
-		mx.Channel += "/stable"
+	if mx.Channel != "" {
+		ch, err := snap.ParseChannel(mx.Channel, "")
+		if err != nil {
+			if _, ok := err.(snap.DeprecatedChannelStringError); !ok {
+				return err
+			}
+			// TODO: get escapes in here so we can bold the Warning
+			head := i18n.G("Warning:")
+			warn := fill(fmt.Sprintf(i18n.G("Specifying a channel %q is relying on undefined behaviour. Interpreting it as %q for now, but this will change.\n"), mx.Channel, ch.Full()), utf8.RuneCountInString(head)+1) // +1 for the space
```
pedronis: along this path ch.Name and ch.Full() are the same, right?

cmd/snap/cmd_snap_op_test.go
```diff
@@ -1302,6 +1292,22 @@ func (s *SnapOpSuite) TestRefreshOneRebooting(c *check.C) {
 
 }
 
+func (s *SnapOpSuite) TestRefreshOneChanDeprecated(c *check.C) {
+	s.RedirectClientToTestServer(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, `{"type": "error", "result": {"message": "snap not found", "value": "foo", "kind": "snap-not-found"}, "status-code": 404}`)
```
pedronis: we should check the channel we get here from the client, no?

cmd/snap/cmd_snap_op_test.go
```diff
@@ -1302,6 +1292,22 @@ func (s *SnapOpSuite) TestRefreshOneRebooting(c *check.C) {
...
+	})
+
+	for in, out := range map[string]string{
+		"/foo":    "foo/stable",
+		"/stable": "latest/stable",
```
mvo5: Should we add something like `		"///foo/stable//": "foo/stable",` here too ? I mean, something that also has extra slashes at the end or something like this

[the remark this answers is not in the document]
chipaca: yeah, i'll fix that

cmd/snap/cmd_snap_op_test.go
chipaca: no, this'll print `latest/stable` instead of `stable` for `/stable`; I'm fixing that now.

cmd/snap/cmd_snap_op.go
chipaca: no, this'll print `latest/stable` instead of `stable` for `/stable`; I'm fixing that now.

cmd/snap/cmd_snap_op_test.go
chipaca: done

cmd/snap/cmd_snap_op_test.go
chipaca: done

cmd/snap/cmd_snap_op.go
chipaca: no, given `"/stable"`, `Name` is `"stable"` and `Full()` is `"latest/stable"`. After discussing (and misunderstanding, and realising and fixing it) on IRC, we want `Full()` everywhere for this case.

chipaca:
sigh. The git shenanigans I pulled to rebase this have confused github. Giving up on it and pushing anew.

PR #7379, opened 2019-08-30
tests: add version-tool for comparing versions

zyga:
I need to start comparing bash and kernel versions. I tried to hack
around in shell but it was murky and probably buggy as well. I decided
to thus write a small version comparison tool in python.

The tool can be used somewhat like the builtin "test" command. It takes
two versions, a relationship operator and an algorithm to use. An
example shell snippet using it might look like this:

    if version-tool --naive "$(uname -r)" -ge 4.13; then
        ...
    fi

Currently only one algorithm exist: naive, which is just enough to
compare kernel version numbers or bash version numbers. It looks through
the segments of each version, separated by dot, converts as much of the
segment to an integer as possible and computes the delta. Missing
segments are like segments reading zero, so 1.0 is equal to 1 but 1.2 is
smaller than 1.2.1. The tool comes with full array of unit tests and
documentation and is compatible with Python 3 and Python 2 alike.

### DIFF
--- a/tests/lib/bin/version-tool
+++ b/tests/lib/bin/version-tool
@@ -122,95 +122,6 @@ class ConsistentRelationTests(unittest.TestCase):
             consistent_relation("???", 0)
 
 
-def numeric_part_of(text):
-    # type: (Text) -> int
-    """
-    numeric_part_of returns the integer representing the numeric portion of text.
-
-    If text does not start with an natural number the result is zero. The
-    function never fails. Negative numbers are not supported because they are
-    meaningless for version numbers.
-    """
-    match = re.match("^([0-9]+)", text)
-    if match:
-        return int(match.group(1))
-    return 0
-
-
-class NumericPartOfTests(unittest.TestCase):
-    def test_simple(self):
-        # type: () -> None
-        self.assertEqual(numeric_part_of("12"), 12)
-        self.assertEqual(numeric_part_of("12-stuff"), 12)
-        self.assertEqual(numeric_part_of("stuff-12"), 0)
-        self.assertEqual(numeric_part_of("-12"), 0)
-
-
-def naive_version_cmp(a, b):
-    # type: (Text, Text) -> int
-    """
-    naive_version_cmp naively compares two version numbers.
-
-    The algorithm considers each version to be a tuple of integers. Non,
-    integer elements or element fragments are discarded.
-
-    Comparison is performed on by considering the leftmost element in each
-    tuple. First pair of numbers that are not equal determine the result of the
-    comparison. Tuples have unequal length then missing elements are
-    substituted with zero.
-
-    The return value is 0 if the version strings are equal, -1 if version a is
-    smaller or +1 if version b is smaller.
-    """
-    a_items = a.split(".")
-    b_items = b.split(".")
-    if PY2:
-        zip_longest_fn = itertools.izip_longest
-    else:
-        zip_longest_fn = itertools.zip_longest
-    for a_item, b_item in zip_longest_fn(a_items, b_items, fillvalue="0"):
-        a_val = numeric_part_of(a_item)
-        b_val = numeric_part_of(b_item)
-        delta = a_val - b_val
-        if delta != 0:
-            return 1 if delta > 0 else -1
-    return 0
-
-
-class NaiveVersionCmpTests(unittest.TestCase):
-    def test_simple(self):
-        # type: () -> None
-        self.assertEqual(naive_version_cmp("10", "10"), 0)
-        self.assertEqual(naive_version_cmp("10", "20"), -1)
-        self.assertEqual(naive_version_cmp("20", "10"), +1)
-
-    def test_many_segments(self):
-        # type: () -> None
-        self.assertEqual(naive_version_cmp("1.2.3", "1.2.3"), 0)
-        self.assertEqual(naive_version_cmp("1.2.3", "1.3.4"), -1)
-        self.assertEqual(naive_version_cmp("1.4.3", "1.2.3"), +1)
-        self.assertEqual(naive_version_cmp("1.0.0", "1.1.0"), -1)
-        self.assertEqual(naive_version_cmp("0.1.2", "1.1.2"), -1)
-
-    def test_unequal_length(self):
-        # type: () -> None
-        self.assertEqual(naive_version_cmp("1", "1.0"), 0)
-        self.assertEqual(naive_version_cmp("1", "1.2"), -1)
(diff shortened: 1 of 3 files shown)


zyga:
@mvo5 mainly because I wanted something standalone and not muddled by the complexity of `snap` startup.

tests/lib/bin/version-tool
```diff
@@ -0,0 +1,323 @@
...
+        self.assertTrue(consistent_relation("<=", 0))
+        self.assertFalse(consistent_relation("<=", +1))
+
+    def test_unknown(self):
+        # type: () -> None
+        with self.assertRaises(ValueError):
+            consistent_relation("???", 0)
+
+
+def numeric_part_of(text):
```
anonymouse64: a comment in this function explaining it doesn't work for negative numbers would be nice

tests/lib/bin/version-tool
```diff
@@ -0,0 +1,323 @@
...
+        # type: () -> None
+        self.assertEqual(naive_version_cmp("10", "10"), 0)
+        self.assertEqual(naive_version_cmp("10", "20"), -1)
+        self.assertEqual(naive_version_cmp("20", "10"), +1)
+
+    def test_many_segments(self):
+        # type: () -> None
+        self.assertEqual(naive_version_cmp("1.2.3", "1.2.3"), 0)
+        self.assertEqual(naive_version_cmp("1.2.3", "1.3.4"), -1)
+        self.assertEqual(naive_version_cmp("1.4.3", "1.2.3"), +1)
```
anonymouse64: I think you should also add test cases where the first elements of versions is the same and the penultimate elements of versions are different, respectively i.e.
```python
        self.assertEqual(naive_version_cmp("1.0.0","1.1.0"), -1)
        self.assertEqual(naive_version_cmp("0.1.2","1.1.2"), -1)
```
I've seen naive version comparison algorithms get confused by these kinds of cases

tests/lib/bin/version-tool
```diff
@@ -0,0 +1,323 @@
...
+        "-ge",
+        action="store_const",
+        const=">=",
+        dest="rel_op",
+        help="test that version-a is greater than or equal to version-b",
+    )
+    rel_op_grp.add_argument(
+        "-lt",
+        action="store_const",
+        const="<=",
```
anonymouse64: this should be `<`
```suggestion
        const="<",
```

tests/lib/bin/version-tool
```diff
@@ -0,0 +1,323 @@
...
+
+def main():
+    # type: () -> None
+    parser = _make_parser()
+    opts = parser.parse_args()
+    if opts.algorithm == "naive":
+        delta = naive_version_cmp(opts.version_a, opts.version_b)
+    else:
+        parser.error("unsupported delta algorithm")
+    is_consistent = consistent_relation(opts.rel_op, delta)
```
anonymouse64: maybe this is just because I haven't read lots of python in a while, but it is a bit odd to declare `delta` within the `if` above, but then use it outside of that branch, no?

tests/lib/bin/version-tool
zyga: This made me realise we're missing one last layer of tests: spread. Added now, will push after. local run.

tests/lib/bin/version-tool
zyga: Added

tests/lib/bin/version-tool
zyga: I don't know, it seems natural to me, this is how python scope works and how I would continue to write code unless I could use Python 3.6+ variable declarations.

tests/lib/bin/version-tool
zyga: Done

tests/lib/bin/version-tool
anonymouse64: sure it's fine as is I'll trust your judgment

zyga:
I discussed this with mvo and I will make two improvements:
 - make the naive version strict
 - add output so that yes-no-error tri-outcome is possible

mvo5:
Thanks @zyga ! I agree, I think we want to check that our versions look like kernel or bash versions and (ideally) error if they don't. Also I think kernel/bash after the first non version fragment because it becomes questionable what things mean then, like `1b1 gt 1a1` is a perfectly valid fragment in the Debian version schema and the result is "true" but in the naive schema its not and I think its great to be strict instead and ideally error with something like "this tool cannot parse the given version" (or somesuch). I like the output idea!

zyga:
@mvo5 I've implemented a strict checker and will push it here shortly, after a pass through spread.

tests/lib/bin/version-tool
```diff
@@ -0,0 +1,413 @@
...
+        self.assertEqual(strict_version_cmp("1", "1.2"), -1)
+        self.assertEqual(strict_version_cmp("1.2", "1"), +1)
+        self.assertEqual(strict_version_cmp("1", "1.0.1"), -1)
+        self.assertEqual(strict_version_cmp("1.1", "1.0.1"), +1)
+
+    def test_version_with_text(self):
+        # type: () -> None
+        with self.assertRaises(ValueError) as cm:
+            strict_version_cmp("1-foo", "1")
+        self.assertEqual(cm.exception.message, "version 1-foo is not purely numeric")
```
bboozzoo: `str(cm.exception)` IIRC message is Python2 only

tests/lib/bin/version-tool
```diff
@@ -0,0 +1,413 @@
...
+class NumericPartOfTests(unittest.TestCase):
+    def test_simple(self):
+        # type: () -> None
+        self.assertEqual(numeric_part_of("12"), 12)
+        self.assertEqual(numeric_part_of("12-stuff"), 12)
+        self.assertEqual(numeric_part_of("stuff-12"), 0)
+        self.assertEqual(numeric_part_of("-12"), 0)
+
+
+def naive_version_cmp(a, b):
```
mvo5: We don't need this anymore with "strict" mode, right? I would like to remove it (we can keep it in the history) so that its not used "accidentally" (and then we may run into the concerns I have with this lenient version).

zyga: I meant to keep it but if you feel strongly about it I will remove it now.

tests/lib/bin/version-tool
zyga: I fixed this, shucks.

zyga:
@sergiocazzolato I removed support for non-strict versions so I can add unit tests for 2.41 and similar but things more fancy (2.41~pre2) will report an error and won't compare anywhere. If there's desire to support that I'd like to do it in a follow-up to unblock my other PR using this

PR #7380, opened 2019-08-30
debian: set GOCACHE dir during build to fix FTBFS on eoan

mvo5:
The eoan sbuild fails with the following error:
```
failed to initialize build cache at /sbuild-nonexistent/.cache/go-build: mkdir /sbuild-nonexistent: permission denied
```

This can be fixed by setting GOCACHE explicitly - if unset it is
derrived from $HOME which is set to an impossible location in
the sbuild environment. 

I also created https://code.launchpad.net/~snappy-dev/+recipe/snapd-vendor-daily-disco-plus to ensure snapd gets build daily on disco, eoan in a PPA.


mvo5:
@zyga Thank you! Yeah, I'm happy to change this.

PR #7383, opened 2019-09-01
cmd/snap: improve help and error msg for snapshot commands

galgalesh:
*I have signed the Canonical CLA, I'm [merlijn-sebrechts](https://launchpad.net/~merlijn-sebrechts) on launchpad.*

This PR

* improves the help text for the snapshot commands; explaining the `id` parameter and linking to the `snap saved` command, which is used to find these id's.
* improves the error message when the set id is incorrect; linking to the `snap saved` command which is used to find these id's.

I created this PR because I had no idea what `id` I was supposed to use and where to get those id's, so @chipaca had to [help me on the forum](https://forum.snapcraft.io/t/accidentally-removed-snap-package-how-to-revert-to-older-revision/12982/6?u=galgalesh).

Note: `HACKING.md` says to run `make fmt` before sending patches, but this make target doesn't exist.

```md
## Submitting patches

Please run `make fmt` before sending your patches.
```


cmd/snap/cmd_snapshot.go
```diff
@@ -336,14 +336,36 @@ func init() {
...
 			"users": i18n.G("Restore data of only specific users (comma-separated) (default: all users)"),
-		}), nil)
+		}), []argDesc{
+			{
+				name: "<snap>",
+				// TRANSLATORS: This should not start with a lowercase letter.
+				desc: i18n.G("The snap for which data will be restored"),
+			}, {
+				// TRANSLATORS: This needs to begin with < and end with >
+				name: i18n.G("<id>"),
```
chipaca: I don't think we want to translate `<id>`

cmd/snap/complete.go
```diff
@@ -508,7 +508,7 @@ func (snapshotID) Complete(match string) []flags.Completion {
 func (s snapshotID) ToUint() (uint64, error) {
 	setID, err := strconv.ParseUint((string)(s), 10, 64)
 	if err != nil {
-		return 0, fmt.Errorf(i18n.G("invalid argument for set id: expected a non-negative integer argument"))
+		return 0, fmt.Errorf(i18n.G("invalid argument for set id: expected a non-negative integer argument (see 'snap help saved')"))
```
chipaca: I think this message can be improved further by changing 'set' to 'snapshot set' :-)

PR #7385, opened 2019-09-02
tests/cross/go-build: use go list rather than shell trickery

bboozzoo:
Use go list with format spec rather than second guessing based on files existing
or not.

Followup on #7259

### DIFF
--- a/tests/cross/go-build/task.yaml
+++ b/tests/cross/go-build/task.yaml
@@ -57,6 +57,7 @@ prepare: |
 
 execute: |
     cd /tmp/cross-build/src/github.com/snapcore/snapd
-    for cmd in $( find ./cmd -mindepth 2 -maxdepth 2 -name main.go -printf '%h\n' | sort -u ); do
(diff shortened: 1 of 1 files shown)


tests/cross/go-build/task.yaml
```diff
@@ -57,6 +57,7 @@ prepare: |
 
 execute: |
     cd /tmp/cross-build/src/github.com/snapcore/snapd
-    for cmd in $( find ./cmd -mindepth 2 -maxdepth 2 -name main.go -printf '%h\n' | sort -u ); do
+    # grab only packages whose name is 'main'
+    for cmd in $( GOPATH=/tmp/cross-build go list -f '{{if eq .Name "main"}}{{.ImportPath}}{{end}}' ./cmd/...); do
```
zyga: Can we do it the other way around? Instead of a shell loop that gives go test stuff to do one by one, can we just call go test on a thing returned by that `go list` output? This will be much more parallel in practice (on an 8 way machine it's clearly noticeable).

PR #7387, opened 2019-09-02
packaging/fedora, tests/lib/prepare-restore: helper tool for packing sources for RPM

bboozzoo:
Couple of patches I kept in a branch that adds support for https://packit.dev/ 

The purpose of the helper tool is to create a source archives in a format that Fedora RPM spec (and openSUSE one too) expect. Eventually I would like to use that inside a packit declaration file, which looks like this:

```
specfile_path: packaging/fedora/snapd.spec
synced_files:
  - .packit.yaml
  - packaging/fedora/snapd.spec
upstream_project_name: snapd
downstream_package_name: snapd
create_tarball_command: ["packaging/fedora/pack-source", "-o", "packaging/fedora", "-g"]
current_version_command: ["git", "describe"]
```

The tool is also useful when building binary RPM on a distro where rpmbuild does not have `--build-in-place`, eg. CentOS/RHEL 7 and Amazon Linux 2.


packaging/fedora/pack-source
```diff
@@ -0,0 +1,62 @@
...
+
+if [ ! -e packaging/fedora/snapd.spec ]; then
+    echo "must be executed at the top of srcdir"
+    exit 1
+fi
+
+outdir=.
+single=0
+
+while getopts "v:o:gsh" arg; do
```
zyga: Such a chance to support `go:sh` ;)

packaging/fedora/pack-source
```diff
@@ -0,0 +1,62 @@
...
+        h|*)
+            show_help
+            ;;
+    esac
+done
+
+set -x
+
+if [ -z "$version" ]; then
+	# try to read if tform spec
```
zyga:
```suggestion
	# Try to read version from the spec file.
```

packaging/fedora/pack-source
```diff
@@ -0,0 +1,62 @@
...
+	version="$(awk '/^Version: / { print $2}' packaging/fedora/snapd.spec)"
+fi
+
+tmpdir=$(mktemp -d)
+
```
zyga: Nitpick: `trap`-based cleanup missing?

packaging/fedora/pack-source
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
zyga: Do you need bash for real or just a habit?

packaging/fedora/pack-source
```diff
@@ -0,0 +1,62 @@
...
+            show_help
+            ;;
+    esac
+done
+
+set -x
+
+if [ -z "$version" ]; then
+	# try to read if tform spec
+	version="$(awk '/^Version: / { print $2}' packaging/fedora/snapd.spec)"
```
zyga: Hmm, this is hard-coding fedora spec in the world where we have an openSUSE spec as well. This feels wrong especially that we also use this script on openSUSE.

bboozzoo: As discussed on IRC, this can be dropped now.

packaging/fedora/pack-source
bboozzoo: Added

packaging/fedora/pack-source
bboozzoo: Needed for `trap .. EXIT` now.

PR #7388, opened 2019-09-02
boot, bootloader, o/snapstate: boot env manip goes in bootloader

chipaca:
This change moves almost everything that looks at boot environment
variables to bootloader itself. The exception (for now at least) is
`image`, which for now is special; thus the env var bits still need to
be exported for now.

It also includes a bunch of little cleanups / refactorings that were pending from previous PRs.


chipaca:
(oops, boot env manip goes in boot!)

PR #7392, opened 2019-09-02
tests: rewrite "retry" command as retry-tool

zyga:
Having pushed Sergio towards writing the command in Python instead of
shell I think it's on me to make sure it lands and looks right. This
patch rewrites the utility, with extra usability touches and, mainly,
correct semantics.

I added a quick spread test to showcase the features as well.

### DIFF
--- a/tests/lib/bin/retry-tool
+++ b/tests/lib/bin/retry-tool
@@ -1,0 +1,100 @@
+#!/usr/bin/env any-python
+from __future__ import print_function, absolute_import, unicode_literals
+
+import argparse
+import subprocess
+import sys
+import time
+
+
+# Define MYPY as False and use it as a conditional for typing import. Despite
+# this declaration mypy will really treat MYPY as True when type-checking.
+# This is required so that we can import typing on Python 2.x without the
+# typing module installed. For more details see:
+# https://mypy.readthedocs.io/en/latest/common_issues.html#import-cycles
+MYPY = False
+if MYPY:
+    from typing import List, Text
+
+
+def _make_parser():
+    # type: () -> argparse.ArgumentParser
+    parser = argparse.ArgumentParser(
+        description="""
+Retry executes COMMAND at most N times, waiting for SECONDS between each
+attempt. On failure the exit code from the final attempt is returned.
+"""
+    )
+    parser.add_argument(
+        "-n",
+        "--attempts",
+        metavar="N",
+        type=int,
+        default=3,
+        help="number of attempts (default %(default)s)",
+    )
+    parser.add_argument(
+        "--wait",
+        metavar="SECONDS",
+        type=float,
+        default=1,
+        help="grace period between attempts (default %(default)ss)",
+    )
+    parser.add_argument(
+        "cmd", metavar="COMMAND", nargs="...", help="command to execute"
+    )
+    return parser
+
+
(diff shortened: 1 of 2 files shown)


tests/main/retry-tool/task.yaml
```diff
@@ -0,0 +1,11 @@
+summary: smoke test for the retry test tool
```
sergiocazzolato: This is really nice, but, we are currently using the main test suite also for testing our tools, I think should be usefull to use a different test suite for that, in fact we dont need most of the preparation for this. @zyga, What do you think?

zyga: I was thinking about a suite that doesn't need to build snapd, that would be a nice improvement from current system. Note that this is not new: mountinfo-tool and version-tool are similarly tested. If anything I'd like to explore changing this in a separate PR.

tests/lib/bin/retry-tool
```diff
@@ -0,0 +1,100 @@
...
+        parser.print_usage()
+        parser.exit(0)
+    # Return the last exit code as the exit code of this process.
+    try:
+        retcode = run_cmd(ns.cmd, ns.attempts, ns.wait)
+    except OSError as exc:
+        print(
+            "retry: cannot execute command {}: {}".format(" ".join(ns.cmd), exc),
+            file=sys.stderr,
+        )
```
zyga:
```suggestion
        )
        raise SystemExit(1)
```

tests/lib/bin/retry-tool
```diff
@@ -0,0 +1,101 @@
...
+    for i in range(1, n + 1):
+        retcode = subprocess.call(cmd)
+        if retcode == 0:
+            break
+        print(
+            "retry: command {} failed with code {}".format(" ".join(cmd), retcode),
+            file=sys.stderr,
+        )
+        if i < n:
+            print(
```
stolowski: Can we show it only with `--verbose`?

tests/main/retry-tool/task.yaml
stolowski: It's great you have a test for this ❤️ ; I agree it's would be nice to move it elsewhere, but that can be done in a followup indeed.

tests/lib/bin/retry-tool
zyga: I was thinking about it but considered not to for the following reason. I would be unsure, while writing test code, where to put verbose and where not to. The whole reason for using retry-tool is to prevent something that sometimes fails from nuking the build by giving it a few more tries. When it happens we may not even get any trace from the actual command we were invoking. By printing the output from retry, we at least see some activity and can use it to debug issues later.

I would be okay with a `--quiet` switch for specific cases, with verbose being the default. Perhaps there's a use case for that and I'm okay with it.

stolowski: I see, that's a valid concern. My main motivation was to avoid excessive output for cases where we *know* things are retried a lot by design (e.g. waiting for some async activity or for something to show up in the logs), but not because occasional errors or flakiness. I agree `--quiet` is a better idea and we are noisy by default, so +1.

zyga: I've added `--quiet` in https://github.com/snapcore/snapd/pull/7397

PR #7396, opened 2019-09-03
tests: don't guess in is_classic_confinement_supported

zyga:
Snapd can tell us if the system supports classic confinement. We don't
need to inspect SPREAD_SYSTEM to do that.

### DIFF
--- a/tests/lib/snaps.sh
+++ b/tests/lib/snaps.sh
@@ -72,21 +72,8 @@ install_generic_consumer() {
 }
 
 is_classic_confinement_supported() {
-    case "$SPREAD_SYSTEM" in
-        ubuntu-core-*)
-            return 1
-            ;;
-        ubuntu-*|debian-*)
-            return 0
-            ;;
-        fedora-*|centos-*)
-            return 1
-            ;;
-        opensuse-*)
-            return 0
-            ;;
-        *)
-            return 0
-            ;;
-    esac
+    if snap debug sandbox-features --required=confinement-options:classic; then
+        return 0
+    fi
+    return 1
 }


tests/lib/snaps.sh
```diff
@@ -72,21 +72,8 @@ install_generic_consumer() {
...
-            return 1
-            ;;
-        opensuse-*)
-            return 0
-            ;;
-        *)
-            return 0
-            ;;
-    esac
+    if snap debug sandbox-features --required=confinement-options:classic; then
```
bboozzoo: That requires snapd to be running though? Given that the travis job complete I guess it's not really called when snapd is down.

Why not:
```
is_classic_confinement_supported() {
    snap debug sandbox-features --required=confinement-options:classic
}
```

zyga: Hmm, because I assumed it would break `set -e`.

sergiocazzolato:
Nice change, thanks!!

tests/lib/snaps.sh
mvo5: It seems like we stumbled into another shell pitfall - `set -e` does not happen inside shell function, see below. However Maciej suggestion works because its the last command inside the function so the exit status of the function is set correctly.
```
$ sh /tmp/lala.sh 
#!/bin/sh -e

cat $0

func() {
    echo in_func_header
    false
    echo in_func_footer
}

func
echo $?
echo "after func"

# OUTPUT:
in_func_header
in_func_footer
0
after func
```
It seems other people stumbled on this as well: https://stackoverflow.com/questions/4072984/how-do-i-get-effect-and-usefuless-of-set-e-inside-a-shell-function - shell is just not a good language it seems :/

PR #7397, opened 2019-09-03
tests: add --quiet switch to retry-tool

zyga:
Pawel made a case for using retry-tool in situations where we know and
anticipate the first few attempts of a command to fail, e.g. while
waiting for some asynchronous operation to finish. In that case the
default output doesn't help much, as we routinely fail on the road to
eventual success.

For this use case, add the --quiet switch, to silence all output from
the tool itself.

### DIFF
--- a/tests/lib/bin/retry-tool
+++ b/tests/lib/bin/retry-tool
@@ -41,32 +41,41 @@ attempt. On failure the exit code from the final attempt is returned.
         help="grace period between attempts (default %(default)ss)",
     )
     parser.add_argument(
+        "--quiet",
+        dest="verbose",
+        action="store_false",
+        default=True,
+        help="refrain from printing any output",
+    )
+    parser.add_argument(
         "cmd", metavar="COMMAND", nargs="...", help="command to execute"
     )
     return parser
 
 
-def run_cmd(cmd, n, wait):
-    # type: (List[Text], int, float) -> int
+def run_cmd(cmd, n, wait, verbose):
+    # type: (List[Text], int, float, bool) -> int
     retcode = 0
     for i in range(1, n + 1):
         retcode = subprocess.call(cmd)
         if retcode == 0:
             break
-        print(
-            "retry: command {} failed with code {}".format(" ".join(cmd), retcode),
-            file=sys.stderr,
-        )
-        if i < n:
+        if verbose:
             print(
-                "retry: next attempt in {} second(s) (attempt {} of {})".format(
-                    wait, i, n
-                ),
+                "retry: command {} failed with code {}".format(" ".join(cmd), retcode),
                 file=sys.stderr,
             )
+        if i < n:
+            if verbose:
+                print(
+                    "retry: next attempt in {} second(s) (attempt {} of {})".format(
+                        wait, i, n
+                    ),
+                    file=sys.stderr,
+                )
             time.sleep(wait)
     else:
-        if n > 1:
+        if verbose and n > 1:
             print(
                 "retry: command {} keeps failing after {} attempts".format(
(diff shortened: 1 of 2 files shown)


tests/lib/bin/retry-tool
```diff
@@ -40,33 +40,42 @@ attempt. On failure the exit code from the final attempt is returned.
         default=1,
         help="grace period between attempts (default %(default)ss)",
     )
+    parser.add_argument(
+        "--quiet",
+        dest="verbose",
```
stolowski: IMHO it'd be better to have quiet flag for consistency and reverse the conditionals below. This will also make it cleaner if we ever want to introduce --verbose for being really verbose.

zyga: Perhaps I misunderstand something. This adds the flag `--quiet` that stores `False` to an internal variable called `verbose`. This lets us avoid negation in front of each verbose print we do. If we want to have `--verbose` for consistency we just add it as another argument via ```
parser.add_argument("--verbose", action="store_true", help="...")
```

File: tests/lib/bin/retry-tool [hunk 2]
```diff
@@ -40,33 +40,42 @@ attempt. On failure the exit code from the final attempt is returned.
...
+                print(
+                    "retry: next attempt in {} second(s) (attempt {} of {})".format(
+                        wait, i, n
+                    ),
+                    file=sys.stderr,
+                )
             time.sleep(wait)
     else:
-        if n > 1:
+        if verbose and n > 1:
```
bboozzoo: We expected the command to succeed in n attempts, so reaching here is unexpected. Wonder if we should print the message anyway. Makes the switch named `--quiet` a bit akward, so maybe `--no-log` then?

tests/lib/bin/retry-tool
stolowski: Sorry: I mean `--quiet` flag for the parser, but also quiet variable internally with negations where needed. 

And `--verbose` could be reserved for something really verbose (more output than by default without `--quiet`) in the future, but you can ignore this remark, the main nitpick was really about '--quiet` and mixing it with "verbose" internally (not a biggie really).

tests/lib/bin/retry-tool
zyga: I was thinking about it as well, oh oh :) Such a hard choice.

So here's my proposal instead. I was thinking about this earlier but decided not to do it because YAGNI but this shows it may have been useful:

The `-n` switch can now be both a number and a range `n-m`. When it is a range it indicates that we expect the first `n` attempts to fail and those automatically become quiet. After those the remaining up-to-`m` attempts log as usual.

tests/lib/bin/retry-tool
zyga: But what's the point of that? The extra negations just add bulk and given that both `--verbose` and `--quiet` can be introduced there's no variable that is more correct. I chose it this way around to avoid the negation that just makes the logic harder to read.

stolowski: Ok, it's probably a matter of personal preference; my only point is that *if* being verbose/chatty is the default state of the tool, then I probably wouldn't create a flag that's positive by default. But again, it's a nitpick and a personal preference, not a blocker by any means, and I think it's not worth further discussion ;)

tests/lib/bin/retry-tool
bboozzoo: Sounds like a followup material.

zyga: Yeah, I'm happy to iterate on this some more. We can always respect `--quiet` even if we have the range syntax implemented.

PR #7399, opened 2019-09-03
tests: verify retry-tool not retrying missing commands

zyga:
It's a test case I didn't add for a bug I didn't notice after posting
the initial version of retry-tool. While I did fix the bug in a quick
github edit I was still meaning to add a small test so here it is.


zyga:
Ah, I knew printing the error directly from python would bit :) Fixing.

PR #7400, opened 2019-09-03
cmd/snap-update-ns: don't propagate detaching changes

zyga:
When snap-update-ns is unmounting something that can host additional
mount points inside, it is now using MNT_DETACH to detach the mount
point (recursively) and lazily unmount anything that is no longer used.
It can be though about like removing a reference in a garbage collected
memory management system. Unlike such systems, however, the kernel also
implements mount and unmount event *propagation*. This can replicate the
change in peer groups that receive propagation events. This can, oddly,
cause unmount loops where the kernel returns EBUSY, because the unmount
operation (even with MNT_DETACH) affects itself, in a way.

One such example is added as a test case in this patch, a fragment of which is
presented here:

    // but snapd's mount profile contains only two entries:
    //
    // 1) recursive-bind $SNAP/a -> $SNAP/b/c  (ie, mount --rbind $SNAP/a $SNAP/b/c)
    // 2) recursive-bind $SNAP/b -> $SNAP/d    (ie, mount --rbind $SNAP/b $SNAP/d)
    //
    // Both mount operations are performed under a substrate that is MS_SHARED.
    // Therefore, due to the rules that decide upon propagation of bind mounts
    // the propagation of the new mount entries is also shared. This is
    // documented in section 5b of
    // https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt.
    //
    // Interactive experimentation shows that the following three mount points exist
    // after this operation, as illustrated by findmnt:
    //
    // TARGET                                SOURCE         FSTYPE      OPTIONS
    // ...
    // └─/snap/test-snapd-layout/x1          /dev/loop1     squashfs    ro,nodev,relatime
    //   ├─/snap/test-snapd-layout/x1/b/c    /dev/loop1[/a] squashfs    ro,nodev,relatime
    //   └─/snap/test-snapd-layout/x1/d      /dev/loop1[/b] squashfs    ro,nodev,relatime
    //     └─/snap/test-snapd-layout/x1/d/c  /dev/loop1[/a] squashfs    ro,nodev,relatime
    //
    // Note that after the first mount operation only one mount point is created, namely
    // $SNAP/a -> $SNAP/b/c. The second recursive bind mount not only creates
    // $SNAP/b -> $SNAP/d, but also replicates $SNAP/a -> $SNAP/b/c as
    // $SNAP/a -> $SNAP/d/c.

When this mount profile is undone, the unmount --lazy (aka MNT_DETACH)
operation on x1/d fails with EBUSY, as explained above. The patch solves that
by changing event propagation, in the unmounted tree, to private. This idea is
hinted at in the umount(2) manual page.

### DIFF
--- a/cmd/snap-update-ns/change.go
+++ b/cmd/snap-update-ns/change.go
@@ -330,10 +330,20 @@ func (c *Change) lowLevelPerform(as *Assumptions) error {
 			flags := umountNoFollow
 			if c.Entry.XSnapdDetach() {
 				flags |= syscall.MNT_DETACH
+				// If we are detaching something then before performing the actual detach
+				// switch the entire hierarchy to private event propagation (that is,
+				// none). This works around a bit of peculiar kernel behavior when the
+				// kernel reports EBUSY during a detach operation, because the changes
+				// propagate in a way that conflicts with itself. This is also documented
+				// in umount(2).
+				err = sysMount("none", c.Entry.Dir, "", syscall.MS_REC|syscall.MS_PRIVATE, "")
+				logger.Debugf("mount --make-rprivate %q (error: %v)", c.Entry.Dir, err)
 			}
 
 			// Perform the raw unmount operation.
-			err = sysUnmount(c.Entry.Dir, flags)
+			if err == nil {
+				err = sysUnmount(c.Entry.Dir, flags)
+			}
 			if err == nil {
 				as.AddChange(c)
 			}
--- a/cmd/snap-update-ns/change_test.go
+++ b/cmd/snap-update-ns/change_test.go
@@ -642,6 +642,7 @@ func (s *changeSuite) TestPerformFilesystemMountWithoutMountPointAndReadOnlyBase
 
 		{C: `lstat "/rofs"`, R: testutil.FileInfoDir},
 		{C: `mount "tmpfs" "/rofs" "tmpfs" 0 "mode=0755,uid=0,gid=0"`},
+		{C: `mount "none" "/tmp/.snap/rofs" "" MS_REC|MS_PRIVATE ""`},
 		{C: `unmount "/tmp/.snap/rofs" UMOUNT_NOFOLLOW|MNT_DETACH`},
 
 		// Perform clean up after the unmount operation.
@@ -805,6 +806,7 @@ func (s *changeSuite) TestPerformFilesystemDetch(c *C) {
 	synth, err := chg.Perform(s.as)
 	c.Assert(err, IsNil)
 	c.Assert(s.sys.RCalls(), testutil.SyscallsEqual, []testutil.CallResultError{
+		{C: `mount "none" "/target" "" MS_REC|MS_PRIVATE ""`},
 		{C: `unmount "/target" UMOUNT_NOFOLLOW|MNT_DETACH`},
 
 		// Perform clean up after the unmount operation.
@@ -1151,6 +1153,7 @@ func (s *changeSuite) TestPerformDirectoryBindMountWithoutMountPointAndReadOnlyB
 
 		{C: `lstat "/rofs"`, R: testutil.FileInfoDir},
 		{C: `mount "tmpfs" "/rofs" "tmpfs" 0 "mode=0755,uid=0,gid=0"`},
+		{C: `mount "none" "/tmp/.snap/rofs" "" MS_REC|MS_PRIVATE ""`},
 		{C: `unmount "/tmp/.snap/rofs" UMOUNT_NOFOLLOW|MNT_DETACH`},
 
 		// Perform clean up after the unmount operation.
@@ -1298,6 +1301,7 @@ func (s *changeSuite) TestPerformDirectoryBindMountWithoutMountSourceAndReadOnly
 		{C: `close 4`},
 		{C: `lstat "/rofs"`, R: testutil.FileInfoDir},
 		{C: `mount "tmpfs" "/rofs" "tmpfs" 0 "mode=0755,uid=0,gid=0"`},
+		{C: `mount "none" "/tmp/.snap/rofs" "" MS_REC|MS_PRIVATE ""`},
 		{C: `unmount "/tmp/.snap/rofs" UMOUNT_NOFOLLOW|MNT_DETACH`},
 
 		// Perform clean up after the unmount operation.
@@ -1690,6 +1694,7 @@ func (s *changeSuite) TestPerformFileBindMountWithoutMountPointAndReadOnlyBase(c
 
 		{C: `lstat "/rofs"`, R: testutil.FileInfoDir},
 		{C: `mount "tmpfs" "/rofs" "tmpfs" 0 "mode=0755,uid=0,gid=0"`},
+		{C: `mount "none" "/tmp/.snap/rofs" "" MS_REC|MS_PRIVATE ""`},
 		{C: `unmount "/tmp/.snap/rofs" UMOUNT_NOFOLLOW|MNT_DETACH`},
 
 		// Perform clean up after the unmount operation.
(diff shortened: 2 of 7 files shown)


zyga:
NOTE: I cherry-picked this out of the `MS_SHARED` branch. To preempt your question about adding a spread test: I plan to but before we have propagation fixed this is not easy because we just don't have any propagation yet. I want to propose it separately to discuss and review and asses the impact of this way.

zyga:
NOTE: This will fail as I didn't include the apparmor parts of the patch that were included in the original, sorry. I'll fix this shortly.

zyga:
This should now incorporate all the fixes for the stuff recently broken on master.

cmd/snap-update-ns/change.go
```diff
@@ -330,10 +330,20 @@ func (c *Change) lowLevelPerform(as *Assumptions) error {
 			flags := umountNoFollow
 			if c.Entry.XSnapdDetach() {
 				flags |= syscall.MNT_DETACH
+				// If we are detaching something then before performing the actual detach
```
bboozzoo: I think I'd make this comment more terse:
```
// Do not propagate unmount events. Prevents detach from conflicting with
// itself and getting EBUSY when hierarchies are nested. See umount(2) for details.
```

zyga: I can adjust that if you want :-)
I *may* do so in the follow-up, if that is green and it gets approved.

cmd/snap-update-ns/change_test.go
```diff
@@ -2424,3 +2431,80 @@ func (s *changeSuite) TestPerformedChangesAreTracked(c *C) {
...
+	// Note that after the first mount operation only one mount point is created, namely
+	// $SNAP/b/c -> $SNAP/a. The second recursive bind mount not only creates
+	// $SNAP/d -> $SNAP/b, but also replicates $SNAP/b/c -> $SNAP/a as
+	// $SNAP/d/c -> $SNAP/a.
+	//
+	// The test will simulate a refresh of the snap from revision x1 to revision
+	// x2. When this happens the mount profile associated with x1 must be undone
+	// and the mount profile associated with x2 must be constructed. Because
+	// ordering matters, let's first consider the order of construction of x1
+	// itself. Starting from nothing, apply x1 as follows:
```
jdstrand: This comment is very descriptive and overall excellent, but more difficult to follow than needed since it is written from the perspective of findmnt output, which lists the source and target in reverse order to mountinfo, /etc/mtab, /etc/fstab, mount alone, various mount commands/APIs, apparmor rules and even snapd's own fstab files. To make this clear since findmnt output is non-standard, perhaps reword like so:
```
	// This problem is more subtle. It is a variant of the regression test
	// implemented in tests/regression/lp-1831010. Here, we have four directories:
	//
	// - $SNAP/a
	// - $SNAP/b
	// - $SNAP/b/c
	// - $SNAP/d
	//
	// but snapd's mount profile contains only two entries:
	//
	// 1) recursive-bind $SNAP/a -> $SNAP/b/c  (ie, mount --rbind $SNAP/a $SNAP/b/c)
	// 2) recursive-bind $SNAP/b -> $SNAP/d    (ie, mount --rbind $SNAP/b $SNAP/d)
	//
	// Both mount operations are performed under a substrate that is MS_SHARED.
	// Therefore, due to the rules that decide upon propagation of bind mounts
	// the propagation of the new mount entries is also shared. This is
	// documented in section 5b of
	// https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt.
	//
	// Interactive experimentation shows that the following three mount points exist
	// after this operation, as illustrated by findmnt:
	//
	// TARGET                                SOURCE         FSTYPE      OPTIONS
	// ...
	// └─/snap/test-snapd-layout/x1          /dev/loop1     squashfs    ro,nodev,relatime
	//   ├─/snap/test-snapd-layout/x1/b/c    /dev/loop1[/a] squashfs    ro,nodev,relatime
	//   └─/snap/test-snapd-layout/x1/d      /dev/loop1[/b] squashfs    ro,nodev,relatime
	//     └─/snap/test-snapd-layout/x1/d/c  /dev/loop1[/a] squashfs    ro,nodev,relatime
	//
	// Note that after the first mount operation only one mount point is created, namely
	// $SNAP/a -> $SNAP/b/c. The second recursive bind mount not only creates
	// $SNAP/b -> $SNAP/d, but also replicates $SNAP/a -> $SNAP/b/c as
	// $SNAP/a -> $SNAP/d/c.
...
```

cmd/snap-update-ns/change_test.go
```diff
@@ -2424,3 +2431,80 @@ func (s *changeSuite) TestPerformedChangesAreTracked(c *C) {
...
+	// 1) recursive-bind $SNAP/b/c -> $SNAP/a
+	// 2) recursive-bind $SNAP/d -> $SNAP/b
+	//
+	// NB: The notation "foo -> bar" can be read as "foo is a view of bar",
+	// and can be implemented by mount "--rbind bar foo".
+	//
+	// Both mount operations are performed under a substrate that is MS_SHARED.
+	// Therefore, due to the rules that decide upon propagation of bind mounts
+	// the propagation of the new mount entries is also shared. This is
+	// documented in https://lwn.net/Articles/159092/ (section 5b).
```
jdstrand: Please use the official documentation, even though it is more unwieldy: https://www.kernel.org/doc/Documentation/filesystems/sharedsubtree.txt (me suggested change below adjusts for this).

cmd/snap-update-ns/change.go
```diff
@@ -330,10 +330,20 @@ func (c *Change) lowLevelPerform(as *Assumptions) error {
...
+				// switch the entire hierarchy to private event propagation (that is,
+				// none). This works around a bit of peculiar kernel behavior when the
+				// kernel reports EBUSY during a detach operation, because the changes
+				// propagate in a way that conflicts with itself. This is also documented
+				// in umount(2).
+				err = sysMount("none", c.Entry.Dir, "", syscall.MS_REC|syscall.MS_PRIVATE, "")
+				logger.Debugf("mount --make-rprivate %q (error: %v)", c.Entry.Dir, err)
```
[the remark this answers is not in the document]
zyga: `/snap/test-napd-layout/x1/b/c` is not affected:

I performed this experiment manually and this is what came out:
```
zyga@yantra:~/snapd/cmd/snap-update-ns/snap/test-snapd-layout/x1> python3 ~/snapd/tests/lib/bin/mountinfo-tool $(realpath .)/...
3669 2969 7:92 /a /home/zyga/go/lnx-src/github.com/snapcore/snapd/cmd/snap-update-ns/snap/test-snapd-layout/x1/b/c ro,relatime shared:1102 - squashfs /dev/loop92 ro
3701 2969 7:92 /b /home/zyga/go/lnx-src/github.com/snapcore/snapd/cmd/snap-update-ns/snap/test-snapd-layout/x1/d ro,relatime - squashfs /dev/loop92 ro
3702 3701 7:92 /a /home/zyga/go/lnx-src/github.com/snapcore/snapd/cmd/snap-update-ns/snap/test-snapd-layout/x1/d/c ro,relatime - squashfs /dev/loop92 ro
```

cmd/snap-update-ns/change_test.go
```diff
@@ -2441,37 +2441,31 @@ func (s *changeSuite) TestComplexPropagatingChanges(c *C) {
...
 	// Note that after the first mount operation only one mount point is created, namely
-	// $SNAP/b/c -> $SNAP/a. The second recursive bind mount not only creates
-	// $SNAP/d -> $SNAP/b, but also replicates $SNAP/b/c -> $SNAP/a as
-	// $SNAP/d/c -> $SNAP/a.
-	//
-	// The test will simulate a refresh of the snap from revision x1 to revision
-	// x2. When this happens the mount profile associated with x1 must be undone
-	// and the mount profile associated with x2 must be constructed. Because
-	// ordering matters, let's first consider the order of construction of x1
-	// itself. Starting from nothing, apply x1 as follows:
```
jdstrand: FYI, I didn't mean for you to remove the 'The test will simulate...' paragraph. Feel free to add it back if desired.

zyga:
Done now, this is good for landing when green.

PR #7401, opened 2019-09-03
tests: add unstable stage for travis execution

sergiocazzolato:
This new travis stage is used for systems which are being stabilized.

The idea of this stage is to continue executing on those systems but
avoid blocking the results.

For idea is to run tests on ubuntu-19.10 until the system is finally
stable, then it should go to google backend again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 git:
   quiet: true
-jobs:
+matrix:
   include:
     - stage: quick
       name: go 1.9/xenial static and unit test suites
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -71,7 +71,18 @@ jobs:
...
+      dist: xenial
+      addons:
+        apt:
+          packages:
+          - xdelta3
+      install:
+        # override the default install for language:go
+        - true
+      script:
+        - ./run-checks --spread-unstable || true
```
mvo5: With that we always see "green" as the result, no? I wonder if there is a way in travis to somehow flag if something needs attention or if travis is really only binary (failed xor succeed).

mvo5: Maybe https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail is helpful here?

sergiocazzolato: @mvo5 change done, tests running right now, thanks for the suggestion!!

PR #3162, opened 2017-04-10
cmd: use libtool for the internal library

morphis:
This just cuts some hassle integration with how the helper library is
built. I could never get to pass the right -fPIC/-fpie combination
otherwise.

Signed-off-by: Zygmunt Krynicki <me@zygoon.pl>

PR #7403, opened 2019-09-04
cmd: use libtool for the internal library

bboozzoo:
This just cuts some hassle integration with how the helper library is
built. I could never get to pass the right -fPIC/-fpie combination
otherwise.

The patch is currently used in snapd packaging for Fedora/EPEL. Let's see if it breaks anything. According to notes form the distro spec, it comes from #3162

cc @Conan-Kudo


Conan-Kudo:
@bboozzoo 

```
2019-09-04 10:01:52 Failed tasks: 1
    - google:ubuntu-19.10-64:tests/main/interfaces-timeserver-control
error: unsuccessful run
```

I don't seem to see any details of the failure, though?

bboozzoo:
I don't think there's a way we can include the patch and keep things working as they are now.

The patch breaks partial static linking on Ubuntu, so `snap-confine` is not statically linked with libapparmor and libcap. This is not caught by the tests (we should have a test for this really).

I have tried different workarounds, but none of them works. 
Tried so far:
- adding static libs to LDADD, errors out on libtoolize
- adding staic libs to LDFLAGS, doesn't work flags too early
- using `-l:libapparmor.a -l:libcap.a` in LDADD, doesn't work as it gets expanded to some `--static` garbage
- adding target specific variable assignment `snap-confine/snap-confine$(EXEEXT): LIBS += ..` fails as automake produces garbage, and treats this as a definition of a rule for snap-confine

For reference, this is the rule generated by automake & libtoolize:
```
snap-confine/snap-confine$(EXEEXT): $(snap_confine_snap_confine_OBJECTS) $(snap_confine_snap_confine_DEPENDENCIES) $(EXTRA_snap_confine_snap_confine_DEPENDENCIES) snap-confine/$(am__dirstamp)
        @rm -f snap-confine/snap-confine$(EXEEXT)
        $(AM_V_CCLD)$(snap_confine_snap_confine_LINK) $(snap_confine_snap_confine_OBJECTS) $(snap_confine_snap_confine_LDADD) $(LIBS)
```
Where actual call to libtool is done in:
```
snap_confine_snap_confine_LINK = $(LIBTOOL) $(AM_V_lt) --tag=CC \
        $(AM_LIBTOOLFLAGS) $(LIBTOOLFLAGS) --mode=link $(CCLD) \
        $(snap_confine_snap_confine_CFLAGS) $(CFLAGS) \
        $(snap_confine_snap_confine_LDFLAGS) $(LDFLAGS) -o $@
```

This is our hacked rule:
```
snap-confine/snap-confine$(EXEEXT): $(snap_confine_snap_confine_OBJECTS) $(snap_confine_snap_confine_DEPENDENCIES) $(EXTRA_snap_confine_snap_confine_DEPENDENCIES) libsnap-confine-private/$(am__dirstamp)
	@rm -f snap-confine/snap-confine$(EXEEXT)
	$(AM_V_CCLD)$(snap_confine_snap_confine_LINK) $(snap_confine_snap_confine_OBJECTS) $(snap_confine_snap_confine_LDADD) $(LIBS)

snap-confine/snap-confine$(EXEEXT): LIBS += -Wl,-Bstatic $(snap_confine_snap_confine_STATIC) -Wl,-Bdynamic -pthread
```
Without libtool, `snap_confine_snap_confine_LINK` just calls out to gcc.

Including the patch, or some form thereof, breaks the build basically. Not including it, I have not observed any failures building on Koji for Fedora or epel7 targets. I suggest dropping the patch from distro packaging too. If there's a flag that is added by libtool and not advertised in LDFLAGS, then we'd have to address it separately.

FWIW, the particular build mode with statically linked apparmor and libcap is only used on Ubuntu for the purpose of supporting reexec. None of the other distro packages do that.

PR #7404, opened 2019-09-04
cmd/snap: don't append / to snap name just because a dir exists

chipaca:
Before this change, when a user did `snap info foo` and a directory
existed, a `/` was added to `foo` as part of path normalization.

That should only be done to the function that sets up the 'disk' snap,
not for every use of the snap name.

This change fixes that.

### DIFF
--- a/tests/main/snap-info/task.yaml
+++ b/tests/main/snap-info/task.yaml
@@ -8,6 +8,11 @@ prepare: |
     snap pack "$TESTSLIB"/snaps/basic
     snap install test-snapd-tools
     snap install --channel beta --devmode test-snapd-devmode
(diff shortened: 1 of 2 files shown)


tests/main/snap-info/task.yaml
```diff
@@ -8,6 +8,11 @@ prepare: |
     snap pack "$TESTSLIB"/snaps/basic
     snap install test-snapd-tools
     snap install --channel beta --devmode test-snapd-devmode
+    # ensures that an empty directory doesn't confuse things
+    mkdir -v core
```
anonymouse64: :+1:

chipaca:
This is (also) https://bugs.launchpad.net/snapd/+bug/1845533

PR #7406, opened 2019-09-04
cmd/snap-confine: keep track of snap instance name and the snap name

bboozzoo:
Snap instance name is composed of the snap name and an optional instance key.
Keep track of both, for future use in parallel instance support code.

Extracted from #7302

### DIFF
--- a/cmd/snap-confine/snap-confine-invocation.c
+++ b/cmd/snap-confine/snap-confine-invocation.c
@@ -61,8 +61,8 @@ void sc_init_invocation(sc_invocation *inv, const struct sc_args *args, const ch
         die("cannot run with NULL executable");
     }
 
-    /* Instance name length + 1 overflow + 1 NULL */
-    char snap_name[SNAP_INSTANCE_LEN + 1 + 1] = {0};
+    /* Instance name length + NULL termination */
+    char snap_name[SNAP_NAME_LEN + 1] = {0};
     sc_snap_drop_instance_key(snap_instance, snap_name, sizeof snap_name);
 
     /* Invocation helps to pass relevant data to various parts of snap-confine. */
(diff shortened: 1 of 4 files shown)


cmd/snap-confine/snap-confine-invocation.c
```diff
@@ -61,13 +61,18 @@ void sc_init_invocation(sc_invocation *inv, const struct sc_args *args, const ch
         die("cannot run with NULL executable");
     }
 
+    /* Instance name length + 1 overflow + 1 NULL */
+    char snap_name[SNAP_INSTANCE_LEN + 1 + 1] = {0};
+    sc_snap_drop_instance_key(snap_instance, snap_name, sizeof snap_name);
```
jdstrand: This is an unexpected allocation. snap_name should never be SNAP_INSTANCE_LEN since up above in sc_snap_instance_name_validate() we made sure (it calls sc_snap_name_validate() on `*snap_name = strsep(&t, "_");` and verifies that snap_name is not > 40 (hardcoded and oddly not SNAP_NAME_LEN)). This should be:

```
// snap_name length plus null termination
char snap_name[SNAP_NAME_LEN + 1] = {0};
sc_snap_drop_instance_key(snap_instance, snap_name, sizeof snap_name - 1;
```

Due to the call to sc_snap_instance_name_validate(), this is not a bug and simply extra allocation, but the allocation is confusing when auditing and I wouldn't want to see it copy/pasted somewhere else where sc_snap_instance_name_validate() wasn't called since the extra allocation without check could allow a longer snap name than expected (this is also why we don't use `SNAP_NAME_LEN + 1 + 1` since then sc_snap_split_instance_name() will only check for a >= 41 char snap_name_len when we want to fail on >= 40.

bboozzoo: I think you mean this?
```
char snap_name[SNAP_NAME_LEN + 1] = {0};
sc_snap_drop_instance_key(snap_instance, snap_name, sizeof snap_name);
```

Also, fixed the validation code to use SNAP_NAME_LEN, fixed the tests so that `die()` isn't masked and added some extra unit tests for sc_snap_drop_instance_key()

jdstrand: I did, yes.

PR #7408, opened 2019-09-04
tests: fix interfaces-timeserver-control on 19.10

mvo5:
On 19.10 the GCE images have the "chrony" package installed. This
will interfere with the systemd-timesyncd.service which has:
```
ConditionFileIsExecutable=!/usr/sbin/chronyd
```

This PR will remove the chrony package and restart the service
to fix the test.


tests/main/interfaces-timeserver-control/task.yaml
```diff
@@ -17,6 +16,13 @@ prepare: |
     # Install a snap declaring a plug on timeserver-control
     install_local test-snapd-timedate-control-consumer
 
+    if [[ "$SPREAD_SYSTEM" == ubuntu-19.10-* ]]; then
+        # ensure chrony is not installed as it interferes with
+        # the systemd-timesyncd.service
+        apt remove -y chrony
```
bboozzoo: Slightly uneasy about removing the package, perhaps we should install it in restore?

bboozzoo: Experimenting a bit with automatically restoring packages in #7414

PR #7409, opened 2019-09-04
interfaces/wayland: allow a confined server running in a user session to work with Qt, GTK3 & SDL2 clients

AlanGriffiths:
Needed to enable a confined Wayland desktop session.

Some Wayland based toolkits (Qt, GTK3, SDL2) create shm buffer pools using files in $XDG_RUNTIME_DIR to pass to the server. While these are passed by FD we still need rw access to the file.


AlanGriffiths:
The need for this can be verified by:

1. Running on X11: `snap install mir-kiosk && mir-kiosk`
2. `snap install mir-kiosk-apps && snap connect mir-kiosk-apps:wayland && mir-kiosk-apps`

Without this, it errors with "invalid arguments for wl_shm@create_pool".

interfaces/builtin/wayland.go
```diff
@@ -56,6 +56,10 @@ owner /run/user/[0-9]*/wayland-[0-9]* rwk,
 # Allow access to common client Wayland sockets from non-snap clients
 /run/user/[0-9]*/{mesa,mutter,sdl,wayland-cursor,weston,xwayland}-shared-* rw,
 
+# Some Wayland based toolkits (Qt, GTK3, SDL2) create shm files to pass to the
+# server. Although they are passed by FD we still need rw access to the file.
+owner /run/user/[0-9]*/snap.*/{,wayland-*} rw,
```
jdstrand: Is owner match correct here? Is there ever a time the passed fd will be opened by one user and accessed by another (eg, a non-root client and a root server).

AlanGriffiths: I don't envisage that being useful with snapped server & client. A daemon server and user client have unrelated $XDG_RUNTIME_DIRs, so they wouldn't connect.

It is, of course, possible to set up that configuration in the Classic environment.

AlanGriffiths: ..although, it looks like the rule can be a lot more specific.

WIll update.

interfaces/builtin/wayland.go
```diff
@@ -55,6 +55,9 @@ capability sys_tty_config,
 owner /run/user/[0-9]*/wayland-[0-9]* rwk,
 # Allow access to common client Wayland sockets from non-snap clients
 /run/user/[0-9]*/{mesa,mutter,sdl,wayland-cursor,weston,xwayland}-shared-* rw,
+# Some Wayland based toolkits (Qt, GTK3, SDL2) create shm files to pass cursor images
+# to the server. Although they are passed by FD we still need rw access to the file.
+/run/user/[0-9]*/snap.*/wayland-cursor-shared-* rw,
```
zyga: should this say `owner ...`?

AlanGriffiths: > should this say `owner ...`?

I don't think that gains much (other rules don't specify "owner" either) and _could_ be a limitation. (When `su`ing to another user before starting a client application this would prevent it appearing on a confined desktop.)

AlanGriffiths:
Anything more I should be doing to get this landed?

PR #7412, opened 2019-09-05
tests: run dbus-launch inside a systemd unit

mvo5:
The test sometimes fails because something tries to write to
/tmp/xdg/ after the dbus-session is terminated. This PR puts
the dbus-launch command into a systemd-run created unit so that
we can kill everything that was run as part of the dbus session
which hopefully ensures we have no stray processes left when we
stop the test.

### DIFF
--- a/tests/main/interfaces-contacts-service/task.yaml
+++ b/tests/main/interfaces-contacts-service/task.yaml
@@ -48,7 +48,7 @@ execute: |
     mkdir -p "$XDG_CONFIG_HOME" "$XDG_DATA_HOME" "$XDG_CACHE_HOME"
 
     echo "Setting up D-Bus session bus in a systemd unit"
-    systemd-run --unit=dbus-session --property=Type=forking -E "$(env | tr '\n' ' ')"  -r /bin/sh -c 'dbus-launch --sh-syntax > /tmp/dbus-sh'
+    systemd-run --unit=dbus-session --property=Type=forking -r /bin/sh -c "XDG_CONFIG_HOME=$XDG_CONFIG_HOME XDG_DATA_HOME=$XDG_DATA_HOME XDG_CACHE_HOME=$XDG_CACHE_HOME dbus-launch --sh-syntax > /tmp/dbus-sh"
     for _ in $(seq 20); do
         if test -e /tmp/dbus-sh; then
             break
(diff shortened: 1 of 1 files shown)


tests/main/interfaces-contacts-service/task.yaml
```diff
@@ -51,9 +47,14 @@ execute: |
     fi
     mkdir -p "$XDG_CONFIG_HOME" "$XDG_DATA_HOME" "$XDG_CACHE_HOME"
 
-    echo "Setting up D-Bus session bus"
-    eval "$(dbus-launch --sh-syntax)"
-    echo "$DBUS_SESSION_BUS_PID" > dbus-launch.pid
+    echo "Setting up D-Bus session bus in a systemd unit"
+    systemd-run --unit=dbus-session --property=Type=forking -E "$(env | tr '\n' ' ')"  -r /bin/sh -c 'dbus-launch --sh-syntax > /tmp/dbus-sh'
+    for _ in $(seq 20); do
```
zyga: How about using retry tool?
```
retry-tool -n 20 test -e /tmp/dbus-sh
```

tests/main/interfaces-contacts-service/task.yaml
```diff
@@ -51,9 +47,14 @@ execute: |
     fi
     mkdir -p "$XDG_CONFIG_HOME" "$XDG_DATA_HOME" "$XDG_CACHE_HOME"
 
-    echo "Setting up D-Bus session bus"
-    eval "$(dbus-launch --sh-syntax)"
-    echo "$DBUS_SESSION_BUS_PID" > dbus-launch.pid
+    echo "Setting up D-Bus session bus in a systemd unit"
+    systemd-run --unit=dbus-session --property=Type=forking -E "$(env | tr '\n' ' ')"  -r /bin/sh -c 'dbus-launch --sh-syntax > /tmp/dbus-sh'
```
zyga: What is the environment part for? Can you explain what do we need from the local environment to make the service operate the way we want?

In addition, can the service run `exec dbus-launch` instead, to make sure there's just enough processes as necessary, not more.

mvo5: Its needed for the XDG_* vars.

zyga: I think it warrants a comment. We could even constrain it to just `XDG_*=*` for clarity

mvo5: I had to change this as older versions of systemd do not support "-E" - now its explicit.

PR #7421, opened 2019-09-06
cmd/snap-confine: unmount /writable from snap view

zyga:
On core systems we establish an elaborate mount namespace, comprised
of the bootable base snap and a lot of bind mounts to the writable
partition.

The writable partition, in its full glory is visible both /writable
and /var/lib/snapd/hostfs/writable. Each view contains /writable/snap/...
coupled with all the snaps visible there.

None of the snap interfaces grant access to actual /writable itself. In
addition the data visible there is totally redundant, because the snaps
are already visible in /snap anyway.

The snaps are visible there because of propagation established by systemd
in the initial mount namespace.

This patch has two changes:
 - It adjust snap-confine to detach /writable/
   and /var/lib/snapd/hostfs/writable/ and anything underneath
   from the mount namespace used by snaps. There's a small permission
   change to make this possible.
 - It adjusts the test data in the mount-ns test to reflect this.

Overall, coupled with other changes to this area, this patch reduces the
complexity of the mount namespace used on ubuntu-core.

### DIFF
--- a/cmd/libsnap-confine-private/string-utils-test.c
+++ b/cmd/libsnap-confine-private/string-utils-test.c
@@ -74,27 +74,6 @@ static void test_sc_startswith(void)
 	g_assert_false(sc_startswith("ba", "bar"));
 }
 
-static void test_sc_is_beneath(void)
-{
-	// Essential difference from sc_startswith
-	g_assert_true(sc_is_beneath("/foo/bar", "/foo"));
-	g_assert_false(sc_is_beneath("/foo/bar", "/foo2"));
-	g_assert_true(sc_is_beneath("foo/bar", "foo"));
-	// True cases
-	g_assert_true(sc_is_beneath("/foo", "/foo"));
-	g_assert_true(sc_is_beneath("", ""));
-	g_assert_true(sc_is_beneath("/foo", "/"));
-	g_assert_true(sc_is_beneath("/foo", ""));
-	// False cases
-	g_assert_false(sc_is_beneath("/foo", "/bar"));
-	g_assert_false(sc_is_beneath("/foo", "/foo/bar"));
-	// NULL is just perpetually false.
-	g_assert_false(sc_is_beneath(NULL, ""));
-	g_assert_false(sc_is_beneath("", NULL));
-	g_assert_false(sc_is_beneath(NULL, NULL));
-
-}
-
 static void test_sc_must_snprintf(void)
 {
 	char buf[5] = { 0 };
@@ -837,7 +816,6 @@ static void __attribute__((constructor)) init(void)
 	g_test_add_func("/string-utils/sc_streq", test_sc_streq);
 	g_test_add_func("/string-utils/sc_endswith", test_sc_endswith);
 	g_test_add_func("/string-utils/sc_startswith", test_sc_startswith);
-	g_test_add_func("/string-utils/sc_is_beneath", test_sc_is_beneath);
 	g_test_add_func("/string-utils/sc_must_snprintf",
 			test_sc_must_snprintf);
 	g_test_add_func("/string-utils/sc_must_snprintf/fail",
--- a/cmd/libsnap-confine-private/string-utils-test.c
+++ b/cmd/libsnap-confine-private/string-utils-test.c
@@ -79,10 +79,12 @@ static void test_sc_is_beneath(void)
 	// Essential difference from sc_startswith
 	g_assert_true(sc_is_beneath("/foo/bar", "/foo"));
 	g_assert_false(sc_is_beneath("/foo/bar", "/foo2"));
+	g_assert_true(sc_is_beneath("foo/bar", "foo"));
 	// True cases
 	g_assert_true(sc_is_beneath("/foo", "/foo"));
 	g_assert_true(sc_is_beneath("", ""));
 	g_assert_true(sc_is_beneath("/foo", "/"));
+	g_assert_true(sc_is_beneath("/foo", ""));
 	// False cases
 	g_assert_false(sc_is_beneath("/foo", "/bar"));
 	g_assert_false(sc_is_beneath("/foo", "/foo/bar"));
--- a/cmd/libsnap-confine-private/string-utils-test.c
+++ b/cmd/libsnap-confine-private/string-utils-test.c
@@ -74,6 +74,25 @@ static void test_sc_startswith(void)
 	g_assert_false(sc_startswith("ba", "bar"));
 }
 
+static void test_sc_is_beneath(void)
+{
+	// Essential difference from sc_startswith
+	g_assert_true(sc_is_beneath("/foo/bar", "/foo"));
+	g_assert_false(sc_is_beneath("/foo/bar", "/foo2"));
+	// True cases
+	g_assert_true(sc_is_beneath("/foo", "/foo"));
+	g_assert_true(sc_is_beneath("", ""));
+	g_assert_true(sc_is_beneath("/foo", "/"));
+	// False cases
+	g_assert_false(sc_is_beneath("/foo", "/bar"));
+	g_assert_false(sc_is_beneath("/foo", "/foo/bar"));
+	// NULL is just perpetually false.
+	g_assert_false(sc_is_beneath(NULL, ""));
+	g_assert_false(sc_is_beneath("", NULL));
+	g_assert_false(sc_is_beneath(NULL, NULL));
+
+}
+
 static void test_sc_must_snprintf(void)
 {
 	char buf[5] = { 0 };
@@ -816,6 +835,7 @@ static void __attribute__((constructor)) init(void)
 	g_test_add_func("/string-utils/sc_streq", test_sc_streq);
 	g_test_add_func("/string-utils/sc_endswith", test_sc_endswith);
 	g_test_add_func("/string-utils/sc_startswith", test_sc_startswith);
+	g_test_add_func("/string-utils/sc_is_beneath", test_sc_is_beneath);
 	g_test_add_func("/string-utils/sc_must_snprintf",
 			test_sc_must_snprintf);
 	g_test_add_func("/string-utils/sc_must_snprintf/fail",
--- a/cmd/libsnap-confine-private/string-utils.c
+++ b/cmd/libsnap-confine-private/string-utils.c
@@ -72,33 +72,6 @@ bool sc_startswith(const char *str, const char *prefix)
 	return strncmp(str, prefix, xlen) == 0;
 }
 
-bool sc_is_beneath(const char *path, const char *parent_dir)
-{
-	if (!path || !parent_dir) {
-		return false;
-	}
-	// The path "path" is beneath the directory "parent_dir" if "path" is a
-	// prefix of "parent_dir" with the following additional restrictions.
-	if (sc_startswith(path, parent_dir)) {
-		size_t parent_len = strlen(parent_dir);
-		// If "parent_dir" ends with a directory separator then it means that
-		// "path" must be of the form "$parent_dir/$ANYTHING" because of the
-		// trailing path separator is explicitly present.
-		if (parent_len > 0 && parent_dir[parent_len - 1] == '/') {
-			return true;
-		}
-		// If "path" either is either identical to "parent_dir" or it contains
-		// a path separator character "/" exactly where at the location where
-		// "parent_dir" ends. For example "/foo/bar" is beneath "/foo" because
-		// of "/foo/bar"[4] is "/". This eliminates "/foo2" from matching
-		// "/foo".
-		if (path[parent_len] == '\0' || path[parent_len] == '/') {
-			return true;
-		}
-	}
-	return false;
-}
-
 char *sc_strdup(const char *str)
 {
 	size_t len;
--- a/cmd/libsnap-confine-private/string-utils.c
+++ b/cmd/libsnap-confine-private/string-utils.c
@@ -72,6 +72,33 @@ bool sc_startswith(const char *str, const char *prefix)
 	return strncmp(str, prefix, xlen) == 0;
 }
 
+bool sc_is_beneath(const char *path, const char *parent_dir)
+{
+	if (!path || !parent_dir) {
+		return false;
+	}
+	// The path "path" is beneath the directory "parent_dir" if "path" is a
+	// prefix of "parent_dir" with the following additional restrictions.
+	if (sc_startswith(path, parent_dir)) {
+		size_t parent_len = strlen(parent_dir);
+		// If "parent_dir" ends with a directory separator then it means that
+		// "path" must be of the form "$parent_dir/$ANYTHING" because of the
+		// trailing path separator is explicitly present.
+		if (parent_len > 0 && parent_dir[parent_len - 1] == '/') {
+			return true;
+		}
+		// If "path" either is either identical to "parent_dir" or it contains
+		// a path separator character "/" exactly where at the location where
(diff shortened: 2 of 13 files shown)


cmd/snap-confine/mount-support.c
```diff
@@ -505,6 +505,39 @@ static void sc_bootstrap_mount_namespace(const struct sc_mount_config *config)
 	// mount table and software inspecting the mount table may become confused.
 	sc_must_snprintf(src, sizeof src, "%s/proc", SC_HOSTFS_DIR);
 	sc_do_umount(src, UMOUNT_NOFOLLOW | MNT_DETACH);
+	// Detach the both views into /writable: the one from hostfs and the one
+	// directly visible in /writable. Interfaces don't grate access to this
```
bboozzoo: s/grate/grant/ ?

cmd/snap-confine/mount-support.c
```diff
@@ -505,6 +505,39 @@ static void sc_bootstrap_mount_namespace(const struct sc_mount_config *config)
...
+	if (config->distro != SC_DISTRO_CLASSIC) {
+		sc_must_snprintf(src, sizeof src, "%s/writable", SC_HOSTFS_DIR);
+		sc_mountinfo *sm SC_CLEANUP(sc_cleanup_mountinfo) = NULL;
+		// We use a crude loop to avoid the problem of ordering the mount
+		// points correctly from mostly nested to least nested. This way we
+		// just unmount the first viable candidate we see and start over. This
+		// guarantees that we either didn't want to unmount anything and
+		// finished or we unmounted at least one thing before trying the
+		// sequence again.
+ again:
```
bboozzoo: I have some mixed feelings about goto for loop like constructs, rather than direct jump to failure/success handling path.

How about something like this: 
```
bool needs_scan = true;
while (needs_scan) {
  sc_mountinfo *sm SC_CLEANUP(sc_cleanup_mountinfo) = NULL;
  sm = sc_parse_mountinfo(NULL);
  if (sm == NULL) {
    die("cannot parse /proc/self/mountinfo");
  }
  
  sc_mountinfo_entry *entry = sc_first_mountinfo_entry(sm);
  bool unmount = false;
  while (entry != NULL) {
    const char *mount_dir = entry->mount_dir;
    ummount = sc_startswith(mount_dir, "/writable") || sc_startswith(mount_dir, src);
    if (umount) {
      sc_do_mount("none", mount_dir, NULL,
                  MS_REC | MS_PRIVATE, NULL);
      sc_do_umount(mount_dir,
                   UMOUNT_NOFOLLOW | MNT_DETACH);
      break;
    }
    entry = sc_next_mountinfo_entry(entry);
  }
  
  /* do another pass if we unmounted anything */
  needs_scan = unmount;
}
```

cmd/snap-confine/mount-support.c
zyga: Fixed, thank you.

cmd/snap-confine/mount-support.c
zyga: Applied with a small tweak.

cmd/snap-confine/mount-support.c
```diff
@@ -509,6 +509,71 @@ static void sc_bootstrap_mount_namespace(const struct sc_mount_config *config)
...
+			while (entry != NULL) {
+				const char *mount_dir = entry->mount_dir;
+				should_unmount =
+				    sc_startswith(mount_dir, "/writable")
+				    || sc_startswith(mount_dir, src);
+				if (should_unmount) {
+					sc_do_mount("none", mount_dir, NULL,
+						    MS_REC | MS_PRIVATE, NULL);
+					sc_do_umount(mount_dir,
+						     UMOUNT_NOFOLLOW |
```
jdstrand: It is probably worth adding a comment that a mountpoint which is now a symlink should not happen under normal conditions since a) the mount syscall follows 'man path_resolution' and resolves the target mountpoint such that mountinfo will not have a symlink in its pathand b) if the sc_do_umount is raced with sc_parse_mountinfo (eg, mountpoint is moved with MS_MOVE/unmounted/replaced with a symlink) then we are under attack and want to die.

cmd/snap-confine/mount-support.c
```diff
@@ -509,6 +509,71 @@ static void sc_bootstrap_mount_namespace(const struct sc_mount_config *config)
...
+				sc_do_mount("none", mount_dir, NULL,
+					    MS_REC | MS_PRIVATE, NULL);
+				sc_do_umount(mount_dir,
+					     UMOUNT_NOFOLLOW | MNT_DETACH);
+				sc_cleanup_mountinfo(&sm);
+				goto again;
+			}
+			entry = sc_next_mountinfo_entry(entry);
+		}
+	}
```
jdstrand: What is this hunk of code for after 'again'? It seems like it should've been removed when you resolved @bboozzoo's comment?

cmd/snap-confine/mount-support.c
```diff
@@ -509,6 +509,71 @@ static void sc_bootstrap_mount_namespace(const struct sc_mount_config *config)
...
+			}
+
+			sc_mountinfo_entry *entry =
+			    sc_first_mountinfo_entry(sm);
+			bool should_unmount = false;
+			while (entry != NULL) {
+				const char *mount_dir = entry->mount_dir;
+				should_unmount =
+				    sc_startswith(mount_dir, "/writable")
+				    || sc_startswith(mount_dir, src);
```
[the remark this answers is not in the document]
zyga: I'll fix this and add a helper as it seems a common mistake pattern to do

cmd/snap-confine/mount-support.c
zyga: Ouh, sorry. Not sure how this happened now.

cmd/snap-confine/mount-support.c
zyga: Please re-visit the PR. I think this comment is no longer necessary because we only unmount exactly known paths.

cmd/snap-confine/mount-support.c
```diff
@@ -509,6 +509,71 @@ static void sc_bootstrap_mount_namespace(const struct sc_mount_config *config)
...
+				if (should_unmount) {
+					sc_do_mount("none", mount_dir, NULL,
+						    MS_REC | MS_PRIVATE, NULL);
+					sc_do_umount(mount_dir,
+						     UMOUNT_NOFOLLOW |
+						     MNT_DETACH);
+					break;	// See the comment above on why we do this.
```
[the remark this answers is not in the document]
zyga: Please re-visit the PR. I think this comment is no longer necessary because we no longer loop.

zyga:
@jdstrand please re-review this branch. I kept two patches: the original idea and the 2nd patch that simplifies it. As a side effect `sc_is_beneath` is no longer necessary but I decided to keep it since it seems like something we may need in the future. If you feel strongly about it I can push a small third patch that removes it while keeping it in history for easy revert.

cmd/libsnap-confine-private/string-utils-test.c
```diff
@@ -74,6 +74,25 @@ static void test_sc_startswith(void)
...
 
+static void test_sc_is_beneath(void)
+{
+	// Essential difference from sc_startswith
+	g_assert_true(sc_is_beneath("/foo/bar", "/foo"));
+	g_assert_false(sc_is_beneath("/foo/bar", "/foo2"));
+	// True cases
+	g_assert_true(sc_is_beneath("/foo", "/foo"));
+	g_assert_true(sc_is_beneath("", ""));
+	g_assert_true(sc_is_beneath("/foo", "/"));
```
bboozzoo: For the record, I think you can add:
```
g_assert_true(sc_is_beneath("/foo", ""));
```

cmd/libsnap-confine-private/string-utils-test.c
```diff
@@ -74,6 +74,25 @@ static void test_sc_startswith(void)
 	g_assert_false(sc_startswith("ba", "bar"));
 }
 
+static void test_sc_is_beneath(void)
+{
+	// Essential difference from sc_startswith
+	g_assert_true(sc_is_beneath("/foo/bar", "/foo"));
+	g_assert_false(sc_is_beneath("/foo/bar", "/foo2"));
```
bboozzoo: I think you can add relative paths too:
```
g_assert_true(sc_is_beneath("foo/bar", "foo"));
```

cmd/snap-confine/mount-support.c
```diff
@@ -509,6 +511,38 @@ static void sc_bootstrap_mount_namespace(const struct sc_mount_config *config)
...
+	// exposes writable, with a structure specific to ubuntu-core. 
+	debug("detaching %s", hostfs_writable_dir);
+	sc_do_mount("none", hostfs_writable_dir, NULL,
+		    MS_REC | MS_PRIVATE, NULL);
+	sc_do_umount(hostfs_writable_dir, UMOUNT_NOFOLLOW | MNT_DETACH);
+
+	// On ubuntu-core 16, when the executed snap uses core as base we also see
+	// the /writable that we directly inherited from the initial mount
+	// namespace.
+	if (distro == SC_DISTRO_CORE16 && !normal_mode) {
```
bboozzoo: Obviously not to be considered in this PR, but it'd be great if we had more descriptive name for `normal_mode`. Each time I see it, I need to go back and check what it means.

zyga: I was hoping to be able to remove the legacy mode and have no more mode at all but this idea seems futile. Perhaps you have some suggestions on where and how to describe those. In my eyes those are:

normal mode: rootfs constructed from base snap _only_
legacy mode: rootfs constructed from booted filesystem with anything visible there

cmd/libsnap-confine-private/string-utils.c
```diff
@@ -72,6 +72,33 @@ bool sc_startswith(const char *str, const char *prefix)
 	return strncmp(str, prefix, xlen) == 0;
 }
 
+bool sc_is_beneath(const char *path, const char *parent_dir)
+{
+	if (!path || !parent_dir) {
+		return false;
+	}
+	// The path "path" is beneath the directory "parent_dir" if "path" is a
+	// prefix of "parent_dir" with the following additional restrictions.
```
jdstrand: This comment isn't right. This is the API: `sc_startswith(const char *str, const char *prefix)` and you call `if (sc_startswith(path, parent_dir)) {` in the next line, so the comment should read:
```
// The path "path" is beneath the directory "parent_dir" if "parent_dir"
// is a prefix of "path" with the following additional restrictions.
```

cmd/libsnap-confine-private/string-utils-test.c
```diff
@@ -74,6 +74,27 @@ static void test_sc_startswith(void)
...
+	g_assert_true(sc_is_beneath("foo/bar", "foo"));
+	// True cases
+	g_assert_true(sc_is_beneath("/foo", "/foo"));
```
jdstrand: Hmm? API-wise, this seems like it should be false.

cmd/libsnap-confine-private/string-utils.c
```diff
@@ -72,6 +72,33 @@ bool sc_startswith(const char *str, const char *prefix)
...
+			return true;
+		}
+		// If "path" either is either identical to "parent_dir" or it contains
+		// a path separator character "/" exactly where at the location where
+		// "parent_dir" ends. For example "/foo/bar" is beneath "/foo" because
+		// of "/foo/bar"[4] is "/". This eliminates "/foo2" from matching
+		// "/foo".
+		if (path[parent_len] == '\0' || path[parent_len] == '/') {
+			return true;
+		}
```
jdstrand: This is more complicated to read than it should be IMHO. 'If "path" either is either identical to "parent_dir"' suggests to me `sc_streq()`. Also, why is "/foo" beneath "/foo"? "/foo" is by definition at the same level as "/foo".

cmd/snap-confine/mount-support.c
```diff
@@ -509,6 +511,38 @@ static void sc_bootstrap_mount_namespace(const struct sc_mount_config *config)
 	// mount table and software inspecting the mount table may become confused.
 	sc_must_snprintf(src, sizeof src, "%s/proc", SC_HOSTFS_DIR);
 	sc_do_umount(src, UMOUNT_NOFOLLOW | MNT_DETACH);
+	// Detach the both views of /writable: the one from hostfs and the one
```
jdstrand: s/the both/both/

cmd/snap-confine/mount-support.c
```diff
@@ -509,6 +511,38 @@ static void sc_bootstrap_mount_namespace(const struct sc_mount_config *config)
...
+
+	// Writable only exists on ubuntu-core.
+	if (distro == SC_DISTRO_CLASSIC) {
+		return;
+	}
+	// On all core distributions we see /var/lib/snapd/hostfs/writable that
+	// exposes writable, with a structure specific to ubuntu-core. 
+	debug("detaching %s", hostfs_writable_dir);
+	sc_do_mount("none", hostfs_writable_dir, NULL,
+		    MS_REC | MS_PRIVATE, NULL);
```
jdstrand: You are using `MS_REC|MS_PRIVATE` here and below due to https://github.com/snapcore/snapd/blob/master/cmd/snap-update-ns/change.go#L333. Can you add a top-level comment for `sc_detach_views_of_writable()` to say why here as well?

jdstrand:
> Thanks for the PR and the cleanup! It is great to see all of those 'writable' entries fall out of the testsuite. I'm am marking this as Approved conditional on removing `sc_is_beneath()` since a) I don't particularly care for unused APIs in the setuid snap-confine and b) there are discussion points wrt the API and implementation that distract from what this PR is meant to be. When you revert, please either link back to this PR so posterity can see those discussion points or alternatively address all of them then revert the whole batch.

@mvo5 and @zyga - please note my approval is conditional on removing `sc_is_beneath()`. So long as it is removed and other changes aren't made, no need to circle back to me. If other changes are made and the API is kept, please have @alexmurray review.

cmd/libsnap-confine-private/string-utils-test.c
```diff
@@ -74,6 +74,27 @@ static void test_sc_startswith(void)
...
+	g_assert_true(sc_is_beneath("", ""));
+	g_assert_true(sc_is_beneath("/foo", "/"));
+	g_assert_true(sc_is_beneath("/foo", ""));
```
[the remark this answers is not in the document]
zyga: I agree that it feels like a tri-state answer is desired. Yes, no, error. I ended up removing this function without addressing this.

cmd/libsnap-confine-private/string-utils.c
zyga: I've removed this function now. The commit message left a remark about unaddressed concerns in case anyone wants to bring it back.

cmd/libsnap-confine-private/string-utils-test.c
zyga: Perhaps the name "beneath" is incorrect, in my original intent it was about a subset of paths that are rooted at /foo. This obviously includes /foo in this sense (it's not a proper subset). The name did not convey this though.

cmd/libsnap-confine-private/string-utils.c
zyga: /foo is indeed at the same level as /foo but for the purpose I intended this the function would include that case. It wasn't really beneath, as I said in the comment above, but rather, "sharing parent directory with".

cmd/snap-confine/mount-support.c
zyga: Done now

PR #7422, opened 2019-09-06
interfaces: allow reading mutter Xauthority file

zyga:
On openSUSE Tumbleweed at least on version VERSION_ID="20190829",
but possible earlier, when running GNOME in Wayland mode, graphical
applications are not starting correctly.

This was traced to a missing rule describing an apparently new place for
Xauthority file.

Fixes: https://bugs.launchpad.net/snapd/+bug/1842615

### DIFF
--- a/interfaces/builtin/wayland.go
+++ b/interfaces/builtin/wayland.go
@@ -59,6 +59,7 @@ owner /run/user/[0-9]*/wayland-[0-9]* rwk,
 # Allow reading an Xwayland Xauth file
 # (see https://gitlab.gnome.org/GNOME/mutter/merge_requests/626)
 /run/user/[0-9]*/.mutter-Xwaylandauth.* r,
+/run/user/[0-9]*/mutter/Xauthority r,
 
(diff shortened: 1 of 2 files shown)


interfaces/builtin/wayland.go
```diff
@@ -59,6 +59,7 @@ owner /run/user/[0-9]*/wayland-[0-9]* rwk,
 # Allow reading an Xwayland Xauth file
 # (see https://gitlab.gnome.org/GNOME/mutter/merge_requests/626)
 /run/user/[0-9]*/.mutter-Xwaylandauth.* r,
+/run/user/[0-9]*/mutter/Xauthority r,
```
bboozzoo: Looks like the path was introduced 2 months ago:
https://gitlab.gnome.org/GNOME/mutter/commit/a8984a81c2e887623d69ec9989ae8a5025f7bd47

bboozzoo: Duh, different path.

PR #7429, opened 2019-09-09
wrappers/services: add ServicesEnableState + unit tests

anonymouse64:
ServicesEnableState returns a map of service name to a boolean of whether the service is enabled or not. It is useful for when we start saving disabled services when refreshing, reverting, disabling and enabled snaps with disabled services.

Also add basic happy path unit test.

Just like https://github.com/snapcore/snapd/pull/7428, this helper is meant as part of a larger PR to prevent disabled services from being re-enabled after a refresh, revert or enable/disable cycle, broken out to be easier and quicker to review.


wrappers/services_test.go
```diff
@@ -189,6 +189,67 @@ apps:
...
+	c.Assert(err, IsNil)
+
+	c.Assert(states, DeepEquals, map[string]bool{
+		"svc1": false,
+		"svc2": true,
+	})
+
+	// the calls could be out of order in the list, since iterating over a map
+	// is non-deterministic, so manually check each call
+	for _, call := range r.Calls() {
```
mvo5: I think it makes sense to add something like `c.Check(r.Calls(), HasLen, 2)` here to ensure that this does not silently works if e.g. no calls at all happen.

wrappers/services_test.go
```diff
@@ -189,6 +189,67 @@ apps:
...
+		c.Assert(call, HasLen, 5)
+		c.Assert(call[:4], DeepEquals, []string{"systemctl", "--root", s.tempdir, "is-enabled"})
+		switch call[4] {
+		case svc1File:
```
mvo5: (nitpick^2) This could be written in a single like: `case svc1File, svc2File:`

wrappers/services.go
```diff
@@ -335,7 +335,27 @@ func StopServices(apps []*snap.AppInfo, reason snap.ServiceStopReason, inter int
 	}
 
 	return nil
+}
+
+// CurrentSnapServiceStates builds a map of service names and whether the snap
+// is enabled or disabled
+func CurrentSnapServiceStates(s *snap.Info, inter interacter) (map[string]bool, error) {
```
bboozzoo: I think you can drop the `Snap` part. How about `CurrentServicesEnableState` or just `ServicesEnableState`?

wrappers/services.go
```diff
@@ -335,7 +335,27 @@ func StopServices(apps []*snap.AppInfo, reason snap.ServiceStopReason, inter int
...
+	// loop over all services in the snap, querying systemd for the current
+	// systemd state of the snaps
+	snapSvcsState := make(map[string]bool, len(s.Apps))
+	for name, app := range s.Apps {
+		if !app.IsService() {
+			continue
+		}
+		state, err := sysd.IsEnabled(app.ServiceName())
+		if err != nil {
+			return nil, err
```
bboozzoo: Missing unit tests for this code path

wrappers/services.go
anonymouse64: I like `ServicesEnableState`, I'll go with that I think

wrappers/services.go
anonymouse64: ok, will add

wrappers/services_test.go
anonymouse64: good catch

wrappers/services_test.go
anonymouse64: yes that's a good idea, thanks

wrappers/services.go
```diff
@@ -335,7 +335,27 @@ func StopServices(apps []*snap.AppInfo, reason snap.ServiceStopReason, inter int
 	}
 
 	return nil
+}
+
+// ServicesEnableState builds a map of service names and whether the snap
```
bboozzoo: Jumping bettwen this and #7437 made me notice that the comment needs a bit of a change, probably something in the lines of:
```
// ServicesEnableState returns a map of service names from the given snap, together with their enable/disable status.
```

wrappers/services_test.go
```diff
@@ -189,6 +189,120 @@ apps:
...
+	esac
+
+	exit 1
+	`)
+	defer r.Restore()
+
+	_, err := wrappers.ServicesEnableState(info, progress.Null)
+	c.Assert(err, ErrorMatches, ".*is-enabled snap.hello-snap.svc1.service\\] failed with exit status 1: whoops\n.*")
+
+	// the calls could be out of order in the list, since iterating over a map
```
mvo5: This comment is no longer accurate :) Its just a single call so the code below can be simplified (no need to loop anymore)

wrappers/services_test.go
```diff
@@ -189,6 +189,120 @@ apps:
...
+	}
+}
+
+func (s *servicesTestSuite) TestServicesEnableStateFail(c *C) {
+	info := snaptest.MockSnap(c, packageHello, &snap.SideInfo{Revision: snap.R(12)})
+	svc1File := "snap.hello-snap.svc1.service"
+
+	s.systemctlRestorer()
+	r := testutil.MockCommand(c, "systemctl", `#!/bin/sh
+	if [ "$1" = "--root" ]; then
```
mvo5: Given that this is only called once and we always want to return an error we can maybe simplify this to something like:
```
#!/bin/sh
echo "whoops"
exit 1
```
wdyt?

anonymouse64: well I still think it's useful to catch programming errors, for example if for some reason we don't call `systemctl` with `--root`, this test would fail with "unexpected op ..." and it would be clear that there's something more unexpected going on here, so I think I'll keep it as is

wrappers/services_test.go
anonymouse64: good catch, will update

anonymouse64:
The tests seem to keep hitting timeouts with the whole class of `ubuntu-core-transition` tasks, so I merged master with the hopes that resolves the issues...

PR #7430, opened 2019-09-09
overlord/snapstate: add SetTaskSnapSetup helper + unit tests

anonymouse64:
This change is part of how I ended up re-implementing https://github.com/snapcore/snapd/pull/7270.

One of the suggestions was that I use SnapSetup as short-term storage for the list of disabled services, then later on in unlink-snap commit the list to SnapState for long-term, cross-change persistence. However, a problem I encountered was that when I go to write to SnapSetup to save the disabled services from a task like `stop-snap-services`, all future tasks don't automatically get the changes that I wrote to SnapSetup. I needed a way to commit changes to SnapSetup not just to the current `snap-setup` task, but to future tasks as well, so rather than try to figure out all of the future tasks and write `snap-setup` to all of them, I figured an easy way out is to rewrite the `snap-setup` for the first task (which is where most all tasks inherit their `snap-setup` from in TaskSnapSetup). The first task is given by `snap-setup-task`, so I implemented this helper function SetSnapSetupTask to do this.

Also added some unit tests for this helper function, but since this function is generic it didn't really fit into any of the handlers_*_test.go files, so I created a new handlers_test.go file for these generic tests

### DIFF
--- a/overlord/snapstate/handlers.go
+++ b/overlord/snapstate/handlers.go
@@ -77,6 +77,32 @@ func TaskSnapSetup(t *state.Task) (*SnapSetup, error) {
 	return &snapsup, nil
 }
 
+// SetTaskSnapSetup writes the given SnapSetup to the provided task's
+// snap-setup-task Task, or to the task itself if the task does not have a
+// snap-setup-task (i.e. it _is_ the snap-setup-task)
+func SetTaskSnapSetup(t *state.Task, snapsup *SnapSetup) error {
+	if t.Has("snap-setup-task") {
+		// this task isn't the snap-setup-task, so go get that and write to that
+		// one
+		var id string
+		err := t.Get("snap-setup-task", &id)
+		if err != nil {
+			return err
+		}
+
+		ts := t.State().Task(id)
+		if ts == nil {
+			return fmt.Errorf("internal error: tasks are being pruned")
+		}
+		ts.Set("snap-setup", snapsup)
+	} else {
+		// this is the snap-setup-task so just write to the task directly
+		t.Set("snap-setup", snapsup)
+	}
+
+	return nil
+}
+
 func snapSetupAndState(t *state.Task) (*SnapSetup, *SnapState, error) {
 	snapsup, err := TaskSnapSetup(t)
 	if err != nil {
--- a/overlord/snapstate/handlers.go
+++ b/overlord/snapstate/handlers.go
@@ -81,7 +81,10 @@ func TaskSnapSetup(t *state.Task) (*SnapSetup, error) {
 // snap-setup-task Task, or to the task itself if the task does not have a
 // snap-setup-task (i.e. it _is_ the snap-setup-task)
 func SetTaskSnapSetup(t *state.Task, snapsup *SnapSetup) error {
-	if t.Has("snap-setup-task") {
+	if t.Has("snap-setup") {
+		// this is the snap-setup-task so just write to the task directly
+		t.Set("snap-setup", snapsup)
+	} else {
 		// this task isn't the snap-setup-task, so go get that and write to that
 		// one
 		var id string
@@ -95,9 +98,6 @@ func SetTaskSnapSetup(t *state.Task, snapsup *SnapSetup) error {
 			return fmt.Errorf("internal error: tasks are being pruned")
 		}
 		ts.Set("snap-setup", snapsup)
-	} else {
-		// this is the snap-setup-task so just write to the task directly
-		t.Set("snap-setup", snapsup)
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


anonymouse64:
See https://github.com/snapcore/snapd/pull/7431/files#diff-576e1c18f7e56b2aefd219a83265967eR1597-R1631 for an example of how this change is used in my bigger PR.

pedronis:
I think the idea was always that there would be one task in a change per snap with SnapSetup on it, and the others just with the id pointing to it. That was not respected/done consistently.  The main issue with changing any of this is what to do with already existing tasks when the new version of snapd restarts.

anonymouse64:
@pedronis it sounds like from what you said that as long as we save the
list of disabled services in a task other than the task we actually disable
them in that we will need some new patch to fix old tasks, correct?

anonymouse64:
From SU, this approach is okay to move forward with, but as pointed out we will need some form or patch level or something that to make tasks with the overall changes work when coming online in a new snapd with tasks that were from an old snapd.

anonymouse64:
This is ready for full review now, I've made it so that all the first tasks now define `snap-setup-task` and have expanded some tests to make sure that snap-setup-task is defined for the first task and is expected.

There's no tests for `rewriteSetupTaskSnapSetup`, but I'm not sure how to test that here without the rest of the changes so I think adding the tests for that in the bigger PR makes sense, but considering that if folks want me to drop that function here and add it where it's used I can do that too.

overlord/snapstate/handlers_test.go
```diff
@@ -0,0 +1,114 @@
...
+	snapsup.Channel = "edge"
+	err := snapstate.SetTaskSnapSetup(t, snapsup)
+	c.Assert(err, IsNil)
+
+	// get a fresh version of this task from state to check that the task's
+	/// snap-setup has the changes in it now
+	newT := s.state.Task(t.ID())
+	c.Assert(newT, Not(IsNil))
+	newsnapsup := &snapstate.SnapSetup{}
+	newT.Get("snap-setup", newsnapsup)
```
pedronis: the codebase tends to have this kind of code as:

      var newsnapsup snapstate.SnapSetup
      err := newT.Get("snap-setup", &newsnapsup)
      c.Assert(err, IsNil)

overlord/snapstate/handlers_test.go
```diff
@@ -0,0 +1,114 @@
...
+	// mutate it and rewrite it with the helper
+	snapsup.Channel = "edge"
+	err := snapstate.SetTaskSnapSetup(t2, snapsup)
+	c.Assert(err, IsNil)
+
+	// check that the original task's snap-setup is different now
+	newT := s.state.Task(t.ID())
+	c.Assert(newT, Not(IsNil))
+	newsnapsup := &snapstate.SnapSetup{}
+	newT.Get("snap-setup", newsnapsup)
```
pedronis: same

overlord/snapstate/snapstate.go
```diff
@@ -1892,9 +1892,7 @@ func Remove(st *state.State, name string, revision snap.Revision, flags *RemoveF
 		removeHook := SetupRemoveHook(st, snapsup.InstanceName())
 		addNext(state.NewTaskSet(removeHook))
 		prev = removeHook
-	}
 
-	if removeAll {
```
pedronis: for next time: this is a drive-by which is ok,  we tend to put "drive-by" in the commit message for those, also if possible to do them first in the commit set

overlord/snapstate/handlers.go
```diff
@@ -77,6 +77,32 @@ func TaskSnapSetup(t *state.Task) (*SnapSetup, error) {
 	return &snapsup, nil
 }
 
+// SetTaskSnapSetup writes the given SnapSetup to the provided task's
+// snap-setup-task Task, or to the task itself if the task does not have a
+// snap-setup-task (i.e. it _is_ the snap-setup-task)
+func SetTaskSnapSetup(t *state.Task, snapsup *SnapSetup) error {
+	if t.Has("snap-setup-task") {
```
pedronis: I would have flipped the check to:

```
    if t.Has("snap-setup") {
         t.Set("snap-setup", snapsup)
         return
    }
    ...
```

overlord/snapstate/handlers_test.go
anonymouse64: okay, done

overlord/snapstate/handlers_test.go
anonymouse64: done

overlord/snapstate/snapstate.go
anonymouse64: okay, I will try to have the commit mention drive-by and put this commit first in the PR

overlord/snapstate/handlers.go
anonymouse64: done

PR #7433, opened 2019-09-09
systemd: detach rather than unmount .mount units

zyga:
Mount units are a thin wrapper on top of libmount mount operation.
Mounting is relatively straightforward in the sense that it doesn't give
us many options on that fundamentally change the mount operation.

Unmounting on the other hand, is more interesting. Unmounting requires
the filesystem to be no longer used. At all. This, however, is
incredibly difficult to guarantee. A process can change the current
working directory to block a filesystem. The user session can run a
filesystem indexing daemon that will open a number of file descriptors
to the files present in a mounted snap.

We've recently seen a report from a user who had a failed refresh of a
number of snaps. In both the "do" path as well as the "undo" path,
unmounting a revision of the core snap has failed with the EBUSY error,
indicating that something is still keeping the file system busy.

We have an alternative though. We can ask libmount to ask the kernel to
detach a filesystem instead. Unlike real unmounting, the associated
block device and driver is still busy. There can still be an open file
descriptor reading or writing data. At some point the file system will
become truly unused at which point the kernel will perform the regular
unmount automatically.

Traditionally using lazy unmounting (or detaching, as it has two names
in the wild) has carried the risk of breaking the filesystem as it is
possible for it to never really be idle enough to unmount for real.

Fortunately snaps don't have this problem since the underlying squashfs
is never written to. As such, using this mode has the chance of
improving upgrade reliability in face of unknown, concurrent
interactions from user-space.

Related-To: https://bugs.launchpad.net/snappy/+bug/1843286

### DIFF
--- a/overlord/snapstate/backend/mountunit_test.go
+++ b/overlord/snapstate/backend/mountunit_test.go
@@ -89,6 +89,7 @@ What=/var/lib/snapd/snaps/foo_13.snap
 Where=%s/foo/13
 Type=squashfs
 Options=nodev,ro,x-gdu.hide
+LazyUnmount=yes
 
 [Install]
 WantedBy=multi-user.target
--- a/systemd/systemd.go
+++ b/systemd/systemd.go
@@ -674,6 +674,7 @@ What=%s
 Where=%s
 Type=%s
 Options=%s
+LazyUnmount=yes
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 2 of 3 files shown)


zyga:
@mvo5 I'll investigate how to write such a test.

zyga:
@bboozzoo perhaps but I think (fortunately) this bug is self-fixing in a way. As soon as anything refreshes it uses the new semantics already. I think the problem is that it's unclear if we can do this in-flight without side-effects.

I think we'd have to rewrite the mount units and call `systemd daemon-reload` if there were any actual changes applied, right? I'd like to consider exploring that but not in this branch as I think it's important to land without the extra mechanism soon rather than later.

PR #7435, opened 2019-09-09
tests: explicitly restore after using LXD

zyga:
The origin of the bug this patch fixes is in misunderstanding how spread
restore works. Spread has project, suite and task level commands. In
addition, it has "-each" variant of some of those, like
restore-suite-each and restore-project-each that run for each task in a
suite and project respectively. The problem is ordering.

I had incorrectly assumed that the order is that project-each commands
run immediately after the commands specified in a particular task.yaml.
This is false. Those actually run last. The result was that the leak
detector ran before some of the restore logic, detecting a leak for
something that we already know about!

Instead of putting the restore code in both project each and suite each,
so that at least one copy runs "early enough", I decided to undo the
whole magic restore property by bringing this special-case fix to five
tests that actually need it.

I moved the restore code to "lxd-tool" under the sub-command name
"undo-lxd-mount-changes". The tool is invoked in project-prepare as well
as in the tests that use LXD.

I prefer to do it this way to show the cost of restore right in the
tests that actually need it and to save it from all the tests that don't
require special restore operations.

### DIFF
--- a/tests/lib/bin/lxd-tool
+++ b/tests/lib/bin/lxd-tool
@@ -1,0 +1,35 @@
+#!/bin/sh -e
+case "${1:-}" in
+    undo-lxd-mount-changes)
+        # Vanilla systems have /sys/fs/cgroup/cpuset without clone_children option.
+        # Using LXD to create a container enables this option, as can be seen here:
+        #
+        # -37 32 0:32 / /sys/fs/cgroup/cpuset rw,nosuid,nodev,noexec,relatime shared:15 - cgroup cgroup rw,cpuset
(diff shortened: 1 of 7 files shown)


tests/lib/bin/lxd-tool
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh -e
```
anonymouse64: I like this, but I have one question for you ... why not python? :stuck_out_tongue_winking_eye:

zyga: Challenge accepted!

PR #7437, opened 2019-09-09
 wrappers/services.go: add disabled svc list arg to AddSnapServices

anonymouse64:
* AddSnapServices now takes a list of disabled services that shouldn't be enabled at all, so don't enable any of the services from that list. Also check the disabled services provided to AddSnapServices to log messages if any of the services are not services anymore or if they don't exist anymore. We should be noticing these things before it gets here, but in case it doesn't, we shouldn't fail here because those things aren't fatal.
* Add unit tests for happy path of AddSnapServices and for the unhappy cases which just check that we log a message about the inconsistency. In the full bugfix where we use this function, this will be verified before being passed here, so it shouldn't hit that codepath under normal circumstances.
* Also drive-by update to use progress.Null consistently everywhere we call AddSnapServices in the tests

This currently is unused anywhere, but will eventually be used to prevent disabled services from being re-enabled in snapstate.

Re-opening of https://github.com/snapcore/snapd/pull/7428

### DIFF
--- a/wrappers/services.go
+++ b/wrappers/services.go
@@ -34,6 +34,7 @@ import (
 	"github.com/snapcore/snapd/osutil"
 	"github.com/snapcore/snapd/osutil/sys"
 	"github.com/snapcore/snapd/snap"
+	"github.com/snapcore/snapd/strutil"
 	"github.com/snapcore/snapd/systemd"
 	"github.com/snapcore/snapd/timeout"
 	"github.com/snapcore/snapd/timeutil"
@@ -199,9 +200,20 @@ func StartServices(apps []*snap.AppInfo, inter interacter, tm timings.Measurer) 
 }
 
 // AddSnapServices adds service units for the applications from the snap which are services.
-func AddSnapServices(s *snap.Info, inter interacter) (err error) {
+func AddSnapServices(s *snap.Info, disabledSvcs []string, inter interacter) (err error) {
 	if s.GetType() == snap.TypeSnapd {
 		return writeSnapdServicesOnCore(s, inter)
+	}
+
+	// check if any previously disabled services are now no longer services and
+	// log messages about that
+	for _, svc := range disabledSvcs {
+		app, ok := s.Apps[svc]
+		if !ok {
+			logger.Noticef("previously disabled service %s no longer exists", svc)
+		} else if !app.IsService() {
+			logger.Noticef("previously disabled service %s is now an app and not a service", svc)
+		}
 	}
 
 	sysd := systemd.New(dirs.GlobalRootDir, systemd.SystemMode, inter)
@@ -277,6 +289,12 @@ func AddSnapServices(s *snap.Info, inter interacter) (err error) {
 		}
 
 		svcName := app.ServiceName()
+
+		if strutil.ListContains(disabledSvcs, app.Name) {
+			// service is disabled, nothing to do
+			continue
+		}
+
 		if err := sysd.Enable(svcName); err != nil {
 			return err
 		}
--- a/wrappers/services_test.go
+++ b/wrappers/services_test.go
@@ -218,7 +218,7 @@ func (s *servicesTestSuite) TestServicesEnableState(c *C) {
 				exit 0
 				;;
 			*)
-				echo "unexpected service $*"
+				echo "unexpected is-enabled of service $2"
 				exit 2
 				;;
 			esac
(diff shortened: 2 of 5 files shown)


anonymouse64:
See how this is used here (in #7431):

https://github.com/snapcore/snapd/blob/3d7e0d833e1fc60553a09872e7f3b0dd1960fc7f/overlord/snapstate/backend/link.go#L160-L174

wrappers/services.go
```diff
@@ -199,7 +199,7 @@ func StartServices(apps []*snap.AppInfo, inter interacter, tm timings.Measurer)
 }
 
 // AddSnapServices adds service units for the applications from the snap which are services.
-func AddSnapServices(s *snap.Info, inter interacter) (err error) {
+func AddSnapServices(s *snap.Info, disabledSvcs []*snap.AppInfo, inter interacter) (err error) {
```
bboozzoo: I think it'd be a bit easier if we pass a list of app names here. Then it'd match the map retuned by `wrappers.ServicesEnableState` and what is kept in snap-setup/snap-state.
Eg:
```
func AddSnapServices(s *snap.Info, disabledSvcs []string, inter interacter) (err error) {
```

wrappers/services.go
```diff
@@ -277,10 +277,27 @@ func AddSnapServices(s *snap.Info, inter interacter) (err error) {
...
+			// services
+			if !disabledSvc.IsService() {
+				return fmt.Errorf("snap app %s is not a service, cannot disable", disabledSvc.Name)
+			}
+			if disabledSvc.Name == app.Name {
+				shouldEnable = false
+			}
+		}
+
+		if shouldEnable {
```
bboozzoo: then you could do the following here:
```
if strutil.ListContains(disabledSvcs, app.Name) {
    // service is disabled, nothing to do
    continue
}

if err := sysd.Enable(svcName); err != nil {
    return err
}
enabled = append(enabled, svcName)
```

wrappers/services.go
```diff
@@ -277,10 +277,27 @@ func AddSnapServices(s *snap.Info, inter interacter) (err error) {
...
-			return err
+
+		// if this service is in the list of services to be left disabled, don't
+		// enable it
+		shouldEnable := true
+		for _, disabledSvc := range disabledSvcs {
+			// sanity check that all apps in the disabledSvcs list are in fact
+			// services
+			if !disabledSvc.IsService() {
+				return fmt.Errorf("snap app %s is not a service, cannot disable", disabledSvc.Name)
```
bboozzoo: This should be done earlier, probably before this line https://github.com/snapcore/snapd/pull/7437/files#diff-08088787360fb3ca74a0aed4c6103b22R207

Also, AFAIK it's perfectly legal if an app that was a service in the previous revision of the snap (as in previously installed, not exactly the n-1 revision), becomes a regular app in the current revision. Perhaps, we should log it, but not throw an error. Unless, those apps are filtered out earlier.

anonymouse64: hmm that's an interesting point, it is possible for services to become apps, given this I think you're right we shouldn't fail here then

wrappers/services.go
anonymouse64: I was trying to match the other methods here which all take AppInfo as an arg when referring to a service, but agreed it would be a fair bit simpler to just use`[]string` instead

anonymouse64: I was trying to match the other functions taking in `AppInfo` rather than a service name, but you're right that would simplify the logic. I will adjust this

wrappers/services.go
anonymouse64: what I ended up doing here is to just log messages about previously disabled services that are now apps and previously disabled services that are missing from the snap, but the larger change will now check for both of those cases when handling that before providing it here, so in practice I don't think this codepath will be hit, but still good to be defensive here and log a notice if we do somehow hit it.

wrappers/services.go
anonymouse64: yes this is simpler, I ended up going with this, thanks

anonymouse64:
The changes here ended up needing a non-trivial merge/rebase with master, so I just rebased on master and force-pushed, but the delta was small so I don't think it's a lot to re-review here.

wrappers/services_test.go
```diff
@@ -294,6 +295,163 @@ func (s *servicesTestSuite) TestServicesEnableStateFail(c *C) {
...
+	case "$1" in
+		enable)
+			case "$2" in 
+				"snap.hello-snap.svc1.service")
+					echo "unexpected enable"
+					exit 1
+				;;
+				"snap.hello-snap.svc2.service")
+					exit 0
+					;;
```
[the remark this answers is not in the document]
anonymouse64: good catch, will do!

PR #7439, opened 2019-09-10
packaging: remove obsolete usr.lib.snapd.snap-confine in postinst

mvo5:
We renamed the apparmor profile for snap-confine to
usr.lib.snapd.snap-confine.real a while ago. Unfortunately due to
a buggy upgrade some people still have the old file and that can
cause real issues like:
https://forum.snapcraft.io/t/telegram-snap-fails-to-start/13132
when this file is used instead of the correct one.

This PR uses a rather aggressive approach to get rid of the file.
Fwiw, I wish we could ship these files in something else than /etc
they are not really conffiles anyway IMO, ppl fiddling with them
have an extremly high chance of breaking snapd this way.


packaging/ubuntu-16.04/snapd.postinst
```diff
@@ -45,7 +45,7 @@ case "$1" in
         # the old usr.lib.snap-confine file. This seems to be loaded instead
         # of the correct usr.lib.snap-confine.real profile. To fix this we
         # use the rather blunt approach to remove the file forcefully.
-        if test -f /etc/apparmor.d/usr.lib.snapd.snap-confine && test "$(md5sum /etc/apparmor.d/usr.lib.snapd.snap-confine | cut -f1 -d' ')" = "2a38d40fe662f46fedd0aefbe78f23e9"; then
+        if test -f /etc/apparmor.d/usr.lib.snapd.snap-confine && test -f /etc/apparmor.d/usr.lib.snapd.snap-confine.real; then
```
anonymouse64: I think I understand this change, it looks like we are now checking (in python psuedo-code):
```python
if exists("/etc/apparmor.d/usr.lib.snapd.snap-confine") and exists("/etc/apparmor.d/usr.lib.snapd.snap-confine.real"):
    remove("/etc/apparmor.d/usr.lib.snapd.snap-confine")
```
which makes sense if the new file we always want is `/etc/apparmor.d/usr.lib.snapd.snap-confine.real` no matter what.

However previously, the logic was just that
```python
if exists("/etc/apparmor.d/usr.lib.snapd.snap-confine") and hasSpecificContent("/etc/apparmor.d/usr.lib.snapd.snap-confine"):
    remove("/etc/apparmor.d/usr.lib.snapd.snap-confine")
```
It seems like the previous logic operated under the assumption that the file should only be removed if it had specific content, so that begs the question is there ever a situation where the file `/etc/apparmor.d/usr.lib.snapd.snap-confine` should _not_ be removed? It seems like the only situation we wouldn't want to remove that file is when we don't have the new file available, so maybe instead we could just do:

```suggestion
        # always delete the old version if we have the new version available
        if test -f /etc/apparmor.d/usr.lib.snapd.snap-confine.real; then
```

zyga: I think that as long as we have the `.real` file in the package we should remove the other one. Having said that, perhaps at some point we will be able to drop the `.real` suffix again.

anonymouse64: Gosh I really hope we wouldn't rename the file again, all of the bugs from the first rename seems to suggest that renaming this file is very sticky and messy

PR #7442, opened 2019-09-10
tests: extend mount-ns test to handle mimics

zyga:
The mount namespace test is very useful for exploring the layout of the
mount namespace used by snapd under various circumstances. The test
already covered several key combinations of the mount namespace: Ubuntu
16.04 and 18.04 both as classic and core hosts also as base snaps. The
test explored the vanilla behavior of the mount namespace, as
constructed by snap-confine.

This patch extends that test to cover the properties of writable mimic,
as created by snap-update-ns. This test is meant as a prerequisite for
the upcoming change to propagation in mount namespaces.

NOTE: This patch is also showing the effect of bug https://bugs.launchpad.net/snapd/+bug/1843423 - namely that /usr/share holds the mimic tmpfs, rather than what one might expect, /usr/share/X11.

### DIFF
--- a/tests/main/mount-ns/google.ubuntu-16.04-64/PER-SNAP-18.expected.txt
+++ b/tests/main/mount-ns/google.ubuntu-16.04-64/PER-SNAP-18.expected.txt
@@ -54,6 +54,9 @@
 0:0 /tmp /tmp rw,relatime master:1 - ext4 /dev/sda1 rw,data=ordered
 0:0 /tmp/snap.test-snapd-mountinfo-core18/tmp /tmp rw,relatime - ext4 /dev/sda1 rw,data=ordered
 2:0 /usr/lib/snapd /usr/lib/snapd ro,nodev,relatime master:16 - squashfs /dev/loop0 ro
+1:34 / /usr/share/gdb rw,relatime - tmpfs tmpfs rw,mode=755
+2:1 /usr/share/gdb/auto-load /usr/share/gdb/auto-load ro,nodev,relatime master:17 - squashfs /dev/loop1 ro
+1:35 / /usr/share/gdb/test rw,relatime - tmpfs tmpfs rw
 0:0 /usr/src /usr/src rw,relatime master:1 - ext4 /dev/sda1 rw,data=ordered
 0:0 /var/lib/snapd /var/lib/snapd rw,relatime master:1 - ext4 /dev/sda1 rw,data=ordered
 0:0 / /var/lib/snapd/hostfs rw,relatime master:1 - ext4 /dev/sda1 rw,data=ordered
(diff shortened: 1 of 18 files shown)


tests/main/mount-ns/google.ubuntu-16.04-64/PER-SNAP-18.expected.txt
```diff
@@ -54,21 +54,75 @@
...
-1068 1060 2:1 / /var/lib/snapd/hostfs/snap/core18/1 ro,nodev,relatime master:17 - squashfs /dev/loop1 ro
-1069 1060 2:2 / /var/lib/snapd/hostfs/snap/test-snapd-mountinfo-classic/1 ro,nodev,relatime master:18 - squashfs /dev/loop2 ro
-1070 1060 2:3 / /var/lib/snapd/hostfs/snap/test-snapd-mountinfo-core16/1 ro,nodev,relatime master:19 - squashfs /dev/loop3 ro
-1071 1060 2:4 / /var/lib/snapd/hostfs/snap/test-snapd-mountinfo-core18/1 ro,nodev,relatime master:20 - squashfs /dev/loop4 ro
-1072 1001 0:0 /var/log /var/log rw,relatime master:1 - ext4 /dev/sda1 rw,data=ordered
-1073 1001 0:0 /var/snap /var/snap rw,relatime master:1 - ext4 /dev/sda1 rw,data=ordered
-1074 1001 0:0 /var/tmp /var/tmp rw,relatime master:1 - ext4 /dev/sda1 rw,data=ordered
+1057 1001 1:34 / /usr/share rw,relatime - tmpfs tmpfs rw,mode=755
+1058 1057 1:35 / /usr/share/X11/test rw,relatime - tmpfs tmpfs rw
+1059 1057 2:1 /usr/share/adduser /usr/share/adduser ro,nodev,relatime master:17 - squashfs /dev/loop1 ro
```
anonymouse64: why is it that with a snap that has core18 base there are all of these additional paths bind mounted from the mimic where as with a core base there is only the first 2? Is that because the mimic bug only is exhibited on snaps using the core18 base snap and not on core base snap? I didn't get that from the PR description or the bug description which is why I'm confused to see all these additional bind mounts

zyga: @anonymouse64 I'll double check but the contents of `/usr/share` is replicated as distinct mount points. It is possible that `core16` and `core18` simply differ in this regard.

Note that I will change this branch to drop the `mount_id` and `parent_id` fields to make the diffs much more readable. Then we can properly evaluate the behaviour of mimics.

zyga: I've opened https://github.com/snapcore/snapd/pull/7448 to this effect. I'll merge it into this branch as soon as I have confidence that it can be merged.

zyga: I've merged 7448 here as I described. As branches land I will take action to reduce the diff in this branch to its essence. As a separate comment, the question raised by @anonymouse64 is valid and stands.

zyga: I think this mystery is now resolved.

PR #7444, opened 2019-09-11
overlord/snapstate/policy, etc: introduce policy, move canRemove to it

chipaca:
This is the first step towards having a fully fledged snapstate/policy
package where logic around type and model decisions can be placed.

For now it's connected in a rather hackish way (which we're already
using elsewhere) although the expectation is that this could change
with some major refactoring of snapstate itself.

This first bit of work moves canRemove.

The tests are a little messed up, because they are straight
translations of the existing ones which were similarly confused. I'd
like to reorganise them so they make more sense though.


overlord/snapstate/policy/policy.go
```diff
@@ -0,0 +1,184 @@
...
+
+func coreInUse(st *state.State) bool {
+	snapStates, err := snapstate.All(st)
+	if err != nil {
+		return err != state.ErrNoState
+	}
+	for name, snapst := range snapStates {
+		for _, si := range snapst.Sequence {
+			if snapInfo, err := snap.ReadInfo(name, si); err == nil {
+				if snapInfo.GetType() != snap.TypeApp || snapInfo.GetType() == snap.TypeSnapd {
```
chipaca: yes the second part of this `if` makes no sense. It's verbatim what was there before... but I'll probably fix it anyway.
in any case the whole `if` doesn't make sense.

overlord/snapstate/policy/policy.go
```diff
@@ -0,0 +1,184 @@
...
+			}
+		}
+	}
+	return false
+}
+
+func baseInUse(st *state.State, baseName string) bool {
+	snapStates, err := snapstate.All(st)
+	if err != nil {
+		return false
```
chipaca: note how this and `coreInUse` react differently to errors that are not `ErrNoState`. I'm not sure I like that.

chipaca:
closing for reorg

PR #7445, opened 2019-09-11
overlord/snapstate/policy, etc: introduce policy, move canRemove to it

chipaca:
This is the first step towards having a fully fledged snapstate/policy
package where logic around type and model decisions can be placed.

For now it's connected in a rather hackish way (which we're already
using elsewhere) although the expectation is that this could change
with some major refactoring of snapstate itself.

This first bit of work moves canRemove.

The tests are a little messed up, because they are straight
translations of the existing ones which were similarly confused. I'd
like to reorganise them so they make more sense though.

In refactoring this we caught several now-obvious bugs.  Also thanks
to this it was very easy to change the API to return the *reason* you
couldn't remove something, so I did; it's not made things grow too
big. So now you get e.g.

    ~$ snap remove core18
    error: cannot remove "core18": snap "core18" is not removable: snap
           is being used by snaps black, bombsquad, cool-retro-term,
           gtk-common-themes, icdiff, shellcheck and word-salad.

### DIFF
--- a/overlord/snapstate/policy/base.go
+++ b/overlord/snapstate/policy/base.go
@@ -39,15 +39,20 @@ func (p *basePolicy) CanRemove(st *state.State, snapst *snapstate.SnapState, rev
 		return errNoName
 	}
 
-	if !rev.Unset() {
-		if boot.InUse(name, rev) {
-			return errInUseForBoot
+	if p.modelBase == name {
+		if !rev.Unset() {
+			// TODO: tweak boot.InUse so that it DTRT when rev.Unset, call
+			// it unconditionally as an extra precaution
+			if boot.InUse(name, rev) {
+				return errInUseForBoot
+			}
+			return nil
 		}
-		return nil
+		return errIsModel
 	}
 
-	if p.modelBase == name {
-		return errIsModel
+	if !rev.Unset() {
+		return nil
 	}
 
 	// a core system could have core18 required in the model due to dependencies for ex
@@ -71,9 +76,15 @@ func baseUsedBy(st *state.State, baseName string) ([]string, error) {
 	}
 	alsoCore16 := false
 	if baseName == "" {
-		// core -> core16 aliasing
-		if snapst, ok := snapStates["core16"]; !ok || !snapst.IsInstalled() {
-			// this base is not installed
+		// if core is installed, a snap having base: core16 will not
+		// pull in core16 itself but use core instead. So if we are
+		// looking at core (a base of ""), and core16 is not installed,
+		// then we need to look out for things having base: core16 as
+		// well as "".
+		//
+		// TODO: if we ever do the converse, using core16 for snaps
+		//       having a base of "", then this needs a tweak.
+		if _, ok := snapStates["core16"]; !ok {
 			alsoCore16 = true
 		}
 	}
@@ -81,9 +92,6 @@ func baseUsedBy(st *state.State, baseName string) ([]string, error) {
 	var usedBy []string
 	for name, snapst := range snapStates {
 		if typ, err := snapst.Type(); err == nil && typ != snap.TypeApp && typ != snap.TypeGadget {
-			continue
-		}
-		if !snapst.IsInstalled() {
 			continue
 		}
 
--- a/overlord/snapstate/policy/base.go
+++ b/overlord/snapstate/policy/base.go
@@ -1,0 +1,73 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package policy
+
+import (
+	"github.com/snapcore/snapd/boot"
+	"github.com/snapcore/snapd/overlord/snapstate"
+	"github.com/snapcore/snapd/overlord/state"
+	"github.com/snapcore/snapd/snap"
+)
+
+type basePolicy struct{}
+
+func (basePolicy) CanRemove(st *state.State, snapst *snapstate.SnapState, all bool) bool {
+	name := snapst.InstanceName()
+	if name == "" {
+		panic("name unset")
+		// not installed, or something. What are you even trying to do.
+		return false
+	}
+
+	if boot.InUse(name, snapst.Current) {
+		return false
+	}
+
+	if !all {
+		return true
+	}
+
+	if snapst.Required {
+		return false
+	}
+
+	// here we use that bases can't be instantiated (InstanceName == SnapName always)
+	return !baseInUse(st, name)
+}
+
+func baseInUse(st *state.State, baseName string) bool {
+	snapStates, err := snapstate.All(st)
+	if err != nil {
+		return false
+	}
+	for name, snapst := range snapStates {
+		for _, si := range snapst.Sequence {
+			if snapInfo, err := snap.ReadInfo(name, si); err == nil {
+				if snapInfo.GetType() != snap.TypeApp {
+					continue
+				}
+				if snapInfo.Base == baseName {
+					return true
+				}
+			}
+		}
+	}
+	return false
+}
--- a/overlord/snapstate/policy/base.go
+++ b/overlord/snapstate/policy/base.go
@@ -94,9 +94,9 @@ func baseUsedBy(st *state.State, baseName string) ([]string, error) {
 				if baseName != snapInfo.Base && !(alsoCore16 && snapInfo.Base == "core16") {
 					continue
 				}
+				usedBy = append(usedBy, snapInfo.InstanceName())
+				break
 			}
-			usedBy = append(usedBy, snapInfo.InstanceName())
-			break
 		}
 	}
 	return usedBy, nil
--- a/overlord/snapstate/policy/base.go
+++ b/overlord/snapstate/policy/base.go
@@ -101,7 +101,7 @@ func baseUsedBy(st *state.State, baseName string) ([]string, error) {
 				if typ := snapInfo.GetType(); typ != snap.TypeApp && typ != snap.TypeGadget {
 					continue
 				}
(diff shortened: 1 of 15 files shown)


overlord/snapstate/policy/os.go
```diff
@@ -0,0 +1,77 @@
...
+
+func coreInUse(st *state.State) bool {
+	snapStates, err := snapstate.All(st)
+	if err != nil {
+		return err != state.ErrNoState
+	}
+	for name, snapst := range snapStates {
+		for _, si := range snapst.Sequence {
+			if snapInfo, err := snap.ReadInfo(name, si); err == nil {
+				if snapInfo.GetType() != snap.TypeApp || snapInfo.GetType() == snap.TypeSnapd {
```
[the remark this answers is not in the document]
chipaca: me neither, that's copied straight through (but I'm dropping this whole function now)

overlord/snapstate/policy/gadget.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"github.com/snapcore/snapd/overlord/snapstate"
+	"github.com/snapcore/snapd/overlord/state"
+)
+
+type gadgetPolicy struct{}
+
+func (gadgetPolicy) CanRemove(_ *state.State, _ *snapstate.SnapState, all bool) bool {
+	return !all
```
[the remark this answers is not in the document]
chipaca: a question about something i'm not 'getting': in what scenario would a snap be blocked from removal because of the model, but not by boot.InUse ?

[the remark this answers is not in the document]
chipaca: I guess it boils down to 'do we want to allow removing the model base on classic'?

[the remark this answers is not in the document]
chipaca: (still not sure how that applies to kernels and gadgets)

overlord/snapstate/policy/base.go
```diff
@@ -0,0 +1,103 @@
...
+			snapInfo, err := snap.ReadInfo(name, si)
+			if err == nil {
+				if typ := snapInfo.GetType(); typ != snap.TypeApp && typ != snap.TypeGadget {
+					continue
+				}
+				if baseName != snapInfo.Base && !(alsoCore16 && snapInfo.Base == "core16") {
+					continue
+				}
+			}
+			usedBy = append(usedBy, snapInfo.InstanceName())
```
chipaca: I suspect I bungled this: I'm using snapInfo when the read errored out

overlord/snapstate/policy/os.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/snapcore/snapd/snap"
+)
+
+type osPolicy struct {
+	modelName string
+}
+
+func (p *osPolicy) CanRemove(st *state.State, snapst *snapstate.SnapState, all bool) bool {
+	name := snapst.InstanceName()
+	if name == "" {
```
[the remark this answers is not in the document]
chipaca: I think I covered them now.

overlord/snapstate/policy/base.go
```diff
@@ -0,0 +1,106 @@
...
+
+	if !rev.Unset() {
+		if boot.InUse(name, rev) {
+			return errInUseForBoot
+		}
+		return nil
+	}
+
+	if p.modelBase == name {
+		return errIsModel
```
pedronis: this branch doesn't seem exercized

overlord/snapstate/policy/base.go
```diff
@@ -0,0 +1,106 @@
...
+	snapStates, err := snapstate.All(st)
+	if err != nil {
+		// note snapstate.All doesn't currently return ErrNoState
+		return nil, err
+	}
+	alsoCore16 := false
+	if baseName == "" {
+		// core -> core16 aliasing
+		if snapst, ok := snapStates["core16"]; !ok || !snapst.IsInstalled() {
+			// this base is not installed
```
pedronis: this comment could be expendaded a bit

overlord/snapstate/policy/gadget.go
```diff
@@ -0,0 +1,48 @@
...
+
+	if !rev.Unset() {
+		return nil
+	}
+
+	if p.modelGadget == name {
+		return errIsModel
+	}
+
+	return nil
```
pedronis: we don't seem to have tests in policy that reach here

overlord/snapstate/policy/kernel.go
```diff
@@ -0,0 +1,52 @@
...
+			return errInUseForBoot
+		}
+		return nil
+	}
+
+	if p.modelKernel == name {
+		return errIsModel
+	}
+
+	return nil
```
pedronis: we don't seem to have tests in policy that reach here

overlord/snapstate/policy/canremove_test.go
```diff
@@ -0,0 +1,315 @@
...
+		Current:  snap.R(1),
+		Sequence: []*snap.SideInfo{{Revision: snap.R(1), RealName: "kernel"}},
+	}
+
+	s.bootloader.SetBootKernel("kernel_1.snap")
+
+	c.Check(policy.NewKernelPolicy("kernel").CanRemove(s.st, snapst, snap.R(1)), check.Equals, policy.ErrInUseForBoot)
+}
+
+func (s *canRemoveSuite) TestOsInUseIsKept(c *check.C) {
```
pedronis: should this name say Base ?

overlord/snapstate/policy/canremove_test.go
```diff
@@ -0,0 +1,315 @@
...
+	c.Check(policy.NewBasePolicy("core18").CanRemove(s.st, snapst, snap.R(1)), check.Equals, policy.ErrInUseForBoot)
+}
+
+func (s *canRemoveSuite) TestRemoveNonModelKernelIsOk(c *check.C) {
+	snapst := &snapstate.SnapState{
+		Current:  snap.R(1),
+		Sequence: []*snap.SideInfo{{Revision: snap.R(1), RealName: "other-non-model-kernel"}},
+	}
+
+	c.Check(policy.NewKernelPolicy("kernel").CanRemove(s.st, snapst, snap.R(1)), check.IsNil)
```
pedronis: this needs to have snap.R(0) ?

overlord/snapstate/snapstate_test.go
```diff
@@ -10969,246 +10967,6 @@ func (snapStateSuite) TestDefaultContentPlugProviders(c *C) {
...
-func (s *canRemoveSuite) TestRemoveNonModelKernelIsOk(c *C) {
-	kernel := &snap.Info{
-		SnapType: snap.TypeKernel,
-	}
-	kernel.RealName = "other-non-model-kernel"
-
-	c.Check(snapstate.CanRemove(s.st, kernel, &snapstate.SnapState{}, true, s.deviceCtx), Equals, true)
-}
-
-func (s *canRemoveSuite) TestRemoveNonModelKernelStillInUseNotOk(c *C) {
```
pedronis: we don't seem to have a test like this anymore

overlord/snapstate/policy/kernel.go
pedronis: maybe we do but something is off and this doesn't get tested

overlord/snapstate/policy/canremove_test.go
chipaca: well caught

overlord/snapstate/policy/canremove_test.go
chipaca: Yeah

overlord/snapstate/snapstate_test.go
chipaca: because you convinced me that the test was testing an impossible situation?

this is the "the kernel used for boot is not the kernel in the model" scenario

pedronis: I see you are right, but then the current code independently of whether that's plausible or not as a strange property:

it won't let you remove the single revision of a kernel not in the model
but it will let you remove the whole thing
that is confusing on face value. Maybe we do need the more complicated code for consistency reasons, and belt and suspender reasons.

overlord/snapstate/policy/errors.go
```diff
@@ -0,0 +1,46 @@
...
+
+func (e inUseByErr) Error() string {
+	switch len(e) {
+	case 0:
+		// how
+		return "snap is being used"
+	case 1:
+		return "snap is being used by snap " + e[0] + "."
+	default:
+		return "snap is being used by snaps " + strings.Join(e[:len(e)-1], ", ") + " and " + e[len(e)-1] + "."
```
stolowski: This can get very long on popey's system ;). Not a serious issue, but perhaps this error could be shortened (e.g. "snap is being used by snaps a, b ,c and 30 more snaps"). Food for thought.

overlord/snapstate/policy/base.go
```diff
@@ -0,0 +1,114 @@
...
+	if snapst.Required {
+		return errRequired
+	}
+
+	// here we use that bases can't be instantiated (InstanceName == SnapName always)
+	usedBy, err := baseUsedBy(st, name)
+	if len(usedBy) == 0 || err != nil {
+		return err
+	}
+	return inUseByErr(usedBy)
```
stolowski: See my remeark below re this error.

overlord/snapstate/policy/base.go
```diff
@@ -0,0 +1,114 @@
...
+			continue
+		}
+
+		for _, si := range snapst.Sequence {
+			snapInfo, err := snap.ReadInfo(name, si)
+			if err == nil {
+				if typ := snapInfo.GetType(); typ != snap.TypeApp && typ != snap.TypeGadget {
+					continue
+				}
+				if baseName != snapInfo.Base && !(alsoCore16 && snapInfo.Base == "core16") {
```
stolowski: This conditional is giving me slight headache, multiple negations read a bit hard... Anyway I think it's ok.

overlord/snapstate/policy/policy.go
```diff
@@ -0,0 +1,61 @@
...
+		return &kernelPolicy{modelKernel: model.Kernel()}
+	case snap.TypeGadget:
+		return &gadgetPolicy{modelGadget: model.Gadget()}
+	case snap.TypeOS:
+		return &osPolicy{modelBase: model.Base()}
+	case snap.TypeBase:
+		return &basePolicy{modelBase: model.Base()}
+	default:
+		return appPolicy{}
+	}
```
stolowski: \o/, Nice

overlord/snapstate/policy/errors.go
chipaca: Yeah, I considered that. I thought I'd ask popey to try once this is on master, and if it's too bad we can do this kind of trimming.

overlord/snapstate/policy/base.go
chipaca: would
```go
!(basename == snapInfo.Base || (alsoCore16 && snapInfo.Base == "core16"))
```
have been clearer?

stolowski: Yes, looks better.

PR #7447, opened 2019-09-11
snapstate: do not allow removal of the snapd snap

mvo5:
We currently allow removing the snapd snap. I think we should
ensure we keep it on the system now that we want to transition
from core->core16+snapd.


anonymouse64:
This change makes sense to me because it matches the behavior of the core snap - even when the core snap is the only snap installed, it still can't be removed, so I think it makes sense for the snapd snap as well.

mvo5:
Closing until the policy work from John has landed

PR #7451, opened 2019-09-11
sandbox/cgroup: introduce cgroup wrappers package

bboozzoo:
Introduce a package wrapping various paths or operations regarding cgroups.


sandbox/cgroup/cgroup.go
```diff
@@ -0,0 +1,145 @@
...
+	"path/filepath"
+	"strings"
+	"syscall"
+
+	"github.com/snapcore/snapd/dirs"
+)
+
+const (
+	// from golang.org/x/sys/unix
+	CGROUP2_SUPER_MAGIC = 0x63677270
```
zyga: We have slight preference for cgroup2SuperMagic and a comment referencing the original name.

sandbox/cgroup/cgroup.go
```diff
@@ -0,0 +1,145 @@
...
+// ProcGroup finds the path of a given cgroup controller for provided process
+// id.
+func ProcGroup(pid int, controller string) (string, error) {
+	f, err := os.Open(ProcPath(pid))
+	if err != nil {
+		return "", err
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
```
zyga: Can I suggest that this is separate from the opening, slightly similar to what we do with mountinfo parsing?

bboozzoo: This was moved from `usersession/userd/helpers.go`. I can split it into a separate patch and include with the actual change in userd perhaps.

zyga: Sounds good.

sandbox/cgroup/cgroup.go
```diff
@@ -0,0 +1,145 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"syscall"
+
+	"github.com/snapcore/snapd/dirs"
+)
+
+const (
+	// from golang.org/x/sys/unix
```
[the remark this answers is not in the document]
bboozzoo: IIRC we dropped it because of build issues on some architectures. #4489

sandbox/cgroup/cgroup.go
```diff
@@ -0,0 +1,145 @@
...
+	return filepath.Join(dirs.GlobalRootDir, defaultMountdir, controller)
+}
+
+func probeCgroupVersion() (version int, err error) {
+	cgroupMount := filepath.Join(dirs.GlobalRootDir, defaultMountdir)
+	typ, err := fsTypeForPath(cgroupMount)
+	if err != nil {
+		return Unknown, fmt.Errorf("cannot determine filesystem type: %v", err)
+	}
+	if typ == CGROUP2_SUPER_MAGIC {
```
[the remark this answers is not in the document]
bboozzoo: I think I'll push back on this a little. We don't care about hybrid mode, neither here nor in snap-confine.

sandbox/cgroup/cgroup.go
```diff
@@ -0,0 +1,145 @@
...
+
+// ProcPath returns the path to the cgroup file under /proc for the given
+// process id.
+func ProcPath(pid int) string {
+	return filepath.Join(dirs.GlobalRootDir, fmt.Sprintf("proc/%v/cgroup", pid))
+}
+
+// ControllerPathV1 returns the path to given controller assuming cgroup v1
+// hierarchy
+func ControllerPathV1(controller string) string {
```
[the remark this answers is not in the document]
bboozzoo: The doc comment may not be clear enough. I'm not looking for a way to describe arbitrary path under a hierarchy. It's more of a way of getting the default/expected mount point of a hierarchy with given controller, given the assumptions we're already making (default /sys/fs/cgroup, hierarchy per controller under /sys/fs/cgroup/<controller>). It's used like this:

```diff
diff --git a/cmd/snap-update-ns/freezer.go b/cmd/snap-update-ns/freezer.go
index 9cce4557c..802d0c2be 100644
--- a/cmd/snap-update-ns/freezer.go
+++ b/cmd/snap-update-ns/freezer.go
@@ -26,9 +26,11 @@ import (
        "os"
        "path/filepath"
        "time"
+
+       "github.com/snapcore/snapd/sandbox/cgroup"
 )
 
-var freezerCgroupDir = "/sys/fs/cgroup/freezer"
+var freezerCgroupDir = cgroup.ControllerPathV1("freezer")
 
 var (
        freezeSnapProcesses = freezeSnapProcessesImpl
```

sandbox/cgroup/cgroup.go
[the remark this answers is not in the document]
zyga: I think it's cheap enough to do that we should just do it but I'll double check something and reconsider.

sandbox/cgroup/cgroup_test.go
```diff
@@ -0,0 +1,124 @@
...
+		return 0, errors.New("statfs fail")
+	})
+	defer restore()
+	v, err := cgroup.ProbeCgroupVersion()
+	c.Assert(err, ErrorMatches, "cannot determine filesystem type: statfs fail")
+	c.Assert(v, Equals, cgroup.Unknown)
+}
+
+func (s *cgroupSuite) TestVersion(c *C) {
+	restore := cgroup.MockVersion(cgroup.V2, nil)
```
mvo5: This test is really not testing much - not that I have a better suggestion but because of mocking we really just test if the mocks work here.

sandbox/cgroup/cgroup.go
```diff
@@ -0,0 +1,110 @@
...
+	"syscall"
+
+	"github.com/snapcore/snapd/dirs"
+)
+
+const (
+	// from golang.org/x/sys/unix
+	cgroup2SuperMagic = 0x63677270
+
+	// the only cgroup path we expect, for v2 this is where the unified
```
zyga:
```suggestion
	// The only cgroup path we expect, for v2 this is where the unified
```

PR #7453, opened 2019-09-11
tests/classic-custom-device-reg: disable on opensuse-15 and fedora-30

anonymouse64:
These tests have frequently failed to prepare with the following output:
```
+ snap download --edge core
Fetching snap "core"

<kill-timeout reached>
```

See full output: https://pastebin.ubuntu.com/p/cbPkYj3TS2/

### DIFF
--- a/tests/main/classic-custom-device-reg/task.yaml
+++ b/tests/main/classic-custom-device-reg/task.yaml
@@ -1,7 +1,8 @@
 summary: |
-   Test gadget customized device initialisation and registration also on classic
+    Test gadget customized device initialisation and registration also on classic
 
-systems: [-ubuntu-core-*]
+# the test is unable to download the core snap on opensuse and fedora 30
+systems: [-ubuntu-core-*, -opensuse-15.1-64, -fedora-30-64]
 
 environment:
     SEED_DIR: /var/lib/snapd/seed


tests/main/classic-custom-device-reg/task.yaml
```diff
@@ -1,7 +1,8 @@
 summary: |
-   Test gadget customized device initialisation and registration also on classic
+    Test gadget customized device initialisation and registration also on classic
 
-systems: [-ubuntu-core-*]
+# the test is unable to download the core snap on opensuse and fedora 30
```
zyga: If this indicates actual failure to operate we're actually shooting ourselves in the foot. I'm +1 to disable tests in emergency situations so that we are not paralysed in addressing the root of the problem. I'm -1 on disabling tests without further pursuit. I'm not suggesting you meant the latter, just making a general remark.

I would like to understand what is going on now.  Is it just a store hiccup or is there a deeper pattern here?

anonymouse64: I don't think there's a store hiccup here, at least there doesn't appear to be for me or for other tests, this just seems to happen for the same two systems with the same conditions. The test logs don't have enough information in them, so I'm still investigating

PR #7458, opened 2019-09-12
interfaces/kubernetes-support: allow systemd-run to ptrace read unconfined

jdstrand:
Certain invocations of systemd-run require this access for interacting
with the systemd daemon


jdstrand:
Thanks! :)

PR #7460, opened 2019-09-12
interfaces/kubernetes-support: allow systemd-run to ptrace read unconfined - 2.41

jdstrand:
https://github.com/snapcore/snapd/pull/7458 for 2.41


jdstrand:
> @jdstrand I will merge it into the 2.41 branch. We have no 2.41.1 release planned right now though. Is this critical enough so that we should?

@mvo5 - no, so long as it ends up in edge, the k8s team will be ok, and that will happen as of 7458. If there happens to be a 2.41.1, then this will be nice.

PR #7463, opened 2019-09-13
tests/main/interfaces-{calendar,contacts}-service: disable on 19.10

bboozzoo:
The snap used for testing - test-snapd-eds, is built with evolution libraries
that automatically trigger dbus to spawn services with
org.gnome.evolution.dataserver.Calendar7 and
org.gnome.evolution.dataserver.AddressBook9 names (the last number probably
represents some API version).

The tests fail with the following log:
```
+ test-snapd-eds.calendar load test-calendar
error: g-dbus-error-quark[2] Unable to connect to 'test-calendar': The name org.gnome.evolution.dataserver.Calendar7 was not provided by any .service files

+ test-snapd-eds.contacts load test-address-book
error: g-dbus-error-quark[2] Unable to connect to 'test-address-book': The name org.gnome.evolution.dataserver.AddressBook9 was not provided by any .service files
```

The evolution-data-server 3.33.91-3 shipped in Ubuntu 19.10 comes with dbus
service files for names org.gnome.evolution.dataserver.AddressBook10 and
org.gnome.evolution.dataserver.Calendar8. The test snap needs an update.

@jhenstridge can you look into updating test-snapd-eds to?

### DIFF
--- a/tests/main/interfaces-calendar-service/task.yaml
+++ b/tests/main/interfaces-calendar-service/task.yaml
@@ -5,7 +5,8 @@ summary: Ensure that the calendar-service interface works
 #
 # FIXME: disable opensuse-tumbleweed until
 # https://github.com/snapcore/snapd/pull/7230 is landed
-systems: [-ubuntu-core-*, -ubuntu-14.04-*, -amazon-*, -centos-*, -opensuse-tumbleweed-*]
+# ubuntu-19.10: test-snapd-eds is incompatible with eds version shipped with the distro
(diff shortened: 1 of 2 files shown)


jhenstridge:
> @jhenstridge can you look into updating test-snapd-eds to?

If I update the test snap to work with 19.10, it will stop working with every release before 19.10 (and likely all the non-Ubuntu systems those tests currently pass on).  We're still looking into whether there is an acceptable way to move forward with these interfaces.

PR #7466, opened 2019-09-13
[RFC] cmd/snap: special handling of snap set refresh.hold

bboozzoo:
The refresh.hold settings describes the moment in time, until which refreshes
will be held. The value is formatted according to RFC3339, eg.
2019-09-13T13:00:32+02:00 and isn't especially user friendly.

Add special casing for `snap set system refresh.hold=<d>` which accespts a
duration string that can be parsed by time.Duration(), and it automatically
replaced with computed now + duration time before sending out the request.

Because it's too hard to explain that one needs to use `date --iso-8601=seconds -d '<describe when it ends>'` to produce a usable string.

Note, the change is client side only.

### DIFF
--- a/cmd/snap/cmd_set.go
+++ b/cmd/snap/cmd_set.go
@@ -22,6 +22,7 @@ package main
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/jessevdk/go-flags"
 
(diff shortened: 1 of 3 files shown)


cmd/snap/cmd_set.go
```diff
@@ -69,7 +70,26 @@ func init() {
 	})
 }
 
+func fixupRefreshHold(snap, key, value string) string {
```
mvo5: I wouldn't use "fixup" here, maybe something along the lines of "convert" or "compute" or "calculate" or somesuch

bboozzoo:
As agreed during the sprint, we'll have a parameter for `snap refresh`, eg. `snap refresh --hold=5d`

PR #7475, opened 2019-09-17
sandbox/cgroup, usersession/userd: move cgroup related helper to a dedicated package

bboozzoo:
A helper for finding out the cgroup of a given process under a particular cgroup     controller. The code is hijacked from usersession/userd and made more generic.

Plus a drive by fix for usersession/userd unit tests.

Split from #7451


sandbox/cgroup/cgroup.go
```diff
@@ -100,6 +103,43 @@ func Version() (int, error) {
...
+		//   ...
+		// See cgroup(7) for details about the /proc/[pid]/cgroup
+		// format.
+		l := strings.Split(scanner.Text(), ":")
+		if len(l) < 3 {
+			continue
+		}
+		controllerList := l[1]
+		cgroupPath := l[2]
+		if !strings.Contains(controllerList, controller) {
```
zyga: Can you please add a test for the `cpu` controller and the `cpuset` controller being present.

Hint: there is a bug.

sandbox/cgroup/cgroup.go
```diff
@@ -125,9 +126,9 @@ func ProcGroup(pid int, controller string) (string, error) {
 		if len(l) < 3 {
 			continue
 		}
-		controllerList := l[1]
+		controllerList := strings.Split(l[1], ",")
```
zyga: ❤️

PR #7477, opened 2019-09-17
packaging: use snapfuse_ll to speed up snapfuse performance

mvo5:
While looking into why fuse is slow inside lxd we discovered
there are two different binaries for snapfuse. One which uses
a more platform agonstic but slower approach and one using
a linux specific approach which is 30x times faster.

This PR switches us to use the faster approach.


Conan-Kudo:
@mvo5 We're not using snapfuse on RH/Fedora, since `squashfuse` is packaged there. The `squashfuse` package has `/usr/bin/squashfuse_ll`... What needs to be done to use that?

mvo5:
@Conan-Kudo I would suggest that RH/Fedora switch to squashfuse_ll by default (i.e. /usr/bin/squashfuse should be squashfuse either via symlink or rename). The speedup is dramatic (~20-50x faster) and AFAICT the only downside of using _ll is that its less portable, i.e. that it won't work on macos which we don't really care about :)

Conan-Kudo:
@kyrofa See the message above. Can you make this change in the squashfuse package?

PR #7489, opened 2019-09-19
store, ..., client: add a "website" field

chipaca:
The snap store exposes a "website" field, and we don't expose it. GUI
desktop stores need it to be at parity with the web store.
See https://bugs.launchpad.net/snapd/+bug/1838787 for details.


robert-ancell:
Support added to snapd-glib - https://github.com/snapcore/snapd-glib/commit/1496089b8ebcb0d5ceebed20720bb2c657035198

PR #7495, opened 2019-09-23
[RFC] tests: add gh action for static checks

mvo5:
This adds github actions for the static check.

Just to give you guys a taste. Not sure its switching at this point yet, some things are more complicated than in travis. But it seems the actions run more often than travis and we have 6h runtime limits (instead of 45min).


mvo5:
Closing again, people had time to review this now and we have no obvious benefit from going to actions at this point (and it is going to be a bunch of work).

PR #7499, opened 2019-09-24
many: move AppArmor probing code under sandbox/apparmor

bboozzoo:
Part of the changes to move the sandbox related pieces under a common github.com/snapcore/snapd/sandbox/<backend> package. This PR moves the AppArmor related bits and updates the relevant call sites.

### DIFF
--- a/interfaces/backends/backends_test.go
+++ b/interfaces/backends/backends_test.go
@@ -23,7 +23,7 @@ import (
 	. "gopkg.in/check.v1"
 
 	"github.com/snapcore/snapd/interfaces/backends"
-	"github.com/snapcore/snapd/release"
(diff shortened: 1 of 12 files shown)


interfaces/backends/backends_test.go
```diff
@@ -23,7 +23,7 @@ import (
 	. "gopkg.in/check.v1"
 
 	"github.com/snapcore/snapd/interfaces/backends"
-	"github.com/snapcore/snapd/release"
+	"github.com/snapcore/snapd/sandbox/apparmor"
```
mvo5: (nitpick) Given that pretty much everything else imports this as apparmor_sandbox I wonder if we should do it here too?

PR #7503, opened 2019-09-24
tests: restart the journald service while preparing the test

sergiocazzolato:
The idea of the test is to restart the systemd-journald.service while
the test is being prepared to avoid issues which could be cause by any
test executed before as happens runnign on ubuntu 14.04 with the tests:
spread2 -order -workers 1 -debug
google:ubuntu-14.04-64:tests/main/refresh-all-undo
google:ubuntu-14.04-64:tests/main/validate-container-failures

See full log:
https://paste.ubuntu.com/p/jyBtYMbqS3/


bboozzoo:
Opened #7505 to fix the failure in `google:ubuntu-18.04-64:tests/main/mount-ns:inherit`.

PR #7511, opened 2019-09-25
tests: when the backend is external skip the loop waiting for snap version

sergiocazzolato:
In case the backend is external it is not needed to wait until an
specific version due to all the snaps are preinstalled in teh image

Debuging on external systems I saw that we are 2 minutes iterating until
the loop finishes when this loop is not needed at all.

### DIFF
--- a/tests/lib/prepare.sh
+++ b/tests/lib/prepare.sh
@@ -608,7 +608,7 @@ prepare_ubuntu_core() {
     fi
 
     # Wait for the snap command to become available.
-    if not [ "$SPREAD_BACKEND" = "external" ]; then
+    if [ "$SPREAD_BACKEND" != "external" ]; then
         for i in $(seq 120); do
             if [ "$(command -v snap)" = "/usr/bin/snap" ] && snap version | grep -q 'snapd +1337.*'; then
                 break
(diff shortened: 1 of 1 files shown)


sergiocazzolato:
@anonymouse64, right external backend is used when the setup is done in a different process and all the snaps which are previously installed are used. So in this case, as we are not building snapd, we don't need to wait for the snap command and also is not needed to check any version because the version could be anything.

tests/lib/prepare.sh
```diff
@@ -608,12 +608,14 @@ prepare_ubuntu_core() {
     fi
 
     # Wait for the snap command to become available.
-    for i in $(seq 120); do
-        if [ "$(command -v snap)" = "/usr/bin/snap" ] && snap version | grep -q 'snapd +1337.*'; then
-            break
-        fi
-        sleep 1
-    done
+    if not [ "$SPREAD_BACKEND" = "external" ]; then
```
mvo5: Using `if [ "$SPREAD_BACKEND" != "external" ]; then` would also work here and is (to me) slightly more canonical but its a nitpick :)

bboozzoo: Yeah, `[ "$SPREAD_BACKEND" != "external" ]` is the right way to do it.

bboozzoo: Pushed a little patch to address this.

PR #7512, opened 2019-09-25
interfaces/docker-support,kubernetes-support: misc updates for strict k8s

jdstrand:
* interfaces/kubernetes-support: add missing /bin/mount ixr
* interfaces/kubernetes-support: use /{,usr/}bin/mount for portability
* interfaces/docker-support: various runc/containerd accesses for k8s

### DIFF
--- a/interfaces/builtin/docker_support.go
+++ b/interfaces/builtin/docker_support.go
@@ -175,8 +175,22 @@ ptrace (read, trace) peer=cri-containerd.apparmor.d,
 # needed by runc for mitigation of CVE-2019-5736
 # For details see https://bugs.launchpad.net/apparmor/+bug/1820344
 / ix,
+/bin/busybox ixr,
 /bin/runc ixr,
 /pause ixr,
+
+# When kubernetes drives containerd, containerd needs access to CNI services,
+# like flanneld's subnet.env for DNS. This would ideally be snap-specific (it
+# could if the control plane was a snap), but in deployments where the control
+# plane is not a snap, it will tell flannel to use this path.
+/run/flannel/{,**} rk,
+
+# When kubernetes drives containerd, containerd needs access to various
+# secrets for the pods which are overlayed at /run/secrets/....
+# This would ideally be snap-specific (it could if the control plane was a
+# snap), but in deployments where the control plane is not a snap, it will tell
+# containerd to use this path for various account information for pods.
+/run/secrets/kubernetes.io/{,**} rk,
 `
 
 const dockerSupportConnectedPlugSecComp = `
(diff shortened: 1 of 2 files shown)


interfaces/builtin/docker_support.go
```diff
@@ -175,8 +175,22 @@ ptrace (read, trace) peer=cri-containerd.apparmor.d,
 # needed by runc for mitigation of CVE-2019-5736
 # For details see https://bugs.launchpad.net/apparmor/+bug/1820344
 / ix,
+/bin/busybox ixr,
```
anonymouse64: is this also for mitigation of CVE-2019-5736 ?

if not, maybe add a space between the above `/ ix` and this to delineate the different accesses?

interfaces/builtin/docker_support.go
```diff
@@ -175,8 +175,22 @@ ptrace (read, trace) peer=cri-containerd.apparmor.d,
...
 /bin/runc ixr,
 /pause ixr,
+
+# When kubernetes drives containerd, containerd needs access to CNI services,
+# like flanneld's subnet.env for DNS. This would ideally be snap-specific (it
+# could if the control plane was a snap), but in deployments where the control
+# plane is not a snap, it will tell flannel to use this path.
+/run/flannel/{,**} rk,
```
anonymouse64: I understand why we needed this in kubernetes-support for microk8s, but why do we also need it in docker-support?

jdstrand:
> It's not clear to me why we need to duplicate some of the same accesses in k8s-support in docker-support.

The accesses aren't the same. kubelet writes and containerd reads.

interfaces/builtin/docker_support.go
jdstrand: Based on the denial (`[runc:INIT]`) it looked to be in the same vein as pause, so I put it there. I can move pause and busybox a line below, sure.

interfaces/builtin/docker_support.go
jdstrand: The control plane tells containerd where to find information when flanneld is in use and containerd needs to be able to read it. In this case with flannel, it is to help with DNS.

interfaces/builtin/docker_support.go
jdstrand: Done.

interfaces/builtin/docker_support.go
anonymouse64: ah I see, that makes sense I didn't read closely enough

jdstrand:
Thanks for the reviews and merge!

PR #7519, opened 2019-09-26
tests: use `snap debug boot-vars` in "ubuntu-core-upgrade" test

mvo5:
This PR uses the new `snap debug boot-vars` comand to get the
boot environment. This enables us to get the boot environment
on core18 ARM systems where fw_printenv is missing, hence the
main/ubuntu-core-upgrade test can be re-enabled.

This reverts PR#7508.

### DIFF
--- a/tests/lib/boot.sh
+++ b/tests/lib/boot.sh
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+bootenv() {
+    if [ $# -eq 0 ]; then
+        snap debug boot-vars
+    else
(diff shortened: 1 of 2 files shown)


sergiocazzolato:
@mvo5 should we remove the check on prepare.sh (echo "Ensure that the bootloader environment output does not contain any of the snap_* variables on classic") right
We could check all the variables are empty, something like "^snapd_*=$" ?

mvo5:
Closing for now until this was further discussed with Samuele

PR #7521, opened 2019-09-26
[RFC] many: use --root as first arg to systemctl everywhere

anonymouse64:
AFAICT there's no reason to not use --root everywhere, and we do need it some places, so for consistencies sake, just use --root everywhere we call systemctl.

Also update many, many, many unit tests for this behavior.

I opened this as a RFC in case there is some good reason why some systemctl commands shouldn't  use --root (user session services maybe?) 

I couldn't request a review through GitHub from you, but @jhenstridge could you take a look and comment if there's something amiss here? Thanks


bboozzoo:
Looks like the shim on 14.04 doesn't like `--root` too much.
```
+ snap install --edge core
- Mount snap "core" (7882) ([--root / start snap-core-7882.mount] failed with exit status 1: Failed to get D-Bus connection: No connection to service manager.
)
```

Also, after reading systemctl manpage I'm not sure if `--root` is allowed, or even makes sense, with commands that aren't enable/disable/is-enabled:
```
--root=
    When used with enable/disable/is-enabled (and related commands), use the specified root path when looking for unit files. If this option is present, systemctl
    will operate on the file system directly, instead of communicating with the systemd daemon to carry out changes.
```
If I'm reading this right, it only affects commands (functionality) that can otherwise be implemented by systemctl. In this case `stop`/`start` is something that must always go through systemd.

chipaca:
Yeah, the reason we don't is 14.04 (and possibly also centos)

stolowski:
> Yeah, the reason we don't is 14.04 (and possibly also centos)

Ah...

anonymouse64:
I guess I read 
```
enable/disable/is-enabled (and related commands),
```

as start/stop being related commands but there I go trying to understand systemd docs again ...

I'll close this then if the reason we don't is 14.04

bboozzoo:
> but there I go trying to understand systemd docs again ...

At least they don't define what MAY, SHOULD, MUST, WILL means :smile:

PR #7522, opened 2019-09-26
httputil: set user agent for CONNECT

chipaca:
Without this change, https proxies would get a generic
`Go-http-client/1.1` user agent from snapd, making filtering by user
agent in the proxy harder than it needs to be.

From Go 1.8 we can ask go's http client to set that header, so this
change does that.

Fixes: lp:1779190


chipaca:
does-not-fix: insomnia

PR #7524, opened 2019-09-27
tests: add unit test for gadget defaults with a multiline string

stolowski:
Add a unit test for multiline string in gadget defaults. I've created this to verify and understand https://bugs.launchpad.net/snapd/+bug/1820060 (non-issue afaict).

### DIFF
--- a/gadget/gadget_test.go
+++ b/gadget/gadget_test.go
@@ -121,6 +121,17 @@ defaults:
       bar: baz
 `)
 
+var mockClassicGadgetMultilineDefaultsYaml = []byte(`
+defaults:
+  system:
+    something: true
+  otheridididididididididididididi:
+    foosnap:
+      multiline: |
+        foo
+        bar
+`)
+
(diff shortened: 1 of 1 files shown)


gadget/gadget_test.go
```diff
@@ -121,6 +121,17 @@ defaults:
       bar: baz
 `)
 
+var mockClassicGadgetMultilineDefaultsYaml = []byte(`
+defaults:
+  system:
+    something: true
+  otheridididididididididididididi:
+    foosnap:
+      multiline: |
```
bboozzoo: For the sake of pushing the YAML parser, would this be accepted?
```
var mock = []byte(`
defaults:
  other:
    multiline: |
      # this is a multiine string with valid YAML inside
      foo: 2
      bar: "123"
      baz: |
        this is a test
`)
```

stolowski: Good question, hard to say without trying ;). Will give it a shot.

PR #7526, opened 2019-09-27
cmd/snap-confine: allow digits in hook names

zyga:
Interface hooks react to interface connection events. Interface names
can contain digits, for example "i2c" is a valid plug or slot name.

It was reported that interface hooks for the i2c interface were not
working. This was traced to an overly-strict validation of the snap
security tag computed for the hook by snapd, performed by snap-confine.

Fixes: https://bugs.launchpad.net/snapd/+bug/1638988

### DIFF
--- a/cmd/libsnap-confine-private/snap-test.c
+++ b/cmd/libsnap-confine-private/snap-test.c
@@ -90,6 +90,10 @@ static void test_verify_security_tag(void)
 	g_assert_true(verify_security_tag("snap.123test.123test", "123test"));
 	g_assert_true(verify_security_tag
 		      ("snap.123test.hook.configure", "123test"));
+
+	// regression test snap.eon-edg-shb-pulseaudio.hook.connect-plug-i2c
+	g_assert_true(verify_security_tag
+		      ("snap.foo.hook.connect-plug-i2c", "foo"));
 }
 
 static void test_sc_is_hook_security_tag(void)
--- a/cmd/libsnap-confine-private/snap.c
+++ b/cmd/libsnap-confine-private/snap.c
@@ -31,7 +31,7 @@
 bool verify_security_tag(const char *security_tag, const char *snap_name)
 {
 	const char *whitelist_re =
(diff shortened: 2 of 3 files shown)


kubiko:
LGTM

cmd/libsnap-confine-private/snap-test.c
```diff
@@ -90,6 +90,10 @@ static void test_verify_security_tag(void)
 	g_assert_true(verify_security_tag("snap.123test.123test", "123test"));
 	g_assert_true(verify_security_tag
 		      ("snap.123test.hook.configure", "123test"));
+
+	// regression test snap.eon-edg-shb-pulseaudio.hook.connect-plug-i2c
+	g_assert_true(verify_security_tag
```
bboozzoo: While at it, can you also add a test case with `connect-plug-i2c` in ValidateSuite.TestValidateHookName ?

cmd/libsnap-confine-private/snap.c
```diff
@@ -31,7 +31,7 @@
 bool verify_security_tag(const char *security_tag, const char *snap_name)
 {
 	const char *whitelist_re =
-	    "^snap\\.([a-z0-9](-?[a-z0-9])*(_[a-z0-9]{1,10})?)\\.([a-zA-Z0-9](-?[a-zA-Z0-9])*|hook\\.[a-z](-?[a-z])*)$";
+	    "^snap\\.([a-z0-9](-?[a-z0-9])*(_[a-z0-9]{1,10})?)\\.([a-zA-Z0-9](-?[a-zA-Z0-9])*|hook\\.[a-z](-?[a-z0-9])*)$";
```
stolowski: Looks good and matches the characters we allow in `hooktypes.go`; nb, we could probably tweak the pattern to be even more consistent and say "....(-?[a-z0-9])+" instead of "....(-?[a-z0-9])*" (note "+" vs "*") cause we will never get a one-letter hook name here, but that doesn't improve/affect anything, so fine to leave as is.

zyga: I'm happy to do it in another pass so that this is not a part of the stable release.

anonymouse64:
@zyga the bug mentioned in your PR description doesn't seem to be related to the fix here? did you mean that this fixes: https://bugs.launchpad.net/snapd/+bug/1845636 instead?

zyga:
Indeed, I must have pasted the wrong URL somehow. Oh well :/

PR #7527, opened 2019-09-27
interfaces/input: support confined snaps accessing keyboard and mouse directly

AlanGriffiths:
With X11 it is normal to use udev for input, and when snapped apps derive this access from the `x11` interface.

In theory, when migrated from X11 to Wayland, they should switch to using Wayland for input.
In practice, some apps & toolkits continue to use udev.

This interface allows these to be snapped without abusing the much wider `x11` interface simply to get udev access for keyboard and mouse input.

### DIFF
--- a/interfaces/builtin/input.go
+++ b/interfaces/builtin/input.go
@@ -48,7 +48,7 @@ const inputConnectedSlotAppArmor = `
 
 const inputConnectedPlugAppArmor = `
 # Description: Allow input from keyboard/mouse devices.
-# raw' rule is not finely mediated by apparmor so we mediate with seccomp arg
+# raw rule is not finely mediated by apparmor so we mediate with seccomp arg
 # filtering.
 network netlink raw,
 /run/udev/data/c13:[0-9]* r,
--- a/interfaces/builtin/input.go
+++ b/interfaces/builtin/input.go
@@ -1,0 +1,90 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package builtin
+
+import (
+	"github.com/snapcore/snapd/interfaces"
+	"github.com/snapcore/snapd/interfaces/udev"
+)
+
+const inputSummary = `allows input from keyboard/mouse devices`
+
+const inputBaseDeclarationSlots = `
+  input:
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
+`
+
+const inputPermanentSlotAppArmor = `
+# Description: Allow input from keyboard/mouse devices.
+`
+
+const inputPermanentSlotSecComp = `
+# Description: Allow input from keyboard/mouse devices.
+`
+
+const inputConnectedSlotAppArmor = `
+# Description: Allow input from keyboard/mouse devices.
+`
+
+const inputConnectedPlugAppArmor = `
+# Description: Allow input from keyboard/mouse devices.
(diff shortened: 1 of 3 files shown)


interfaces/builtin/input.go
```diff
@@ -0,0 +1,90 @@
...
+# Description: Allow input from keyboard/mouse devices.
+`
+
+const inputConnectedSlotAppArmor = `
+# Description: Allow input from keyboard/mouse devices.
+`
+
+const inputConnectedPlugAppArmor = `
+# Description: Allow input from keyboard/mouse devices.
+# raw' rule is not finely mediated by apparmor so we mediate with seccomp arg
```
zyga:
```suggestion
# raw rule is not finely mediated by apparmor so we mediate with seccomp arg
```

AlanGriffiths:
@zyga FYA:
```bash
$ find . -name \*.go | xargs grep -R "# raw' rule is not finely mediated by apparmor so we mediate with seccomp arg" | wc -l
19
```

AlanGriffiths:
I've discovered that the `hardware-observe` interface also provides the access these toolkits seek.

Would it be preferable to use that instead of adding this?

zyga:
@AlanGriffiths I think a more precise interface is more preferred. Specifically interfaces encode intent rather than the raw rules they implement.

interfaces/builtin/input.go
```diff
@@ -0,0 +1,90 @@
...
+  input:
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
+`
+
+const inputPermanentSlotAppArmor = `
+# Description: Allow input from keyboard/mouse devices.
+`
```
jdstrand: This isn't used anywhere and can be dropped.

interfaces/builtin/input.go
```diff
@@ -0,0 +1,90 @@
...
+
+const inputPermanentSlotSecComp = `
+# Description: Allow input from keyboard/mouse devices.
+`
```
jdstrand: This isn't used anywhere and can be dropped.

interfaces/builtin/input.go
```diff
@@ -0,0 +1,90 @@
...
+
+const inputConnectedSlotAppArmor = `
+# Description: Allow input from keyboard/mouse devices.
+`
```
jdstrand: This isn't used anywhere and can be dropped.

AlanGriffiths:
> If these were moved to the existing wayland interface (assuming inline questions indicate that is appropriate), then I would Approve this interface.

I don't want to add this to the Wayland interface as toolkits *should not* be using udev for input events they can and should receive via Wayland. Using udev directly has the potential to let them watch input that would not be routed to them by the compositor.

interfaces/builtin/input.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (iface *inputInterface) UDevConnectedPlug(spec *udev.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
+	spec.TriggerSubsystem("input")
+	spec.TagDevice(`KERNEL=="tty[0-9]*"`)
+	spec.TagDevice(`KERNEL=="mice"`)
+	spec.TagDevice(`KERNEL=="mouse[0-9]*"`)
+	spec.TagDevice(`KERNEL=="event[0-9]*"`)
+	spec.TagDevice(`KERNEL=="ts[0-9]*"`)
+	return nil
+}
```
jdstrand: This is also unused.

jdstrand:
> > If these were moved to the existing wayland interface (assuming inline questions indicate that is appropriate), then I would Approve this interface.
> 
> I don't want to add this to the Wayland interface as toolkits _should not_ be using udev for input events they can and should receive via Wayland. Using udev directly has the potential to let them watch input that would not be routed to them by the compositor.

Well, all of this is already allowed in the manually connected hardware-observe interface and, like you said, the auto-connected x11 interface. If this is something they shouldn't really be doing, but are, perhaps they should plugs one of these instead of accommodating a scenario that shouldn't really be blessed. When they balk at having to use x11 with a pure wayland app and come to the forum to ask for auto-connection for hardware-observe for just needing udev input info, we can guide them towards not using udev.

Another way to think about it (looking long term) is that x11 is a transitional interface, and for practical reasons people plugs both wayland and x11 until the day that x11 is no longer needed. Similarly we can look at the toolkits themselves as in transition while they still use udev on a pure wayland system, so it makes sense for them to plugs a transitional interface until they are fixed.

AlanGriffiths:
> When they balk at having to use x11 with a pure wayland app and come to the forum to ask for auto-connection for hardware-observe for just needing udev input info, we can guide them towards not using udev.

Sadly for this scenario, I suspect the "they" doing the snap packaging and the "they" maintaining the toolkit will be different folks.

But I'll go along with using `hardware-observe` (as that's available for Core) in the hope of better days to come.

PR #7532, opened 2019-09-30
interfaces/openvswitch: allow access to other openvswitch sockets

dosaboy:
Commands such as ovs-ofctl require access to
/var/run/openvswitch/<bridge name>.mgmt.

### DIFF
--- a/interfaces/builtin/openvswitch.go
+++ b/interfaces/builtin/openvswitch.go
@@ -29,8 +29,13 @@ const openvswitchBaseDeclarationSlots = `
     deny-auto-connection: true
 `
 
+// List of sockets we want to allow access to. This list currently includes
+// sockets needed by ovs-vsctl and ovs-ofctl commands. The latter requires
+// access to per-bridge sockets e.g. for bridge br-data you would need access
+// to /run/openvswitch/br-data.mgmt.
 const openvswitchConnectedPlugAppArmor = `
 /run/openvswitch/db.sock rw,
+/run/openvswitch/*.mgmt rw,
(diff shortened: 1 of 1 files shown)


interfaces/builtin/openvswitch.go
```diff
@@ -31,6 +31,7 @@ const openvswitchBaseDeclarationSlots = `
 
 const openvswitchConnectedPlugAppArmor = `
 /run/openvswitch/db.sock rw,
+/var/run/openvswitch/*.mgmt,
```
javacruft: Drop the /var prefix and you'll need to have rw permissions (as in line 33)

interfaces/builtin/openvswitch.go
```diff
@@ -31,6 +31,7 @@ const openvswitchBaseDeclarationSlots = `
 
 const openvswitchConnectedPlugAppArmor = `
 /run/openvswitch/db.sock rw,
+/run/openvswitch/*.mgmt rw,
```
jdstrand: A comment about what this mgmt files would be nice (and enumerating what some of them are). Please feel free to commit once an appropriate comment is present.

dosaboy: done.

jdstrand: Thanks for the added comment. Future note: once comments have been applied to the PR, we don't force push any more.

mvo5:
This is cherry-picked to 2.42 now. If we do a 2.42.2 it will be part of this release.

PR #7537, opened 2019-09-30
tests: fix ubuntu-core-device-reg test for arm devices on core18

sergiocazzolato:
The test is currently failing on arm devices for core18.

### DIFF
--- a/tests/main/ubuntu-core-device-reg/task.yaml
+++ b/tests/main/ubuntu-core-device-reg/task.yaml
@@ -31,6 +31,9 @@ execute: |
         ubuntu-core-18-64)
             snap known serial | MATCH "model: ubuntu-core-18-amd64"
             ;;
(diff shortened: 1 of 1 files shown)


tests/main/ubuntu-core-device-reg/task.yaml
```diff
@@ -31,6 +31,9 @@ execute: |
         ubuntu-core-18-64)
             snap known serial | MATCH "model: ubuntu-core-18-amd64"
             ;;
+        ubuntu-core-18-arm-*)
+            snap known serial | MATCH "model: ubuntu-core-18-$gadget_name"
```
anonymouse64: not for this PR, but maybe we could use the newly minted `snap model` command for this kind of thing instead, I happen to know it's of _excellent_ quality :grinning:

PR #7540, opened 2019-10-01
interfaces/seccomp: query apparmor sandbox helper rather than aggregate info

bboozzoo:
Some snippets are added conditionally depending on AppAmor being fully supported
by the host system. Instead of querying the aggregate forced-dev-mode status,
poke the AppArmor sandbox helpers in sandbox/apparmor directly.

Make the unit test more specific.

### DIFF
--- a/interfaces/seccomp/backend.go
+++ b/interfaces/seccomp/backend.go
@@ -46,6 +46,7 @@ import (
 	"github.com/snapcore/snapd/interfaces"
 	"github.com/snapcore/snapd/osutil"
 	"github.com/snapcore/snapd/release"
+	"github.com/snapcore/snapd/sandbox/apparmor"
 	"github.com/snapcore/snapd/sandbox/seccomp"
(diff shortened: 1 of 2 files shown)


interfaces/seccomp/backend.go
```diff
@@ -313,10 +314,10 @@ func generateContent(opts interfaces.ConfinementOptions, snippetForTag string, a
 	buffer.WriteString(snippetForTag)
 	buffer.WriteString(uidGidChownSyscalls)
 
-	// For systems with force-devmode we need to apply a workaround
-	// to avoid failing hooks. See description in template.go for
-	// more details.
-	if release.ReleaseInfo.ForceDevMode() {
+	// For systems with partial or missing AppArmor support we need to apply
```
zyga: This is interesting in light of cgroup v2 work, where we are no longer in full confinement mode when v2 is used. I think this is a step in the correct direction.

PR #7543, opened 2019-10-01
release: make forced dev mode look at cgroupv2 support

bboozzoo:
Not all security backends are fully supported on a system using unified cgroup
hierarchy. Make sure this is reflected in the release package.


release/release_test.go
```diff
@@ -149,8 +151,34 @@ func (s *ReleaseTestSuite) TestReleaseInfo(c *C) {
...
+		{apparmor.Full, cgroup.V2, true},
+		{apparmor.Partial, cgroup.V2, true},
+	} {
+		runTest(tc.apparmorLevel, tc.cgroupVersion, tc.exp)
+	}
+}
+
+func (s *ReleaseTestSuite) TestMockForceDevMode(c *C) {
 	for _, devmode := range []bool{true, false} {
-		release.MockForcedDevmode(devmode)
```
zyga: Ouch. Was this missing call to defer before?

bboozzoo: Yeah. At the same time we were lucky that each package is tested in a separate process.

release/release_test.go
```diff
@@ -149,8 +151,34 @@ func (s *ReleaseTestSuite) TestReleaseInfo(c *C) {
...
+		defer release.MockForcedDevmode(devmode)()
```
mvo5: Not sure what else we do in this package but I prefer the slightly more explicit "restore := MockForceDevMode(..)\ndefer restre" ?

PR #7544, opened 2019-10-01
tests: fix snapd-failover test for core18 tests on boards

sergiocazzolato:
This change fixes 2 issues:
1. The snapd.failure.service remains active after execute the test, so
when the test is executed using -repeat option it fails.
2. On boards the snap command takes a while to be active, it happens
because while the snapd service is activating. The idea is to wait a bit
until the snapd service is active again and the check for the snap
command.

### DIFF
--- a/tests/core18/snapd-failover/task.yaml
+++ b/tests/core18/snapd-failover/task.yaml
@@ -4,6 +4,12 @@ debug: |
     # dump failure data
     journalctl -u snapd.failure.service
 
+restore: |
+    # Stop snapd.failure.service in case it is active
+    if systemctl is-active snapd.failure.service | MATCH active; then
+        systemctl stop snapd.failure.service
+    fi
+
 execute: |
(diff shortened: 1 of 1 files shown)


tests/core18/snapd-failover/task.yaml
```diff
@@ -27,6 +33,12 @@ execute: |
     fi
 
     echo "And verify that snap commands still work and snapd is reverted"
+    for _ in $(seq 30); do
```
bboozzoo: this bit can use the `retry-tool`:
```
retry-tool -n 30 --wait 5 systemctl is-active snapd.failure.service
```

sergiocazzolato: sure, thanks for the suggestion

sergiocazzolato: @bboozzoo change done

PR #7545, opened 2019-10-02
cmd/model: don't show model with display-name inline w/ opts

anonymouse64:
With `snap model` we want to show the display-name with the actual model in paranthesis for easy operator (i.e. human) inspection, but with `snap model --verbose` we want the model key to be the verbatim model so it is easy to inspect the output programatically in scripts and such.

Also add a unit test for --verbose output when there's a display-name in the model assertion.

Also fix a typo in the diff output for one of the spread tests in the error condition, and simplify the spread test to always use --verbose to ensure that the model in the model assertion from `snap known` matches that from `snap model`.

### DIFF
--- a/cmd/snap/cmd_model.go
+++ b/cmd/snap/cmd_model.go
@@ -198,9 +198,11 @@ func (x *cmdModel) Execute(args []string) error {
 		serial = serialAssertion.HeaderString("serial")
 	}
 
-	// handle brand/brand-id (the former is only on `snap model` w/o opts)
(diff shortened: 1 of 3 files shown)


anonymouse64:
hmm spread is unhappy, for some reason `snap model --verbose` is exiting with status code 1 :-/

anonymouse64:
Fixed the spread test (hopefully :crossed_fingers:), forgot a colon in the grep regex

PR #7548, opened 2019-10-02
tests: update snap logs to match for multiple lines for  "running"

sergiocazzolato:
The test assumes the only lines in the log come from the unit, but they
can also be from systemd, so the change makes sure the logs for the last
line will match for the service name which is allways included also in
systemd log lines.

### DIFF
--- a/tests/main/snap-logs/task.yaml
+++ b/tests/main/snap-logs/task.yaml
@@ -16,7 +16,7 @@ execute: |
     snap logs test-snapd-service.test-snapd-other-service | MATCH "running"
 
     echo "check output lines for the logs"
-    snap logs -n=1 test-snapd-service | MATCH "running"
+    snap logs -n=1 test-snapd-service | MATCH "test-snapd-service"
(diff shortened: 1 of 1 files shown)


tests/main/snap-logs/task.yaml
```diff
@@ -16,7 +16,7 @@ execute: |
     snap logs test-snapd-service.test-snapd-other-service | MATCH "running"
 
     echo "check output lines for the logs"
-    snap logs -n=1 test-snapd-service | MATCH "running"
+    snap logs -n=1 test-snapd-service | MATCH "systemd|running"
```
mvo5: Should we simply check for more lines than one?

sergiocazzolato: @mvo5, the test is currently checking with -n=all and -n=1, the problem is that looking at the logs sometimes we get about  5~10 consecutive lines for the systemd unit. If it is ok for you I could update this to check with -n=20 and I think that should be enough to prevent any error.

PR #7552, opened 2019-10-03
tests: add new option "invert" to retry-tool

sergiocazzolato:
This is util to retry until a command returns 0, and keeps the command
passed as parameter easier to understand.

This is an example where the command will be retried until it fails:
retry-tool -n 50 --wait 1 --invert systemctl restart systemd-journald


zyga:
I'm mildly -1 on this because we can just say `retry not false` and because of what we're building towards (commands rather than shell functions) it will actually work okay.

sergiocazzolato:
@zyga you are right, it was just to make the retry-tool command independent so then I can reuse it other projects like (spread-images or for beta validation scripts).

PR #7559, opened 2019-10-07
tests: change regex to validate access to cdn during snap download

sergiocazzolato:
Store can't waranty the fastly cdn is going to be use 100% of the
downloads. This is why after a sync it was defined that the best option
is to check that the cdn is being used but not which one.

Currently we see sporadically that canonical-lgw01.cdn.snapcraft.io is
used instead of fastly making the proxy-no-core test fails.

Full log:
https://travis-ci.org/snapcore/snapd/jobs/593659116#L4396

### DIFF
--- a/tests/main/proxy-no-core/task.yaml
+++ b/tests/main/proxy-no-core/task.yaml
@@ -52,4 +52,4 @@ execute: |
 
     # shellcheck source=tests/lib/journalctl.sh
     . "$TESTSLIB/journalctl.sh"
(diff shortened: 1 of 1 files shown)


bboozzoo:
Yeah, just observed this in my PR too:
```
Oct 07 14:03:25 oct071349-480229 snapd[11422]: logger.go:67: DEBUG: > "GET /download-origin/canonical-lgw01/99T7MUlRhtI3U0QFgl5mXXESAiSwt776_7713.snap?token=1570471200_62198a362d5b77561d313bc2f5414301df313589 
```

PR #7560, opened 2019-10-07
daemon: change /v2/download API to take "snap-name" as input

mvo5:
This changes the /v2/download API as discussed during the standup
and on irc to take a single "snap-name" and no action.

No backward compatiblity code is provided. This should be ok
because the interface was never advertised and we don't have any
matching client support yet.

Providing backward compatiblity is trivial and I am happy to
add it (requires a bit more testing code though).


robert-ancell:
Adding support for this to snapd-glib in https://github.com/snapcore/snapd-glib/pull/78/files

PR #7562, opened 2019-10-07
tests: check the apparmor_parser when the file exists on snap-confine test

sergiocazzolato:
This change is because with the new version of the apparmor_parser, when
the file does not exist then the exit code is 2 and it makes the test
fail.

This error is blocking the new updates for the opensuse leap 15.1 image

New AppArmor parser
-------------------

google:opensuse-15.1-64 .../tests/main/snap-confine# apparmor_parser
-QTK /etc/apparmor.d/usr.lib.snapd.snap-confine*
File /etc/apparmor.d/usr.lib.snapd.snap-confine* not found, skipping...
google:opensuse-15.1-64 .../tests/main/snap-confine# echo $?
2
google:opensuse-15.1-64 .../tests/main/snap-confine# apparmor_parser -h
AppArmor parser version 2.12.3

Prev AppArmor parser
--------------------
google:opensuse-15.1-64 .../tests/main/snap-confine# apparmor_parser
-QTK /etc/apparmor.d/usr.lib.snapd.snap-confine*
File /etc/apparmor.d/usr.lib.snapd.snap-confine* not found, skipping...
google:opensuse-15.1-64 .../tests/main/snap-confine# echo $?
0
google:opensuse-15.1-64 .../tests/main/snap-confine# apparmor_parser -h
AppArmor parser version 2.12.2


tests/main/snap-confine/task.yaml
```diff
@@ -43,6 +43,10 @@ execute: |
 
     echo "Ensure apparmor profile for snap-confine is parsable"
     for f in /etc/apparmor.d/usr.lib.snapd.snap-confine*; do
+        # Breake the loop when the file does not exist
+        if [ ! -f "$f" ]; then
```
bboozzoo: I don't think this is supposed to happen on openSUSE. @zyga we don't ship s-c apparmor profiles on opensuse?

bboozzoo:
The test is passing on the latest master, without the changes proposed in the PR.

tests/main/snap-confine/task.yaml
zyga: Yes, that's correct. Unless something broke in packaging and suddenly we do.

Perhaps the loop should change to a find-style loop. Unfortunately shell shows the sharp edges all too often.

sergiocazzolato:
@bboozzoo It works in the latest master but it doesn't with the image test-os-1 which contains the latest changes for opensuse 15.1

anonymouse64:
With this, on ubuntu, the test is picking up `etc/apparmor.d/local/usr.lib.snapd.snap-confine` (note the `local`) and failing:

```
++ find /etc/apparmor.d -name 'usr.lib.snapd.snap-confine*'
+ for f in '$(find /etc/apparmor.d -name '\''usr.lib.snapd.snap-confine*'\'')'
+ command -v apparmor_parser
/sbin/apparmor_parser
+ apparmor_parser -QTK /etc/apparmor.d/local/usr.lib.snapd.snap-confine
+ command -v aa-enforce
/usr/sbin/aa-enforce
+ aa-enforce /etc/apparmor.d/local/usr.lib.snapd.snap-confine
Setting /etc/apparmor.d/local/usr.lib.snapd.snap-confine to enforce mode.

ERROR: /etc/apparmor.d/local/usr.lib.snapd.snap-confine contains no profile
```

This seems unintended to me, since that file is empty on my machine. Should we be filtering out the `local` dir here?

sergiocazzolato:
@anonymouse64 thanks for the review, It is ready for another round!

anonymouse64:
looks like an issue with shell scripting: 

```
ERROR:root:shellcheck failed for file tests/main/snap-confine/task.yaml in sections: execute; error log follows
ERROR:root:tests/main/snap-confine/task.yaml: section 'execute':

In - line 28:
for f in $(find /etc/apparmor.d -maxdepth 1 -name 'usr.lib.snapd.snap-confine*'); do
         ^-- SC2044: For loops over find output are fragile. Use find -exec or a while read loop.

For more information:
  https://www.shellcheck.net/wiki/SC2044 -- For loops over find output are fr...

ERROR:root:validation failed for the following non-whitelisted files:
 - tests/main/snap-confine/task.yaml
```

Also this seems to have hit the snapd+lxd apt dependency problem mentioned this morning.

sergiocazzolato:
@anonymouse64, thanks for reviewing! Which comment?

anonymouse64:
@sergiocazzolato the one right above my review :smile: (sorry I meant to make that the review comment but I got confused...)

https://github.com/snapcore/snapd/pull/7562#issuecomment-542273722

PR #7563, opened 2019-10-08
interfaces: bump system-key version (and keep on bumping)

chipaca:
System key's type declaration had a comment rightfully stating

    // IMPORTANT: when adding new inputs bump this version

and this sage advice was then ignored by everything else.

This change:

* bumps the version to approximately the right number, as if it had
  been there from the start and bumped every time (this done by
  inspection of git log, so I might have missed something, or counted
  something twice).

* adds a static check that should catch us if we forget to bump it in
  the future.

* fixes an existing test that would fail every time we bumped the
  version :-)

HTH, HAND.

### DIFF
--- a/interfaces/system_key.go
+++ b/interfaces/system_key.go
@@ -56,7 +56,7 @@ var (
 // Note that this key gets generated on *each* `snap run` - so it
 // *must* be cheap to calculate it (no hashes of big binaries etc).
 type systemKey struct {
-	// IMPORTANT: when adding new inputs bump this version
+	// IMPORTANT: when adding/removing/changing inputs bump this version (see below)
 	Version int `json:"version"`
 
 	// This is the build-id of the snapd that generated the profiles.
@@ -74,6 +74,9 @@ type systemKey struct {
 	SecCompActions         []string `json:"seccomp-features"`
 	SeccompCompilerVersion string   `json:"seccomp-compiler-version"`
 }
+
+// IMPORTANT: when adding/removing/changing inputs bump this
+const systemKeyVersion = 9
 
 var (
 	isHomeUsingNFS  = osutil.IsHomeUsingNFS
@@ -93,7 +96,7 @@ func generateSystemKey() (*systemKey, error) {
 	}
 
 	sk := &systemKey{
-		Version: 1,
+		Version: systemKeyVersion,
 	}
 	snapdPath, err := cmd.InternalToolPath("snapd")
 	if err != nil {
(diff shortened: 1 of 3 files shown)


interfaces/system_key.go
```diff
@@ -75,6 +75,9 @@ type systemKey struct {
 	SeccompCompilerVersion string   `json:"seccomp-compiler-version"`
 }
 
+// IMPORTANT: when adding/removing/changing inputs bump this
+const systemKeyVersion = 9
```
bboozzoo:
```
11:51 <mborzecki> Chipaca: why 9 though? :)
11:52 <Chipaca> mborzecki: git log -p interfaces/system_key.go
11:52 <Chipaca> mborzecki: and count fields added/removed
```

interfaces/system_key_test.go
```diff
@@ -226,3 +238,23 @@ func (s *systemKeySuite) TestInterfaceSystemKeyMismatchVersions(c *C) {
...
 	c.Assert(err, Equals, interfaces.ErrSystemKeyVersion)
 }
+
+func (s *systemKeySuite) TestStaticVersion(c *C) {
+	// this is a static check to ensure we remember to bump the
+	// version when we add fields
+	//
+	// *** IF THIS FAILS, YOU NEED TO BUMP THE VERSION BEFORE "FIXING" THIS ***
+	var sk interfaces.SystemKey
+	c.Check(reflect.ValueOf(sk).NumField(), Equals, interfaces.SystemKeyVersion)
```
bboozzoo: Wouldn't this fail when we remove a field? Also, when we drop a field and then add one, version should be bumped twice IMO.

chipaca: yes, it would fail when we remove a field, and when we change a field. We'll have to fix the test for that.
And yes, about bumping it twice. This check is intended to remind us to bump things, not necessarily to catch every instance of needing to bump things --- that's why the next check is also part of the test.

chipaca: I think we should add comments to this effect.
spread's already underway. If it succeeds, and nothing needs actually fixing, can you add them in your followup?

bboozzoo: Yeah, will do.

interfaces/system_key_test.go
```diff
@@ -226,3 +238,23 @@ func (s *systemKeySuite) TestInterfaceSystemKeyMismatchVersions(c *C) {
...
+		"Version:0",
+		"BuildID:",
+		"AppArmorFeatures:[]",
+		"AppArmorParserMtime:0",
+		"AppArmorParserFeatures:[]",
+		"NFSHome:false",
+		"OverlayRoot:",
+		"SecCompActions:[]",
+		"SeccompCompilerVersion:",
+	}, " ")+"}")
```
stolowski: Nice test!

PR #7567, opened 2019-10-08
snap-repair: error if run as non-root

mvo5:
Running snap-repair as a non-root user is not erroring but the
outputs are really confusing, e.g.:
- snap-repair list will not have a summary
- snap-repair show will show for script and output:
    "error: open .../1/r0.script: permission denied"

So instead of confusing users this PR adds an early check if
snap-repair is really run as root.

### DIFF
--- a/cmd/snap-repair/cmd_run_test.go
+++ b/cmd/snap-repair/cmd_run_test.go
@@ -39,6 +39,8 @@ func (r *repairSuite) TestNonRoot(c *C) {
 	restore = release.MockOnClassic(false)
 	defer restore()
 
+	origArgs := os.Args
+	defer func() { os.Args = origArgs }()
 	os.Args = []string{"snap-repair", "run"}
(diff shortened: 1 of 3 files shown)


cmd/snap-repair/cmd_run_test.go
```diff
@@ -33,8 +33,22 @@ import (
...
 )
 
+func (r *repairSuite) TestNonRoot(c *C) {
+	restore := repair.MockOsGetuid(func() int { return 1000 })
+	defer restore()
+	restore = release.MockOnClassic(false)
+	defer restore()
+
+	os.Args = []string{"snap-repair", "run"}
+	err := repair.Run()
```
bboozzoo: Should there be a `defer func() { os.Args = origArgs }()` right above this line?

mvo5: thank you! nice catch.

PR #7576, opened 2019-10-09
spread.yaml,.gitignore: sync ignored files

zyga:
Having seen more build failures against debian-sid-64 I decided to
garden the ignore lists properly. This patch updates both ignore files
to have the same things.

Some duplicates were removed. Notably one entry was also removed,
"share" caused undesired side-effects when used in spread.yaml's
exclusion list. It removed the "share" directory used in some of the
test snaps. It seems of little value in daily coding, git logs indicate
it is from the age of "git-buildpackage".

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -402,13 +402,80 @@ path: /home/gopath/src/github.com/snapcore/snapd
 
 exclude:
     - .git
-    - cmd/snap/snap
-    - cmd/snapd/snapd
-    - cmd/snapctl/snapctl
-    - cmd/snap-exec/snap-exec
-    - cmd/autom4te.cache
-    - "*.o"
+
+    - tags
+    - .coverage
+    - cmd/version_generated.go
+    - cmd/VERSION
+    - "*~"
+    - "*.swp"
+    - ".*.swp"
+    - "vendor/*/"
+    - ".spread-reuse*.yaml"
+    - po/snappy.pot
+
+    # snap-confine bits
     - "*.a"
+    - "*.o"
+    - .dirstamp
+    # Locally built binaries
+    - cmd/decode-mount-opts/decode-mount-opts
+    - cmd/libsnap-confine-private/unit-tests
+    - cmd/snap-confine/snap-confine
+    - cmd/snap-confine/snap-confine-debug
+    - cmd/snap-confine/snap-confine.apparmor
+    - cmd/snap-confine/unit-tests
+    - cmd/snap-discard-ns/snap-discard-ns
+    - cmd/snap-gdb-shim/snap-gdb-shim
+    - cmd/snap-mgmt/snap-mgmt
+    - cmd/snap-seccomp/snap-seccomp
+    - cmd/snap-update-ns/snap-update-ns
+    - cmd/snapd-env-generator/snapd-env-generator
+    - cmd/snapd-generator/snapd-generator
+    - cmd/system-shutdown/system-shutdown
+    - cmd/system-shutdown/unit-tests
+
+    # manual pages
+    - "cmd/*/*.[1-9]"
+
+    # auto-generated systemd units
+    - "data/systemd/*.service"
+
+    # auto-generated dbus services
+    - "data/dbus/*.service"
+
+    - data/info
+    - data/env/snapd.sh
+
+    # test-driver
+    - "*.log"
+    - "*.trs"
+
+    - # Automake for the cmd/ parts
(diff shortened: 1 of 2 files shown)


spread.yaml
```diff
@@ -402,13 +402,81 @@ path: /home/gopath/src/github.com/snapcore/snapd
...
 exclude:
     - .git
-    - cmd/snap/snap
-    - cmd/snapd/snapd
-    - cmd/snapctl/snapctl
-    - cmd/snap-exec/snap-exec
-    - cmd/autom4te.cache
-    - "*.o"
+
+    - share
```
bboozzoo: won't this drop `tests/lib/snaps/test-snapd-layout/usr/share` too? :smiling_imp:

zyga: 💣

spread.yaml
```diff
@@ -402,13 +402,80 @@ path: /home/gopath/src/github.com/snapcore/snapd
...
+    - cmd/snap-confine/unit-tests
+    - cmd/snap-discard-ns/snap-discard-ns
+    - cmd/snap-gdb-shim/snap-gdb-shim
+    - cmd/snap-mgmt/snap-mgmt
+    - cmd/snap-seccomp/snap-seccomp
+    - cmd/snap-update-ns/snap-update-ns
+    - cmd/snapd-env-generator/snapd-env-generator
+    - cmd/snapd-generator/snapd-generator
+    - cmd/system-shutdown/system-shutdown
+    - cmd/system-shutdown/unit-tests
```
bboozzoo: These should probably be prefixed with `./`

spread.yaml
```diff
@@ -402,13 +402,80 @@ path: /home/gopath/src/github.com/snapcore/snapd
...
+
+    - tags
+    - .coverage
+    - cmd/version_generated.go
+    - cmd/VERSION
+    - "*~"
+    - "*.swp"
+    - ".*.swp"
+    - "vendor/*/"
+    - ".spread-reuse*.yaml"
```
bboozzoo: This is already added by spread

spread.yaml
```diff
@@ -402,13 +402,80 @@ path: /home/gopath/src/github.com/snapcore/snapd
...
-    - "*.o"
+
+    - tags
+    - .coverage
+    - cmd/version_generated.go
+    - cmd/VERSION
+    - "*~"
+    - "*.swp"
+    - ".*.swp"
+    - "vendor/*/"
```
bboozzoo: `./vendor` should work too

spread.yaml
```diff
@@ -402,13 +402,80 @@ path: /home/gopath/src/github.com/snapcore/snapd
...
-    - cmd/snapd/snapd
-    - cmd/snapctl/snapctl
-    - cmd/snap-exec/snap-exec
-    - cmd/autom4te.cache
-    - "*.o"
+
+    - tags
+    - .coverage
+    - cmd/version_generated.go
+    - cmd/VERSION
```
bboozzoo: Probably with `./` prefix

spread.yaml
```diff
@@ -402,13 +402,80 @@ path: /home/gopath/src/github.com/snapcore/snapd
...
+    - cmd/config.h
+    - cmd/config.h.in
+    - cmd/config.status
+    - cmd/config.sub
+    - cmd/configure
+    - cmd/depcomp
+    - cmd/install-sh
+    - cmd/missing
+    - cmd/stamp-h1
+    - cmd/test-driver
```
bboozzoo: Probably the same `./` prefix thing

spread.yaml
```diff
@@ -402,13 +402,80 @@ path: /home/gopath/src/github.com/snapcore/snapd
...
 exclude:
     - .git
-    - cmd/snap/snap
-    - cmd/snapd/snapd
-    - cmd/snapctl/snapctl
-    - cmd/snap-exec/snap-exec
-    - cmd/autom4te.cache
-    - "*.o"
+
+    - tags
```
bboozzoo: can you add
```
- GRTAGS
- GTAGS
- GPATH
- cscope.out
```
as well?

zyga:
I experimented with this some more but I'll call it a day now. I need to look at how git implements the ignore logic but it seems that we cannot get perfectly same behaviour with both tools. We may need to settle for less (similar but not identical) for the sake of practicality.

zyga:
I'm going to close this now. Boy was this a journey.

PR #7578, opened 2019-10-09
sandbox/cgroup, overlord/snapstate: move helper for listing pids in group to the cgroup package

bboozzoo:
Part of cgroup cleanup. Move and update the helper for listing PIDs of a cgroup
from overlord/snapstate/snapstate to sandbox/cgroup.

### DIFF
--- a/overlord/snapstate/refresh.go
+++ b/overlord/snapstate/refresh.go
@@ -20,18 +20,17 @@
 package snapstate
 
 import (
-	"bufio"
 	"fmt"
-	"io"
-	"os"
-	"path/filepath"
 	"sort"
-	"strconv"
 	"strings"
 
 	"github.com/snapcore/snapd/cmd/snaplock"
-	"github.com/snapcore/snapd/dirs"
+	"github.com/snapcore/snapd/sandbox/cgroup"
 	"github.com/snapcore/snapd/snap"
+)
+
+var (
+	pidsCgroupDir = cgroup.ControllerPathV1("pids")
 )
 
 func genericRefreshCheck(info *snap.Info, canAppRunDuringRefresh func(app *snap.AppInfo) bool) error {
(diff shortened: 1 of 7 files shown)


overlord/snapstate/refresh.go
```diff
@@ -20,20 +20,19 @@
...
 	"strings"
 
 	"github.com/snapcore/snapd/cmd/snaplock"
-	"github.com/snapcore/snapd/dirs"
+	"github.com/snapcore/snapd/sandbox/cgroup"
 	"github.com/snapcore/snapd/snap"
 )
 
+var (
+	pidsCgroupDir = cgroup.ControllerPathV1("pids")
```
zyga: Nice

Once the name=snapd cgroup lands this will change to just `/run/snapd/cgroup` right?

sandbox/cgroup/cgroup.go
```diff
@@ -146,6 +148,49 @@ func ProcGroup(pid int, controller string) (string, error) {
 	return "", fmt.Errorf("cannot find cgroup controller %q path for pid %v", controller, pid)
 }
 
+// PidsInGroup returns the list of process ID currently registered in a given cgroup
+func PidsInGroup(hierarchyMount, groupPath string) ([]int, error) {
+	// TODO: check whether hierarchyMount looks like a valid cgroup root
```
zyga: Aren't you doing that a few lines below already. If the file doesn't exist then you return. What am I missing?

bboozzoo: That's about adding a low tech check that the hierarchy mount point looks like a cgroup hierarchy (eg. there's at least a cgroup.procs, or maybe look at statfs). Otherwise you can pass an arbitrary path and the code will happily return nil, nil, even though you may not be looking at cgroups hierarchy at all.

overlord/snapstate/refresh.go
bboozzoo: Yes

PR #7580, opened 2019-10-10
spread.yaml: exclude vendor dir

anonymouse64:
This specific dir was breaking local runs of spread with QEMU images of Ubuntu Core for me, because the mounted raw image that we flash to the QEMU device has limited space and couldn't fit all of the vendor dir in it.

There's more we could do with this by syncing .gitignore with the spread.yaml's exclude, but it seems that's a bit more complicated than it seems given #7576 

Feel free to close this if #7576 is fixed/updated properly first.

### DIFF
--- a/spread.yaml
+++ b/spread.yaml
@@ -409,7 +409,7 @@ exclude:
     - cmd/autom4te.cache
     - "*.o"
     - "*.a"
-
+    - ./vendor
 
 debug-each: |
     if [ "$SPREAD_DEBUG_EACH" = 1 ]; then


bboozzoo:
There's bit more stuff in #7576

bboozzoo:
Looks like it may take a while for #7576 to land.

zyga:
This is now handled by my updated branches:
 - https://github.com/snapcore/snapd/pull/7585
 - https://github.com/snapcore/spread/pull/89

PR #7581, opened 2019-10-10
tests: add netplan test on ubuntu core

anonymouse64:
This new, ubuntu-core specific test is expected to fail on Ubuntu Core 18 until a new revision of the core18 snap is published which has netplan.io version `0.98-0ubuntu1~18.04.1` in it. 

See https://bugs.launchpad.net/ubuntu/+source/netplan.io/+bug/1840832 for the 18.04 SRU and https://bugs.launchpad.net/ubuntu/xenial/+source/nplan/+bug/1842511 for the 16.04 SRU.

The old netplan-apply test has been renamed to make it clear that test uses a fake netplan D-Bus service. The new UC test is separate from that test because that test ensures the interface is working, while this integration test is more meant to make sure the full end-to-end netplan works from inside a strict snap on Ubuntu Core.


anonymouse64:
Specifically, the UC18 test fails thusly:

```
2019-10-09 21:22:28 Error executing qemu:ubuntu-core-18-64:tests/main/ubuntu-core-netplan-apply : 
-----
+ echo 'The interface is disconnected by default'
The interface is disconnected by default
+ MATCH 'network-setup-control +test-snapd-netplan-apply:network-setup-control +- +-'
+ snap connections test-snapd-netplan-apply
+ echo 'Running netplan apply without network-setup-control fails'
Running netplan apply without network-setup-control fails
+ test-snapd-netplan-apply.netplan apply
The name io.netplan.Netplan was not provided by any .service files
+ echo 'When the interface is connected'
When the interface is connected
+ snap connect test-snapd-netplan-apply:network-setup-control
+ echo 'Running netplan apply now works'
Running netplan apply now works
+ test-snapd-netplan-apply.netplan apply
The name io.netplan.Netplan was not provided by any .service files
+ echo 'Unexpected error running netplan apply'
Unexpected error running netplan apply
+ exit 1
-----
```

tests/main/ubuntu-core-netplan-apply/task.yaml
```diff
@@ -0,0 +1,39 @@
...
+    fi
+
+    echo "When the interface is connected"
+    snap connect test-snapd-netplan-apply:network-setup-control
+
+    echo "Running netplan apply now works"
+    if ! test-snapd-netplan-apply.netplan apply; then
+        echo "Unexpected error running netplan apply"
+        exit 1
+    fi
```
mvo5: Can we somehow check that netplan did actually run here? Does it create some log or state or something that we might check if it got an updated timestamp or similar?

anonymouse64: Yes, we could check that the Network service was stopped and started again. I'll add this

anonymouse64:
Actually, we can enable this now for UC18, as the edge channel of the core18 snap was just rebuilt today and has the required version of netplan. Since this is still small and hasn't been looked at much, I will just force push the commit that disables UC18 out of existence

tests/main/ubuntu-core-netplan-apply/task.yaml
mvo5: It's fine to do this in a followup if you prefer it this way

anonymouse64: No, sorry I was just busy with other things the PR is now updated to do this.

mvo5:
@sergiocazzolato Can you please have a look?

tests/main/ubuntu-core-netplan-apply/task.yaml
```diff
@@ -0,0 +1,57 @@
...
+environment:
+    NETPLAN: io.netplan.Netplan
+
+# only run on ubuntu core systems
+systems:
+    - ubuntu-core-*
+
+prepare: |
+    #shellcheck source=tests/lib/snaps.sh
+    . "$TESTSLIB"/snaps.sh
```
sergiocazzolato: it is not needed to source it

anonymouse64: Thanks, added this

anonymouse64:
Now that #7579 has landed, I will update this to add a "real" test with the Info D-Bus API.

tests/main/ubuntu-core-netplan-apply/task.yaml
```diff
@@ -0,0 +1,55 @@
...
+    snap connections test-snapd-netplan-apply | MATCH 'network-setup-control +test-snapd-netplan-apply:network-setup-control +- +-'
+
+    echo "Running netplan apply without network-setup-control fails"
+    if test-snapd-netplan-apply.netplan apply; then
+        echo "Expected access denied error for netplan apply"
+        exit 1
+    fi
+
+    echo "Count how many network service restarts happened before calling netplan apply"
+    wait_for_network_config_stopped_before="$(get_journalctl_log -u systemd-networkd | grep -c 'Stopped Network Service.' || true)"
```
zyga: This variable is a little mouthful. Perhaps just `stopped_before` and `started_before`

tests/main/ubuntu-core-netplan-apply/task.yaml
```diff
@@ -0,0 +1,55 @@
...
+        wait_for_network_config_started_after="$(get_journalctl_log -u systemd-networkd | grep -c 'Started Network Service.' || true)"
+        if [ "$wait_for_network_config_stopped_after" -gt "$wait_for_network_config_stopped_before" ] && \
+           [ "$wait_for_network_config_started_after" -gt "$wait_for_network_config_started_before" ] ; then
+            break
+        fi
+        sleep 1
+    done
+
+    echo "Ensure that the number of network restarts is greater after netplan apply was run"
+    [ "$wait_for_network_config_stopped_after" -gt "$wait_for_network_config_stopped_before" ] && \
```
zyga: Nitpick, given that you have `&&` between them, IMO this would read more easily in this form:
```
test "$wait_for_network_config_stopped_after" -gt "$wait_for_network_config_stopped_before"
test "$wait_for_network_config_started_after" -gt "$wait_for_network_config_started_before"
```

tests/main/ubuntu-core-netplan-apply/task.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+environment:
+    NETPLAN: io.netplan.Netplan
+
+# only run on ubuntu core systems
+systems:
+    - ubuntu-core-*
+
+prepare: |
+    snap install test-snapd-netplan-apply --edge
```
zyga: Is this snap something that we must get from the store for correctness or just for convenience?

zyga:
@anonymouse64 this branch has conflicts now.

tests/main/ubuntu-core-netplan-apply/task.yaml
anonymouse64: It has to be built with snapcraft, so we could try to build it in the test but this seems a bit heavy for a test job to need to install snapcraft, etc. especially since we are now on core for this test and this snap is re-used from the fake-netplan test so it would need to be built there too.

tests/main/ubuntu-core-netplan-apply/task.yaml
anonymouse64: ok, renamed

tests/main/ubuntu-core-netplan-apply/task.yaml
anonymouse64: with the rename I think this is okay

anonymouse64:
> @anonymouse64 this branch has conflicts now.

Yes they have been resolved

anonymouse64:
Note that I updated this PR to also test the Info D-Bus method on UC18 only because the Info D-Bus method has not been SRU'd to xenial and AFAIK there are no plans to do this.

PR #7586, opened 2019-10-11
interfaces/optical-drive: additional permission needed by mir-kiosk-kodi

AlanGriffiths:
In addition to the existing `optical-drive` permissions Kodi (and possibly other apps) needs read access to `/var/lib/snapd/hostfs/media/<user>/DVDVolume/` in order to play DVDs.

### DIFF
--- a/interfaces/builtin/optical_drive.go
+++ b/interfaces/builtin/optical_drive.go
@@ -46,6 +46,7 @@ const opticalDriveConnectedPlugAppArmor = `
 # Allow read access to optical drives
 /dev/sr[0-9]* r,
 /dev/scd[0-9]* r,
+owner /var/lib/snapd/hostfs/media/*/DVDVolume/{,**} r,
 # allow all generic scsi devices here and use the device cgroup to
 # differentiate optical drives
 /dev/sg[0-9]* r,


interfaces/builtin/optical_drive.go
```diff
@@ -46,6 +46,7 @@ const opticalDriveConnectedPlugAppArmor = `
 # Allow read access to optical drives
 /dev/sr[0-9]* r,
 /dev/scd[0-9]* r,
+owner /var/lib/snapd/hostfs/media/*/DVDVolume/{,**} r,
```
zyga: Why cannot this be accessed through `/media/*` using the `removable-media` interface?

AlanGriffiths: Is there some trick to using `/media`? I've connected `removable-media` but when I try to access a DVD I get:
```plain
[28703.307105] audit: type=1400 audit(1570809000.510:1406): apparmor="DENIED" operation="open" profile="snap.mir-kiosk-kodi.mir-kiosk-kodi" name="/var/lib/snapd/hostfs/media/alan/DVDVolume/" pid=24650 comm="kodi-wayland" requested_mask="r" denied_mask="r" fsuid=1000 ouid=1000
```

AlanGriffiths: (I can find no obvious reason Kodi uses this path to open the DVD, nor any way to change it.)

AlanGriffiths: @zyga the snap already uses the `removable-media`interface. Would that be a better place to add this rule?

I don't particularly want to go down a rabbit-hole trying to find out what happens here. And the rule I'm suggesting here doesn't seem particularly problematic?

AFAICS Kodi isn't itself trying to open `/var/lib/snapd/hostfs/media/alan/DVDVolume/`I don't think it is even aware of that as a meaningful path. (I'm no expert on the Kodi codebase, so I may have missed something.) Perhaps an FD that was originally opened with that path is being migrated into the process? (I've seen similar issues with the `wayland` interface.) If true, that could also affect other media players.

If you want to investigate further I've pushed a version of mir-kiosk-kodi with the `optical-drive` interface and a script to apply this patch to a branch on the store. (You'll need a Wayland desktop or mir-kiosk.)
```plain
snap install --channel edge/dvd-support mir-kiosk-kodi
/snap/mir-kiosk-kodi/current/bin/setup.sh
```
If you insert a DVD it fails to play (you'll see the above error in dmesg).
```plain
/snap/mir-kiosk-kodi/current/bin/hack-optical-drive.sh
```
Now it "works". (Code here: https://github.com/MirServer/mir-kiosk-kodi/pull/5)

zyga: I will check this out. Thank you for sharing the snap.

AlanGriffiths: @zyga did you ever get anywhere with this?

jdstrand: If an additional access were to be added, it should be to the removable-media interface, yes, but the fact that it is showing up in hostfs is really strange. This should be handled by these rules:
```
/{,run/}media/*/ r,
/{,run/}media/*/** rwkl,
```

Perhaps there is a symlink in /media out to hostfs? There seems to be a problem with the snap that it is finding the mount point in hostfs. I wonder if it is doing something with the mount table?

NAK for applying to optical-drive.

I would prefer this not be added to removable-media until we understand what the application is doing, otherwise I think we are papering over an issue that might affect other snaps. @zyga said he'd look at this, did anything come of that? If not, an strace might be helpful.

AlanGriffiths:
@mardy I can't currently reproduce with the current version of mir-kiosk-kodi (this functionality fails before the point where problems arose here), nor easily rebuild with the recipe cited above (the xbmc PPA has changed).

PR #7601, opened 2019-10-15
overlord/ifacestate: use SetupMany in setupSecurityByBackend

stolowski:
Based on #7355 

Use SetupMany helper in setupSecurityByBackend (=setupProfilesForSnap), i.e. benefit from apparmor optimization introduced in #7355 also when generating profiles for a single snap and it's affected (i.e. connected) snaps.


overlord/ifacestate/helpers.go
```diff
@@ -357,22 +357,27 @@ func (m *InterfaceManager) removeConnections(snapName string) error {
...
+		return fmt.Errorf("internal error: setupSecurityByBackend received an unexpected number of snaps (expected: %d, got %d)", len(opts), len(snaps))
+	}
+	confOpts := make(map[string]interfaces.ConfinementOptions, len(snaps))
+	for i, snapInfo := range snaps {
+		confOpts[snapInfo.InstanceName()] = opts[i]
+	}
+
 	st := task.State()
+	st.Unlock()
+	defer st.Lock()
```
jdstrand: The lock ends up being in place for a pretty long time in one go, as opposed to before with several smaller locks. I can't see a way around it and am not sure it is a problem, but it might be worth verifying it is not a problem.

overlord/ifacestate/ifacestate_test.go
```diff
@@ -3078,6 +3078,34 @@ func (s *interfaceManagerSuite) TestSetupProfilesHonorsDevMode(c *C) {
...
+	})
+	s.settle(c)
+
+	s.state.Lock()
+	defer s.state.Unlock()
+
+	c.Check(change.Status(), Equals, state.ErrorStatus)
+	c.Check(change.Err(), ErrorMatches, `cannot perform the following tasks:\n-  \(fail\)`)
+}
+
```
jdstrand: A test for `if len(snaps) != len(opts) {` would be good since you are doing the check. Presumably it was added because it is conceivable that the case could be hit, so having a test to ensure that check isn't dropped down the line would be good.

stolowski: Test added. It can only be tested directly though, so I exported this helper for tests.

overlord/ifacestate/helpers.go
stolowski: This is actually not the case ;), but I see how this can be confusing. The function is called with lock in place, and we  unlock the state before calling backend's setup. The state needs to be locked back before returning from the function.

jdstrand: Right, I understand that but typed it wrong (erf). I meant to say 'This *un*lock ends up being in place for a pretty long time in one go, as opposed to before with several smaller *un*locks.'

My point was that a) there is more code in each call to SetupMany() than to the previous Setup() in the loop iteration and b) now all iterations of the loop must be gone through before Unlock() whereas before, it was an Unlock()/Lock() on each iteration. I'm not saying this is a problem, just calling it out since I don't know what the side-effects of being unlocked for so long are, if any.

stolowski: Ok, I see. I was thinking if the unlocking/locking there was deliberate or not. Semantically it looks irrelevant to me and if it had any sideffects it would be a bit fishy tbh. Either way it's impossible to test / prove.

PR #7603, opened 2019-10-15
interfaces: add dpdk interface

alfonsosanchezbeato:
Implements dpdk-control interface.

This is a follow up from @wililupy's PR https://github.com/snapcore/snapd/pull/5885, but now including support for Mellanox NICs.


interfaces/builtin/hugepages_control.go
```diff
@@ -0,0 +1,49 @@
...
+
+const hugepagesControlBaseDeclarationSlots = `
+  hugepages-control:
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
+`
+
+const hugepagesControlConnectedPlugAppArmor = `
```
zyga: Is it desired that this snippet provides _write_ access without _read_ access to `/dev/` and `/run/`?

interfaces/builtin/hugepages_control.go
```diff
@@ -0,0 +1,49 @@
...
+# Description: Allow control to hugepages.
+/sys/kernel/mm/hugepages/{,**} rw,
```
anonymouse64: should this perhaps be

```suggestion
/sys/kernel/mm/{hugepages,transparent_hugepage}/{,**} rw,
```

For example see some denials from the docker snap where it seems like docker is now trying to look at hugepages: https://forum.snapcraft.io/t/docker-18-09-9-snap-crashes-with-apparmor-denial-on-sys-kernel-mm-transparent-hugepage-hpage-pmd-size/14218.

Also from my basic reading of https://www.kernel.org/doc/Documentation/vm/transhuge.txt it sounds like transparent_hugepage is a better (for some definition of better) hugepage implementation so it may be used more often nowadays?

interfaces/builtin/hugepages_control.go
wililupy: When testing the interface, I found that with the write access, read was implied as well and it worked correctly.

interfaces/builtin/hugepages_control.go
wililupy: I do not see an issue with adding transparent_hugepage, this interface should be able to be used with multiple snaps that request/require hugepages, like docker or kvm as well.

interfaces/builtin/hugepages_control.go
alfonsosanchezbeato: I have set read explicitly here, it is not clear to me by looking at apparmor docs if this is implicitly set by "wk" in some way.

interfaces/builtin/hugepages_control.go
alfonsosanchezbeato: added

alfonsosanchezbeato:
@jdstrand could you please take a look at this one?

interfaces/builtin/dpdk_control.go
```diff
@@ -0,0 +1,80 @@
...
+
+const dpdkControlConnectedPlugSecComp = `
+# Description: Controls DPDK SecComp.
+move_pages
+`
+
+// Network drivers with DPDK support used by different vendors.
+var dpdkConnectedPlugKmod = []string{
+	"mlx4_core",
+	"mlx5_core",
```
jdstrand: Lots of the above drivers need other modules loaded. Eg, vfio needs vfio-pci, uio needs igb_uio or uio_pci_generic, etc.

interfaces/builtin/dpdk_control.go
```diff
@@ -0,0 +1,80 @@
...
+/dev/uio[0-9]* rw,
+/dev/vfio/vfio rw,
+/dev/infiniband/rdma_cm rw,
+/dev/infiniband/uverbs[0-9]* rw,
+/sys/bus/pci/drivers{,/**} rw,
+`
+
+const dpdkControlConnectedPlugSecComp = `
+# Description: Controls DPDK SecComp.
+move_pages
```
jdstrand: This doesn't provide enough information. It also seems to require CAP_SYS_NICE on 4.12 and later and PTRACE_MODE_READ_REALCREDS prior (see man move_pages).

interfaces/builtin/dpdk_control.go
```diff
@@ -0,0 +1,80 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package builtin
+
+const dpdkControlSummary = `allows using dpdk`
```
jdstrand: Please adjust to be:
```
// http://doc.dpdk.org/guides/linux_gsg/
const dpdkControlSummary = `allows configuring dpdk drivers`
```

interfaces/builtin/dpdk_control.go
```diff
@@ -0,0 +1,80 @@
...
+  dpdk-control:
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
+`
+
+const dpdkControlConnectedPlugAppArmor = `
+# Description: Allow control to dpdk.
+/run/dpdk{,/**} rwk,
```
jdstrand: What files are in this directory? What do they do?

interfaces/builtin/hugepages_control.go
```diff
@@ -0,0 +1,48 @@
...
+/sys/kernel/mm/{hugepages,transparent_hugepage}/{,**} rw,
+/dev/hugepages/{,**} rwk,
+/run/hugepages/{,**} rwk,
+`
+
+func init() {
+	registerIface(&commonInterface{
+		name:                  "hugepages-control",
+		summary:               hugepagesControlSummary,
+		implicitOnCore:        true,
```
jdstrand: What is providing /run/dpdk on the host on core? Are you expecting snaps to ship stuff that write to there? Would that not be properly implemented as a slot?

interfaces/builtin/dpdk_control.go
```diff
@@ -0,0 +1,80 @@
...
+/sys/bus/pci/drivers/e1000/bind rw,
+/sys/bus/pci/drivers/igb_uio/bind rw,
+/sys/bus/pci/drivers/rte_kni/bind rw,
+/sys/bus/pci/drivers/uio_pci_generic/bind rw,
+/sys/devices/pci*{,/**} rw,
+/proc/ioports r,
+/dev/uio[0-9]* rw,
+/dev/vfio/vfio rw,
+/dev/infiniband/rdma_cm rw,
+/dev/infiniband/uverbs[0-9]* rw,
```
jdstrand: All of this /dev accesses are missing the corresponding UDev backend setup.

interfaces/builtin/dpdk_control.go
```diff
@@ -0,0 +1,80 @@
...
+/sys/bus/pci/drivers{,/**} rw,
```
jdstrand: This is way too lenient.

jdstrand:
While dpdk remains in this PR, marking as Blocked and Complex. If dpdk is moved out, hugepages would be relatively straightforward to iterate on and land.

alfonsosanchezbeato:
I have created a new MP just for the hugepages interface: https://github.com/snapcore/snapd/pull/8271

alfonsosanchezbeato:
@jdstrand @pedronis I am not a DPDK expert, but let me try to explain how it works.

DPDK uses polling instead of interrupts to maximize network throughput. For this, it changes the usual network driver for an interface and replaces it with UIO or VFIO, which lets DPDK handle most of the packet processing in user space.

Therefore, it needs to be able to unbind/bind different ethernet drivers and also have access to UIO/VFIO devices.

The `/run/dpdk/rte` folder contains a socket and some configuration files that are created by dpdk's Run Time Environment (RTE). The socket is for multi-process support in dpdk: dpdk allows to create a group of processes that work together in a simple transparent manner to perform packet processing.

To minimize the permissions allowed by the interface, we would need a way to give permissions just for a given network device: for instance, if we work with intel ethernet cards, we would need permissions to unbind igb and access VFIO devices.

So, which direction do you think we should follow to support confined dpdk?

I do not think there will be much difference on making dpdk work confined between core and classic.

interfaces/builtin/hugepages_control.go
jdstrand: FYI, 'w' does *not* imply read, but it is possible that another interface already granted read.

interfaces/builtin/dpdk_control.go
```diff
@@ -0,0 +1,80 @@
...
+/sys/bus/pci/drivers/rte_kni/unbind rw,
+/sys/bus/pci/drivers/uio_pci_generic/unbind rw,
+/sys/bus/pci/drivers/virtio-pci/bind rw,
+/sys/bus/pci/drivers/e1000/bind rw,
+/sys/bus/pci/drivers/igb_uio/bind rw,
+/sys/bus/pci/drivers/rte_kni/bind rw,
+/sys/bus/pci/drivers/uio_pci_generic/bind rw,
+/sys/devices/pci*{,/**} rw,
+/proc/ioports r,
+/dev/uio[0-9]* rw,
```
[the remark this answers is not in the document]
alfonsosanchezbeato: Right, but afaiu the `uio` interface must be implemented by a gadget snap. What dpdk does is changing the eth driver for a card and lets userspace access the HW more directly, via the uio_pci_generic driver. We would be requiring a gadget for any server that uses dpdk, and I am not sure if that makes sense.

interfaces/builtin/dpdk_control.go
```diff
@@ -0,0 +1,80 @@
...
+/sys/bus/pci/drivers/e1000/unbind rw,
+/sys/bus/pci/drivers/igb_uio/unbind rw,
+/sys/bus/pci/drivers/rte_kni/unbind rw,
+/sys/bus/pci/drivers/uio_pci_generic/unbind rw,
+/sys/bus/pci/drivers/virtio-pci/bind rw,
+/sys/bus/pci/drivers/e1000/bind rw,
+/sys/bus/pci/drivers/igb_uio/bind rw,
+/sys/bus/pci/drivers/rte_kni/bind rw,
+/sys/bus/pci/drivers/uio_pci_generic/bind rw,
+/sys/devices/pci*{,/**} rw,
```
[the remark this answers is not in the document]
alfonsosanchezbeato: Removed for the moment, being able to bind/unbind the driver should be enough.

interfaces/builtin/dpdk_control.go
```diff
@@ -0,0 +1,80 @@
...
+    allow-installation:
+      slot-snap-type:
+        - core
+    deny-auto-connection: true
+`
+
+const dpdkControlConnectedPlugAppArmor = `
+# Description: Allow control to dpdk.
+/run/dpdk{,/**} rwk,
+owner @{PROC}/@{pid}/pagemap r,
```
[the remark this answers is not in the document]
alfonsosanchezbeato: DPDK uses this to get the physical address of any mapped virtual address in the current process. As it controls directly the PHY layer of the ethernet device, I guess it needs it some times to track the packet data.

interfaces/builtin/dpdk_control.go
```diff
@@ -0,0 +1,80 @@
...
+/sys/bus/pci/drivers/igb_uio/unbind rw,
+/sys/bus/pci/drivers/rte_kni/unbind rw,
+/sys/bus/pci/drivers/uio_pci_generic/unbind rw,
+/sys/bus/pci/drivers/virtio-pci/bind rw,
+/sys/bus/pci/drivers/e1000/bind rw,
+/sys/bus/pci/drivers/igb_uio/bind rw,
+/sys/bus/pci/drivers/rte_kni/bind rw,
+/sys/bus/pci/drivers/uio_pci_generic/bind rw,
+/sys/devices/pci*{,/**} rw,
+/proc/ioports r,
```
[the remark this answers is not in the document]
alfonsosanchezbeato: Moved

alfonsosanchezbeato:
I have rebased the branch now that hugepages is merged, and addressed part of the comments. I think that maybe the main blocker for this interface is that it provides access to different HW devices: one for intel NICs, other for Mellanox ones, etc. @jdstrand would it help to have different interfaces for each? Like, having a `dpdk-control` interface and then `dpdk-intel`, `dpdk-mellanox`, etc.?

Also, did my previous comment clarify things a bit on what we need?

interfaces/builtin/dpdk_control.go
alfonsosanchezbeato: Added now

PR #7605, opened 2019-10-15
tests: configure the journald service for core systems

sergiocazzolato:
This change makes the same journal configuration for classic and core
systems.

Previously the configuration for the journald service was being done on
prepare_project function (prepare-restore.sh), but when the target
system is core, the prepare_project function is executed on
ubuntu-16.04-64 for core16 or ubuntu-18.04-64 for core18. It happens
because the target ubuntu-core system is created during the first test
suite preparation which happens after the prepare_project function
execution.

The idea is to configure the journald service after the ubuntu-core
system is booted.

### DIFF
--- a/tests/lib/prepare.sh
+++ b/tests/lib/prepare.sh
@@ -25,6 +25,21 @@ disable_kernel_rate_limiting() {
     #sysctl -w kernel.printk_ratelimit=0
 }
 
+disable_journald_rate_limiting() {
+    # Disable journald rate limiting
+    mkdir -p /etc/systemd/journald.conf.d
+    # The RateLimitIntervalSec key is not supported on some systemd versions causing
+    # the journal rate limit could be considered as not valid and discarded in concecuence.
+    # RateLimitInterval key is supported in old systemd versions and in new ones as well,
+    # maintaining backward compatibility.
+    cat <<-EOF > /etc/systemd/journald.conf.d/no-rate-limit.conf
+    [Journal]
+    RateLimitInterval=0
+    RateLimitBurst=0
+EOF
+    systemctl restart systemd-journald.service
+}
+
 ensure_jq() {
     if command -v jq; then
         return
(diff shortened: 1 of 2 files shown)


tests/lib/prepare.sh
```diff
@@ -25,6 +25,21 @@ disable_kernel_rate_limiting() {
     #sysctl -w kernel.printk_ratelimit=0
 }
 
+disable_journald_rate_limiting() {
+    # Disable journald rate limiting
+    mkdir -p /etc/systemd/journald.conf.d
```
bboozzoo: This fails on core16 because /etc/systemd/journal.conf.d is not a writable location. However, on core18, the whole /etc/systemd is writable.

tests/lib/prepare.sh
```diff
@@ -25,6 +25,21 @@ disable_kernel_rate_limiting() {
...
+    # The RateLimitIntervalSec key is not supported on some systemd versions causing
+    # the journal rate limit could be considered as not valid and discarded in concecuence.
+    # RateLimitInterval key is supported in old systemd versions and in new ones as well,
+    # maintaining backward compatibility.
+    cat <<-EOF > /etc/systemd/journald.conf.d/no-rate-limit.conf
+    [Journal]
+    RateLimitInterval=0
+    RateLimitBurst=0
+EOF
+    systemctl restart systemd-journald.service
```
bboozzoo: As discussed yesterday, we may need an override for the actual service too like this:
```
     cat <<-EOF > /etc/systemd/journald.conf.d/no-rate-limit.conf
     [Journal]
     RateLimitInterval=0
     RateLimitBurst=0
 EOF
     systemctl restart systemd-journald.service
```
but that depends on whether the tests actually fail because of journald reaching the start retry limit.

anonymouse64:
It would be great to get this, I see this somewhat regularly (maybe a couple times a week)

anonymouse64:
Still seeing this kind of failure unfortunately... @sergiocazzolato is there anything I can do to help you with this PR?

tests/lib/prepare.sh
mvo5: https://github.com/snapcore/core/pull/109 <- should fix this.

PR #7613, opened 2019-10-16
snap-recovery: add minimal binary so that we can use spread on it

mvo5:
This adds a minimal snap-recovery binary and spread spread tests for creating an initial partition and append to an existing partition.

Based on https://github.com/snapcore/snapd/pull/7609

### DIFF
--- a/cmd/snap-recovery/partition/sfdisk.go
+++ b/cmd/snap-recovery/partition/sfdisk.go
@@ -100,16 +100,12 @@ func (sf *SFDisk) Create(pv *gadget.LaidOutVolume) (map[string]string, error) {
 	}
 	buf, deviceMap := buildPartitionList(sf.partitionTable, pv)
 
-	// Write the partition table
+	// Write the partition table, note that sfdisk will re-read the
+	// partition table by itself: see disk-utils/sfdisk.c:write_changes()
 	cmd := exec.Command("sfdisk", sf.device)
 	cmd.Stdin = buf
 	if output, err := cmd.CombinedOutput(); err != nil {
 		return deviceMap, osutil.OutputErr(output, err)
-	}
-
-	// Reload the partition table using blockdev which is part of the initramfs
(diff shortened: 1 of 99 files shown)


gadget/partition.go
```diff
@@ -113,3 +113,31 @@ func runSfdisk(image string, script string) error {
 	}
 	return nil
 }
+
+// PositionedVolumeFromGadget takes a gadget rootdir and positions the
+// partitions as specified.
+func PositionedVolumeFromGadget(gadgetRoot string) (*LaidOutVolume, error) {
```
bboozzoo: This should probably be moved to layout.go or even gadget.go

cmd/snap-recovery/partition/sfdisk.go
```diff
@@ -100,18 +100,14 @@ func (sf *SFDisk) Create(pv *gadget.LaidOutVolume) (map[string]string, error) {
...
+	// Write the partition table, note that sfdisk will re-read the
+	// partition table by itself: see disk-utils/sfdisk.c:write_changes()
 	cmd := exec.Command("sfdisk", sf.device)
 	cmd.Stdin = buf
 	if output, err := cmd.CombinedOutput(); err != nil {
 		return deviceMap, osutil.OutputErr(output, err)
 	}
 
-	// Reload the partition table using blockdev which is part of the initramfs
-	if output, err := exec.Command("blockdev", "--rereadpt", sf.device).CombinedOutput(); err != nil {
```
bboozzoo: Right, this only makes sense when we operate on the actual device and want to make the kernel reload the partition table. At some point perhaps osutil would even be a better place for this.

mvo5: Yeah, we don't need it here, I checked the sfdisk code and it is doing this by itself after it has written the changes to the disk.

mvo5: I also added a comment close to the sfdisk call that this already rescanns the partition table.

PR #7614, opened 2019-10-16
cmd/snap-confine: implement snap-device-helper internally

zyga:
This patch begins the cleanup of snap-confine's handling of udev
tagging. The script snap-device-helper had two roles:

1) Invoked by udev to dynamically add and remove device group ACLs
in response to udev hotplug events.
2) Invoked by snap-confine to configure the device cgroup.

The second role is discarded. Instead snap-confine now implements the
relevant logic itself. This "logic" is literally:

    dprintf(devices_allow_fd, "%c %u:%u", block_device ? 'b' : 'c',
	dev_major, dev_minor);

Doing so in C allows us to spare countless fork + exec combos, which
also fork for various parts of the script, making snap execution a
little bit faster.

In addition, since the snap-device-helper script is now only invoked by
udev in the initial mount namespace, we no longer have to worry about
the script interpreter. As such the packaging for some distributions is
adjusted to disable our previous workarounds.

This step opens the path towards cgroup v2 support, although much more
work is still required there.


cmd/snap-confine/udev-support.c
```diff
@@ -18,290 +18,355 @@
...
+		const char *path = udev_list_entry_get_name(entry);
+		if (path == NULL) {
+			die("udev_list_entry_get_name failed");
+		}
+		struct udev_device *device =
+		    udev_device_new_from_syspath(udev, path);
+		if (device == NULL) {
+			die("cannot find device from syspath %s", path);
+		}
+		dev_t devnum = udev_device_get_devnum(device);
```
bboozzoo: As discussed on IRC, on my system a bunch of devices that do not have a device node are tagged right now, eg. video ports that apppear as separate devices under sys, but have no device node. The old code did a silly thing as well:
```
DEBUG: run_snappy_app_dev_add: /sys/devices/pci0000:00/0000:00:03.1/0000:23:00.0/0000:24:00.0/0000:25:00.0/drm/card0 snap_ohmygiraffe_ohmygiraffe
DEBUG: running snap-device-helper add snap_ohmygiraffe_ohmygiraffe /sys/devices/pci0000:00/0000:00:03.1/0000:23:00.0/0000:24:00.0/0000:25:00.0/drm/card0 226:0
DEBUG: run_snappy_app_dev_add: /sys/devices/pci0000:00/0000:00:03.1/0000:23:00.0/0000:24:00.0/0000:25:00.0/drm/card0/card0-DP-1 snap_ohmygiraffe_ohmygiraffe
>>>> see one below 
DEBUG: running snap-device-helper 
  add 
  snap_ohmygiraffe_ohmygiraffe 
  /sys/devices/pci0000:00/0000:00:03.1/0000:23:00.0/0000:24:00.0/0000:25:00.0/drm/card0/card0-DP-1
  0:0 <-- major:minor
...

DEBUG: running snap-device-helper add snap_ohmygiraffe_ohmygiraffe /sys/devices/pci0000:00/0000:00:03.1/0000:23:00.0/0000:24:00.0/0000:25:00.0/drm/renderD128 226:128
```

By assuming all devices have a devnum, we ended up adding `c 0:0 rwm` to allowed devices. This probably has no effect, since there's no block/char devices with major 0.

cmd/snap-confine/udev-support.c
```diff
@@ -18,290 +18,355 @@
...
+		char type_c = strstr(path, "/block/") != NULL ? 'b' : 'c';
```
bboozzoo: Maybe we should use `udev_device_get_devnode()` and then `stat()` the resulting path to get the type or check whether subsystem is `"block"`?

cmd/snap-confine/udev-support.c
```diff
@@ -18,290 +18,355 @@
...
+		sc_cleanup_close(&fds->devices_allow_fd);
+		sc_cleanup_close(&fds->devices_deny_fd);
+		sc_cleanup_close(&fds->cgroup_procs_fd);
+	}
+}
+
+void sc_setup_device_cgroup(const char *security_tag)
+{
+	if (sc_cgroup_is_v2()) {
+		return;
```
bboozzoo: Perhaps leave a TODO note here:
```
// TODO: currently unsupported
// cgroupv2 uses BPF for mediating device access
```

cmd/snap-confine/udev-support.c
bboozzoo: The 0:0 really ended up in devices list:
```
$ sudo cat /sys/fs/cgroup/devices/snap.ohmygiraffe.ohmygiraffe/devices.list
[sudo] password for maciek: 
c 1:3 rwm
c 1:7 rwm
c 1:5 rwm
c 1:8 rwm
c 1:9 rwm
c 5:0 rwm
c 5:1 rwm
c 5:2 rwm
c 136:* rwm
c 137:* rwm
c 138:* rwm
c 139:* rwm
c 140:* rwm
c 141:* rwm
c 142:* rwm
c 143:* rwm
c 10:239 rwm
c 226:0 rwm
c 0:0 rwm   <---
c 226:128 rwm
```

zyga:
Thank you for the review. I will iterate on this shortly.

cmd/snap-confine/udev-support.c
zyga: This is now fixed by 1353abe3bbe442860fe8837439996742d0d30398

cmd/snap-confine/udev-support.c
zyga: Done in 02b48d5645b0e4169dead186cb1b2a0a7e816b1e

bboozzoo:
Looked at the cmd/Makefile.a and check-unit-tests target needs an update, we can probably drop some tests in cmd/snap-confine/snap-device-helper-tests.c too.

zyga:
@bboozzoo thank you, I didn't think about looking there. I'll make sure it's correct.

zyga:
@bboozzoo I had a look at the code you mentioned and after some analysis I didn't find any code to alter or remove (apart from a typo). The remaining unit tests work and exercise the snap-device-helper script which is stil used by udev.

bboozzoo:
Yeah, forgot about udev use case.

cmd/snap-confine/udev-support.c
```diff
@@ -18,290 +18,355 @@
...
-	}
-	if (waitpid(pid, &status, 0) < 0)
-		die("waitpid failed");
-	if (WIFEXITED(status) && WEXITSTATUS(status) != 0)
-		die("child exited with status %i", WEXITSTATUS(status));
-	else if (WIFSIGNALED(status))
-		die("child died with signal %i", WTERMSIG(status));
+	/* The devices we add here have static number allocation.
+	 * https://www.kernel.org/doc/html/v4.11/admin-guide/devices.html */
+	dprintf(devices_allow_fd, "c 1:3 rwm");	// /dev/null
```
bboozzoo: Looked at `snap-device-helper` and back here, I think the current code is not entirely equivalent/ The `s-d-h` runs with `set -e`, and the C code that ran it called `die()` on any error. We probably need a wrapper that does more less this:

```
void dprintf_or_die(int fd, const char *format, ...) {
    va_list args;
    va_start(args, format);
    if (vdprintf(fd, format, args) == -1) {
        die("failed to write");
    }
    va_end(list);
}
```

zyga: Done

cmd/snap-confine/udev-support.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015-2016 Canonical Ltd
+ * Copyright (C) 2015-2019 Canonical Ltd
```
bboozzoo: 2020 now :smile: 
```suggestion
 * Copyright (C) 2015-2020 Canonical Ltd
```

cmd/snap-confine/udev-support.c
```diff
@@ -18,290 +18,368 @@
...
-			       "/usr/libexec/snapd/snap-device-helper", "add",
-			       udev_s->tagname, path, buf, NULL, env);
-		else
-			execle("/lib/udev/snappy-app-dev",
-			       "/lib/udev/snappy-app-dev", "add",
-			       udev_s->tagname, path, buf, NULL, env);
-		die("execl failed");
+	va_list ap;
+	va_start(ap, format);
+	if (vdprintf(fd, format, ap) == -1) {
```
bboozzoo: A scenario that I'm not sure can happen in practice, but what if we have short write?

Say, we write `c 5:2 rwm`, and it's cut short and we effectively write `c 5:2 rw`, then there's fewer permissions than we wanted, but since the setup denies everything upfront, and only `rw` gets enabled, it should be fine. Another scenario, if we write `c 5:`, which is incorrect, the kernel will return an error, so the operation fails. I suppose we could add a note here to describe the scenario that a short write which produced a valid string could go undetected, but the permission set would limited to those flags that were written.

OTOH, it's probably unlikely to get a short write.

cmd/snap-confine/udev-support.c
```diff
@@ -18,290 +18,368 @@
...
- * cleanup.
- */
-int snappy_udev_init(const char *security_tag, struct snappy_udev *udev_s)
+/** Allow access to current and future PTY slaves.
+ *
+ * We unconditionally add them since we use a devpts newinstance. Unix98 PTY
+ * slaves major are 136-143.
+ *
+ * See also:
+ * https://github.com/torvalds/linux/blob/master/Documentation/admin-guide/devices.txt
```
bboozzoo:
```suggestion
 * https://www.kernel.org/doc/Documentation/admin-guide/devices.txt
```

cmd/snap-confine/udev-support.c
```diff
@@ -18,290 +18,368 @@
...
+ **/
+static void sc_udev_allow_pty_slaves(int devices_allow_fd)
+{
+	for (unsigned pty_major = 136; pty_major <= 143; pty_major++) {
+		sc_dprintf(devices_allow_fd, "c %u:* rwm", pty_major);
+	}
+}
+
+/** Allow access to NVidia devices.
```
bboozzoo:
```suggestion
/** Allow access to Nvidia devices.
```

cmd/snap-confine/udev-support.c
```diff
@@ -18,290 +18,368 @@
...
+ *
+ * NVidia modules are proprietary and therefore aren't in sysfs and can't be
```
bboozzoo:
```suggestion
 * Nvidia modules are proprietary and therefore aren't in sysfs and can't be
```

cmd/snap-confine/udev-support.c
```diff
@@ -18,290 +18,368 @@
...
+ * udev tagged. For now, just add existing nvidia devices to the cgroup
+ * unconditionally (AppArmor will still mediate the access).  We'll want to
+ * rethink this if snapd needs to mediate access to other proprietary devices.
+ *
+ * Device major and minor numbers are described in (though nvidia-uvm currently
+ * isn't listed):
+ *
+ * https://github.com/torvalds/linux/blob/master/Documentation/admin-guide/devices.txt
```
bboozzoo:
```suggestion
 * https://www.kernel.org/doc/Documentation/admin-guide/devices.txt
```

cmd/snap-confine/udev-support.c
```diff
@@ -18,290 +18,368 @@
...
+	/* Allow access to various devices. */
+	sc_udev_allow_common(devices_allow_fd);
+	sc_udev_allow_pty_slaves(devices_allow_fd);
+	sc_udev_allow_nvidia(devices_allow_fd);
+	sc_udev_allow_uhid(devices_allow_fd);
+	sc_udev_allow_assigned(devices_allow_fd, udev, assigned);
 }
 
-void setup_devices_cgroup(const char *security_tag, struct snappy_udev *udev_s)
+static char *sc_udev_mangle_security_tag(const char *security_tag)
```
bboozzoo: How about `sc_security_to_udev_tag`?

cmd/snap-confine/udev-support.c
```diff
@@ -18,290 +18,368 @@
...
+		debug("cgroup v1 unavailable, skipping device cgroup setup");
+		return;
+	}
+	/* Setup the device group access control list */
+	sc_udev_setup_acls(fds.devices_allow_fd, fds.devices_deny_fd,
+			   udev, assigned);
+
+	/* Move ourselves to the device cgroup */
+	sc_dprintf(fds.cgroup_procs_fd, "%i", getpid());
+	debug("associated snap application process with device cgroup %s",
```
bboozzoo: Shall we include the pid here too?
```
debug("associated snap application process %ld with device cgroup %s", (long)getpid(), security_tag);
```

cmd/snap-confine/udev-support.h
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015-2016 Canonical Ltd
+ * Copyright (C) 2015-2019 Canonical Ltd
```
bboozzoo:
```suggestion
 * Copyright (C) 2015-2020 Canonical Ltd
```

zyga: But I wrote it in 2019 :)

cmd/snap-confine/udev-support.c
zyga: Sure, do you mind adding a suggestion? I can look later today as well.

cmd/snap-confine/udev-support.c
zyga: Sure, just add a suggestion please.

cmd/snap-confine/udev-support.c
zyga: I don't think it can happen in practice when writing to this kernel interface. The kernel scans the whole request, which must be a complete request line and must parse entirely.

jdstrand:
@zyga - I was expecting to see /usr/lib/snapd/snap-device-helper rewritten in C as part of that. Is that a separate PR?

zyga:
@jdstrand that is a separate PR. The current goal is to get 90% of the way there by making the code in C. The next steps reuse that code in the standalone binary and finally implement the v2 support.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,388 @@
...
-		// manner, we'll add /dev/nvidia0 and /dev/nvidia1 but stop
-		// trying to find nvidia3 - nvidia254 if nvidia2 is not found.
-		if (stat(nv_path, &sbuf) != 0) {
-			break;
+	/* Open /sys/fs/cgroup */
+	const char *cgroup_path = "/sys/fs/cgroup";
+	int SC_CLEANUP(sc_cleanup_close) cgroup_fd = -1;
+	cgroup_fd = open(cgroup_path,
+			 O_PATH | O_DIRECTORY | O_CLOEXEC | O_NOFOLLOW);
+	if (cgroup_fd < 0 && errno == ENOENT) {
```
anonymouse64: should this be 

```suggestion
	if (cgroup_fd < 0 || errno == ENOENT) {
```

or even just

```suggestion
	if (cgroup_fd < 0) {
```

? 
I'm a bit confused how we could ever get to the `die("cannot open ...)` code below due to the `&& errno ==ENOENT` and the subsequent `if (errno == ENOENT) {` inside that if.

zyga: Thanks for spotting this. I'll adjust this code to be less surprising.

zyga:
I need to change this after snap-confine has lost setgid bit, should be pretty simple but needs a small tweak.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
+	if (fds.cgroup_procs_fd < 0) {
+		debug("cgroup v1 unavailable, skipping device cgroup setup");
+		return;
+	}
+	/* Setup the device group access control list */
+	sc_udev_setup_acls(fds.devices_allow_fd, fds.devices_deny_fd,
+			   udev, assigned);
+
+	/* Move ourselves to the device cgroup */
+	sc_dprintf(fds.cgroup_procs_fd, "%i", getpid());
```
jdstrand, line 404: Note, this is a change in behavior. In the old code we moved ourselves into the device cgroup before modifying devices.deny and devices.allow. Is there a reason this isn't before sc_udev_setup_acls()?

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
-					       major(sbuf.st_rdev),
-					       minor(sbuf.st_rdev));
+	/* Everything worked so pack the result and "move" the descriptors over so
+	 * that they are not closed by the cleanup functions. */
+	fds.devices_allow_fd = devices_allow_fd;
+	fds.devices_deny_fd = devices_deny_fd;
+	fds.cgroup_procs_fd = cgroup_procs_fd;
+	devices_allow_fd = -1;
+	devices_deny_fd = -1;
+	cgroup_procs_fd = -1;
```
jdstrand, line 336: Putting these back to -1 is fine but unneeded since they are scoped to this function and you are immediately returning.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
+	return fds;
```
jdstrand, line 337: I can't help but feel we should be closing cgroup_fd, devices_fd and security_tag_fd for cleanliness. Sure, we have O_CLOEXEC, but that means these stay open a relatively long time when they don't have to.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
+		return;
+	}
+	/* Setup the device group access control list */
+	sc_udev_setup_acls(fds.devices_allow_fd, fds.devices_deny_fd,
+			   udev, assigned);
+
+	/* Move ourselves to the device cgroup */
+	sc_dprintf(fds.cgroup_procs_fd, "%i", getpid());
+	debug("associated snap application process with device cgroup %s",
+	      security_tag);
```
jdstrand, line 406: For cleanliness sake, can we close fds before returning?

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
+	/* The devices we add here have static number allocation.
+	 * https://www.kernel.org/doc/html/v4.11/admin-guide/devices.html */
+	sc_dprintf(devices_allow_fd, "c 1:3 rwm");	// /dev/null
+	sc_dprintf(devices_allow_fd, "c 1:5 rwm");	// /dev/zero
+	sc_dprintf(devices_allow_fd, "c 1:7 rwm");	// /dev/full
+	sc_dprintf(devices_allow_fd, "c 1:8 rwm");	// /dev/random
+	sc_dprintf(devices_allow_fd, "c 1:9 rwm");	// /dev/urandom
+	sc_dprintf(devices_allow_fd, "c 5:0 rwm");	// /dev/tty
+	sc_dprintf(devices_allow_fd, "c 5:1 rwm");	// /dev/console
+	sc_dprintf(devices_allow_fd, "c 5:2 rwm");	// /dev/ptmx
```
jdstrand, line 62: These major:minors are correct.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
+ * We unconditionally add them since we use a devpts newinstance. Unix98 PTY
+ * slaves major are 136-143.
+ *
+ * See also:
+ * https://www.kernel.org/doc/Documentation/admin-guide/devices.txt
+ **/
+static void sc_udev_allow_pty_slaves(int devices_allow_fd)
+{
+	for (unsigned pty_major = 136; pty_major <= 143; pty_major++) {
+		sc_dprintf(devices_allow_fd, "c %u:* rwm", pty_major);
```
jdstrand, line 76: This matches previous behavior.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
+		sc_dprintf(devices_allow_fd, "c %u:%u rwm", major(sbuf.st_rdev),
+			   minor(sbuf.st_rdev));
+	}
+	if (stat("/dev/nvidia-uvm", &sbuf) == 0) {
+		sc_dprintf(devices_allow_fd, "c %u:%u rwm", major(sbuf.st_rdev),
+			   minor(sbuf.st_rdev));
+	}
+	if (stat("/dev/nvidia-modeset", &sbuf) == 0) {
+		sc_dprintf(devices_allow_fd, "c %u:%u rwm", major(sbuf.st_rdev),
+			   minor(sbuf.st_rdev));
```
jdstrand, line 122: This matches previous behavior.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
-	udev_device_unref(d);
-
-	_run_snappy_app_dev_add_majmin(udev_s, path, major(devnum),
-				       minor(devnum));
+	struct stat sbuf;
+
+	if (stat("/dev/uhid", &sbuf) == 0) {
+		sc_dprintf(devices_allow_fd, "c %u:%u rwm", major(sbuf.st_rdev),
+			   minor(sbuf.st_rdev));
+	}
```
jdstrand, line 139: This matches previous behavior.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
-	udev_s->tagname_len = sc_must_snprintf(udev_s->tagname, MAX_BUF,
-					       "%s", security_tag);
-	for (size_t i = 0; i < udev_s->tagname_len; i++)
-		if (udev_s->tagname[i] == '.')
-			udev_s->tagname[i] = '_';
+	if (stat("/dev/net/tun", &sbuf) == 0) {
+		sc_dprintf(devices_allow_fd, "c %u:%u rwm", major(sbuf.st_rdev),
+			   minor(sbuf.st_rdev));
+	}
+}
```
jdstrand, line 159: This matches previous behavior.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
+		const char *path = udev_list_entry_get_name(entry);
+		if (path == NULL) {
+			die("udev_list_entry_get_name failed");
+		}
+		struct udev_device *device =
+		    udev_device_new_from_syspath(udev, path);
+		if (device == NULL) {
+			die("cannot find device from syspath %s", path);
+		}
+		dev_t devnum = udev_device_get_devnum(device);
```
jdstrand, line 182: `man udev_device_get_devnum` has this to say: "On success, udev_device_get_devnum() returns the device type of the passed device. On failure, a device type with minor and major number set to 0 is returned."

We should be checking if major(devnum) != 0 and minor(devnum) != 0. I suggest:
```
dev_t devnum = udev_device_get_devnum(device);
unsigned int major = major(devnum);
unsigned int minor = minor(devnum);
if (major != 0 || minor != 0) {
    char type_c = strstr(path, "/block/") != NULL ? 'b' : 'c';
    sc_dprintf(devices_allow_fd, "%c %u:%u rwm", type_c, major(devnum), minor(devnum));
}
udev_device_unref(device);
```

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
+		char type_c = strstr(path, "/block/") != NULL ? 'b' : 'c';
```
jdstrand, line 183: Nitpick: `type_c` makes me think of 'c' as in 'character device'. Perhaps another name for the variable?

Not to mention, this doesn't contain the expected S_IFBLK/S_IFCHR switch from https://github.com/snapcore/snapd/commit/02b48d5645b0e4169dead186cb1b2a0a7e816b1e

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
+		debug("no devices tagged with %s, skipping device cgroup setup",
+		      udev_tag);
+		return;
 	}
+
+	sc_cgroup_fds SC_CLEANUP(sc_cleanup_cgroup_fds) fds = { -1, -1, -1 };
+	fds = sc_udev_open_cgroup_v1(security_tag);
+	if (fds.cgroup_procs_fd < 0) {
+		debug("cgroup v1 unavailable, skipping device cgroup setup");
+		return;
```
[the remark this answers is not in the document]
zyga: This is expected. If you are on a v2 system you are in forced partial confinement.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
+	security_tag_fd = openat(devices_fd, security_tag_relpath,
+				 O_RDONLY | O_DIRECTORY | O_CLOEXEC |
+				 O_NOFOLLOW);
+	if (security_tag_fd < 0) {
+		die("cannot open %s/%s/%s", cgroup_path, devices_relpath,
+		    security_tag_relpath);
+	}
+	if (fchown(security_tag_fd, 0, 0) < 0) {
+		die("cannot chown %s/%s/%s to root:root", cgroup_path,
+		    devices_relpath, security_tag_relpath);
```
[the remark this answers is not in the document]
zyga: Yeah, leftover from before we had that code. I'll drop it.

cmd/snap-confine/udev-support.c
zyga: It is necessary. Please observe that each of the non-structure variables assigned above has a cleanup function so without resetting them to a safe value of -1, the return value would return stale descriptors because of the compiler-guaranteed call to `close()`

cmd/snap-confine/udev-support.c
zyga: We always close them: ` 	sc_cgroup_fds SC_CLEANUP(sc_cleanup_cgroup_fds) fds = { -1, -1, -1 };`

cmd/snap-confine/udev-support.c
zyga: We are closing them. Please see my comment below. The return value returned here is closed with a cleanup function.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
+
+	/* Open devices relative to /sys/fs/cgroup */
+	const char *devices_relpath = "devices";
+	int SC_CLEANUP(sc_cleanup_close) devices_fd = -1;
+	devices_fd = openat(cgroup_fd, devices_relpath,
+			    O_PATH | O_DIRECTORY | O_CLOEXEC | O_NOFOLLOW);
+	if (devices_fd < 0) {
+		if (errno == ENOENT) {
+			/* This system does not support the device cgroup. */
+			return fds;
```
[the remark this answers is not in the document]
zyga: The point is to be able to support v2 mode so we cannot just die outright when we're not in v1 mode. Note that _today_ snap-confine on v2 system just doesn't do anything in this code. This is not changing that, we are _not_ die()ing on v2 systems today.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
+
+	/* Open /sys/fs/cgroup */
+	const char *cgroup_path = "/sys/fs/cgroup";
+	int SC_CLEANUP(sc_cleanup_close) cgroup_fd = -1;
+	cgroup_fd = open(cgroup_path,
+			 O_PATH | O_DIRECTORY | O_CLOEXEC | O_NOFOLLOW);
+	if (cgroup_fd < 0) {
+		if (errno == ENOENT) {
+			/* This system does not support cgroups. */
+			return fds;
```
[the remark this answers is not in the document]
zyga: See my comment below about v2 mode.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
-	udev_s->assigned = udev_enumerate_get_list_entry(udev_s->devices);
-	if (udev_s->assigned == NULL)
-		rc = -1;
+static char *sc_udev_mangle_security_tag(const char *security_tag)
+{
+	char *udev_tag = sc_strdup(security_tag);
+	for (char *c = strchr(udev_tag, '.'); c != NULL; c = strchr(c, '.')) {
+		*c = '_';
+	}
+	return udev_tag;
```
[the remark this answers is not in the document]
zyga: I'll look into this.

cmd/snap-confine/udev-support.c
zyga: Applied

cmd/snap-confine/udev-support.c
zyga: Done. Thanks for spotting this. The cost of leaving a branch open for so long.

cmd/snap-confine/udev-support.c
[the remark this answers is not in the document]
zyga: Done

cmd/snap-confine/udev-support.c
zyga: It felt more correct to move a process only after completing the operations.

cmd/snap-confine/udev-support.c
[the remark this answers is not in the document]
zyga: Done

cmd/snap-confine/udev-support.c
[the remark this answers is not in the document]
zyga: I didn't read the diff right.

I did have a point here but I'd rather iterate on something rather than discuss it without making progress in the code. I will revert to die() and then open a follow-up that explains my point clearly.

This does make sense in an environment where you don't have cgroups at all, but also in that mode you don't have apparmor or other confinement.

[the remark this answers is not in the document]
zyga: This is done

cmd/snap-confine/udev-support.c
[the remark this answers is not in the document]
zyga: I changed my mind, this is done now.

cmd/snap-confine/udev-support.c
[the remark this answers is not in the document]
zyga: This is done now.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
+{
+	if (sc_cgroup_is_v2()) {
+		return;
 	}
-	// /dev/nvidia-uvm
-	if (stat(nvuvm_path, &sbuf) == 0) {
-		_run_snappy_app_dev_add_majmin(udev_s, nvuvm_path,
-					       major(sbuf.st_rdev),
-					       minor(sbuf.st_rdev));
+
```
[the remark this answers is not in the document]
zyga: I added a debug message now.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,390 @@
...
-		run_snappy_app_dev_add(udev_s, path);
-		udev_s->assigned = udev_list_entry_get_next(udev_s->assigned);
+	/* NOTE: udev_list_entry is bound to life-cycle of the used udev_enumerate */
+	struct udev_list_entry *assigned;
+	assigned = udev_enumerate_get_list_entry(devices);
+	if (assigned == NULL) {
+		/* NOTE: Nothing is assigned, don't create or use the device cgroup. */
+		debug("no devices tagged with %s, skipping device cgroup setup",
+		      udev_tag);
+		return;
```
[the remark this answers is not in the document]
zyga: The manual page says:
>        On success, udev_enumerate_get_list_entry() returns a pointer to the first entry in the list of found devices. If the list is empty, or on failure, NULL is returned.

Unfortunately two distinct cases are bundled into one: no devices tagged or unspecified error.
Do you suppose there's something we should change here? I think the code is correct as-is.

zyga:
@jdstrand can you please re-review this. I think we are close and it could land today.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,403 @@
...
+		}
+		struct udev_device *device =
+		    udev_device_new_from_syspath(udev, path);
+		if (device == NULL) {
+			die("cannot find device from syspath %s", path);
+		}
+		dev_t devnum = udev_device_get_devnum(device);
+		unsigned int major = major(devnum);
+		unsigned int minor = minor(devnum);
+		if (major != 0 || minor != 0) {
```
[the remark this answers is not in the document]
zyga: Yeah, think you are right. I'll send a patch.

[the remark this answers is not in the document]
zyga: Done

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,403 @@
...
-		die("snappy_udev->tagname has invalid length");
-
-	// create devices cgroup controller
-	char cgroup_dir[PATH_MAX] = { 0 };
-
-	sc_must_snprintf(cgroup_dir, sizeof(cgroup_dir),
-			 "/sys/fs/cgroup/devices/%s/", security_tag);
+	if (enumerate != NULL && *enumerate != NULL) {
+		udev_enumerate_unref(*enumerate);
+		*enumerate = NULL;
```
jdstrand, line 255: The man page for `udev_enumerate_new` doesn't mention anything about opaqueness like the one for `udev_new` did, but the udev_enumerate_unref() should be sufficient AIUI (which is why the old code also didn't set the pointer to NULL. Unless this is fixing an observed problem (eg, are you doing this cause of valgrind?), I think I'd prefer to stay with the old method (I'm not sure that setting this to NULL isn't breaking the opaqueness rule of what is returned by udev_new since in order to allocate the udev_enumerate, we must call udev_enumerate_new() passing in the udev struct from udev_new()).

cmd/snap-confine/udev-support.c
jdstrand: Note that snap-device-helper will write in a trailing newline for each item written, but this does not and none of the callers add `\n`. The upstream documentation in https://www.kernel.org/doc/Documentation/cgroup-v1/devices.txt uses `echo` in its example (like snap-device-helper does). This is probably ok (it certainly is on my 5.4 kernel), but it is a change over the current behavior. I think it would be much easier to add the newline in our code rather then verifying omitting it is safe over all kernels snapd might run on.

To @bboozzoo's point, it is never wrong to verify things are correct and I would prefer we check for a short write. Please note that snap-device-helper has `set -e`. OTOH I'm not sure what dash will do with a short write via `echo "$acl" > "$app_dev_cgroup/devices.allow"` when `set -e` is in use, but it sure seems plausible it would exit with error, and the old `_run_snappy_app_dev_add_majmin()` would `die()` in this case.

Lastly, honestly, I think that since sc_dprintf() is local to this file, why not simplify it to just be:
```
// sc_dprintf() writes the string (adding a trailing newline) to a file descriptor
static void sc_dprintf(int fd, const char *s)
{
	int n = dprintf(fd, "%s\n", s);
	if (n < 0 || n < strlen(s) + 1) {
		die("cannot write to fd %d", fd);
	}
}
```

If at some future date you want it to conditionally add the newline, just add a third arg.

jdstrand (dismissed):
Thanks for the updates! This is really close. See inline.

cmd/snap-confine/udev-support.c
zyga: Sure, I'll handle both.

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,403 @@
...
+			die("udev_list_entry_get_name failed");
+		}
+		struct udev_device *device =
+		    udev_device_new_from_syspath(udev, path);
+		if (device == NULL) {
+			die("cannot find device from syspath %s", path);
+		}
+		dev_t devnum = udev_device_get_devnum(device);
+		unsigned int major = major(devnum);
+		if (major != 0) {
```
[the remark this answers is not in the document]
zyga: Done

zyga:
I will return to this PR this week as it came up in a related bug in snap-confine handling of udev.

cmd/snap-confine/udev-support.c
zyga: I've hardened `sc_dprintf` but I dropped the strlen check. I think it's just the wrong way to implement this. Think about `%s` and empty string argument. Or about `%u` called with zero.

A while ago I implemented a full set of APIs that allow us to printf to open file descriptors, with nicer error checking but I never proposed that as it was a bit too NIH for snap-confine. I think the check we have now is enough.

cmd/snap-confine/udev-support.c
zyga: Done

cmd/snap-confine/udev-support.c
zyga: Done

cmd/snap-confine/udev-support.c
```diff
@@ -18,301 +18,403 @@
...
 {
-	// udev_s->assigned does not need to be unreferenced since it is a
-	// pointer into udev_s->devices
-	if (udev_s->devices != NULL)
-		udev_enumerate_unref(udev_s->devices);
-	if (udev_s->udev != NULL)
-		udev_unref(udev_s->udev);
+	if (udev != NULL && *udev != NULL) {
+		udev_unref(*udev);
+		*udev = NULL;
```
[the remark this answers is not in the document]
zyga: Because we're not accessing that pointer. This is a cleanup handler that gets called through a double pointer. This tactic is used in many other places and gives us some extra sanity in case we somehow cleanup twice (because then this is a no-op).

cmd/snap-confine/udev-support.c
zyga: My comment above for `sc_cleanup_udev` stands. This is the double pointer cleanup function.

cmd/snap-confine/udev-support.c
zyga: Done :)

cmd/snap-confine/udev-support.c
zyga: I had an idea how to harden this function without involving the format string length. Have a look please. :-)

cmd/snap-confine/udev-support.c
alexmurray: From what I can see I concur with @zyga here - the cleanup function will run on scope exit (ie when this function returns) at which point it will `close()` these fds if they are not already -1 - in this case, we don't want them to get closed since they are now 'owned' by the caller within the fds struct so setting them to -1 ensures they remain open for the caller to close later.

alexmurray (approved):
@jdstrand asked that I do the security review for this PR - LGTM - thanks @zyga .

cmd/snap-confine/udev-support.c
bboozzoo: Perhaps that warrants a comment before `devices_allow_fd = -1` for the benefit of anyone who ends up wondering about this piece of code during future reviews/refactors.

bboozzoo (approved):
Suggested adding a comment about a part that clearly caused some confusion during reviews. Otherwise LGTM.

zyga:
Removed the "needs security review" because Alex +1 this.

cmd/snap-confine/udev-support.c
```diff
@@ -18,317 +18,436 @@
...
 #include "../libsnap-confine-private/utils.h"
 #include "udev-support.h"
 
-static void
-_run_snappy_app_dev_add_majmin(struct snappy_udev *udev_s,
-			       const char *path, unsigned major, unsigned minor)
+__attribute__((format(printf, 2, 3)))
+static void sc_dprintf(int fd, const char *format, ...);
+
+static void sc_dprintf(int fd, const char *format, ...)
```
anonymouse64, line 40: /me idly notes that perhaps the udev-support file is not the best place for this generic utility function given that we already have a `utils.h`, but that's a nitpick, better to land this PR before it's first birthday and handle in followups

cmd/snap-confine/udev-support.c
```diff
@@ -18,317 +18,436 @@
...
-		die("execl failed");
+	va_end(ap2);
+	va_end(ap1);
+}
+
+/* Allow access to common devices. */
+static void sc_udev_allow_common(int devices_allow_fd)
+{
+	/* The devices we add here have static number allocation.
+	 * https://www.kernel.org/doc/html/v4.11/admin-guide/devices.html */
```
anonymouse64, line 61: just a note that the other comments reference https://www.kernel.org/doc/Documentation/admin-guide/devices.txt while this one is a different link, would be kinda nice if they all referenced the same link, I have no strong preference for which link we choose just that we are consistent

again for a followup

anonymouse64 (approved):
A couple small nitpicks that should be followup material I think, but otherwise lgtm thanks!

zyga:
I discussed this with @mvo5 today and it should land after core20 is released.

cmd/snap-confine/udev-support.c
zyga: This is here for a reason but that has been lot to time. This is the current version, in v2 support mode this changes to a more complex machinery that is still local to this file. I would keep it here for now, as it is a step along the path.

mvo5:
I checked with @bboozzoo that it already has a security review so let's merge it (given that it also has two plus +1).

PR #7615, opened 2019-10-16
policy: implement CanRemove policy for the snapd type

mvo5:
This PR implements a CanRemove policy for the snapd snap type
so that this snap can only be removed if its safe to do so.

The policy is:
- if on core devices: never allow removal
- if on classic: allow removal if snapd is the last snap

### DIFF
--- a/overlord/snapstate/policy/errors.go
+++ b/overlord/snapstate/policy/errors.go
@@ -30,6 +30,7 @@ var (
 	errInUseForBoot = errors.New("snap is being used for boot")
 	errRequired     = errors.New("snap is required")
 	errIsModel      = errors.New("snap is used by the model")
+	errNotRemovable = errors.New("snap cannot be removed")
 )
 
 type inUseByErr []string
--- a/overlord/snapstate/policy/errors.go
+++ b/overlord/snapstate/policy/errors.go
@@ -30,7 +30,9 @@ var (
 	errInUseForBoot = errors.New("snap is being used for boot")
 	errRequired     = errors.New("snap is required")
 	errIsModel      = errors.New("snap is used by the model")
-	errNotRemovable = errors.New("snap cannot be removed")
+
+	errSnapdNotRemovableOnCore       = errors.New("snapd required on core devices")
+	errSnapdNotYetRemovableOnClassic = errors.New("remove all other snaps first")
 )
 
 type inUseByErr []string
--- a/overlord/snapstate/policy/snapd.go
+++ b/overlord/snapstate/policy/snapd.go
@@ -17,46 +17,42 @@
  *
  */
 
-// Package policy implements fine grained decision-making for snapstate
 package policy
 
 import (
-	"github.com/snapcore/snapd/asserts"
 	"github.com/snapcore/snapd/overlord/snapstate"
 	"github.com/snapcore/snapd/overlord/state"
+	"github.com/snapcore/snapd/release"
 	"github.com/snapcore/snapd/snap"
 )
 
-func init() {
-	snapstate.PolicyFor = For
+type snapdPolicy struct {
+	modelBase string
 }
 
-func For(typ snap.Type, model *asserts.Model) snapstate.Policy {
-	switch typ {
-	case snap.TypeKernel:
-		return &kernelPolicy{modelKernel: model.Kernel()}
-	case snap.TypeGadget:
-		return &gadgetPolicy{modelGadget: model.Gadget()}
-	case snap.TypeOS:
-		return &osPolicy{modelBase: model.Base()}
-	case snap.TypeBase:
-		return &basePolicy{modelBase: model.Base()}
-	case snap.TypeSnapd:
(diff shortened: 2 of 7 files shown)


overlord/snapstate/policy/errors.go
```diff
@@ -30,6 +30,7 @@ var (
 	errInUseForBoot = errors.New("snap is being used for boot")
 	errRequired     = errors.New("snap is required")
 	errIsModel      = errors.New("snap is used by the model")
+	errNotRemovable = errors.New("snap cannot be removed")
```
chipaca: this is not a good error message: it will result in «snap "snapd" is not removable: snap cannot be removed».
Granted "snap is required" isn't much better (if you don't know what required means in this context).

On core, I'd go maybe with saying it is "essential"; on classic, I'd mention that it needs to be the last thing to go?

overlord/snapstate/policy/snapd.go
```diff
@@ -0,0 +1,59 @@
...
+		// not installed, or something. What are you even trying to do.
+		return errNoName
+	}
+
+	if !rev.Unset() {
+		return nil
+	}
+
+	// snapd cannot be removed on core
+	if !release.OnClassic {
```
chipaca: hmm, wouldn't using `model.Classic()` make more sense here?

overlord/snapstate/policy/snapd.go
```diff
@@ -0,0 +1,59 @@
...
+		return errNotRemovable
+	}
+
+	// only allow snapd removal if its the last snap on a (classic) system
+	numSnaps, err := snapstate.NumSnaps(st)
```
chipaca: _can_ you get there, on classic? You'd have to be able to remove the base first... and the model base is not removable...?

mvo5: Yes, classic is fine with this. It is possible to remove bases if they are unused. We disallow to rmeove core but that could be considered a bug I think, we should probably allow it on classic just like we do for bases.

overlord/snapstate/policy/snapd.go
mvo5: Maybe, it should be equivalent - or am I missing something?

overlord/snapstate/policy/errors.go
mvo5: I updated the code now:
+       errSnapdNotRemovableOnCore       = errors.New("snapd required on core devices")
+       errSnapdNotYetRemovableOnClassic = errors.New("remove all other snaps first")
happy about suggestions of course :)

overlord/snapstate/policy/snapd.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"github.com/snapcore/snapd/overlord/snapstate"
+	"github.com/snapcore/snapd/overlord/state"
+	"github.com/snapcore/snapd/release"
+	"github.com/snapcore/snapd/snap"
+)
+
+type snapdPolicy struct {
+	modelBase string
```
chipaca: I don't think snapdPolicy needs to know the model base

overlord/snapstate/policy/snapd.go
chipaca: I think it fits better with policy if you grab classic from the model, as we do with base in other policies.
I'm not sure there's a stronger reason than that, as I don't think you can remodel from classic to non-, but the fact that I need to ask myself this points you at why it's saner to get it from the model that was passed in...

overlord/snapstate/policy/snapd.go
chipaca: I just double-checked and indeed it all works.
yeah, agreed about it being a bug for core, we can probably fix that soon though
we need to find a way to restart snapd after removing the snapd snap, probably, but that's not for this PR

overlord/snapstate/policy/snapd.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func (p *snapdPolicy) CanRemove(st *state.State, snapst *snapstate.SnapState, rev snap.Revision) error {
+	name := snapst.InstanceName()
+	if name == "" {
+		// not installed, or something. What are you even trying to do.
+		return errNoName
+	}
+
+	if !rev.Unset() {
```
zyga: Hmmm

so if revision _is_ set then you can remove the snap? I must be missing something but this feels wrong.

overlord/snapstate/policy/snapd.go
mvo5: Yeah, this was from an earlier iteration. Updated the code now.

overlord/snapstate/policy/snapd.go
mvo5: This is done like this in all removal policies, this means that a non-current snap revision is removed.

zyga: Can you add a comment to the code please.

chipaca: I'll add them in a separate pass, if this is green and GTG

PR #7616, opened 2019-10-16
interfaces/many: allow k8s/systemd-run to mount volume subPaths plus cleanups

jdstrand:
* interfaces/kubernetes-support: allow systemd-run to mount volume subPaths
* interfaces/kubernetes-support: update to use SNAP_INSTANCE_NAME consistently
* apparmor: add @{SNAP_COMMAND_NAME} variable
* interfaces/kubernetes-support: use @{SNAP_COMMAND_NAME} with systemd-run

### DIFF
--- a/interfaces/apparmor/backend_test.go
+++ b/interfaces/apparmor/backend_test.go
@@ -688,6 +688,7 @@ const commonPrefix = `
 @{SNAP_NAME}="samba"
 # This is a snap name with instance key
 @{SNAP_INSTANCE_NAME}="samba"
+@{SNAP_COMMAND_NAME}="smbd"
 @{SNAP_REVISION}="1"
 @{PROFILE_DBUS}="snap_2esamba_2esmbd"
 @{INSTALL_DIR}="/{,var/lib/snapd/}snap"`
@@ -845,6 +846,7 @@ func (s *backendSuite) TestParallelInstallCombineSnippets(c *C) {
 @{SNAP_NAME}="samba"
 # This is a snap name with instance key
 @{SNAP_INSTANCE_NAME}="samba_foo"
+@{SNAP_COMMAND_NAME}="smbd"
 @{SNAP_REVISION}="1"
 @{PROFILE_DBUS}="snap_2esamba_5ffoo_2esmbd"
 @{INSTALL_DIR}="/{,var/lib/snapd/}snap"
--- a/interfaces/apparmor/backend_test.go
+++ b/interfaces/apparmor/backend_test.go
@@ -864,6 +864,44 @@ profile "snap.samba_foo.smbd" (attach_disconnected,mediate_deleted) {
 	s.RemoveSnap(c, snapInfo)
 }
 
+func (s *backendSuite) TestTemplateVarsWithHook(c *C) {
+	restore := apparmor_sandbox.MockLevel(apparmor_sandbox.Full)
+	defer restore()
+	restore = apparmor.MockIsHomeUsingNFS(func() (bool, error) { return false, nil })
+	defer restore()
+	restore = apparmor.MockIsRootWritableOverlay(func() (string, error) { return "", nil })
+	defer restore()
+	// NOTE: replace the real template with a shorter variant
+	restoreTemplate := apparmor.MockTemplate("\n" +
+		"###VAR###\n" +
+		"###PROFILEATTACH### (attach_disconnected,mediate_deleted) {\n" +
+		"###SNIPPETS###\n" +
+		"}\n")
+	defer restoreTemplate()
+
+	expected := `
+# This is a snap name without the instance key
+@{SNAP_NAME}="foo"
+# This is a snap name with instance key
+@{SNAP_INSTANCE_NAME}="foo"
+@{SNAP_COMMAND_NAME}="hook.configure"
+@{SNAP_REVISION}="1"
+@{PROFILE_DBUS}="snap_2efoo_2ehook_2econfigure"
+@{INSTALL_DIR}="/{,var/lib/snapd/}snap"
+profile "snap.foo.hook.configure" (attach_disconnected,mediate_deleted) {
+
+}
+`
+	snapInfo := s.InstallSnap(c, interfaces.ConfinementOptions{}, "", ifacetest.HookYaml, 1)
+	c.Assert(snapInfo, NotNil)
+	profile := filepath.Join(dirs.SnapAppArmorDir, "snap.foo.hook.configure")
+	stat, err := os.Stat(profile)
+	c.Assert(err, IsNil)
+	c.Check(profile, testutil.FileEquals, expected)
+	c.Check(stat.Mode(), Equals, os.FileMode(0644))
+	s.RemoveSnap(c, snapInfo)
+}
+
 func mockPartalAppArmorOnDistro(c *C, kernelVersion string, releaseID string, releaseIDLike ...string) (restore func()) {
 	restore1 := apparmor_sandbox.MockLevel(apparmor_sandbox.Partial)
 	restore2 := release.MockReleaseInfo(&release.OS{ID: releaseID, IDLike: releaseIDLike})
--- a/interfaces/apparmor/template_vars.go
+++ b/interfaces/apparmor/template_vars.go
@@ -27,14 +27,15 @@ import (
 	"github.com/snapcore/snapd/snap"
 )
 
-// templateVariables returns text defining apparmor variables that can be used in the
-// apparmor template and by apparmor snippets.
-func templateVariables(info *snap.Info, securityTag string) string {
+// templateVariables returns text defining apparmor variables that can be used
+// in the apparmor template and by apparmor snippets.
+func templateVariables(info *snap.Info, securityTag string, cmdName string) string {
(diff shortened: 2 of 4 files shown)


interfaces/builtin/kubernetes_support.go
```diff
@@ -148,6 +148,17 @@ const kubernetesSupportConnectedPlugAppArmorKubeletSystemdRun = `
   /{,usr/}bin/mount ixr,
   mount fstype="tmpfs" tmpfs -> /var/snap/@{SNAP_INSTANCE_NAME}/common/**,
   deny /run/mount/utab rw,
+
+  # For mounting volume subPaths
+  mount /var/snap/@{SNAP_INSTANCE_NAME}/common/{,**} -> /var/snap/@{SNAP_INSTANCE_NAME}/common/{,**},
```
anonymouse64: do we want to specifically disallow `$SNAP_DATA`? what if someday kubernetes changes to use `$SNAP_DATA` instead of `$SNAP_COMMON` for these volumes?

interfaces/builtin/kubernetes_support.go
```diff
@@ -148,6 +148,17 @@ const kubernetesSupportConnectedPlugAppArmorKubeletSystemdRun = `
...
+  mount options=(rw, remount, bind) -> /var/snap/@{SNAP_INSTANCE_NAME}/common/{,**},
+  umount /var/snap/@{SNAP_INSTANCE_NAME}/common/**,
+  # When mounting a volume subPath, kubelet binds mounts on an open fd (eg,
+  # /proc/.../fd/N) which triggers a ptrace 'trace' denial on the parent
+  # kubelet peer process from this child profile. Note, this child profile
+  # doesn't have 'capability sys_ptrace', so systemd-run is still not able to
+  # ptrace this snap's processes.
+  ptrace (trace) peer=snap.@{SNAP_INSTANCE_NAME}.@{SNAP_COMMAND_NAME},
```
anonymouse64: nice, I've thought to myself before if there was a way to do this

jdstrand: Yeah, the builtin `@{profile_name}` is very handy, but within the context of the child profile that expands to `snap.microk8s.kubelet//systemd_run`. There isn't another builtin for the parent atm so have to resort to this.

interfaces/builtin/kubernetes_support.go
jdstrand: I'm happy to add another commit to for @{SNAP_REVISION}. Let me do that now.

anonymouse64: probably fine as is then, IIRC for greengrass we made it more open allowing `$SNAP_DATA` or `$SNAP_COMMON`, but that's a bit different as that was a closed source app which could change at any time. k8s is probably more predictable and less likely to change randomly

anonymouse64: we responded at the same time, either is fine with me, if you already have it ready for snap_revision that's fine, otherwise without is fine too :smile:

jdstrand: @anonymouse64 - I added it. Let me know if after reading the commit message you prefer me to revert.

interfaces/builtin/kubernetes_support.go
```diff
@@ -133,21 +133,32 @@ deny ptrace (trace) peer=unconfined,
...
-umount /var/snap/@{SNAP_INSTANCE_NAME}/common/**,
+umount /var/snap/@{SNAP_INSTANCE_NAME}/*/**,
 `
 
 const kubernetesSupportConnectedPlugAppArmorKubeletSystemdRun = `
   # kubelet mount rules
   capability sys_admin,
   /{,usr/}bin/mount ixr,
-  mount fstype="tmpfs" tmpfs -> /var/snap/@{SNAP_INSTANCE_NAME}/common/**,
+  mount fstype="tmpfs" tmpfs -> /var/snap/@{SNAP_INSTANCE_NAME}/*/**,
```
anonymouse64: hmm actually I'm not sure that we want `*` here instead of `common`, wouldn't that allow a snap revision to mount stuff on a different revision? i.e. `mount --bind /var/snap/k8s/4 /var/snap/k8s/10`

I was just thinking something like 
```
  mount fstype="tmpfs" tmpfs -> /var/snap/@{SNAP_INSTANCE_NAME}/{common,@{SNAP_REVISION}}/**,
```

interfaces/builtin/kubernetes_support.go
```diff
@@ -133,26 +133,26 @@ deny ptrace (trace) peer=unconfined,
...
   deny /run/mount/utab rw,
 
   # For mounting volume subPaths
-  mount /var/snap/@{SNAP_INSTANCE_NAME}/common/{,**} -> /var/snap/@{SNAP_INSTANCE_NAME}/common/{,**},
-  mount options=(rw, remount, bind) -> /var/snap/@{SNAP_INSTANCE_NAME}/common/{,**},
-  umount /var/snap/@{SNAP_INSTANCE_NAME}/common/**,
+  mount /var/snap/@{SNAP_INSTANCE_NAME}/*/{,**} -> /var/snap/@{SNAP_INSTANCE_NAME}/*/{,**},
+  mount options=(rw, remount, bind) -> /var/snap/@{SNAP_INSTANCE_NAME}/*/{,**},
+  umount /var/snap/@{SNAP_INSTANCE_NAME}/*/**,
```
anonymouse64: again this seems like it would allow an individual revision to modify or do things to a different revision of the snap

interfaces/builtin/kubernetes_support.go
jdstrand: After discussion on IRC, I reverted it. The revert's commit message gives the rationale.

interfaces/builtin/kubernetes_support.go
jdstrand: Yes, but that was intentional considering 'refresh-mode: skip-refresh'. I've reverted it per IRC conversation (see the revert's commit message).

interfaces/builtin/kubernetes_support.go
jdstrand: Yes, see previous comments.

jdstrand:
> perhaps I'm being overly specific here considering that this is a super-privileged interface, but this new commit isn't quite what I expected to support `$SNAP_DATA` and `$SNAP_COMMON`.
> 
> There seems to be a leak between revisions here where say revision 4 could mount stuff on top of revision 5.

Reverted (see revert commit message for rationale for the intended behavior change of the glob rule and why it was reverted).

interfaces/apparmor/backend_test.go
```diff
@@ -688,6 +688,7 @@ const commonPrefix = `
 @{SNAP_NAME}="samba"
 # This is a snap name with instance key
 @{SNAP_INSTANCE_NAME}="samba"
+@{SNAP_COMMAND_NAME}="smbd"
```
zyga: What is the value for hooks?

I like the idea but I wonder if instead we should simply have $SNAP_SECURITY_TAG as a generalization of the concept. I don't mind landing this if we intend to iterate as this is an entirely internal variable.

interfaces/apparmor/template_vars.go
```diff
@@ -27,14 +27,15 @@ import (
...
-func templateVariables(info *snap.Info, securityTag string) string {
+// templateVariables returns text defining apparmor variables that can be used
+// in the apparmor template and by apparmor snippets.
+func templateVariables(info *snap.Info, securityTag string, cmdName string) string {
 	var buf bytes.Buffer
 	fmt.Fprintf(&buf, "# This is a snap name without the instance key\n")
 	fmt.Fprintf(&buf, "@{SNAP_NAME}=\"%s\"\n", info.SnapName())
 	fmt.Fprintf(&buf, "# This is a snap name with instance key\n")
 	fmt.Fprintf(&buf, "@{SNAP_INSTANCE_NAME}=\"%s\"\n", info.InstanceName())
+	fmt.Fprintf(&buf, "@{SNAP_COMMAND_NAME}=\"%s\"\n", cmdName)
```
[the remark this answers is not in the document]
jdstrand: Oh, yes, great catch. Adjusting.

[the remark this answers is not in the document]
jdstrand: Done

interfaces/apparmor/backend_test.go
jdstrand: I implemented it differently than `@{SNAP_SECURITY_TAG}` because unless you are in a child profile, `@{profile_name}` will be the same as `@{SNAP_SECURITY_TAG}`. Because `@{profile_name}` should always be used instead of `@{SNAP_SECURITY_TAG}` outside of a child profile and is almost always the correct choice when in a child profile, the subtlety of when to use `@{SNAP_SECURITY_TAG}` correctly could lead to bugs or hinder readability.

Instead, I continue to use `@{SNAP_COMMAND_NAME}` but prepend `hook.` in addContent(). I think `@{SNAP_COMMAND_NAME}` still works as a variable name, but let me know if there is something better and I can change it.

jdstrand:
Thanks for the reviews!

PR #7617, opened 2019-10-16
snap-confine.apparmor.in: harden pivot_root until we have full mediation

jdstrand:
snap-confine uses pivot_root() (man 2 pivot_root) only when initially
setting up the mount namespace (eg, on first start of any snap's command
or after a snap-discard-ns) and it does so in a very predictable manner,
which allows the snap-confine profile to be more specific with both
new_root and put_old (oldroot) than the current bare pivot_root rule.

While this does not address LP: #1791711, it does provide a bit of
hardening until full pivot_root mediation is available in AppArmor.

References:
- https://bugs.launchpad.net/apparmor/+bug/1791711


jdstrand:
Thanks for the reviews!

PR #7620, opened 2019-10-17
snap-install: add ext4,vfat creation support

mvo5:
We'll need ext4 filesystem creation for ubuntu-data, so add that here.
Note that ubuntu-data will also require more infrastructure to be
properly created, such as cryptsetup and key sealing support.

### DIFF
--- a/cmd/snap-recovery/partition/mkfs.go
+++ b/cmd/snap-recovery/partition/mkfs.go
@@ -18,13 +18,19 @@
  */
 package partition
 
-type LsblkFilesystemInfo = lsblkFilesystemInfo
-type LsblkBlockDevice = lsblkBlockDevice
-type SFDiskPartitionTable = sfdiskPartitionTable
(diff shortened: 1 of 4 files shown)


cmd/snap-recovery/partition/mkfs.go
```diff
@@ -0,0 +1,36 @@
...
+ */
+package partition
+
+import (
+	"fmt"
+
+	"github.com/snapcore/snapd/gadget"
+)
+
+func makeFilesystem(node, label, filesystem, content string) error {
```
zyga: Given that there are four strings here how about using a distinct type for the filesystem type argument?
From the looks of the function alone I don't know what the content argument is. Perhaps a docstring would help here.

mvo5: Yeah, that makes sense, I will add the docstring in a followup to avoid another full spread run.

PR #7629, opened 2019-10-18
snap-recovery: create filesystems as defined in the gadget

mvo5:
This adds filesystem creation to the snap-recovery tool.

Based on https://github.com/snapcore/snapd/pull/7613

### DIFF
--- a/cmd/snap-recovery/main.go
+++ b/cmd/snap-recovery/main.go
@@ -42,12 +42,7 @@ func run(args []string) error {
 	device := os.Args[2]
 	options := &recover.Options{}
 
-	recov := recover.New(gadgetRoot, device, options)
-	if err := recov.Run(); err != nil {
-		return err
-	}
-
-	return nil
+	return recover.Run(gadgetRoot, device, options)
 }
 
 func main() {
--- a/cmd/snap-recovery/partition/mkfs.go
+++ b/cmd/snap-recovery/partition/mkfs.go
@@ -24,10 +24,10 @@ import (
 	"github.com/snapcore/snapd/gadget"
 )
 
-func MakeFilesystems(created map[string]gadget.LaidOutStructure) error {
-	for node, part := range created {
+func MakeFilesystems(created []deviceStructure) error {
+	for _, part := range created {
 		if part.VolumeStructure.Filesystem != "" {
-			if err := mkfs(node, part.VolumeStructure.Label, part.VolumeStructure.Filesystem); err != nil {
+			if err := mkfs(part.Node, part.VolumeStructure.Label, part.VolumeStructure.Filesystem); err != nil {
 				return err
 			}
 		}
--- a/cmd/snap-recovery/partition/mkfs.go
+++ b/cmd/snap-recovery/partition/mkfs.go
@@ -24,16 +24,25 @@ import (
 	"github.com/snapcore/snapd/gadget"
 )
 
-// makeFilesystem will create a filesystem on the given node with
-// the given label and filesystem type. Optionally a directory with
-// filesystem content can be specified and the newly created filesystem
-// will be populated with that.
-func makeFilesystem(node, label, filesystem, content string) error {
+func MakeFilesystems(created map[string]gadget.LaidOutStructure) error {
+	for node, part := range created {
+		if part.VolumeStructure.Filesystem != "" {
+			if err := mkfs(node, part.VolumeStructure.Label, part.VolumeStructure.Filesystem); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// mkfs will create a single filesystem on the given node with
+// the given label and filesystem type.
+func mkfs(node, label, filesystem string) error {
 	switch filesystem {
 	case "vfat":
-		return gadget.MkfsVfat(node, label, content)
+		return gadget.MkfsVfat(node, label, "")
 	case "ext4":
-		return gadget.MkfsExt4(node, label, content)
+		return gadget.MkfsExt4(node, label, "")
 	default:
 		return fmt.Errorf("cannot create unsupported filesystem %q", filesystem)
 	}
--- a/cmd/snap-recovery/partition/mkfs.go
+++ b/cmd/snap-recovery/partition/mkfs.go
@@ -24,6 +24,10 @@ import (
 	"github.com/snapcore/snapd/gadget"
 )
 
+// makeFilesystem will create a filesystem on the given node with
+// the given label and filesystem type. Optionally a directory with
+// filesystem content can be specified and the newly created filesystem
+// will be populated with that.
 func makeFilesystem(node, label, filesystem, content string) error {
 	switch filesystem {
 	case "vfat":
--- a/cmd/snap-recovery/partition/mkfs_test.go
+++ b/cmd/snap-recovery/partition/mkfs_test.go
@@ -22,30 +22,84 @@ import (
 	. "gopkg.in/check.v1"
 
 	"github.com/snapcore/snapd/cmd/snap-recovery/partition"
+	"github.com/snapcore/snapd/gadget"
 	"github.com/snapcore/snapd/testutil"
 )
 
-func (s *partitionTestSuite) TestMkfsUnhappy(c *C) {
-	err := partition.MakeFilesystem("/dev/node", "some-label", "unsupported-filesystem-type", "/dir/with/initial/fs/content")
+type mkfsSuite struct {
+	testutil.BaseTest
+
+	mockMkfsVfat *testutil.MockCmd
+	mockMkfsExt4 *testutil.MockCmd
+}
+
+var _ = Suite(&mkfsSuite{})
+
+func (s *mkfsSuite) SetUpTest(c *C) {
+	s.mockMkfsVfat = testutil.MockCommand(c, "mkfs.vfat", "")
+	s.AddCleanup(s.mockMkfsVfat.Restore)
+	s.mockMkfsExt4 = testutil.MockCommand(c, "mkfs.ext4", "")
+	s.AddCleanup(s.mockMkfsExt4.Restore)
+}
+
+func (s *mkfsSuite) TestMkfsUnhappy(c *C) {
+	err := partition.Mkfs("/dev/node", "some-label", "unsupported-filesystem-type")
 	c.Assert(err, ErrorMatches, `cannot create unsupported filesystem "unsupported-filesystem-type"`)
 }
 
-func (s *partitionTestSuite) TestMkfsVfat(c *C) {
-	mockMkfs := testutil.MockCommand(c, "mkfs.vfat", "")
-	defer mockMkfs.Restore()
-
-	err := partition.MakeFilesystem("/dev/node", "some-label", "vfat", "")
+func (s *mkfsSuite) TestMkfsVfat(c *C) {
+	err := partition.Mkfs("/dev/node", "some-label", "vfat")
 	c.Assert(err, IsNil)
 	// details are already tested in the gadget package
-	c.Assert(mockMkfs.Calls(), HasLen, 1)
+	c.Assert(s.mockMkfsVfat.Calls(), HasLen, 1)
 }
 
-func (s *partitionTestSuite) TestMkfsExt4(c *C) {
-	mockMkfs := testutil.MockCommand(c, "mkfs.ext4", "")
-	defer mockMkfs.Restore()
-
-	err := partition.MakeFilesystem("/dev/node", "some-label", "ext4", "")
+func (s *mkfsSuite) TestMkfsExt4(c *C) {
+	err := partition.Mkfs("/dev/node", "some-label", "ext4")
 	c.Assert(err, IsNil)
 	// details are already tested in the gadget package
-	c.Assert(mockMkfs.Calls(), HasLen, 1)
+	c.Assert(s.mockMkfsExt4.Calls(), HasLen, 1)
+}
+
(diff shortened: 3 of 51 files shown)


cmd/snap-recovery/main.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+func run(args []string) error {
+	if !osutil.GetenvBool("SNAPPY_TESTING") {
+		return fmt.Errorf("cannot use outside of tests yet")
+	}
+	if os.Getuid() != 0 {
+		return fmt.Errorf("please run as root")
+	}
+	if len(os.Args) < 3 {
```
zyga: I'd prefer if this was using a real (tm) argument parser. I realise this is an early stage though so I won't block on it.

cmd/snap-recovery/main.go
```diff
@@ -0,0 +1,59 @@
...
+		// XXX: slightly ugly to return usage as an error but ok for now
+		return fmt.Errorf("usage: %s <gadget root> <block device>\n", os.Args[0])
+	}
+
+	gadgetRoot := os.Args[1]
+	device := os.Args[2]
+	options := &recover.Options{}
+
+	recov := recover.New(gadgetRoot, device, options)
+	if err := recov.Run(); err != nil {
```
zyga: It seems you can just return `recov.Run()` here.

cmd/snap-recovery/partition/mkfs.go
```diff
@@ -24,12 +24,25 @@ import (
 	"github.com/snapcore/snapd/gadget"
 )
 
-func makeFilesystem(node, label, filesystem, content string) error {
+func MakeFilesystems(created map[string]gadget.LaidOutStructure) error {
+	for node, part := range created {
+		if part.VolumeStructure.Filesystem != "" {
+			if err := mkfs(node, part.VolumeStructure.Label, part.VolumeStructure.Filesystem); err != nil {
```
zyga: I don't like the fact that this creates filesystems in non-deterministic order. Could you please consider sorting them by node name so that this function is entirely deterministic.

cmd/snap-recovery/partition/mkfs_test.go
```diff
@@ -22,30 +22,84 @@ import (
...
+			StartOffset: 1260388352,
+			Index:       3,
+		},
+	}
+	err := partition.MakeFilesystems(created)
 	c.Assert(err, IsNil)
-	// details are already tested in the gadget package
-	c.Assert(mockMkfs.Calls(), HasLen, 1)
+	c.Assert(s.mockMkfsExt4.Calls(), HasLen, 1)
+	c.Assert(s.mockMkfsVfat.Calls(), HasLen, 1)
```
zyga: If you follow my recommendation please ensure that the order is tested here.

cmd/snap-recovery/recover/recover.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/snapcore/snapd/cmd/snap-recovery/partition"
+	"github.com/snapcore/snapd/gadget"
+)
+
+type Options struct {
+	// will contain encryption later
+}
+
+type Recover struct {
```
zyga: Package as well as type-level descriptions would be appreciated.

cmd/snap-recovery/recover/recover.go
```diff
@@ -0,0 +1,65 @@
...
+func New(gadgetRoot, device string, options *Options) *Recover {
+	return &Recover{
+		gadgetRoot: gadgetRoot,
+		device:     device,
+		options:    options,
+	}
+}
+
+func (r *Recover) Run() error {
+	// XXX: ensure we test that the current partition table is
```
zyga: CC @bboozzoo

cmd/snap-recovery/recover/recover.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return fmt.Errorf("cannot layout the volume: %v", err)
+	}
+
+	sfdisk := partition.NewSFDisk(r.device)
+	created, err := sfdisk.Create(lv)
+	if err != nil {
+		return fmt.Errorf("cannot create the partitions: %v", err)
+	}
+	if err := partition.MakeFilesystems(created); err != nil {
```
zyga: Are partially created filesystems (where a subset was created successfully but one failed) problematic? Do we need to consider `wipefs` to remove them in this case?

cmd/snap-recovery/recover/recover.go
```diff
@@ -0,0 +1,65 @@
...
+		return err
+	}
+	// XXX: deploy contento
```
zyga:
```suggestion
	// XXX: deploy content
```

packaging/debian-sid/snapd.install
```diff
@@ -5,6 +5,7 @@ usr/bin/snap-exec /usr/lib/snapd/
 usr/bin/snap-update-ns /usr/lib/snapd/
 usr/bin/snapd /usr/lib/snapd/
 usr/bin/snap-seccomp /usr/lib/snapd/
+usr/bin/snap-recovery /usr/lib/snapd/
```
zyga: Do we need this on classic systems? I think it should not be packaged.

packaging/fedora/snapd.spec
```diff
@@ -471,6 +471,7 @@ sed -e "s:github.com/snapcore/bolt:github.com/boltdb/bolt:g" -i advisor/*.go err
 %gobuild -o bin/snapd $GOFLAGS %{import_path}/cmd/snapd
 %gobuild -o bin/snap $GOFLAGS %{import_path}/cmd/snap
 %gobuild -o bin/snap-failure $GOFLAGS %{import_path}/cmd/snap-failure
+%gobuild -o bin/snap-recovery $GOFLAGS %{import_path}/cmd/snap-recovery
```
zyga: Same remark about building and shipping this on classic systems.

tests/main/uc20-snap-recovery/task.yaml
```diff
@@ -0,0 +1,48 @@
...
+systems: [ubuntu-18.04-64]
+
+restore: |
+    if [ -f loop.txt ]; then
+        losetup -d "$(cat loop.txt)"
+    fi
+
+prepare: |
+    echo "Create a fake block device image"
+    dd if=/dev/zero of=fake.img bs=1M seek=2000 count=1
```
zyga: You can also just call `truncate --size=2GB fake.img` which seems "nicer" since it is zero-filled and sparse.

tests/main/uc20-snap-recovery/task.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+    echo "Setup the image as a block device"
+    losetup -fP fake.img
```
zyga: The `-P` part seems useless since we just created an empty file.

cmd/snap-recovery/main.go
mvo5: Yeah, early stages.

cmd/snap-recovery/main.go
mvo5: Yes, but TBH this code may get simplified further in a followup but I'm happy to change if you feel its important.

cmd/snap-recovery/partition/mkfs.go
mvo5: There will be a followup that will use the gpt partition attributes to mark partitions as ready or in progress. I assume your concern is idempotency? This will be addressed. Or is there another reason why this should be deterministic?

cmd/snap-recovery/recover/recover.go
mvo5: I think we do, there is a followup planned that uses the partition attributes to keep track of the state of the operations. I.e. at the start we set a bit that we work on the partition and only clean it once we are fully done.

cmd/snap-recovery/recover/recover.go
mvo5: Yeah, this is all very early though, given the current structure of the Recover struct this may just become a single function and the package may actually be removed and just folded into the cmd/snap-recovery/ code. Sorry that its not quite so clean as usually, this is merging bit of Claudio and my own ideas which makes this a bit less straightforward than usual.

packaging/debian-sid/snapd.install
mvo5: I'm fine with not packaging it, it just means I need to figure out how to not do that on all the distro packaging we do, which is a bit of a headache on its own :/

tests/main/uc20-snap-recovery/task.yaml
mvo5: Pretty sure the outcome is (almost) the same, note the "seek" in the dd.

tests/main/uc20-snap-recovery/task.yaml
mvo5: Indeed, I will remove this.

mvo5:
Thanks for the feedback! I addressed a lot of this in #7613 and once that is in will get back to this one here and address the remaining pieces.

cmd/snap-recovery/partition/mkfs.go
zyga: In this specific case my concern is determinism. If this is used on millions of devices it should not result in sometimes doing partition A ahead of B and sometimes the other way around.

packaging/debian-sid/snapd.install
zyga: I'm happy to help. I think this should not be in the `snapd` package. It also has its own dependencies which we should be aware of. You ship it and people start to use it ;)

tests/main/uc20-snap-recovery/task.yaml
zyga: Ah, I didn't notice the seek is large and count is small.

packaging/debian-sid/snapd.install
mvo5: This is no longer shipped (except for ubuntu 16.04+ where we need it currently so that its available in core/snapd snap)

tests/main/uc20-snap-recovery/task.yaml
mvo5: I updated to use truncate now to make things easier to read.

packaging/snapd.mk
```diff
@@ -50,14 +50,14 @@ snap_mount_dir = /snap
 endif
 
 # The list of go binaries we are expected to build.
-go_binaries = snap snapctl snap-seccomp snap-update-ns snap-exec snapd snap-recovery
```
zyga: Note, you can still keep this part but expand the part below that removes non-core binaries.

cmd/snap-recovery/partition/sfdisk.go
```diff
@@ -216,10 +217,10 @@ func buildPartitionList(ptable *sfdiskPartitionTable, pv *gadget.LaidOutVolume)
 			s.Size/sectorSize, partitionType(ptable.Label, p.Type), s.Name)
 
 		// Are roles unique so we can use it to map nodes? Should we use labels instead?
-		deviceMap[s.Role] = node
+		toBeCreated[node] = p
```
bboozzoo: Wish we had more reliable name to identify device nodes rather than making up the name ourselves. In the context of the followup PR, perhaps we should create, list the partitions and match then with structures, and then operate on the device nodes.

cmd/snap-recovery/partition/sfdisk.go
```diff
@@ -173,12 +173,13 @@ func deviceName(name string, index int) string {
...
-// suitable for sfdisk input and a role to device node map.
-func buildPartitionList(ptable *sfdiskPartitionTable, pv *gadget.LaidOutVolume) (*bytes.Buffer, map[string]string) {
+// buildPartitionList builds a list of partitions based on the current
+// device contents and gadget structure list, in sfdisk dump
+// format. Return a partitioning description suitable for sfdisk input
+// and a list of the partitions to be created
+func buildPartitionList(ptable *sfdiskPartitionTable, pv *gadget.LaidOutVolume) (sfdiskInput *bytes.Buffer, toBeCreated map[string]gadget.LaidOutStructure) {
 	buf := &bytes.Buffer{}
-	deviceMap := map[string]string{}
+	toBeCreated = make(map[string]gadget.LaidOutStructure, len(pv.LaidOutStructure))
```
bboozzoo: Probably a followup material, but we could do:
```
type deviceStructure {
    gadget.LaidOutStructure
    Node string
}
```

cmd/snap-recovery/partition/mkfs.go
bboozzoo: Yeah, I agree with @zyga but a doing that in the next PR sounds good too.

cmd/snap-recovery/partition/sfdisk.go
mvo5: Good suggestion, I implemented this now as []deviceStructure sorted by layout as defined in the gadget.yaml

cmd/snap-recovery/partition/mkfs.go
```diff
@@ -24,10 +24,10 @@ import (
 	"github.com/snapcore/snapd/gadget"
 )
 
-func MakeFilesystems(created map[string]gadget.LaidOutStructure) error {
-	for node, part := range created {
+func MakeFilesystems(created []deviceStructure) error {
```
zyga: Having a public function that takes a private argument is difficult to use externally.

mvo5: Indeed, I will fix in a followup, I expect some tweaks here anyway :)

mvo5: And indeed it did change in the followup 7640

PR #7631, opened 2019-10-18
interfaces/pulseaudio: adjust to manually connect by default

jdstrand:
Snaps should use audio-playback (and audio-record if they need
recording) instead of pulseaudio. Now that audio-record has been
available since 2.41 and all existing snaps in the snap store that plugs
pulseaudio have been grandfathered to have a snap declaration to
auto-connect pulseaudio, it is time to adjust pulseaudio to manually
connect.

References:
- https://forum.snapcraft.io/t/upcoming-pulseaudio-interface-deprecation/13418

### DIFF
--- a/interfaces/builtin/pulseaudio.go
+++ b/interfaces/builtin/pulseaudio.go
@@ -38,6 +38,7 @@ const pulseaudioBaseDeclarationSlots = `
         - core
     deny-connection:
       on-classic: false
+    deny-auto-connection: true
 `
 
 const pulseaudioConnectedPlugAppArmor = `
--- a/interfaces/policy/basedeclaration_test.go
+++ b/interfaces/policy/basedeclaration_test.go
@@ -165,7 +165,6 @@ func (s *baseDeclSuite) TestAutoConnection(c *C) {
 		"online-accounts-service": true,
 		"opengl":                  true,
 		"optical-drive":           true,
-		"pulseaudio":              true,
 		"screen-inhibit-control":  true,
 		"ubuntu-download-manager": true,
 		"unity7":                  true,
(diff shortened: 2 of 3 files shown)


tom-james-watson:
If a new release is published to the snap store that requests pulse-audio and not audio-playback, will they have grandfathering patch auto-applied?

Just wondering what will happen if I publish a new version of my package before https://github.com/electron-userland/electron-builder/issues/4234#issuecomment-543809065 gets resolved. Thanks!

jdstrand:
> If a new release is published to the snap store that requests pulse-audio and not audio-playback, will they have grandfathering patch auto-applied?
> 
> Just wondering what will happen if I publish a new version of my package before [electron-userland/electron-builder#4234 (comment)](https://github.com/electron-userland/electron-builder/issues/4234#issuecomment-543809065) gets resolved. Thanks!

The grandfathering happens store side and yes, for existing snaps, this has already been done and snaps don't have to change to add audio-playback and shouldn't regress. Note also, it is fine to specify audio-playback and pulseaudio at the same time and also fine to specify audio-playback on a system with snapd < 2.41 (it, obviously, just won't be connected since it doesn't exist there, but it will install just fine).

The main thing about https://github.com/electron-userland/electron-builder/issues/4234 is that for *new* snaps without this applied, pulseaudio won't be auto-connected, but they will install just fine. We have a [process](https://forum.snapcraft.io/t/process-for-aliases-auto-connections-and-tracks/455/11) that people can follow, and if/when someone requests pulseaudio to be auto-connected, we can guide them towards using audio-playback instead until electron is updated.

jdstrand:
Thanks for the reviews!

PR #7632, opened 2019-10-18
interfaces: de-duplicate emitted update-ns profiles

zyga:
It was brought to our attention that apparmor_parser cannot de-duplicate
the following profile efficiently, unlike all the other rules:

    mount foo,
    mount foo,

Internally apparmor_parser first de-duplicates some things that are
identical.  This step is efficient in complexity and is not a problem.
The next step constructs a state machine based on all the rules that
remain and minimises it.  Unfortunately this step is exponential in
memory so while the resulting binary profile is efficient the means of
computing one is not.

To combat this, this patch switches apparmor.Specification to use
OrderedSet to store update-ns snippets. Existing buffered emitters are
replaced by a de-duplicating emitter. Since the automatically generated
update-ns rules are very repetitive, that last step ensures that
actually generated profile text does not contain the repetitions that
would stress apparmor_parser.

Fixes: https://bugs.launchpad.net/snapd/+bug/1848567

### DIFF
--- a/interfaces/apparmor/spec.go
+++ b/interfaces/apparmor/spec.go
@@ -42,7 +42,7 @@ type Specification struct {
 	snippets map[string][]string
 	// updateNS describe parts of apparmor policy for snap-update-ns executing
 	// on behalf of a given snap.
-	updateNS []string
+	updateNS strutil.OrderedSet
 
 	// AppArmor deny rules cannot be undone by allow rules which makes
 	// deny rules difficult to work with arbitrary combinations of
@@ -92,7 +92,19 @@ func (spec *Specification) AddSnippet(snippet string) {
 
 // AddUpdateNS adds a new apparmor snippet for the snap-update-ns program.
 func (spec *Specification) AddUpdateNS(snippet string) {
-	spec.updateNS = append(spec.updateNS, snippet)
+	spec.updateNS.Put(snippet)
+}
+
+// EmitUpdateNSFunc returns a function for emitting update-ns snippets.
+func (spec *Specification) EmitUpdateNSFunc() func(f string, args ...interface{}) {
+	return func(f string, args ...interface{}) {
+		spec.AddUpdateNS(fmt.Sprintf(f, args...))
+	}
+}
+
+// UpdateNSIndexOf returns the index of a previously added snippet.
+func (spec *Specification) UpdateNSIndexOf(snippet string) (idx int, ok bool) {
+	return spec.updateNS.IndexOf(snippet)
 }
 
 // AddLayout adds apparmor snippets based on the layout of the snap.
@@ -145,12 +157,11 @@ func (spec *Specification) AddLayout(si *snap.Info) {
 		}
 		sort.Strings(spec.snippets[tag])
 	}
+
+	emit := spec.EmitUpdateNSFunc()
+
 	// Append update-ns snippets that allow constructing the layout.
 	for _, path := range paths {
-		var buf bytes.Buffer
-		emit := func(f string, args ...interface{}) {
-			fmt.Fprintf(&buf, f, args...)
-		}
 		l := si.Layout[path]
 		emit("  # Layout %s\n", l.String())
 		path := si.ExpandSnapVariables(l.Path)
@@ -184,7 +195,6 @@ func (spec *Specification) AddLayout(si *snap.Info) {
 			emit("  %s rw,\n", path)
 			GenWritableProfile(emit, path, 2) // At least / and /some-top-level-directory
 		}
-		spec.AddUpdateNS(buf.String())
 	}
 }
 
@@ -347,9 +357,7 @@ func (spec *Specification) SecurityTags() []string {
 
 // UpdateNS returns a deep copy of all the added snap-update-ns snippets.
 func (spec *Specification) UpdateNS() []string {
-	cp := make([]string, len(spec.updateNS))
-	copy(cp, spec.updateNS)
-	return cp
+	return spec.updateNS.Items()
 }
 
 func snippetFromLayout(layout *snap.Layout) string {
--- a/interfaces/apparmor/spec_test.go
+++ b/interfaces/apparmor/spec_test.go
@@ -20,6 +20,8 @@
 package apparmor_test
 
 import (
+	"strings"
+
 	. "gopkg.in/check.v1"
 
 	"github.com/snapcore/snapd/interfaces"
@@ -139,6 +141,11 @@ func (s *specSuite) TestAddUpdateNS(c *C) {
 	s.spec.AddUpdateNS("s-u-n snippet 1")
 	s.spec.AddUpdateNS("s-u-n snippet 2")
 
+	// Check the order of the snippets can be retrieved.
+	idx, ok := s.spec.UpdateNSIndexOf("s-u-n snippet 2")
+	c.Assert(ok, Equals, true)
+	c.Check(idx, Equals, 1)
+
 	// The snippets were recorded correctly and in the right place.
 	c.Assert(s.spec.UpdateNS(), DeepEquals, []string{
 		"s-u-n snippet 1", "s-u-n snippet 2",
@@ -218,33 +225,21 @@ func (s *specSuite) TestApparmorSnippetsFromLayout(c *C) {
   umount /etc/*,
   umount /etc/*/,
   # Writable mimic /snap/vanguard/42
-  # .. permissions for traversing the prefix that is assumed to exist
-  / r,
   /snap/ r,
   /snap/vanguard/ r,
   # .. variant with mimic at /snap/vanguard/42/
-  # Allow reading the mimic directory, it must exist in the first place.
   /snap/vanguard/42/ r,
-  # Allow setting the read-only directory aside via a bind mount.
   /tmp/.snap/snap/vanguard/42/ rw,
   mount options=(rbind, rw) /snap/vanguard/42/ -> /tmp/.snap/snap/vanguard/42/,
-  # Allow mounting tmpfs over the read-only directory.
   mount fstype=tmpfs options=(rw) tmpfs -> /snap/vanguard/42/,
-  # Allow creating empty files and directories for bind mounting things
-  # to reconstruct the now-writable parent directory.
   /tmp/.snap/snap/vanguard/42/*/ rw,
   /snap/vanguard/42/*/ rw,
   mount options=(rbind, rw) /tmp/.snap/snap/vanguard/42/*/ -> /snap/vanguard/42/*/,
   /tmp/.snap/snap/vanguard/42/* rw,
   /snap/vanguard/42/* rw,
   mount options=(bind, rw) /tmp/.snap/snap/vanguard/42/* -> /snap/vanguard/42/*,
-  # Allow unmounting the auxiliary directory.
-  # TODO: use fstype=tmpfs here for more strictness (LP: #1613403)
   mount options=(rprivate) -> /tmp/.snap/snap/vanguard/42/,
   umount /tmp/.snap/snap/vanguard/42/,
-  # Allow unmounting the destination directory as well as anything
-  # inside.  This lets us perform the undo plan in case the writable
-  # mimic fails.
   mount options=(rprivate) -> /snap/vanguard/42/,
   mount options=(rprivate) -> /snap/vanguard/42/*,
   mount options=(rprivate) -> /snap/vanguard/42/*/,
@@ -252,38 +247,30 @@ func (s *specSuite) TestApparmorSnippetsFromLayout(c *C) {
   umount /snap/vanguard/42/*,
   umount /snap/vanguard/42/*/,
 `
-	c.Assert(updateNS[0], Equals, profile0)
+	// Find the slice that describes profile0 by looking for the first unique
+	// line of the next profile.
+	start := 0
+	end, _ := s.spec.UpdateNSIndexOf("  # Layout /usr/foo: bind $SNAP/usr/foo\n")
+	c.Assert(strings.Join(updateNS[start:end], ""), Equals, profile0)
 
 	profile1 := `  # Layout /usr/foo: bind $SNAP/usr/foo
   mount options=(rbind, rw) /snap/vanguard/42/usr/foo/ -> /usr/foo/,
   mount options=(rprivate) -> /usr/foo/,
   umount /usr/foo/,
   # Writable mimic /usr
-  # .. permissions for traversing the prefix that is assumed to exist
-  / r,
   # .. variant with mimic at /usr/
-  # Allow reading the mimic directory, it must exist in the first place.
   /usr/ r,
-  # Allow setting the read-only directory aside via a bind mount.
   /tmp/.snap/usr/ rw,
   mount options=(rbind, rw) /usr/ -> /tmp/.snap/usr/,
-  # Allow mounting tmpfs over the read-only directory.
   mount fstype=tmpfs options=(rw) tmpfs -> /usr/,
-  # Allow creating empty files and directories for bind mounting things
-  # to reconstruct the now-writable parent directory.
   /tmp/.snap/usr/*/ rw,
   /usr/*/ rw,
   mount options=(rbind, rw) /tmp/.snap/usr/*/ -> /usr/*/,
   /tmp/.snap/usr/* rw,
   /usr/* rw,
   mount options=(bind, rw) /tmp/.snap/usr/* -> /usr/*,
-  # Allow unmounting the auxiliary directory.
-  # TODO: use fstype=tmpfs here for more strictness (LP: #1613403)
   mount options=(rprivate) -> /tmp/.snap/usr/,
   umount /tmp/.snap/usr/,
-  # Allow unmounting the destination directory as well as anything
-  # inside.  This lets us perform the undo plan in case the writable
-  # mimic fails.
   mount options=(rprivate) -> /usr/,
   mount options=(rprivate) -> /usr/*,
   mount options=(rprivate) -> /usr/*/,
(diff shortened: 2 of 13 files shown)


zyga:
NOTE: I'm adding similar change to `interfaces/builtin/content.go`

strutil/set.go
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+package strutil
+
+// OrderedSet is a set of strings that maintains the order of insertion.
+//
+// External synchronization is required for safe concurrent access.
+type OrderedSet struct {
+	items      []string
+	positionOf map[string]int
```
stolowski: Just an observation (in case memory was a concern here, but probably isn't), we could avoid storing items and just reconstruct them when needed from positionOf.

interfaces/apparmor/spec.go
```diff
@@ -92,7 +92,12 @@ func (spec *Specification) AddSnippet(snippet string) {
 
 // AddUpdateNS adds a new apparmor snippet for the snap-update-ns program.
 func (spec *Specification) AddUpdateNS(snippet string) {
-	spec.updateNS = append(spec.updateNS, snippet)
+	spec.updateNS.Put(snippet)
+}
+
+// UpdateNSIndexOf returns the index of a previously added snippet.
+func (spec *Specification) UpdateNSIndexOf(snippet string) (idx int, ok bool) {
```
stolowski: Is is just for the tests? Can it go to exports?

strutil/set.go
```diff
@@ -0,0 +1,99 @@
...
+// If the item was not present then it is stored and ordered after all existing
+// elements. If the item was already present its position is not changed.
+//
+// Put is O(1) in the size of the set.
+func (o *OrderedSet) Put(item string) {
+	if o.positionOf == nil {
+		o.positionOf = make(map[string]int)
+	}
+
+	if _, ok := o.positionOf[item]; ok {
```
stolowski: This check is only needed if o.positionOf was not null, so could be simplifed to:
```
if o.positionOf == nil {
   ...
} else {
     if _, ok := o.positionOf[item]; ok {
       return
    }
}
```

interfaces/apparmor/spec.go
```diff
@@ -181,7 +186,9 @@ func (spec *Specification) AddLayout(si *snap.Info) {
 			fmt.Fprintf(&buf, "  %s rw,\n", path)
 			WritableProfile(&buf, path, 2) // At least / and /some-top-level-directory
 		}
-		spec.AddUpdateNS(buf.String())
+		for _, line := range strings.SplitAfter(buf.String(), "\n") {
```
stolowski: It seems a bit wasteful to construct bytes.Buffer and append lines and then split it back; can we simply collect them in an array and avoid temporary bytes.Buffer?

interfaces/apparmor/spec.go
zyga: Yes and no :)

strutil/set.go
zyga: I know and I did code it exactly like so earlier but I think the way I coded it in the end is clearer for the reader.

interfaces/apparmor/spec.go
zyga: That's true but I wanted to get feedback on the proposal first. The whole buffer can go away and we can simply add each snippet as they are created.

strutil/set.go
zyga: Yeah that is true. I had a TODO that I removed from the implementation of `Items()` that we can essentially look at the map and craft the order.

The complexity of the current `Items()` method is O(N), if we drop the `items` field then items would roughly look like:

```
items := make([]string, len(os.positionOf)
for item, idx := range positionOf {
    items[idx] = item
}
return items
```

Which is also O(N) but at O(N) less memory than originally in the stationary case. I can make the change if you feel like it. I was mainly going for simple-and-correct since the original problem was more interesting but I don't mind optimising this slightly, especially since it seems it would not be hard to do so.

stolowski: No strong opinion, as I said it's more of an observation. Perhaps just add a comment next to items member saying it's for convienience and can be inferred from positionOf.

interfaces/apparmor/spec.go
stolowski: Ok, can you clarify please? At the moment is used in tests only afaict?

strutil/set.go
zyga: I'll just implement it :)

zyga: Done, marking as resolved.

interfaces/apparmor/spec.go
zyga: Yes but it is used from multiple packages.

zyga:
I split the review of ordered set to https://github.com/snapcore/snapd/pull/7633

interfaces/apparmor/spec_test.go
```diff
@@ -139,6 +141,11 @@ func (s *specSuite) TestAddUpdateNS(c *C) {
 	s.spec.AddUpdateNS("s-u-n snippet 1")
 	s.spec.AddUpdateNS("s-u-n snippet 2")
 
+	// Check the order of the snippets can be retrieved.
+	idx, ok := s.spec.UpdateNSIndexOf("s-u-n snippet 2")
```
anonymouse64: do we need this test to run `s.spec.AddUpdateNS("s-u-n snippet 2")` again to show that it doesn't get put in multiple times?

interfaces/apparmor/spec.go
```diff
@@ -42,7 +42,7 @@ type Specification struct {
 	snippets map[string][]string
 	// updateNS describe parts of apparmor policy for snap-update-ns executing
 	// on behalf of a given snap.
-	updateNS []string
+	updateNS strutil.OrderedSet
```
anonymouse64: why does this need to be an _ordered_ set? We only seem to use the ordering in the tests, and even there it seems suspiciously testing just that the order is what is expected, not necessarily some larger bit which requires it to be ordered.

interfaces/apparmor/spec.go
anonymouse64: I think this is problematic if we ever use this code for templated snippets because AppArmor allows multi-line rules for some types of rules, such as mount. For example the following is valid apparmor on my 19.04 Ubuntu AppArmor:
```
mount options=ro
/tmp/blah -> /tmp/blah2
```

If this type of apparmor rule is ever input into the de-duplication code you have here, then these two rules would become messed up/out of order IIUC:

```
mount options=ro
/tmp/blah -> /tmp/blah2,
```
and 
```
mount options=ro
/tmp/blah3 -> /tmp/blah4,
```
would get serialized out as
```
mount options=ro
/tmp/blah -> /tmp/blah2,
/tmp/blah3 -> /tmp/blah4,
```

which is obviously wrong and potentially very bad.

It's not clear to me yet if the code here is such that something else trying to write AppArmor could code accidentally cross paths with this and use the de-duplication. I will look more deeply to see if this is the case, but I think we ought to tread carefully to ensure that only snap-update-ns where the full profile is auto-generated uses this de-duplication and we can easily ensure no multi-line rules get input.

interfaces/apparmor/spec.go
zyga: If we ever want to audit the text we produce I would argue that we want to have this property.

interfaces/apparmor/spec.go
zyga: We don't have any rules like that in this code so it's safe. I agree the split is unnecessary and I'm happy to remove it, it will blow up the patch size slightly since way more code will be affected, but we can remove it entirely in the end.

The annoying part of this pull request was the test changes to reflect the new state of the profiles we generate. We can certainly improve the details now.

interfaces/apparmor/spec.go
anonymouse64: okay that's fair

zyga:
Rebased on the now-merged OrderedSet branch

interfaces/apparmor/spec.go
zyga: This is now changed to emit the lines directly. The buffer has been eliminated.

zyga:
I could further split this into a branch that just does the `emit` changes without using the `OrderedSet` yet. This would cut virtually all of the automatic changes and reduce this branch to just the de-duplication change (a line) and a flurry of cascading test changes in generated profiles.

interfaces/apparmor/spec.go
zyga: This is now addressed.

interfaces/apparmor/spec_test.go
zyga: I can add it though it feels like testing the internal implementation. I'm perhaps biased because this is still fresh in my head.

interfaces/apparmor/spec.go
stolowski: Ah, got it, thanks!

zyga:
I've broken out the automatic changes to https://github.com/snapcore/snapd/pull/7639 and will re-base on top of it once it lands. This should reduce this to just the test changes and a very small change that switches to the use of OrderedSet

interfaces/apparmor/spec.go
bboozzoo: I not happy that we loose the input information. The input snippet is silently dropped in the input, what may cause the output to be confusing. Eg:

```
// snippet 1
spec.AddUpdateNs("/foo rwx\n")
// snippet 2
spec.AddUpdateNS("# the rule below does magic, needed for you-name-it-here\n")
spec.AddUpdateNs("/foo rwx\n")
// snippet 3
spec.AddUpdateNs("/bar rxi\n")
```
If I understand correctly what will happen, the output will look like this:
```
/foo rwx
# the rule below does magic, needed for you-name-it-here\
/bar rxi
```

I think it'd be better to preserve the input snippet but have it commented out, or leave a trace that it's been deduplicated, eg.
```
/foo rwx
# the rule below does magic, needed for you-name-it-here
# (deduplicated block start, prevoious in line 123):
# /foo rwx
# (deduplicated block end)
/bar rxi
```

I do hope that AppArmor parser does not choke on comment lines though.

zyga: It doesn't choke on comments but it was NACKed by Jamie as far as I recall. He asked not to do this.

zyga:
This is now re-based on https://github.com/snapcore/snapd/pull/7639

interfaces/apparmor/spec.go
```diff
@@ -92,7 +92,19 @@ func (spec *Specification) AddSnippet(snippet string) {
 
 // AddUpdateNS adds a new apparmor snippet for the snap-update-ns program.
 func (spec *Specification) AddUpdateNS(snippet string) {
-	spec.updateNS = append(spec.updateNS, snippet)
+	spec.updateNS.Put(snippet)
+}
+
+// EmitUpdateNSFunc returns a function for emitting update-ns snippets.
+func (spec *Specification) EmitUpdateNSFunc() func(f string, args ...interface{}) {
```
stolowski: Following my comment to the earlier PR regarding the indirection here, cannot this simply become:
```
diff --git a/interfaces/apparmor/spec.go b/interfaces/apparmor/spec.go
index 93f5429c69..2c8f933acb 100644
--- a/interfaces/apparmor/spec.go
+++ b/interfaces/apparmor/spec.go
@@ -96,10 +96,8 @@ func (spec *Specification) AddUpdateNS(snippet string) {
 }

 // EmitUpdateNSFunc returns a function for emitting update-ns snippets.
-func (spec *Specification) EmitUpdateNSFunc() func(f string, args ...interface{}) {
-       return func(f string, args ...interface{}) {
-               spec.AddUpdateNS(fmt.Sprintf(f, args...))
-       }
+func (spec *Specification) EmitUpdateNSFunc(f string, args ...interface{}) {
+       spec.AddUpdateNS(fmt.Sprintf(f, args...))
 }

 // UpdateNSIndexOf returns the index of a previously added snippet.
@@ -158,7 +156,7 @@ func (spec *Specification) AddLayout(si *snap.Info) {
                sort.Strings(spec.snippets[tag])
        }

-       emit := spec.EmitUpdateNSFunc()
+       emit := spec.EmitUpdateNSFunc

        // Append update-ns snippets that allow constructing the layout.
        for _, path := range paths {
diff --git a/interfaces/builtin/appstream_metadata.go b/interfaces/builtin/appstream_metadata.go
index ffc2711287..4a235c45b2 100644
--- a/interfaces/builtin/appstream_metadata.go
+++ b/interfaces/builtin/appstream_metadata.go
@@ -75,7 +75,7 @@ func (iface *appstreamMetadataInterface) AppArmorConnectedPlug(spec *apparmor.Sp
        spec.AddSnippet(appstreamMetadataConnectedPlugAppArmor)

        // Generate rules to allow snap-update-ns to do its thing
-       emit := spec.EmitUpdateNSFunc()
+       emit := spec.EmitUpdateNSFunc
        for _, target := range appstreamMetadataDirs {
                source := "/var/lib/snapd/hostfs" + target
                emit("  # Read-only access to %s\n", target)
diff --git a/interfaces/builtin/content.go b/interfaces/builtin/content.go
index b15df7faa9..e8bd912e3e 100644
--- a/interfaces/builtin/content.go
+++ b/interfaces/builtin/content.go
@@ -214,7 +214,7 @@ func mountEntry(plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot,
 func (iface *contentInterface) AppArmorConnectedPlug(spec *apparmor.Specification, plug *interfaces.ConnectedPlug, slot *interfaces.ConnectedSlot) error {
        contentSnippet := bytes.NewBuffer(nil)
        writePaths := iface.path(slot, "write")
-       emit := spec.EmitUpdateNSFunc()
+       emit := spec.EmitUpdateNSFunc
        if len(writePaths) > 0 {
                fmt.Fprintf(contentSnippet, `
 # In addition to the bind mount, add any AppArmor rules so that
diff --git a/interfaces/builtin/desktop.go b/interfaces/builtin/desktop.go
index 7bb8c9c89c..944de43ca1 100644
--- a/interfaces/builtin/desktop.go
+++ b/interfaces/builtin/desktop.go
@@ -231,7 +231,7 @@ func (iface *desktopInterface) AppArmorConnectedPlug(spec *apparmor.Specificatio
        spec.AddSnippet(desktopConnectedPlugAppArmor)

        // Allow mounting document portal
-       emit := spec.EmitUpdateNSFunc()
+       emit := spec.EmitUpdateNSFunc
        emit("  # Mount the document portal\n")
        emit("  mount options=(bind) /run/user/[0-9]*/doc/by-app/snap.%s/ -> /run/user/[0-9]*/doc/,\n", plug.Snap().InstanceName())
        emit("  umount /run/user/[0-9]*/doc/,\n\n")
```

(kept emit := ... in place to reduce the diff and not touch all the emit(..) lines).

The tests seem to be happy.

zyga: Yeah, it can _now_ :-)

As this code was shaping, doing what you asked for would introduce observable differences and de-rail the argument of mechanical change that can be easier to review.

We can streamline this in this PR or in the follow-up.

stolowski: Allright. And as mentioned earlier, the Gen* helpers could become methods of spec and drop _emit_ func argument), .e.g.
```
spec.GenWritableProfile(target, 3)
```

zyga: I wouldn't do that because they generate apparmor rules that could equally well be placed in `snap-update-ns.pkg` as in `snap.pkg.app`. We could make it so that the emit function is simply AddUpdateNS so there's less glue code though.

zyga:
Rebased on prior branches now.

interfaces/builtin/content_test.go
```diff
@@ -962,12 +822,11 @@ slots:
...
-  /var/snap/consumer/common/import/ rw,
-  /var/snap/consumer/common/ rw,
-  /var/snap/consumer/ rw,
 `
-	c.Assert(updateNS[4], Equals, profile4)
-	c.Assert(updateNS, DeepEquals, []string{profile0, profile1, profile2, profile3, profile4})
+	// Find the slice that describes profile4 by looking till the end of the list.
+	start = end
+	c.Assert(strings.Join(updateNS[start:], ""), Equals, profile4)
+	c.Assert(strings.Join(updateNS, ""), DeepEquals, strings.Join([]string{profile0, profile1, profile2, profile3, profile4}, ""))
```
anonymouse64: :+1:

interfaces/builtin/desktop_test.go
```diff
@@ -117,11 +118,10 @@ func (s *DesktopInterfaceSuite) TestAppArmorSpec(c *C) {
...
 	updateNS = spec.UpdateNS()
-	c.Assert(updateNS, HasLen, 4)
-	c.Check(updateNS[0], testutil.Contains, "# Mount the document portal")
-	c.Check(updateNS[1], testutil.Contains, "# Read-only access to /usr/share/fonts")
-	c.Check(updateNS[2], testutil.Contains, "# Read-only access to /usr/local/share/fonts")
-	c.Check(updateNS[3], testutil.Contains, "# Read-only access to /var/cache/fontconfig")
+	c.Check(updateNS, testutil.Contains, "  # Mount the document portal\n")
+	c.Check(updateNS, testutil.Contains, "  # Read-only access to /usr/share/fonts\n")
+	c.Check(updateNS, testutil.Contains, "  # Read-only access to /usr/local/share/fonts\n")
+	c.Check(updateNS, testutil.Contains, "  # Read-only access to /var/cache/fontconfig\n")
```
anonymouse64: same question as above, is it still useful to check the length of the UpdateNS snippet?

tests/regression/lp-1848567/task.yaml
```diff
@@ -0,0 +1,44 @@
...
+  snap install --dangerous test-snapd-app_1_all.snap
+  snap install --dangerous test-snapd-gnome-3-28-1804_1_all.snap
+  snap install --dangerous test-snapd-gtk-common-themes_1_all.snap
+
+  snap connect test-snapd-app:gnome-3-28-1804 test-snapd-gnome-3-28-1804
+  snap connect test-snapd-app:icon-themes test-snapd-gtk-common-themes:icon-themes
+  snap connect test-snapd-app:sound-themes test-snapd-gtk-common-themes:sound-themes
+execute: |
+  # TODO: expand this test to install GNU time or find a suitable replacement.
+  if snap debug sandbox-features --required apparmor:kernel:mount && [ "$(command -v time)" != time ]; then
```
anonymouse64: As Samuele pointed out in SU, perhaps you could use `/usr/bin/time` from the base snaps, but note that I don't actually see `/usr/bin/time` in core or core18, so you would have to get it added there first.

tests/regression/lp-1848567/task.yaml
```diff
@@ -0,0 +1,44 @@
...
+    # Re-compile the apparmor profile for snap-update-ns for the
+    # gnome-system-monitor snap while ensuring that the profile is not loaded
```
anonymouse64: this snap isn't actually named `gnome-system-monitor`?

```suggestion
    # test-snapd-app snap while ensuring that the profile is not loaded
```

tests/regression/lp-1848567/task.yaml
zyga: Ohh, drat. I think I will just NIH a time-tool that does the equivalent without being a pain to use over time.

tests/regression/lp-1848567/task.yaml
zyga: This is a leftover from when I tested the real snap. I since re-wrote that to be a shim of the real thing that won't drift over time. Thanks for catching this.

interfaces/apparmor/spec.go
bboozzoo: As discussed that on IRC, we're aiming for shorter profile files.

tests/regression/lp-1848567/task.yaml
stolowski: This is a very nice test, but I think it's also ok to land it as a followup (given the annoying issue with time).

zyga: The test works as-is on some systems so it's already valuable as-is. We can iterate on it after the bug fix is merged so that it is tested more broadly on other systems.

interfaces/apparmor/spec.go
jdstrand: So, to be clear, the affected profiles in question ended up being 17500+ lines long, taking up 1M each. This takes up space which we don't need to use, but more importantly, and while I appreciate the sentiment, a 17500 line profile, commented or not, is pretty unreadable for auditing. It seems the least worst option is a smaller profile for this *auto-generated* snap-update-ns profile. For the app templates and snippets that are handcrafted, we expressly want to keep everything.

tests/regression/lp-1848567/task.yaml
jdstrand: With hello-world.sh, I noticed that `command -v time` is going to give `time` on UC16 (at least), but `command time -f "%M" ...` will fail with "bash: time: command not found". If in hello-world.sh I use /bin/sh (dash), then `command -v time` doesn't display 'time'. I mention this cause this seems to be working due to use of dash, but it seems brittle.

Furthermore, I don't care for the time check since it means we'll get a pass if the time check shortcircuits everywhere and that would go unnoticed. Normally we use `systems` to set the expectations. Why can't we do something similar here?

zyga:
@jdstrand I didn't want to use systems to constrain this to specific OSes. Apart from apparmor, which has a feature check via snapd, this should work everywhere.

I'll merge this now (it's green) and iterate on the test via follow-up

PR #7633, opened 2019-10-21
strutil: add OrderedSet

zyga:
This patch adds strutil.OrderedSet, a data structure representing a set
of strings which remembers order of inserted elements.

### DIFF
--- a/strutil/set.go
+++ b/strutil/set.go
@@ -28,7 +28,7 @@ type OrderedSet struct {
 
 // Items returns a slice of strings representing insertion order.
 //
-// Contains is O(N) in the size of the set.
+// Items is O(N) in the size of the set.
 func (o *OrderedSet) Items() []string {
 	if o.positionOf == nil {
 		return nil
--- a/strutil/set.go
+++ b/strutil/set.go
@@ -58,20 +58,6 @@ func (o *OrderedSet) IndexOf(item string) (idx int, ok bool) {
 	return idx, ok
 }
 
-// Del removes an item from the set.
-//
-// Del is O(1) in the size of the set.
-func (o *OrderedSet) Del(item string) {
-	if o.positionOf == nil {
-		return
-	}
-
-	delete(o.positionOf, item)
-	if len(o.positionOf) == 0 {
-		o.positionOf = nil
-	}
-}
-
 // Put adds an item into the set.
 //
 // If the item was not present then it is stored and ordered after all existing
--- a/strutil/set.go
+++ b/strutil/set.go
@@ -33,7 +33,7 @@ type OrderedSet struct {
 //
 // Items is O(N) in the size of the set.
 func (o *OrderedSet) Items() []string {
-	if o.positionOf == nil {
+	if len(o.positionOf) == 0 {
 		return nil
 	}
 	items := make([]string, len(o.positionOf))
@@ -47,7 +47,7 @@ func (o *OrderedSet) Items() []string {
 //
 // Contains is O(1) in the size of the set.
 func (o *OrderedSet) Contains(item string) bool {
-	if o.positionOf == nil {
+	if len(o.positionOf) == 0 {
 		return false
 	}
 
--- a/strutil/set.go
+++ b/strutil/set.go
@@ -90,7 +90,7 @@ func (o *OrderedSet) Put(item string) {
 	o.positionOf[item] = len(o.positionOf)
 }
 
-// Size returns the size of the set.
+// Size returns the number of elements in the set.
 func (o *OrderedSet) Size() int {
 	return len(o.positionOf)
 }
--- a/strutil/set.go
+++ b/strutil/set.go
@@ -21,6 +21,9 @@ package strutil
 
 // OrderedSet is a set of strings that maintains the order of insertion.
 //
+// The order of putting items into the set is retained. Putting items "a", "b"
+// and then "a", results in order "a", "b".
+//
 // External synchronization is required for safe concurrent access.
 type OrderedSet struct {
 	positionOf map[string]int
--- a/strutil/set.go
+++ b/strutil/set.go
@@ -1,0 +1,96 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package strutil
+
+// OrderedSet is a set of strings that maintains the order of insertion.
+//
+// External synchronization is required for safe concurrent access.
+type OrderedSet struct {
+	positionOf map[string]int
+}
+
+// Items returns a slice of strings representing insertion order.
+//
+// Contains is O(N) in the size of the set.
+func (o *OrderedSet) Items() []string {
+	if o.positionOf == nil {
+		return nil
+	}
+	items := make([]string, len(o.positionOf))
+	for item, idx := range o.positionOf {
+		items[idx] = item
+	}
+	return items
+}
+
+// Contains returns true if the set contains a given item.
+//
+// Contains is O(1) in the size of the set.
+func (o *OrderedSet) Contains(item string) bool {
+	if o.positionOf == nil {
+		return false
+	}
+
+	_, ok := o.positionOf[item]
+	return ok
+}
+
+// IndexOf returns the position of an item in the set.
+func (o *OrderedSet) IndexOf(item string) (idx int, ok bool) {
+	idx, ok = o.positionOf[item]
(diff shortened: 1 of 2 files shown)


strutil/set.go
```diff
@@ -0,0 +1,96 @@
...
+// OrderedSet is a set of strings that maintains the order of insertion.
+//
+// External synchronization is required for safe concurrent access.
+type OrderedSet struct {
+	positionOf map[string]int
+}
+
+// Items returns a slice of strings representing insertion order.
+//
+// Contains is O(N) in the size of the set.
```
chipaca: copy-paste error :-)

strutil/set.go
```diff
@@ -0,0 +1,96 @@
...
+func (o *OrderedSet) Items() []string {
+	if o.positionOf == nil {
```
chipaca: while you're short-cutting, why not
```suggestion
	if len(o.positionOf) == 0 {
```
?

chipaca: ... and then I spotted that len()==0 <=> it's nil, because you check and nil it. Good.

strutil/set.go
```diff
@@ -0,0 +1,96 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package strutil
+
+// OrderedSet is a set of strings that maintains the order of insertion.
```
chipaca: Typically in this bit I'd expect a note about whether `Put(a); Put(b); Put(a)` ends with `a` coming before, or after, `b`.

strutil/set.go
```diff
@@ -0,0 +1,96 @@
...
+
+// IndexOf returns the position of an item in the set.
+func (o *OrderedSet) IndexOf(item string) (idx int, ok bool) {
+	idx, ok = o.positionOf[item]
+	return idx, ok
+}
+
+// Del removes an item from the set.
+//
+// Delete is O(1) in the size of the set.
```
chipaca: del, or delete?

strutil/set.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	if _, ok := o.positionOf[item]; ok {
+		return
+	}
+
+	o.positionOf[item] = len(o.positionOf)
+}
+
+// Size returns the size of the set.
```
chipaca: This doc is bad and you should feel bad.
What is the size of the set? The number of elements that it currently contains? The number of elements that have been added? The sum of the lengths of all the strings it contains? The number of bytes allocated to it? To them?

Peeking at the implementation, why `Size` and not `Len` or `Length`?

strutil/set.go
```diff
@@ -0,0 +1,96 @@
...
+func (o *OrderedSet) Put(item string) {
+	if o.positionOf == nil {
+		o.positionOf = make(map[string]int)
+	}
+
+	if _, ok := o.positionOf[item]; ok {
+		return
+	}
+
+	o.positionOf[item] = len(o.positionOf)
```
anonymouse64: while potentially clever, I think this is wrong, see https://play.golang.org/p/Pp6vUyS3Uzp specifically, but what happens when we delete enough things that the length goes below something inserted earlier, then insert another thing? the newly inserted thing will be ordered _before_ the old thing and thus the ordering is broken. see:

```go
	s := OrderedSet{}
	s.Put("a") // a is 0
	fmt.Printf("%+v\n", s.positionOf)
	s.Put("b") // b is 1
	fmt.Printf("%+v\n", s.positionOf)
	s.Put("c") // c is 2
	fmt.Printf("%+v\n", s.positionOf)
	s.Del("a")
	fmt.Printf("%+v\n", s.positionOf)
	s.Del("b")
	fmt.Printf("%+v\n", s.positionOf)
	s.Put("d") // d is now len() == 1, but this is less than "c" which is 2
	fmt.Printf("%+v\n", s.positionOf)
```

I think instead an easy patch for this is to just keep a running counter, and increment it for every `Put`, but never decrement it.

chipaca: nice

strutil/set.go
zyga: Good point, I'll include this.

strutil/set.go
zyga: Yeah, I prefer Del because it's shorter and matches the other data structure we have but Delete is more like the go built-in. I'll unify to Del but we can swap to Delete later.

strutil/set.go
zyga: Size is more of a "number of elements" whereas Length is "the physical dimension of something", at least to me. I don't know if there is a commonly used term for this in Go, for beyond-builtin types.

As for the doc string, how about
```
// Size returns the cardinality of the set
```

or `number of elements` instead

strutil/set.go
zyga: Silly me, I'll have a look at fixing this now.

strutil/set.go
anonymouse64: cardinality is very mathy and specific so +1 from me as a nerd, but I don't know how non-nerds would read that, I guess probably first thing they would do is google "cardinality" and find the definition, but perhaps we should try to write docs that don't require folks to google first? how about something simple and obvious like:

```go
// Size returns the number of elements currently in the set
```

(possibly without `currently` I don't know if that's clear)

chipaca:
```golang
func (s OrderedSet) Cardinality() bool
...
s.Cardinality(🐦) == true
```

strutil/set.go
zyga: Done

strutil/set.go
zyga: Del/Delete is gone now

strutil/set.go
zyga: Done

strutil/set.go
zyga: We don't need to delete elements in the initial use case so I'll just skip implementing delete altogether. Since the implementation is private we can change it at any time, if we end up needing it in another spot.

Thank you for spitting this again.

anonymouse64: sure, if we don't need to delete things then the current code is fine

strutil/set.go
```diff
@@ -0,0 +1,85 @@
...
+		items[idx] = item
+	}
+	return items
+}
+
+// Contains returns true if the set contains a given item.
+//
+// Contains is O(1) in the size of the set.
+func (o *OrderedSet) Contains(item string) bool {
+	if len(o.positionOf) == 0 {
```
mvo5: Do we really need this check? It seems like we could just do:
```
       _, ok := o.positionOf[item]
	return ok
```
I don't think we actually safe much instructions with the extra if() (but I'm happy to be proven wrong).

strutil/set.go
```diff
@@ -0,0 +1,85 @@
...
+		return false
+	}
+
+	_, ok := o.positionOf[item]
+	return ok
+}
+
+// IndexOf returns the position of an item in the set.
+func (o *OrderedSet) IndexOf(item string) (idx int, ok bool) {
```
mvo5: Do we need both Contains and IndexOf ? Just wondering if its a case of YAGNI and we could add it later?

strutil/set_test.go
```diff
@@ -0,0 +1,74 @@
...
+func (s *orderedSetSuite) TestZeroValueItems(c *C) {
+	c.Assert(s.set.Items(), HasLen, 0)
+}
+
+func (s *orderedSetSuite) TestZeroValueContains(c *C) {
+	c.Check(s.set.Contains("foo"), Equals, false)
+}
+
+func (s *orderedSetSuite) TestZeroValueIndexOf(c *C) {
+	c.Check(s.set.Contains("foo"), Equals, false)
```
mvo5: This tests "set.Contains()" (just like in the test above) but the test function says "TestZeroValue**IndexOf**"

strutil/set_test.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (s *orderedSetSuite) TestZeroValuePut(c *C) {
+	s.set.Put("foo")
+	c.Check(s.set.Contains("foo"), Equals, true)
+	c.Check(s.set.Items(), DeepEquals, []string{"foo"})
+	idx, ok := s.set.IndexOf("foo")
+	c.Check(ok, Equals, true)
+	c.Check(idx, Equals, 0)
```
mvo5: I think it would be good to add at least one more string to get a non-zero index in the test.

mvo5: Maybe something like:
```
func (s *orderedSetSuite) TestZeroValuePut(c *C) {
	strs := []string{"foo", "bar", "baz"}
	for i, n := range strs {
		s.set.Put(n)
		c.Check(s.set.Contains(n), Equals, true)
		c.Check(s.set.Items(), DeepEquals, strs[:i+1])
		idx, ok := s.set.IndexOf(n)
		c.Check(ok, Equals, true)
		c.Check(idx, Equals, i)
                c.Check(s.set.Size(), Equals, i+1)
	}
}
``` 
?

strutil/set.go
zyga: It feels distinct enough to have both. Especially because of difference in return type.

strutil/set.go
zyga: I wanted to avoid working with nil maps.

zyga: Though on second thought this should work as well.

zyga: Though on second thought this should work as well.

strutil/set_test.go
zyga: I wrote a test modeled after this one. Thanks :)

strutil/set_test.go
zyga: Oops, thank you! Fixed

PR #6666, opened 2019-03-29
Change file names that are breaking go modules

slimjim777:
Fix file names as they break `go mod` when pulling in snapd into another project

Thanks for helping us make a better snapd!
Have you signed the [license agreement](https://www.ubuntu.com/legal/contributors) and read the [contribution guide](https://github.com/snapcore/snapd/blob/master/CONTRIBUTING.md)?

PR #7634, opened 2019-10-21
tests: ignore directories for go modules

slimjim777:
Have you signed the [license agreement](https://www.ubuntu.com/legal/contributors) and read the [contribution guide](https://github.com/snapcore/snapd/blob/master/CONTRIBUTING.md)? yes + yes

This PR has been created by a couple of people over the past couple of years (I accidentally closed #6666).

The background is when creating an application (like an agent that will operate on a device) that uses snapd as a library, it's common to want to use Go modules. However, there are a couple of files in the snapd repo that cause Go modules to stall. The recommended approach is to create an empty go.mod file that will ignore those directories: https://github.com/golang/go/wiki/Modules#can-an-additional-gomod-exclude-unnecessary-content-do-modules-have-the-equivalent-of-a-gitignore-file

### DIFF
--- a/tests/main/chattr/task.yaml
+++ b/tests/main/chattr/task.yaml
@@ -4,6 +4,12 @@ summary: test chattr
 # ppc64el disabled because of https://github.com/snapcore/snapd/issues/2503
 systems: [-ubuntu-core-*, -ubuntu-*-ppc64el]
 
+environment:
+    # an empty $topsrcdir/tests/go.mod seems to break importing or building go
+    # packages referenced by their import paths while under the tests directory,
+    # need to disable go modules supportfor this test
+    GO111MODULE: off
+
 prepare: |
   go build -o toggle ./toggle.go
 
--- a/tests/main/high-user-handling/task.yaml
+++ b/tests/main/high-user-handling/task.yaml
@@ -2,6 +2,12 @@ summary: Check that the refresh data copy works.
 
 systems: [-ubuntu-core-*]
 
+environment:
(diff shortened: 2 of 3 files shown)


slimjim777:
@anonymouse64  CI failed on the name of the PR, which I've changed. Can you trigger the build?

anonymouse64:
@slimjim777 done

bboozzoo:
some failures:
```
travis_time:start:30a04df9
2019-10-21 21:52:52 Error executing google:arch-linux-64:tests/main/static : 
-----
+ CGO_ENABLED=0
+ go build -o snapd.static github.com/snapcore/snapd/cmd/snapd
go: finding github.com/snapcore/snapd latest
can't load package: package github.com/snapcore/snapd/cmd/snapd: no matching versions for query "latest"
-----
2019-10-21 21:51:50 Error executing google:arch-linux-64:tests/main/high-user-handling : 
-----
++ command -v go
+ /usr/sbin/go build -o hightest test.go
go: finding github.com/snapcore/snapd latest
build command-line-arguments: cannot load github.com/snapcore/snapd/osutil/sys: no matching versions for query "latest"
-----
```
we probably do something silly in the tests. It also failed on Arch and Debian Sid only, both running go 1.13?

mvo5:
It seems like this test failure is specific to this PR, master looks ok.

bboozzoo:
Pushed a little workaround for the tests that failed. Let's see if it gets green now.

anonymouse64:
I restarted the jobs because they failed with store 403's here

bboozzoo:
Thank you!

PR #7639, opened 2019-10-22
interfaces: emit update-ns snippets to function

zyga:
The helper function emit collects individual update-ns snippets to
buffer, just like before, but allows us to easily change that. This will
be handy in the upcoming change that de-duplicates the update-ns
snippets to reduce memory usage of apparmor_parser.

### DIFF
--- a/interfaces/apparmor/spec.go
+++ b/interfaces/apparmor/spec.go
@@ -148,38 +148,41 @@ func (spec *Specification) AddLayout(si *snap.Info) {
 	// Append update-ns snippets that allow constructing the layout.
 	for _, path := range paths {
 		var buf bytes.Buffer
+		emit := func(f string, args ...interface{}) {
+			fmt.Fprintf(&buf, f, args...)
+		}
 		l := si.Layout[path]
-		fmt.Fprintf(&buf, "  # Layout %s\n", l.String())
+		emit("  # Layout %s\n", l.String())
 		path := si.ExpandSnapVariables(l.Path)
 		switch {
 		case l.Bind != "":
 			bind := si.ExpandSnapVariables(l.Bind)
 			// Allow bind mounting the layout element.
-			fmt.Fprintf(&buf, "  mount options=(rbind, rw) %s/ -> %s/,\n", bind, path)
-			fmt.Fprintf(&buf, "  mount options=(rprivate) -> %s/,\n", path)
-			fmt.Fprintf(&buf, "  umount %s/,\n", path)
+			emit("  mount options=(rbind, rw) %s/ -> %s/,\n", bind, path)
+			emit("  mount options=(rprivate) -> %s/,\n", path)
+			emit("  umount %s/,\n", path)
 			// Allow constructing writable mimic in both bind-mount source and mount point.
-			WritableProfile(&buf, path, 2) // At least / and /some-top-level-directory
-			WritableProfile(&buf, bind, 4) // At least /, /snap/, /snap/$SNAP_NAME and /snap/$SNAP_NAME/$SNAP_REVISION
+			GenWritableProfile(emit, path, 2) // At least / and /some-top-level-directory
+			GenWritableProfile(emit, bind, 4) // At least /, /snap/, /snap/$SNAP_NAME and /snap/$SNAP_NAME/$SNAP_REVISION
 		case l.BindFile != "":
 			bindFile := si.ExpandSnapVariables(l.BindFile)
 			// Allow bind mounting the layout element.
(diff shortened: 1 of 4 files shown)


interfaces/apparmor/export_test.go
```diff
@@ -108,3 +110,24 @@ func MockParserFeatures(f func() ([]string, error)) (resture func()) {
...
+		fmt.Fprintf(buf, f, args...)
+	}, path, assumedPrefixDepth)
+}
+
+// WritableProfile writes a profile for snap-update-ns for making given directory writable.
+func WritableProfile(buf *bytes.Buffer, path string, assumedPrefixDepth int) {
+	GenWritableProfile(func(f string, args ...interface{}) {
+		fmt.Fprintf(buf, f, args...)
+	}, path, assumedPrefixDepth)
+}
```
stolowski: These three helpers seem unused (replaced by GenWritable* helpers directly)?

interfaces/apparmor/spec.go
```diff
@@ -148,38 +148,41 @@ func (spec *Specification) AddLayout(si *snap.Info) {
 	// Append update-ns snippets that allow constructing the layout.
 	for _, path := range paths {
 		var buf bytes.Buffer
+		emit := func(f string, args ...interface{}) {
```
stolowski: Looking at the original PR and this now I'm slightly confused about what's the added value of this indirection, i.e. having EmitUpdateNSFunc that returns a simple function (that adds snippet to spec with de-duplication); why not simply change AddUpdateNS function to take variable number of args?

interfaces/apparmor/export_test.go
zyga: Those helpers are actually tested :) All the test code calls `WriteProfile` and look at the resulting buffer. I didn't want to change that as those are super tedious to touch.

interfaces/apparmor/spec.go
zyga: The original PR has to split each of the buffers that were multi-line strings into actual strings. This means that each of the `fmt.Fprintf(&buf, ...)` line has to change because we don't want to just `strings.Split()` the contents of the buffer and hope for the best.

The indirection gives us a way to do that with one change, the change of the emit function, across the whole stack. Here in this PR the emit function prints to a buffer, just like before. In the original PR emit adds a single line snippet to the `snap-update-ns` profile.

interfaces/apparmor/export_test.go
zyga: Ooops, they are not tested directly. I removed them now.

interfaces/apparmor/spec.go
stolowski: Ok, let me rephrase: why not simply have a public helper on spec:
`func Emit(f string, args ...interface{})`

+ make `GenWritable*` functions belong to Specification (no need to pass emit func around)?

zyga: We could do that but that would impact the tests. Right now we have certain batches that are added. Tests measure those precisely.

If you look at the other PR, there is something like that there, it's actually `UpdateNSEmitFunc() func(f string, args ...interface{})`. The difference is that only the other PR changes the granularity, along with the huge impact to how tests look like.

interfaces/apparmor/spec.go
```diff
@@ -220,20 +223,20 @@ func isProbablyPresent(path string) bool {
 	return path == "/" || path == "/snap" || path == "/var" || path == "/var/snap" || path == "/tmp" || path == "/usr" || path == "/etc"
 }
 
-// WritableMimicProfile writes apparmor rules for a writable mimic at the given path.
-func WritableMimicProfile(buf *bytes.Buffer, path string, assumedPrefixDepth int) {
-	fmt.Fprintf(buf, "  # Writable mimic %s\n", path)
+// GenWritableMimicProfile generates apparmor rules for a writable mimic at the given path.
+func GenWritableMimicProfile(emit func(f string, args ...interface{}), path string, assumedPrefixDepth int) {
```
bboozzoo: You can use something like:
```
type emitFunc func(f string, args ...interface{})

func GenWritableMimicProfile(emit emitFunc, path string, assumedPrefixDepth int) {
   ...
}
```
Or use whatever name seems fitting given the other PRs.

interfaces/apparmor/spec.go
stolowski: Allright, it's ok for this PR to keep it minimal, but I'll carry my arguments to the followup PR (sorry, don't mean to make your life harder, just trying to understand the reasoning :) )

interfaces/apparmor/spec.go
zyga: I'm happy to iterate on this after the bug fix is merged.

PR #7640, opened 2019-10-22
snap-recovery: deploy gadget content when creating partitions

mvo5:
This will deploy the gadget content to the newly created partitions in snap-recovery.

Based on https://github.com/snapcore/snapd/pull/7629


cmd/snap-recovery/partition/deploy.go
```diff
@@ -0,0 +1,94 @@
...
+	mountpoint := filepath.Join(deployMountpoint, strconv.Itoa(part.Index))
+	if err := os.MkdirAll(mountpoint, 0755); err != nil {
+		return err
+	}
+
+	// temporarily mount the filesystem
+	if err := sysMount(part.Node, mountpoint, part.Filesystem, 0, ""); err != nil {
+		return fmt.Errorf("cannot mount filesystem %q to %q: %v", part.Node, mountpoint, err)
+	}
+	defer sysUnmount(mountpoint, 0)
```
bboozzoo: I think we should catch umount errrors in case it's busy for some reason:
```
func deployFilesystemContent(part DeviceStructure, gadgetRoot string) (err error) {
...
defer func() {
    errUmount := sysUnmount(mountpoint, 0)
    if err == nil {
       err = errUmount
    }
}
...
```
{

cmd/snap-recovery/partition/deploy.go
```diff
@@ -0,0 +1,94 @@
...
+
+	fs, err := gadget.NewMountedFilesystemWriter(gadgetRoot, &part.LaidOutStructure)
+	if err != nil {
+		return fmt.Errorf("cannot create filesystem image writer: %v", err)
+	}
+
+	if err := fs.Write(mountpoint, []string{}); err != nil {
```
bboozzoo: For extra readability you can try:
```
preserveFiles := nil 
if err := fs.Write(mountpoint, preserveFiles); err != nil {
...
```

cmd/snap-recovery/partition/deploy.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func deployNonFSContent(part DeviceStructure, gadgetRoot string) error {
+	f, err := os.OpenFile(part.Node, os.O_RDWR, 0644)
+	if err != nil {
+		return fmt.Errorf("cannot deploy bare content for %q: %v", part.Node, err)
+	}
+	defer f.Close()
+
+	l := gadget.ShiftStructureTo(part.LaidOutStructure, 0)
```
bboozzoo: Could use a comment:
```
// Laid out structures start relative to the beginning of the volume, 
// shift the structure offsets to 0, so that it starts at the beginning of the partition
```

cmd/snap-recovery/partition/deploy.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	return raw.Write(f)
+}
+
+func DeployContent(created []DeviceStructure, gadgetRoot string) error {
+	for _, part := range created {
+		switch {
+		case part.Type == "bare":
+			return fmt.Errorf("cannot deploy type 'bare' yet")
+		case part.Filesystem == "":
```
bboozzoo: || part.Filesystem == "none"

This should really use `part.IsBare()` which we should eventually rename to something more meaningful, like `HasFilesystem() bool`

cmd/snap-recovery/partition/deploy.go
```diff
@@ -0,0 +1,103 @@
...
+	f, err := os.OpenFile(part.Node, os.O_RDWR, 0644)
+	if err != nil {
+		return fmt.Errorf("cannot deploy bare content for %q: %v", part.Node, err)
+	}
+	defer f.Close()
+
+	// Laid out structures start relative to the beginning of the
+	// volume, shift the structure offsets to 0, so that it starts
+	// at the beginning of the partition
+	l := gadget.ShiftStructureTo(part.LaidOutStructure, 0)
```
zyga: It's a little bit dangerous that this is required for correct operation. I understand how we got here but I think this could be folded into NewRawStructureWriter. Is there ever a case where this would not be wanted or required?

cmd/snap-recovery/partition/deploy.go
```diff
@@ -0,0 +1,103 @@
...
+	raw, err := gadget.NewRawStructureWriter(gadgetRoot, &l)
+	if err != nil {
+		return err
+	}
+	return raw.Write(f)
+}
+
+func DeployContent(created []DeviceStructure, gadgetRoot string) error {
+	for _, part := range created {
+		switch {
```
zyga: I'm not sure this switch statement is the natural flow because it may fool the reader into thinking this is handling a one-of-N situation where it seems it's only a if-then-else chain disguised as a switch.

Not a blocker but I'd prefer to see:
```
if !part.IsPartition() {
    return fmt.Errorf("cannot deploy non-partitions yet")
}
if part.HasFilesystem() {
    err = deployFilesystemContent(part, gadgetRoot)
} else {
    err = deployNonFSContent(part, gadgetRoot)
}
if err != nil {
    return err
}
```

PR #7643, opened 2019-10-23
overlord/devicestate: refactor and split into per-functionality files, drop dead code

bboozzoo:
As a prerequisite for the remodel work, attempt to refactor and split the
current code into files based on functionality. This should prevent
uncontrollable growth of devicestate and avoid the issues we have with
snapstate.

The PR is large, but it's just the code being moved around.


zyga:
This test failure looks real:
```
2019-10-23 11:21:56 Error executing google:ubuntu-18.04-64:tests/main/uc20-snap-recovery : 
-----
++ cat loop.txt
+ LOOP=/dev/loop2
+ echo 'Run the snap-recovery tool'
Run the snap-recovery tool
+ /usr/lib/snapd/snap-recovery ./gadget-dir /dev/loop2
+ echo 'And check that the partitions are created'
And check that the partitions are created
+ MATCH '1M\s+BIOS boot'
+ sfdisk -l /dev/loop2
+ MATCH '50M\s+EFI System'
+ sfdisk -l /dev/loop2
+ MATCH 'vfat\s* system-boot'
+ retry-tool lsblk --fs /dev/loop2
NAME      FSTYPE LABEL UUID MOUNTPOINT
loop2                       
ââloop2p1                   
ââloop2p2
```

PR #7646, opened 2019-10-23
tests: update mount-ns after addition of /etc/systemd/user

zyga:
We have recently added a bind mount for /etc/systemd/user to support
user-session services. This was not reflected in the mount-ns test
because core part of those tests are still disabled due to leaking
elements in two core tests.

This patch updates the test to pass if explicitly enabled.


zyga:
I suspect this is a fallout from 4b4cf41fd17951157221f406e6f7a5bb5f2f6fff

PR #7647, opened 2019-10-23
gadget: skip structures with MBR role during remodel

bboozzoo:
When using a remodel update policy, skip structures that do not have a partition
table entry.

### DIFF
--- a/gadget/update.go
+++ b/gadget/update.go
@@ -236,9 +236,10 @@ func defaultPolicy(from, to *LaidOutStructure) bool {
 	return to.Update.Edition > from.Update.Edition
 }
 
-// RemodelUpdatePolicy implements the update policy of a remodel scenario.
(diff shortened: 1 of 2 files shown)


gadget/update.go
```diff
@@ -237,7 +237,10 @@ func defaultPolicy(from, to *LaidOutStructure) bool {
 }
 
 // RemodelUpdatePolicy implements the update policy of a remodel scenario.
-func RemodelUpdatePolicy(_, _ *LaidOutStructure) bool {
+func RemodelUpdatePolicy(from, _ *LaidOutStructure) bool {
+	if !from.IsPartition() {
```
bboozzoo: I can actually think of a use case where this could make sense, eg. when updating a bootloader to a newer/different one. Perhaps we should only limit it to skip MBR role structures?

PR #7655, opened 2019-10-24
interfaces: allow introspecting network-manager on core

zyga:
We were alerted about a program that fails to work on core systems, with
the network-manager plug connected, because said program used
introspection to discover the available DBus methods.

Introspection was not a part of the network manager interface before.
After discussion with security is is now conditionally added on core
systems only, where the peer is scoped to a snap.pkg.app label that
pinpoints a specific snap. On classic system that would be unavoidably
"unconfined" where it would apply to all the classic services.

Fixes: https://bugs.launchpad.net/snapd/+bug/1849291

### DIFF
--- a/interfaces/builtin/network_manager.go
+++ b/interfaces/builtin/network_manager.go
@@ -294,7 +294,7 @@ dbus (receive, send)
     peer=(label=###SLOT_SECURITY_TAGS###),
 `
 
-const networkManagerSendIntrospectionSnippet = `
+const networkManagerConnectedPlugIntrospectionSnippet = `
 # Allow us to introspect the network-manager providing snap
 dbus (send)
     bus=system
@@ -303,7 +303,7 @@ dbus (send)
     peer=(label=###SLOT_SECURITY_TAGS###),
 `
 
-const networkManagerRecvIntrospectionSnippet = `
+const networkManagerConnectedSlotIntrospectionSnippet = `
 # Allow plugs to introspect us
 dbus (receive)
     bus=system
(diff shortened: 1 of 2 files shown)


interfaces/builtin/network_manager.go
```diff
@@ -294,6 +294,24 @@ dbus (receive, send)
     peer=(label=###SLOT_SECURITY_TAGS###),
 `
 
+const networkManagerSendIntrospectionSnippet = `
```
jdstrand: This should have ConnectedPlug in the name somewhere.

interfaces/builtin/network_manager.go
```diff
@@ -294,6 +294,24 @@ dbus (receive, send)
...
+# Allow us to introspect the network-manager providing snap
+dbus (send)
+    bus=system
+    interface="org.freedesktop.DBus.Introspectable"
+    member="Introspect"
+    peer=(label=###SLOT_SECURITY_TAGS###),
+`
+
+const networkManagerRecvIntrospectionSnippet = `
```
jdstrand: This should have ConnectedSlot somewhere in the name.

interfaces/builtin/network_manager.go
zyga: Done

interfaces/builtin/network_manager.go
zyga: Done

jdstrand:
@zyga - thanks for the PR and updates! :)

zyga:
@mvo5 please make sure to pull this into the 2.42.1 release if there is one.

PR #7656, opened 2019-10-24
tests: verify host is not affected by mount-ns tests

zyga:
The mount-ns test performs a set of measurements to mount namespaces.
The test is structured so that the host is measured first, followed by a
number of per-snap and per-user mount namespaces.

What the test doesn't do though, is to check that the test has not
affected the host in any way. This patch closes that loophole.

### DIFF
--- a/tests/main/mount-ns/task.yaml
+++ b/tests/main/mount-ns/task.yaml
@@ -127,7 +127,11 @@ prepare: |
     snap connect test-snapd-mountinfo-core16:mount-observe
     snap connect test-snapd-mountinfo-core18:mount-observe
 
-    echo "Collect mountinfo from the host"
+    echo "Collect mountinfo from the host before the test"
+    echo "but make sure that persistent mount namespaces don't clobber the output"
+    if mountinfo-tool /run/snapd/ns; then
+        umount /run/snapd/ns
+    fi
     cat /proc/self/mountinfo >HOST.raw.txt
 
     if snap debug sandbox-features --required confinement-options:classic; then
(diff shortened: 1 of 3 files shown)


tests/main/mount-ns/task.yaml
```diff
@@ -127,7 +127,9 @@ prepare: |
     snap connect test-snapd-mountinfo-core16:mount-observe
     snap connect test-snapd-mountinfo-core18:mount-observe
 
-    echo "Collect mountinfo from the host"
+    echo "Collect mountinfo from the host before the test"
+    echo "but make sure that persistent mount namespaces don't clobber the output"
```
bboozzoo: I think you can drop this line or make it a comment

tests/main/mount-ns/task.yaml
```diff
@@ -144,6 +148,14 @@ prepare: |
     su root -c "snap run test-snapd-mountinfo-core18" >PER-SNAP-18.raw.txt
     su test -c "snap run test-snapd-mountinfo-core18" >PER-USER-18.raw.txt
 
+    echo "Collect mountinfo from the host after the test"
```
anonymouse64: it's a little bit weird that this says "after the test" when it is being run in the prepare section... not that there's much to be done about this, I understand why this is done this way, I guess more just speaking to the wind as it were about this oddity

zyga: I see what you mean. I can change it to say "after executing snap apps".

zyga:
@anonymouse64 I'll send follow-ups, merging because green.

tests/main/mount-ns/task.yaml
zyga: Updated in https://github.com/snapcore/snapd/pull/7668

tests/main/mount-ns/task.yaml
zyga: Updated in https://github.com/snapcore/snapd/pull/7668

PR #7661, opened 2019-10-25
packaging: tweak handling of usr.lib.snapd.snap-confine

mvo5:
This PR switches the code to use dpkg-maintscript-helper to deal
with usr.lib.snapd.snap-confine. Technically it's a dpkg conffile
(even though I think that is a mistake) so we should let the
dpkg tools handle it. This PR tweaks the call to try to remove it
regardless of version, it can cause quite a bit of harm to have
it, e.g. https://forum.snapcraft.io/t/13132

If it is acceptable to foundations I would like to unconditionally
rename it to ".dpkg-bak" but it's (technically) a violation of
the policy.

Note that this is tested in the tests/main/upgrade-from-2.15
spread test.

### DIFF
--- a/packaging/ubuntu-16.04/snapd.maintscript
+++ b/packaging/ubuntu-16.04/snapd.maintscript
@@ -3,3 +3,4 @@
 rm_conffile /etc/grub.d/09_snappy 1.7.3ubuntu1
 rm_conffile /etc/ld.so.conf.d/snappy.conf 2.0.7~
 rm_conffile /etc/apparmor.d/usr.lib.snapd.snap-confine 2.23.6~ snap-confine
+rm_conffile /etc/apparmor.d/usr.lib.snapd.snap-confine
--- a/packaging/ubuntu-16.04/snapd.maintscript
+++ b/packaging/ubuntu-16.04/snapd.maintscript
@@ -2,5 +2,4 @@
 # we used to ship a custom grub config that is no longer needed
 rm_conffile /etc/grub.d/09_snappy 1.7.3ubuntu1
 rm_conffile /etc/ld.so.conf.d/snappy.conf 2.0.7~
-rm_conffile /etc/apparmor.d/usr.lib.snapd.snap-confine 2.23.6~ snap-confine
-rm_conffile /etc/apparmor.d/usr.lib.snapd.snap-confine
+rm_conffile /etc/apparmor.d/usr.lib.snapd.snap-confine 2.23.6~
--- a/packaging/ubuntu-16.04/snapd.postinst
+++ b/packaging/ubuntu-16.04/snapd.postinst
@@ -39,6 +39,16 @@ case "$1" in
             done
         fi
 
+        # In commit 0dce4704a5d (2017-03-28, snapd v2.23.6) we renamed
+        # /etc/apparmor.d/usr.lib.snap-confine to usr.lib.snap-confine.real
+        # to fix LP: #1673247 - however some people (developers?) still have
+        # the old usr.lib.snap-confine file. This seems to be loaded instead
+        # of the correct usr.lib.snap-confine.real profile. To fix this we
+        # use the rather blunt approach to rename the file forcefully.
+        if test -f /etc/apparmor.d/usr.lib.snapd.snap-confine && test -f /etc/apparmor.d/usr.lib.snapd.snap-confine.real; then
+            mv /etc/apparmor.d/usr.lib.snapd.snap-confine /etc/apparmor.d/usr.lib.snapd.snap-confine.dpkg-bak
+        fi
+
         # Ensure that the void directory has correct permissions.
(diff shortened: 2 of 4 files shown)


packaging/ubuntu-16.04/snapd.maintscript
```diff
@@ -3,3 +3,4 @@
 rm_conffile /etc/grub.d/09_snappy 1.7.3ubuntu1
 rm_conffile /etc/ld.so.conf.d/snappy.conf 2.0.7~
 rm_conffile /etc/apparmor.d/usr.lib.snapd.snap-confine 2.23.6~ snap-confine
+rm_conffile /etc/apparmor.d/usr.lib.snapd.snap-confine
```
vorlonofportland: The line before this is the one that was not obviously correct in the previous SRU, it's even less obvious to me why we would have two different rm_conffile lines here.  As far as I know, the second line is a superset of the first.

packaging/ubuntu-16.04/snapd.postinst
```diff
@@ -39,16 +39,6 @@ case "$1" in
...
 
-        # In commit 0dce4704a5d (2017-03-28, snapd v2.23.6) we renamed
-        # /etc/apparmor.d/usr.lib.snap-confine to usr.lib.snap-confine.real
-        # to fix LP: #1673247 - however some people (upgrades?) still have
-        # the old usr.lib.snap-confine file. This seems to be loaded instead
-        # of the correct usr.lib.snap-confine.real profile. To fix this we
-        # use the rather blunt approach to remove the file forcefully.
-        if test -f /etc/apparmor.d/usr.lib.snapd.snap-confine && test -f /etc/apparmor.d/usr.lib.snapd.snap-confine.real; then
-            rm -f /etc/apparmor.d/usr.lib.snapd.snap-confine
-        fi
```
vorlonofportland: I'm missing an understanding of why this rm -f wasn't already sufficient to solve the reported bugs.  Using rm_conffile may be "cleaner", but how does switching the method help the user-facing runtime behavior?

mvo5: Aha, thank you! If its considered okay to use this method I will change this PR to use rm (or mv to .dpkg-bak) and leave the rm_conffile calls alone. Will push that shortly.

mvo5:
This is now updated to restore back to the previous dpkg-maintscript usage. I did tweak the postinst snippet a little bit to just rename the usr.lib.snapd.snap-confine file (if it exists). Very few people should have this file, I tried many permutations of upgrades (see  https://forum.snapcraft.io/t/telegram-snap-fails-to-start/13132/3) and couldn't reproduce a leftover conffile. So it might just be developers who are affected by this who used "make hack" outside of dpkg. In any case, I think the postinst snippet is robust now and it leaves the file as a .dpkg-bak so *if* someone altered it for whatever reason at least their changes did not get lost.

mvo5:
@xnox If you agree with the PR, could you please +1 it?

packaging/ubuntu-16.04/snapd.maintscript
```diff
@@ -2,4 +2,4 @@
 # we used to ship a custom grub config that is no longer needed
 rm_conffile /etc/grub.d/09_snappy 1.7.3ubuntu1
 rm_conffile /etc/ld.so.conf.d/snappy.conf 2.0.7~
-rm_conffile /etc/apparmor.d/usr.lib.snapd.snap-confine 2.23.6~ snap-confine
+rm_conffile /etc/apparmor.d/usr.lib.snapd.snap-confine 2.23.6~
```
zyga: What is the semantics of this change? I understand it was referring to the snap-confine package before and now it no longer does so. Why?

mvo5: It was changed to snap-confine erroneously by me in a previous version of this that was never uploaded to the ubuntu archive.

jdstrand:
> Note to self - force to move apparmor profiles from /etc to /usr/lib across everywhere.

This has been discussed to death elsewhere and shouldn't distract from this PR, but in general they are in /etc because people do modify them. Unlike normal profiles shipped by deb packaging, snap-confine is special because we don't want people fiddling with it and it is fully managed by snapd (much like the application profiles themselves) and having it in /etc is just a burden, so big +1 for ditching the conffile approach as @mvo5 is doing in this PR.

@xnox - there are other options for apparmor in the general case like storing them in /usr somewhere and adding an override mechanism, perhaps a sensible config file (ala ucf) or similar. Since you are jotting down notes, perhaps add discussing it at the next engineering sprint?

packaging/ubuntu-16.04/snapd.maintscript
xnox: package name specification is redundant is automatically derived, and should normally not be used at all.

PR #7666, opened 2019-10-25
tests: don't depend on GNU time

zyga:
This branch contains two small patches that let us easily measure the maximum
RSS size of a process with a python helper.

This way we no longer have to special-case systems that don't have GNU time
installed. The single existing user of "time" is ported to use memory-tool
instead.


zyga:
@anonymouse64 actually I think I can test it in spread the same way I tested it during development.

zyga:
I'll send follow ups with spread test

PR #7669, opened 2019-10-27
snap, cmd/snap: support (but warn) deprecated multi-slash channel

chipaca:
if `$FOO` was `""`.

of behaviour.

break our users. So...

will come in a separate PR.

---

This is exactly the same as #7367, but if you look at the mess that github made of that...


chipaca:
Closing in favour of #7670

PR #7671, opened 2019-10-27
overlord/patch: normalize tracking channel in state

chipaca:
This PR adds a patch that will normalize the channel in the state.
Snapd was not normalizing the channel so it is possible that the
state contains channels like `/stable` or even `//edge//` without
perceivable ill effects. However with the upcoming change to
"default-tracks" that can be set store side the semantic of those
become less clear. To avoid any ambiguity we normalize the channels
so "/edge" -> "latest/edge".

I've built this on #7670, but it can easily be rebased on #7669 if that's the one we go for.

And yes, this is just an updating of #7377 (and a fix for a bug in it).

### DIFF
--- a/overlord/patch/patch.go
+++ b/overlord/patch/patch.go
@@ -34,7 +34,7 @@ var Level = 6
 // Sublevel is the current implemented sublevel for the Level.
 // Sublevel 0 is the first patch for the new Level, rollback below x.0 is not possible.
 // Sublevel patches > 0 do not prevent rollbacks.
-var Sublevel = 2
+var Sublevel = 3
 
 type PatchFunc func(s *state.State) error
 
--- a/overlord/patch/patch6.go
+++ b/overlord/patch/patch6.go
@@ -25,7 +25,7 @@ import (
 )
 
 func init() {
-	patches[6] = []PatchFunc{patch6, patch6_1, patch6_2}
+	patches[6] = []PatchFunc{patch6, patch6_1, patch6_2, patch6_3}
 }
 
 type patch6Flags struct {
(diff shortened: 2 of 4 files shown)


snap/channel/channel_test.go
```diff
@@ -189,20 +189,26 @@ func (s storeChannelSuite) TestParseErrors(c *C) {
...
 	}{
-		{"", "channel name cannot be empty"},
-		{"1.0////", "channel name has too many components: 1.0////"},
-		{"1.0/cand", "invalid risk in channel name: 1.0/cand"},
-		{"fix//hotfix", "invalid risk in channel name: fix//hotfix"},
-		{"/stable/", "invalid track in channel name: /stable/"},
-		{"//stable", "invalid risk in channel name: //stable"},
-		{"stable/", "invalid branch in channel name: stable/"},
-		{"/stable", "invalid track in channel name: /stable"},
+		{"", "channel name cannot be empty", ""},
```
mvo5: Mixing err and full feels a little bit strange but I guess it makes sense here.

snap/channel/channel_test.go
```diff
@@ -248,6 +254,37 @@ func (s *storeChannelSuite) TestFull(c *C) {
...
+
+	for _, t := range tests {
+		can, err := channel.Full(t.channel)
+		c.Assert(err, IsNil)
+		c.Check(can, Equals, t.str)
+	}
+}
+
+func (s *storeChannelSuite) TestFuncFullErr(c *C) {
+	_, err := channel.Full("foo/bar/baz/quux")
```
mvo5: Do need a test here that checks if we catch invalid risk levels? I.e. "1.0/no-risk-level" and "1.0/no-risk-level/foo" ? Or is that not an error ?

chipaca: it's not an error at this level, at least with the current code :-)

chipaca:
Whew! thank you

PR #7673, opened 2019-10-27
interfaces/desktop: allow access to system prompter interface

alexmurray:
This allows snaps to use pinentry-gnome3 via way of gpg-agent to unlock
private keys etc and to still integrate nicely with the user's existing
desktop.


alexmurray:
(force pushed with correct email address in commit)

alexmurray:
@zyga - this is related to https://github.com/snapcore/snapd/pull/7366 but I don't think there is a specific bug about it

Re safety of methods - I see these methods as akin to the existing DBus Notification methods we already allow - it allows an application to create a 'prompter' and get some input back from it - so this is similar to notifications IMO and so I don't see a safety issue in regards to providing this to untrusted applications. Users can easily dismiss any Prompt which is created using this interface as well.

alexmurray:
@jdstrand I would be keen to get your review on this as well since you are more familiar with what things are generally inside the threat model for snapd or not

alexmurray:
@jdstrand is this still on your radar?

interfaces/builtin/desktop.go
```diff
@@ -131,6 +131,22 @@ dbus (send)
...
+dbus (send)
+    bus=session
+    path=/org/gnome/keyring/Prompter
+    interface=org.gnome.keyring.internal.Prompter
+    member="{BeginPrompting,PerformPrompt,StopPrompting}"
+    peer=(label=unconfined),
+
+dbus (receive)
+    bus=session
+    path=/org/gnome/keyring/Prompt/p[0-9]*
```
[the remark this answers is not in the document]
alexmurray: The DBus API does not allow snooping or for other snaps to potentially receive input which was not destined for them - a prompt is initiated with the `BeginPrompting` method and this takes as an argument the path of the object to callback to - so the caller essentially says - start a prompt and call me back via this object which they provide the path to - this object is already registered via the DBus session bus - so they have full control over that - ie. the caller creates and pre-registers the object at `/org/gnome/keyring/Prompt/p[0-9]*` so they can receive the reply asynchronously via that object. So other than potentially allowing a denial-of-service via a snap continously spamming the interface I do not see that this a vulnerable attack surface to snaps.

alexmurray:
I rebased this on current master and moved it to `desktop-legacy` for now (with updated comment) as per @jdstrand's request - this way we can land support for this via `desktop-legacy` and if / when we are satisfied it does not present an unreasonable attack surface then we can perhaps move it to `desktop`.

jdstrand:
> I rebased this on current master and moved it to `desktop-legacy` for now (with updated comment) as per @jdstrand's request - this way we can land support for this via `desktop-legacy` and if / when we are satisfied it does not present an unreasonable attack surface then we can perhaps move it to `desktop`.

If as a member of the security team you're confident after your review that the API is safe, then leaving it in desktop is ok with me.

alexmurray:
After auditing gcr a bit more closely I am pretty sure this should be safe in terms of one snap being able to intercept credentials destined for another, but I believe there is the chance of a denial of service due to the way Gcr is coded - so the DBus interface itself is fine which I think is all we should really worry about in terms of this PR, but perhaps Gcr as a library could be a bit more defensive to try and avoid this DoS opportunity. The Gcr code is a bit convoluted due to the async nature of the APIs but I will try and summarise it here to explain how I believe this is safe from snooping on credentials:

1. Gcr uses this DBus interface in the [GcrSystemPrompt](https://gitlab.gnome.org/GNOME/gcr/-/blob/master/gcr/gcr-system-prompt.c) object - this is in-turn used to implement the `GcrPrompt` interface which is what clients would actually call
2. This maintains the callback object path in a private variable `prompt_path` which is [initialised when the object is constructed](https://gitlab.gnome.org/GNOME/gcr/-/blob/master/gcr/gcr-system-prompt.c#L424) - at this point there is no checking whether the name is available via DBus  - so it is possible that another snap etc may have already registered the name which the prompt wishes to use
3. [`perform_init_async`](https://gitlab.gnome.org/GNOME/gcr/-/blob/master/gcr/gcr-system-prompt.c#L942) maintains a state machine which is used to gradually initialise the object in an async manner - this is first called after the constructor has run via [gcr_system_prompt_real_init_async](https://gitlab.gnome.org/GNOME/gcr/-/blob/master/gcr/gcr-system-prompt.c#L1014) - at this point since there is no connection yet to DBus at all, the object will attempt to [connect to the session DBus](https://gitlab.gnome.org/GNOME/gcr/-/blob/master/gcr/gcr-system-prompt.c#L952).
4. Once connected to the session bus the [`on_bus_connected`](https://gitlab.gnome.org/GNOME/gcr/-/blob/master/gcr/gcr-system-prompt.c#L818) callback will be called - and this will proceed to try and register the prompt object on the session bus at `prompt_path` by calling [`register_prompt_object`](https://gitlab.gnome.org/GNOME/gcr/-/blob/master/gcr/gcr-system-prompt.c#L846)
5. If there was already another object registered with that name, this will [return an error](https://gitlab.gnome.org/GNOME/gcr/-/blob/master/gcr/gcr-system-prompt.c#L781) and this error will [propagate](https://gitlab.gnome.org/GNOME/gcr/-/blob/master/gcr/gcr-system-prompt.c#L854) back to the original caller who constructed the `GcrSystemPrompt` - ie. the primary interface to this is [`gcr_system_prompt_open_for_prompter_async`](https://gitlab.gnome.org/GNOME/gcr/-/blob/master/gcr/gcr-system-prompt.c#L1413) which will propagate the error which occurs on registration of the path on dbus back to the caller.

So *if* a malicious snap where to go and register a heap of `/org/gnome/keyring/Prompt/p[0-9]*` object paths on the session bus this would stop other snaps being able to create and use this interface via `Gcr` (since when they go to create their prompts it would error out as the path would already exist) - however I am not sure we can protect against all DoS type attacks. `Gcr` could try and be smarter and when creating the `prompt_path` and check to see if this is available on the session bus and if not try and find one that does not exist already - but I don't think we should try and put in place snap interface policy to protect against poor interface implementations. So to my mind I feel we should allow this DBus interface via the `desktop` snap interface since the DBus API is safe, and if there is a possible issue with `Gcr` or `gnome-shell` or any other implementation of the interface that can be followed up separately if needed.

alexmurray:
As per the above audit, I have moved this back to the `desktop` interface and added a comment in the AppArmor policy as to why this is safe.

alexmurray:
The TravisCI failure https://travis-ci.org/snapcore/snapd/builds/656022977?utm_source=github_status appears to be a transient failure unrelated to this PR - is there any way I can retrigger that to run again?

interfaces/builtin/desktop.go
```diff
@@ -131,6 +131,32 @@ dbus (send)
...
+    member="{BeginPrompting,PerformPrompt,StopPrompting}"
+    peer=(label=unconfined),
+
+# Whilst this path is not snap specific (ie. any snap or other application
+# could register it on the session bus) this is safe to allow snaps to
+# receive via this since the snap itself has to provide this path name via
+# the BeginPrompting/PerformPrompt methods above to initiate a prompt in
+# the first place so it has complete control to first register this path on
+# the session bus ensure no other application/snap can do so and possibly
+# intercept their response
```
jdstrand: Since the testsuite passed, that gives us the opportunity to make this slightly less wordy. :) How about:
```
# While the DBus path is not snap-specific, by the time an application
# registers the prompt path via DBus, Gcr will check that it isn't
# already in use and send the client an error. See:
# https://github.com/snapcore/snapd/pull/7673#issuecomment-592229711
```

alexmurray: Thanks - yep this is a bit less wordy :)

jdstrand:
@alexmurray - this is the failure:
```
=== RUN   TestOverlord

FAIL: overlord_test.go:694: overlordSuite.TestEnsureLoopPruneAbortsOld

overlord_test.go:748:
    // change was aborted
    c.Check(chg.Status(), Equals, state.HoldStatus)
... obtained state.Status = 3 ("Doing")
... expected state.Status = 1 ("Hold")

OOPS: 97 passed, 1 FAILED
```

This is in the unit tests and not a transient failure. I suggest merging from master (but not rebasing) and push here. That will trigger a new test run with updated master.

alexmurray:
Thanks for the hint re merging from master @jdstrand  - the unit tests at least have passed this time so :crossed_fingers:

PR #7691, opened 2019-10-29
builtin/browser_support.go: allow monitoring process memory utilizati…

Erick555:
…on (used by chromium)

Blockng this creates denials as chromium tries to write into it
continously. Allowing acces should be safe.

http://man7.org/linux/man-pages/man5/proc.5.html

### DIFF
--- a/interfaces/builtin/browser_support.go
+++ b/interfaces/builtin/browser_support.go
@@ -230,6 +230,9 @@ owner /{dev,run}/shm/WK2SharedMemory.* mrw,
 
 # Chromium content api on (at least) later versions of Ubuntu just use this
 owner /{dev,run}/shm/shmfd-* mrw,
+
+# Allow monitoring process memory utilization (used by chromium)
+owner @{PROC}/@{pid}/clear_refs w,
 `
 
 const browserSupportConnectedPlugSecComp = `


jdstrand:
> Looking at the manual page and at the rule I'm happy with allowing this
> 
> Perhaps it could even go to the default template?
> 
> CC @jdstrand

I'd prefer to not add it to the default template because while the man page says "writable only by owner of the process", it is talking about DAC, not the process tree for this process (ie, adding to template would mean that daemon snaps could write to the file for other snaps/unconfined root processes, etc).

interfaces/builtin/browser_support.go
```diff
@@ -230,6 +230,9 @@ owner /{dev,run}/shm/WK2SharedMemory.* mrw,
 
 # Chromium content api on (at least) later versions of Ubuntu just use this
 owner /{dev,run}/shm/shmfd-* mrw,
+
+# Allow monitoring process memory utilization (used by chromium)
+owner @{PROC}/@{pid}/clear_refs w,
```
jdstrand: This looks fine since while this would allow the snap to 'w'rite to the file for any pid with matching owner (ie, it is not specific to the snap), it is in browserSupportConnectedPlugAppArmorWithSandbox which requires a snap declaration to use.

PR #7697, opened 2019-10-30
interfaces/apparmor: handle pre-seeding mode

stolowski:
Add SetPreseedMode() method to apparmor backend and pass `--skip-kernel-load` flag to apparmor_parser when in pre-seeding mode. This has no effect yet until enabled with later PRs.


interfaces/apparmor/backend.go
```diff
@@ -467,14 +487,22 @@ func (b *Backend) SetupMany(snaps []*snap.Info, confinement func(snapName string
 	}
 
 	if !fallback {
+		aaFlags := skipReadCache | conserveCPU
```
bboozzoo: Can we drop conserveCPU when in preseed mode?

stolowski: We probably could, but I'm not sure it's going to gain much / anything, the pre-bakinkg is done once to prepare an image, gaining a fraction of a second here is mostly irrelevant (contrary to saving it on actual firstboot/every boot).

